PR #208, opened 2013-10-20
WIP: LSF submission

alexbw:
Second try submitting a PR for the LSF batch submission system (messed the first one up while trying to juggle a separate PR).


jsmedmar:
So I guess LSF support was never added?

Issue #222, opened 2013-11-14
Workers quit before all pending tasks are done


jimjh:
... and as a result some tasks are never run and eventually expired.

Consider the following set of tasks:

```python
class Block(luigi.Task):
  def run(self):
    # sleep for 20 seconds
  def complete(self):
    # return true after it has slept for 20 seconds

class Dep(luigi.Task):
  def requires(self):
    return Block()
  def run(self):
    # do something, such as creating a file
```

Submit `Block`, then in another terminal submit `Dep` while `Block` is running.

```bash
$> ./block_task.py Block
$> ./block_task.py Dep
```

```
INFO: There are no more tasks to run at this time
INFO: Block() is currently run by worker worker-428203729
INFO: Worker was stopped. Shutting down Keep-Alive thread
```

Now visit the scheduler's web UI at http://localhost:8082. We see that `Dep` is still pending, but it never gets run because all workers have exited.

![screen shot 2013-11-14 at 9 31 23 am](https://f.cloud.github.com/assets/965782/1543196/ab8c6826-4d52-11e3-8c70-b976fbb863fe.png)

I believe this is related to the following comment in `luigi/worker.py`.

```
# TODO: sleep for a bit and query server again if there are
# pending tasks in the future we might be able to run
```

jimjh:
Oops forgot to close this. Fixed in #225.

PR #225, opened 2013-11-14
Let worker wait for tasks running on other workers to complete.

jimjh:
See #222 and #223.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -47,13 +47,24 @@ class Worker(object):
     """
 
     def __init__(self, scheduler=CentralPlannerScheduler(), worker_id=None,
-                 worker_processes=1, ping_interval=None):
+                 worker_processes=1, ping_interval=None, keep_alive=None,
+                 wait_interval=None):
         if not worker_id:
             worker_id = 'worker-%09d' % random.randrange(0, 999999999)
 
+        config = configuration.get_config()
+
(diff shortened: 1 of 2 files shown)


duckontheweb:
Did this ever make it's way into the documentation? I see that it's referenced in a couple of places, but I didn't see an entry in the `[core]` section of the configuration docs for it.

dlstadther:
@duckontheweb it's in the [`[worker]` section](https://luigi.readthedocs.io/en/stable/configuration.html#worker)

duckontheweb:
It seems like the worker code [here](https://github.com/spotify/luigi/blob/master/luigi/worker.py#L412-L413) is getting that config from the `[core]` section, though. Am I misinterpreting that?

dlstadther:
The legacy method was to provide the `config_path=dict(section='core', name='worker-keep-alive')`. So for legacy purposes, it will look under `[core]` for `worker-keep-alive`. But config classes (`class worker(luigi.config)`) will look under `[worker]` (the name of the class) for the parameter names listed. So in this case `keep_alive`.

So to be verbose in this example, you could provide
```
[core]
worker-keep-alive: True
```
or
```
[worker]
keep_alive: True
```

duckontheweb:
Perfect, thanks for clarifying that for me!

PR #479, opened 2014-09-30
Initial prototype of a "slave" worker

erikbern:
If a worker is created using slave=True, then it will pull *any* tasks from the scheduler,
not just tasks that are assigned to it.

The biggest limitation atm is that the slave needs to be aware of what those tasks map to,
i.e. somehow they have to be imported for the worker to know what to do with the tasks.

It also doesn't handle failure very well.

Changed the semantics of how interface.load_task works a bit. It will now rely on the
Register class to locate task classes rather than importing modules dynamically. There is
still an option to import modules but it's not run by default.

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -32,7 +32,6 @@ from luigi import parameter
 from luigi import rpc
 from luigi import scheduler
 from luigi import task
-from luigi import worker
 from luigi.task import Register
 
 
@@ -40,7 +39,8 @@ def load_task(module, task_name, params_str):
     """
     Imports task dynamically given a module and a task name.
     """
-    __import__(module)
+    if module is not None:
+        __import__(module)
     task_cls = Register.get_task_cls(task_name)
     return task_cls.from_str_params(params_str)
 
@@ -122,6 +122,7 @@ class WorkerSchedulerFactory(object):
         return rpc.RemoteScheduler(host=host, port=port)
 
     def create_worker(self, scheduler, worker_processes):
+        from luigi import worker
         return worker.Worker(
             scheduler=scheduler, worker_processes=worker_processes)
 
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -604,7 +604,7 @@ class CentralPlannerScheduler(Scheduler):
                 return False
         return True
 
-    def get_work(self, worker, host=None, **kwargs):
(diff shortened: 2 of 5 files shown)


erikbern:
Rebased this, getting even simpler. I think I can add the dynamic module loading pretty easy. Will keep experimenting.

My goal is to build the next generation of the scheduler by triggering jobs from the scheduler to a worker pool.

Also did anyone have any thoughts about what word to use instead of "slave"?

Tarrasch:
What about "volunteer"? They are in fact not forced into work, they volunteer and happily take the workload the server assigns to them, so that word would make more sense. Or is my understanding of this new worker type not correct?

daveFNbuck:
How about "dedicated worker" for a name?

I just noticed that some of my larger backfills have been silently failing lately because they use too much memory to fork all the workers and hadoop jobs successfully. Looking forward to this to mitigate the problem.

erikbern:
Maybe deputy :) or assistant, executor, clerk, contractor, laborer, workforce

erikbern:
I think it would be great if we can pick this up an start introducing more and more elements of a purely distributed task execution framework. Unfortunately my time is very constrained right now. But I can at least try to rebase this feature and we can merge it as an experimental feature

Tarrasch:
Sure. As long as it's optional and doesn't break anything. ^^

erikbern:
Of course

daveFNbuck:
If we can get it in as an experimental feature, I'll help iterate on it as a test subject.

erikbern:
That's awesome. I'll take a look at it.

erikbern:
Rebased it. I also renamed "slave" to "assistant"

There's an annoying circular dependency between `luigi.worker` and `luigi.interface` that I solved by using a dynamic import. We should refactor soon (both are huge).

gpoulin:
@erikbern This import loop also caused me issues when porting to python3 (e3ec5f46). Would be nice to refactor this.

erikbern:
Ok, I'll find some time to address it.

erikbern:
Does anyone object if I merge this? It's an undocumented feature (so far) but it's a starting point for a lot of cool things and it's awesome if someone (like @daveFNbuck or @tommyengstrom) has time to look at it

daveFNbuck:
Please do. I'll start using it today.

PR #569, opened 2014-12-11
Times out long-running workers

daveFNbuck:
I woke up this morning to find that very few jobs had run overnight because
something weird happened with our hadoop cluster that caused all of our jobs to
hang permanently. When I checked in the morning, all workers were busy running
jobs that had been going for about 10 hours, and thousands of pending jobs that
would normally have been run overnight had piled up.

After killing and restarting all of my workers, everything started running fine
again because whatever weird issue had caused everything to hang was over. In
order to automate this fix, I've added a worker_timeout property to tasks that
tells the worker how long to attempt to run the task before killing it.

This property defaults to 0, meaning do not time tasks out. It can be changed
individually for a task by override or globally for all non-overridden tasks
via client.cfg. This only applies when running with more than 1 worker process,
as we need to run the task in a separate process in order to be able to kill it
after the timeout.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -126,6 +126,15 @@ worker-ping-interval
   Number of seconds to wait between pinging scheduler to let it know
   that the worker is still alive. Defaults to 1.0.
 
+worker-timeout
+  .. versionadded:: 1.0.20
+  Number of seconds after which to kill a task which has been running
+  for too long. This provides a default value for all tasks, which can
+  be overridden by setting the worker-timeout property in any task. This
+  only works when using multiple workers, as the timeout is implemented
+  by killing worker subprocesses. Default value is 0, meaning no
+  timeout.
+
 worker-wait-interval
   Number of seconds for the worker to wait before asking the scheduler
   for another job after the scheduler has said that it does not have any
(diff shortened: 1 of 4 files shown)


thatneat:
Can I get an example of how to use this? I'd like a particular task to time out after 5 seconds.
I've tried these: 

```python
    class TimeoutTask(luigi.Task):
        worker_timeout = 5

    class TimeoutTask(luigi.Task):
        worker_timeout = luigi.IntParameter(default=5)
```

I've also tried setting the timeout downstream:

```python
    class DownstreamTask(luigi.Task):
        def requires(self):
            return TimeoutTask(worker_timeout=5)
```
in any case though, the task seems to just run perpetually. I'm using the centralized scheduler and TimeoutTask continously shows under Running Tasks and the downtstream tasks show as Pending.

daveFNbuck:
You need to run with multiple workers for timeouts to be enabled. Add `--workers 2` or some higher number to get the behavior you expect. Either method should work, but I usually use the first one of just setting worker_timeout to a constant.

thatneat:
Thanks! PR above should help clarify the situation for the next person coming along.

PR #570, opened 2014-12-11
Groups tasks that supersede each other together in the scheduler

daveFNbuck:
Many tasks supersede older versions of the same job. For example, if I do an
hourly sqoop job to copy data from a production db to a hive table, completing
the 8am copy means I don't need to run the 7am copy. If both the 7am copy and
the 8am copy are pending and ready, I should run the 8am copy and consider them
both complete.

This commit achieves these goals by allowing jobs to define two optional
properties. The first is supersedes_bucket. Tasks with the same supersedes
bucket (if not None) are considered by the scheduler to supersede each other.
The scheduler will mark all earlier versions of the task DONE when one of them
completes. It us up to the user to ensure that earlier versions of the task
become complete when a later version is run.

Earlier tasks are identified by having a lower supersedes_priority. The
scheduler will prefer to schedule tasks with higher supersedes_priority.
Supersedes priority can be any type, but most be comparable between tasks in
the same bucket.

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -94,7 +94,7 @@ class RemoteScheduler(Scheduler):
 
     def add_task(self, worker, task_id, status=PENDING, runnable=False,
                  deps=None, new_deps=None, expl=None, resources={},priority=0,
-                 family='', params={}):
+                 family='', params={}, supersedes_bucket=None, supersedes_priority=None):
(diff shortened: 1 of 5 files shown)


erikbern:
Sorry for the long delay about this one. Let me think about it a bit more and get back to you

PR #584, opened 2014-12-23
Keeps track of tasks in scheduler grouped by status

daveFNbuck:
I've been seeing half-second get_work runtimes recently on my scheduler. The
cause of this has been a large backfill that scheduled a lot of done tasks. With
about 100k done tasks and about 350 pending tasks, a lot of time is wasted
looping over done tasks just to ignore them. This essentially caches these
computations by keeping a dictionary to group tasks by status and updating it
whenever task statuses are changed. This has sped my scheduling up by an order
of magnitude, now taking about 20ms per get_work.

In order to ensure updates happen every time, the task state has to take over
any functions that modify status. This is a necessary for remote task storage
anyway, so it had to be done eventually. There are no new tests here because
there are no functionality changes, only improvements in efficiency and
different internal representations.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -15,6 +15,7 @@
 import collections
 import datetime
 import functools
+import itertools
 import notifications
 import os
 import logging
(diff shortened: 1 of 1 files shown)


erikbern:
Was just going to merge this but now it looks like there's a merge conflict. Would you mind fixing and then I can merge? Thanks!

daveFNbuck:
Fixed

Issue #592, opened 2015-01-02
Provide RDBMS target using SQLAlchemy 


gouthambs:
@rantav @erikbern I have an implementation of a SQLAlchemy target. I created this feature in luigi to be able to test by writing to SQLite.  Using SQLAlchemy will open up other db targets such as Oracle or SQL Server that don't currently exist in luigi.

I am wondering if there is an interest in taking this feature if I provide a pull request.

gouthambs:
My implementation (so far) only requires sqlalchemy, which is already something that luigi requires. Additional dependencies will be left to the user based on what dbapi they chose to use.

erikbern:
I think it sounds great. You can put it in luigi.contrib. Regarding imports I think it's fine to depend on anything as long as it's separate from the core luigi codebase.

PR #593, opened 2015-01-08
Sql alchemy targets

gouthambs:
Provides the feature described in #592

### DIFF
--- a/luigi/contrib/sqla.py
+++ b/luigi/contrib/sqla.py
@@ -160,7 +160,6 @@ class CopyToTable(luigi.Task):
             echo=self.echo
             )
 
-
     def rows(self):
         """Return/yield tuples or lists corresponding to each row to be inserted """
         with self.input().open('r') as fobj:
--- a/luigi/contrib/sqla.py
+++ b/luigi/contrib/sqla.py
@@ -52,6 +52,8 @@ class SQLAlchemyTarget(luigi.Target):
 
     def exists(self):
         row = None
+        if self.marker_table_bound is None:
+            self.create_marker_table()
         try:
             with self.engine.begin() as conn:
                 table = self.marker_table_bound
--- a/luigi/contrib/sqla.py
+++ b/luigi/contrib/sqla.py
@@ -11,6 +11,7 @@ import abc
 import logging
 import luigi
 import datetime
+import itertools
 from sqlalchemy import Table, MetaData, Column, String, DateTime, create_engine, select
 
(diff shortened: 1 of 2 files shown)


gouthambs:
@erikbern what would be a good place to document this addition? I haven't had a chance to do that yet.

erikbern:
I think you can put documentation in the code. It will be auto-generated and uploaded to readthedocs.org

luigi/contrib/sqla.py
```diff
@@ -0,0 +1,215 @@
...
+from sqlalchemy import Table, MetaData, Column, String, DateTime, create_engine, select
+
+logger = logging.getLogger('luigi-interface')
+
+
+class SQLAlchemyTarget(luigi.Target):
+    """Target for a resource in database using SQLAlchemy.
+
+    This will rarely have to be directly instantiated by the user"""
+    marker_table = luigi.configuration.get_config().get('sqlalchemy', 'marker-table', 'table_updates')
```
erikbern: I would prefer it not to hit configuration in the global scope. Can you set it `None` and look it up when needed?

gouthambs:
@erikbern I have removed the configuration usage in the global scope. Take a look and let me know if you have any more thoughts.

erikbern:
LGTM – I'll give it a few days in case anyone else has thoughts (@freider @Tarrasch @daveFNbuck maybe?)

Tarrasch:
Hi @gouthambs , I would like to hear if you have/are used this in production. What motivated you to do this? Also I want to give +100 on the documentation! On a glance, this looks very well written!

gouthambs:
@Tarrasch I am employing this part of a development project that I am woking on. It is not in production yet.

I use SQLAlchemy in my projects which helps me to not worry so much about the database layer. My motivation to write the SQLAlchemyTarget was to have the same freedom in some of my processing jobs. Though this is the first time I am working with luigi.

gouthambs:
@erikbern The docs are not showing up for this module.

erikbern:
Yeah that's pretty weird. Looks like readthedocs rebuilt the docs but didn't include this submodule: http://luigi.readthedocs.org/en/stable/

Anyone has an idea?

gouthambs:
Should there be an entry in doc/api/luigi.contrib.rst?

erikbern:
I think that explains it yeah. Would you mind adding it?

gouthambs:
@erikbern #596 adds this.

PR #594, opened 2015-01-12
Add support for specifying Snakebite effective user

RobCherry:
Adding a configuration option for specifying the effective_user parameter for the Snakebite clients.

### DIFF
--- a/luigi/hdfs.py
+++ b/luigi/hdfs.py
@@ -285,27 +285,21 @@ class SnakebiteHdfsClient(HdfsClient):
         If Luigi has forked, we have a different PID, and need to reconnect.
         """
         if self.pid != os.getpid() or not self._bite:
-            autoconfig_enabled = self.config.getboolean("hdfs", "snakebite_autoconfig", False)
-            if autoconfig_enabled is True:
+            client_kwargs = dict(filter(lambda (k, v): v is not None and v != '',  {
+                'hadoop_version': self.config.getint("hdfs", "client_version", None),
+                'effective_user': self.config.get("hdfs", "effective_user", None)
+            }.items()))
+            if self.config.getboolean("hdfs", "snakebite_autoconfig", False):
                 """
                 This is fully backwards compatible with the vanilla Client and can be used for a non HA cluster as well.
                 This client tries to read ``${HADOOP_PATH}/conf/hdfs-site.xml`` to get the address of the namenode.
                 The behaviour is the same as Client.
                 """
                 from snakebite.client import AutoConfigClient
-                self._bite = AutoConfigClient()
+                self._bite = AutoConfigClient(**client_kwargs)
             else:
                 from snakebite.client import Client
-                try:
-                    ver = self.config.getint("hdfs", "client_version")
-                    if ver is None:
-                        raise RuntimeError()
-                    self._bite = Client(self.config.get("hdfs", "namenode_host"),
-                                       self.config.getint("hdfs", "namenode_port"),
-                                       hadoop_version=ver)
-                except:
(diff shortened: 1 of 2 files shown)


luigi/hdfs.py
```diff
@@ -285,27 +285,21 @@ def get_bite(self):
         If Luigi has forked, we have a different PID, and need to reconnect.
         """
         if self.pid != os.getpid() or not self._bite:
-            autoconfig_enabled = self.config.getboolean("hdfs", "snakebite_autoconfig", False)
-            if autoconfig_enabled is True:
+            client_kwargs = dict(filter(lambda (k, v): v is not None and v != '',  {
+                'hadoop_version': self.config.getint("hdfs", "client_version", None),
+                'effective_user': self.config.get("hdfs", "effective_user", None)
+            }.items()))
```
Tarrasch: Thanks for respecting 2.6 backwards compatibility (by not using dict comprehensions) :)

luigi/hdfs.py
```diff
@@ -285,27 +285,21 @@ def get_bite(self):
...
-                    ver = self.config.getint("hdfs", "client_version")
-                    if ver is None:
-                        raise RuntimeError()
-                    self._bite = Client(self.config.get("hdfs", "namenode_host"),
-                                       self.config.getint("hdfs", "namenode_port"),
-                                       hadoop_version=ver)
-                except:
-                    self._bite = Client(self.config.get("hdfs", "namenode_host"),
-                                       self.config.getint("hdfs", "namenode_port"))
+                self._bite = Client(self.config.get("hdfs", "namenode_host"), self.config.getint("hdfs", "namenode_port"), **client_kwargs)
```
Tarrasch: Quite some changes here. Have you tested that the new stuff works?

luigi/hdfs.py
```diff
@@ -285,27 +285,21 @@ def get_bite(self):
...
+            }.items()))
+            if self.config.getboolean("hdfs", "snakebite_autoconfig", False):
                 """
                 This is fully backwards compatible with the vanilla Client and can be used for a non HA cluster as well.
                 This client tries to read ``${HADOOP_PATH}/conf/hdfs-site.xml`` to get the address of the namenode.
                 The behaviour is the same as Client.
                 """
                 from snakebite.client import AutoConfigClient
-                self._bite = AutoConfigClient()
+                self._bite = AutoConfigClient(**client_kwargs)
```
Tarrasch: Not related to this patch, but do you see why it uses the `AutoConfigClient` when `snakebite_autoconfig==False`?

RobCherry: I don't follow.

luigi/hdfs.py
RobCherry: Unless there was some other exception that is intended to be caught and result in a new Client without the specified version, yes.

Tarrasch: Yea, that code was quite badly written. Nice refactoring! :)

luigi/hdfs.py
Tarrasch: Eh... I think github is buggy or something. Why is the code **different** in the conversation view [1] and the diff view????

[1]: https://github.com/spotify/luigi/pull/594#discussion_r22802079
[2]: https://github.com/spotify/luigi/pull/594/files#r22802079

Anyway. Your code change makes total sense, I just misread the `False` in the if guard to be ` == False`, but of course that's not what it says. :)

erikbern:
LGTM

Tarrasch:
I'll make changes to the snakebite code myself today. So this must merge now (and everyone seems happy with it). :)

Tarrasch:
Thank you very much @RobCherry !

PR #599, opened 2015-01-13
Add bulk_complete to Task to have an explicit API for range tools

ulzha:
All kinds of reusable base classes (e.g. CopyToTable) can implement bulk_complete, and become schedulable with RangeHourly.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -484,6 +484,15 @@ class Task(object):
 
         return all(itertools.imap(lambda output: output.exists(), outputs))
 
+    @classmethod
+    def bulk_complete(cls, parameter_tuples):
+        """Returns those of parameter_tuples for which this Task is complete.
+
+        Override (with an efficient implementation) for efficient scheduling
+        with range tools. Keep the logic consistent with that of complete().
+        """
+        raise NotImplementedError
+
     def output(self):
         """The output that this Task produces.
 
--- a/luigi/tools/range.py
+++ b/luigi/tools/range.py
@@ -130,7 +130,7 @@ class RangeHourlyBase(luigi.WrapperTask):
         self.trigger_event(RangeEvent.COMPLETE_FRACTION, self.of, float(complete_count) / expected_count if expected_count else 1)
 
     def requires(self):
-        # cache because we anticipate lots of tasks
+        # cache because we anticipate a fair amount of computation
         if hasattr(self, '_cached_requires'):
             return self._cached_requires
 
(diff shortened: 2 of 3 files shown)


luigi/task.py
```diff
@@ -484,6 +484,15 @@ def complete(self):
...
         return all(itertools.imap(lambda output: output.exists(), outputs))
 
+    @classmethod
+    def bulk_complete(cls, parameter_tuples):
+        """Returns those of parameter_tuples for which this Task is complete.
+
+        Override (with an efficient implementation) for efficient scheduling
+        with range tools. Keep the logic consistent with that of complete().
+        """
+        raise NotImplementedError
```
erikbern: 1. I'm generally a bit skeptical about adding stuff to the base Task class since it makes the interface more complicated. But I guess in this case it makes sense since it's such a big performance optimization

2. The default behavior could be to return `all([cls(*parameter).complete for parameter in parameter_tuples])` instead of raising a `NotImplementedError`. I would have a mild preference towards doing that.

erikbern: actually nvm about #2 – probably a bad idea to creates lots of Task objects on the fly

luigi/tools/range.py
```diff
@@ -217,101 +217,106 @@ def most_common(items):
...
-
-        # quickly learn everything that's missing
-        missing_datehours = []
-        for d, p in zip(finite_datehours, paths_by_datehour):
-            if not set(p) <= listing:
-                missing_datehours.append(d)
-
-        return missing_datehours
+        try:
+            return set(finite_datehours) - set(task_cls.bulk_complete.__func__(task_cls, finite_datehours))
```
erikbern: what's the `__func__` thing?

Tarrasch:
Good questions from @erikbern, other than that LGTM.

luigi/tools/range.py
ulzha: I thought that calling like task_cls.bulk_complete(finite_datehours) would work, but that causes "TypeError: unbound method bulk_complete() must be called with BulkCompleteTask instance as first argument". I feel quite like lacking a clue on some basic idiom. A classmethod. Why does it. Demand an instance... Well I could make an instance there for an arbitrary datehour, but this was shorter.

luigi/task.py
ulzha: 1. I share that skepticism, inasmuch as I kept all the magic parts (inference from filesystem listings for the common case) out, in range.py. Still this kind of a building block is necessary to have EIBTI and support output to arbitrary fancy systems, all the while enabling performance.

Tarrasch: Yea. Totally agreeing with having EIBTI (explicit is better than implicit).

luigi/tools/range.py
erikbern: OK, let's fix that before merging. Is the issue that classmethods are not possible to overload? Could be something like that

ulzha: Ah it just arose to me - the overload needed @classmethod as well. Done.

Good to squash and merge?

luigi/task.py
erikbern: Glad we're on the same page. The smaller interface the better, but I think in this case the benefits seem to outweigh the added complexity

luigi/tools/range.py
Tarrasch: I think so. @erikbern ?

erikbern:
LGTM

PR #600, opened 2015-01-14
Use MiniCluster for mapreduce jobs as well

erikbern:
HDFS tests have been using the mini cluster for a while. I generalized it a bit and added similar tests for the mapreduce tests.

http://hadoop.apache.org/docs/r2.5.1/hadoop-project-dist/hadoop-common/CLIMiniCluster.html

### DIFF
--- a/luigi/contrib/sqla.py
+++ b/luigi/contrib/sqla.py
@@ -83,7 +83,7 @@ modified example would look as shown below::
 
         def run(self):
             out = self.output().open("w")
-            TASK_LIST = ["item%d\tproperty%d\n" % (i, i) for i in range(10)]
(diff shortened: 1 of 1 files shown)


test/hadoop_test.py
```diff
@@ -158,6 +190,19 @@ def test_unicode_job(self):
         self.assertEqual(c[0], "test\t2\n")
 
 
+class MapreduceLocalTest(unittest.TestCase, MapreduceTestMixin):
```
erikbern: I hate multiple inheritance, but sometimes for unit tests it makes things a lot more convenient

erikbern:
not sure why this closed, will create a new PR

PR #601, opened 2015-01-14
Use the minicluster for Hadoop mapreduce jobs

erikbern:
Seems pretty useful for testing: http://hadoop.apache.org/docs/r2.5.1/hadoop-project-dist/hadoop-common/CLIMiniCluster.html

### DIFF
--- a/test/hadoop_test.py
+++ b/test/hadoop_test.py
@@ -24,13 +24,35 @@ import luigi.mrrunner
 from luigi.mock import MockFile
 import StringIO
 import luigi.notifications
+from nose.plugins.attrib import attr
+import minicluster
+
 luigi.notifications.DEBUG = True
 File = MockFile
 
+luigi.hadoop.attach(minicluster)
 
-class Words(luigi.Task):
+class OutputMixin(luigi.Task):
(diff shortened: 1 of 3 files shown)


erikbern:
Some more notes:
- No functionality in luigi is changed, just in the tests
- I deleted a lot of old tests that mocked out popen etc because the code paths should now be tested
- `unicode_job` is failing on the mini cluster. I disabled it for now. We should figure out what's the issue. Note that we never had that test for this before
- It might be worth to promote some of this code into luigi so that it can be used by people who want to test their Hadoop pipelines

anyman:
Looks great.

Nice with the refactoring breaking out minicluster test. This should be applied to snakebite_test as well.

Tarrasch:
LGTM. I merge since +2 and I might need these changes.

erikbern:
Great!

Why do you need these changes?

Tarrasch:
Thought of making changes to the snakebite client. And maybe will refactor those tests at the same time.

erikbern:
Ok, got it. Maybe we should move the minicluster functionality into its own repo?

erikbern:
Btw I might try to add integration tests for Scalding, depending on how messy it is

PR #602, opened 2015-01-15
Remove hadoop_test.py from gitignore

Tarrasch:
No idea why it was added. My grep-tool ag didn't search in that file
because of it. Hopefully nobody else need to waste time on this.


erikbern:
that's pretty weird

PR #604, opened 2015-01-15
Propagate tox's args to nosetest

Tarrasch:
In particular, this patch allows you to do

    tox test/only_run_this_test.py

By the way, here's a real life example of how I run the tests on Ubuntu
12.04.

    export TOX_ENV=hdp
    export PYTHONPATH=''
    export JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk-amd64
    tox -e $TOX_ENV test/snakebite_test.py

### DIFF
--- a/scripts/ci/run_tests.sh
+++ b/scripts/ci/run_tests.sh
@@ -7,4 +7,4 @@ fi
 
 echo "Using ${HADOOP_DISTRO} distribution of Hadoop from ${HADOOP_HOME}"
 
-python test/runtests.py -v
+python test/runtests.py -v $@
--- a/tox.ini
+++ b/tox.ini
@@ -15,4 +15,4 @@ setenv =
   LUIGI_CONFIG_PATH={toxinidir}/test/testconfig/client.cfg
 commands =
   {toxinidir}/scripts/ci/setup_env.sh
-  {toxinidir}/scripts/ci/run_tests.sh
+  {toxinidir}/scripts/ci/run_tests.sh {posargs:}


Tarrasch:
@anyman @berkerpeksag, you've both been around and touched these parts of the code base. I saw @anyman added this and then @berkerpeksag removed it #562. Was that a mistake or was it intentional?

erikbern:
i don't know how this works but LGTM

tox.ini
```diff
@@ -15,4 +15,4 @@ setenv =
   LUIGI_CONFIG_PATH={toxinidir}/test/testconfig/client.cfg
 commands =
   {toxinidir}/scripts/ci/setup_env.sh
-  {toxinidir}/scripts/ci/run_tests.sh
+  {toxinidir}/scripts/ci/run_tests.sh []
```
berkerpeksag: You can use `{posargs:}` here.

scripts/ci/run_tests.sh
```diff
@@ -7,4 +7,4 @@ fi
 
 echo "Using ${HADOOP_DISTRO} distribution of Hadoop from ${HADOOP_HOME}"
 
-python test/runtests.py -v
+python test/runtests.py -v $@
```
berkerpeksag: I think the `$@` part is not necessary. `nose.main()` should be able to get `sys.argv` if you add `{posargs:}` to `tox.ini`.

berkerpeksag:
> export JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk-amd64
> tox -e $TOX_ENV test/snakebite_test.py

I'm using the same approach for Django development. Here is my `tox.ini` for reference purposes: https://gist.github.com/berkerpeksag/1f644aa3454f99bb4c51

berkerpeksag:
I will test the patch with and without my suggestions this weekend. Sorry for the noise!

Tarrasch:
Thanks for your python and tox expertise @berkerpeksag!

I'll check out what you proposed. Again thanks!

anyman:
@berkerpeksag insights are better than mine. I based my initial commit on the work done in www.github.com/spotify/snakebite but I have no other experience with using tox.

scripts/ci/run_tests.sh
Tarrasch: Nah, I tried this and it doesn't seem to work. Also, I expected it to not work either. How would the arguments be passed if you just execute `python test/runtests.py -v`?

Tarrasch:
I'll merge once travis accepts this.

erikbern:
murgh

scripts/ci/run_tests.sh
berkerpeksag: Yes, you're right. I thought that run_tests.sh is only used by Travis. LGTM. Sorry about the delay.

Tarrasch: No problem at all! I'm really grateful for the `{posargs:}` advice! :)

PR #605, opened 2015-01-16
Add rename2 for snakebite

Tarrasch:
Note that the tests definetly should fail. As `snakebite 2.5.0` is not even released yet.

I wanted to push this out for review still and I checked that the tests I added passed with my local snakebite version with `rename_dont_move()` support.

### DIFF
--- a/test/hdfs_test.py
+++ b/test/hdfs_test.py
@@ -20,6 +20,7 @@ import re
 import functools
 from minicluster import MiniClusterTestCase
 from nose.plugins.attrib import attr
+import helpers
 
 try:
     import unittest2 as unittest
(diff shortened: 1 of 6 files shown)


Tarrasch:
Basically this need to be accepted by the snakebite guys: https://github.com/spotify/snakebite/pull/128

Tarrasch:
-3, Add `rename_dont_move` to the Cascading client or all hell will break loose.

Tarrasch:
So. Now you guys can review this. @erikbern @plinton @ulzha

test/hdfs_test.py
```diff
@@ -32,6 +33,43 @@ class TestException(Exception):
 
 
 @attr('minicluster')
+class ConfigurationTest(MiniClusterTestCase):
+
+    def tezt_rename_dont_move(self, client):
```
erikbern: haha tezt? I have a minor preference for something less confusing like just `def rename_dont_move`

erikbern:
LGTM

test/hdfs_test.py
Tarrasch: Hmm, it's so much like a `test_`, but it's not as I want to run it once for each configuration. Hmm..

PR #607, opened 2015-01-16
Moved luigi.scalding to luigi.contrib.scalding

erikbern:
Also fixed a minor thing with how the 'hadoop' command is run on the command line


Tarrasch:
LGTM.

PR #610, opened 2015-01-18
doc - remove colons from sample decorator usage

hblanks:
Fix two examples in api_overview.rst where decorators are terminated
with a colon.


Tarrasch:
LGTM.

PR #612, opened 2015-01-19
Remove some old/deprecated stuff from interface

erikbern:
- Removed optparse interface (I doubt anyone was using it – we had to use it at Spotify because of legacy reasons)
- Removed the "lock" parameter that's been deprecated for a long time

This is to prepare for a big parameter refactoring I have in another branch. It will be much easier to refactor if I don't have to support optparse. I'm not sure if anyone uses it, but please speak up if you do!

PR #614, opened 2015-01-20
Cleaned up some parameter parsing & dynamic dependencies

erikbern:
Would be great to get your input @tommyengstrom – in particular I'm not sure about the `__main__` stuff you were doing in `load_task`

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -250,34 +250,14 @@ class ArgParseInterface(Interface):
     ''' Takes the task as the command, with parameters specific to it
     '''
     @classmethod
-    def add_parameter(cls, parser, param_name, param, prefix=None):
-        description = []
-        if prefix:
-            description.append('%s.%s' % (prefix, param_name))
-        else:
-            description.append(param_name)
-        if param.description:
-            description.append(param.description)
-        if param.has_value:
-            description.append(" [default: %s]" % (param.value,))
-
-        if param.is_list:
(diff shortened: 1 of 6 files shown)


Tarrasch:
Didn't you just start to remove the optparse stuff in #612? It's confusing that there still are traces of it.

erikbern:
Yeah wasn't super consistent... but this one should be merged first. For the optparse PR I want to give it a few days to see if anyone has any issues. I probably have to rebase the optparse PR later.

erikbern:
I can remove the first commit from this PR if you think it makes more sense? The two commits have very little with each other to do

erikbern:
Added even more tests now. Fairly sure this PR works. Definitely simplifies a lot of code with dynamic dependencies.

In case you haven't noticed, I'm trying to refactor various stuff with `luigi.interface` and `luigi.parameter`. It's all entangled so I have to start in other places.

Tarrasch:
Alright LGTM but porbably give @tommyengstrom a day or so to review. :)

erikbern:
Not sure what's up with Travis, but it looks like it passed

erikbern:
Also does dynamic dependencies not run on other workers? When I tried to run it with multiple workers it doesn't look like it's parallelizing the tasks. I remember talking about this but forgot what we concluded – is there a fundamental reason why you can't put a dynamic dependency on another worker?

PR #618, opened 2015-01-21
Don't have contrib-dependencies

Tarrasch:
Currently, we can't build the luigi debian package. The reason is that
the requirements in setup.py is including dependencies that are for
packages that (at least should) belong to contrib.

I also removed requests as I didn't see it being used anywhere. The test
suite should reveal if it was required or not. :)


erikbern:
LGTM, let's merge

PR #620, opened 2015-01-21
Remove global parameters

erikbern:
This PR makes global parameters obsolete. Instead, there's a mechanism to override any parameter from the command line. By running `--SomeTask-some-parameter 42` you override the default value for `SomeTask`. Previous global parameters still work as before.

### DIFF
--- a/doc/api_overview.rst
+++ b/doc/api_overview.rst
@@ -128,6 +128,33 @@ order to support command line interaction and make sure to convert the
 input to the corresponding type (i.e. datetime.date instead of a
 string).
 
+Setting parameter value for other classes
+-----------------------------------------
+
+All parameters are also exposed on a class level on the command line
+interface. For instance, say you have classes TaskA and TaskB:
+
+.. code:: python
+
+    class TaskA(luigi.Task):
+        x = luigi.Parameter()
+
+    class TaskB(luigi.Task):
+        y = luigi.Parameter()
+
+
+You can run *TaskB* on the command line: *python script.py TaskB --y 42*.
+But you can also set the class value of *TaskA* by running *python script.py
+TaskB --y 42 --TaskA-x 43*. This sets the value of *TaskA.x* to 43 on a
+*class* level. It is still possible to override it inside Python if you
+instantiate *TaskA(x=44)*.
+
+Parameters are resolved in the following order of decreasing priority:
+1. Any value passed to the constructor, or task level value set on the command line
+2. Any class level value set on the command line
+3. Any configuration option (if using the *config_path* argument)
+4. Any default value provided to the parameter
+
 Task.requires
 ^^^^^^^^^^^^^
 
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -227,14 +227,14 @@ def get_task_parameters(task_cls, args):
     # Parse a str->str dict to the correct types
     params = {}
     for param_name, param in task_cls.get_params():
-        param.parse_from_args(param_name, args, params)
+        param.parse_from_args(param_name, task_cls.task_family, args, params)
     return params
 
 
 def set_global_parameters(args):
     # Note that this is not side effect free
     for task_name, param_name, param in Register.get_all_params():
-        param.set_global_from_args(param_name, args)
+        param.set_global_from_args(param_name, task_name, args)
 
 
 class ArgParseInterface(Interface):
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -211,35 +211,30 @@ class ErrorWrappedArgumentParser(argparse.ArgumentParser):
 
 def add_task_parameters(parser, task_cls, optparse=False):
     for param_name, param in task_cls.get_params():
-        if not param.is_global:
-            param.add_to_cmdline_parser(parser, param_name, task_cls.task_family, optparse=optparse)
+        param.add_to_cmdline_parser(parser, param_name, task_cls.task_family, optparse=optparse, glob=False)
 
 
 def add_global_parameters(parser, optparse=False):
     seen_params = set()
     for task_name, param_name, param in Register.get_all_params():
-        if param.is_global:
-            if param in seen_params:
-                continue
-            seen_params.add(param)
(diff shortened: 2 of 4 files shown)


erikbern:
Happy to split it up also. Commit 1-4 were mostly red and simplifies a bunch of stuff. Commit 5 & 6 add a bunch of lines and new functionality

erikbern:
Ok now Travis is complaining... I'll submit commit 1-4 in a separate PR because it's mostly just cleanup

erikbern:
Great if you can let me know if I'm on the right track here

Tarrasch:
Ok. So, you "kinda" can use global variables like this now:

```
class Gobal(Task):
   pool = Parameter(global=True)

class SomeMRTask():
    pool_2 = Global.pool
```

And then you run your program

     luigi --task SomeMRTask --Global-pool production

but not

     luigi --task SomeMRTask --SomeMRTask-pool-2 production   # Won't work

nor

     luigi --task SomeMRTask --pool-2 production   # Won't work

-----

Ok. I think this makes sense, as 


```
class Gobal(Task):
   pool = Parameter(global=True)

class Gobal2(Task):
   pool = Parameter(global=True)
```

Will no longer be an issue as the parameters are namespaced. Or well, I guess you don't even have the `global=True` kwarg anymore. Right? Yea I like this. LGTM.

erikbern:
Yeah you mean `is_global=True` but right, the point is to remove the need for `is_global` by making all parameters global through `--MyTask-my-parameter` as well as "local" through `--parameter`. So there will be no point using `is_global`.

Was thinking maybe it would make sense to transform the task name as well? Eg. `--my-task-my-parameter`. What do you think?

Tarrasch:
No. I'm against that for many reasons:

 1. It's not as greppable
 2. parameter names can collide more easily, so it's less robust.
 3. It'll be less confusing, it's easier to guess that the first part of the param name comes from a Task name if it's capitalized like a task name.

erikbern:
ok sure i won't do it then. Same thing should apply to configuration options btw (I want config options to be picked up automatically based on class & parameter names)

erikbern:
Is this good enough? There's a couple of more things I want to fix, eg. we shouldn't use `is_global` for the environment params now that it's deprecated.

erikbern:
Actually lmk if you have any smart ideas on how to deal with things like `--workers` or `--scheduler-host`. I'm thinking maybe we should allow parameters outside of a `Task` class – this would unify configuration and command line options as well

Tarrasch:
Is there any way to specify the module name as well? I mean, what if you have module `a` and `b` and both have a task in it called `CommonTaskName`?

erikbern:
Then you should use task_namespace to separate them, like how it works for tasks.

But I should definitely make sure the code still works with two classes of the same name and that it doesn't crap out. Let me try.

Tarrasch:
Yea, but how do you specify the task_namespace as a parameter? Something like `--my-module-MyTask-my-parameter myvalue`?

erikbern:
I don't know, how does it work now? I'll make sure to add a test for that as well

Tarrasch:
Such a thing does not exist now. I mean,  you both specify `--task MyTask` and `--module my_module` and then you specify `--my-parameter`. So currently you always specify module and task name. As far as I understand, your PR introduces the possibility to for any Task specify it's parameters from the command line. However, logically you should need to both specify the module *and* task name to uniquely identify a luigi Task. Currently, your PR doesn't support to set parameters for any Task, just tasks that happen to have unique names.

Does this make any sense?

erikbern:
Hm not sure if it's the same thing. The `--module` parameter is used for dynamically loading modules. It doesn't help you if you have multiple tasks with the same name. You still need to use `task_namespace` for that. So it's a different thing

Tarrasch:
Ah, I'm a bit confused then. How is the `task_namespace` decided?. Is it task name and task namespace what identifies a Task uniquely?

erikbern:
Yeah for the purpose of identifying a Task on the command line, it's task namespace and class name

erikbern:
Btw, my latest thinking on merging configuration and global options is that you can define a `Parameter` class on the global scope, i.e. just say

`my_param = luigi.IntParameter()`

And using some Python magic (metaclass, pull out the module of the parameter class, scan the module and see where the class is defined) you can actually get the parameter name as well. This creates a new type of parameter which is taskless. Does that make sense?

Tarrasch:
Wait, I thought parameters where instantiated upon `__init__()` for classes (I never verified this, but I saw this as the only possiblity). You mean that is not the case? I'm not sure I follow your design.

Would 

```

**Comment by Tarrasch** [comment]
Wait, I thought parameters where instantiated upon `__init__()` for classes (I never verified this, but I saw this as the only possiblity). You mean that is not the case? I'm not sure I follow your design.

Would 

```
cat hdfs.py
...
my_param = luigi.IntParameter()
...
```

Correspond to 

```
cat client.cfg
...
[hdfs]
my_param: 123
...
```

Or something?????

**Comment by erikbern** [comment]
You can put parameters in the global scope but they won't be detected
because we scan for all Task classes and find parameters there.

I wasn't yet thinking about how to bind parameters to config options. Right
now I think using config_path is enough. Later let's figure out how to
automate it!

**Comment by erikbern** [comment]
I'm a bit stupid – I don't think it's possible to find out (in Python) where an object is *instantiated*, just where it's *defined*.

Do you have any smart ideas on how to define a taskless parameter? I can only think of stupid way that forces you to repeat the flag many time. Eg.
```python
my_funny_parameter = luigi.Parameter(flag='--my-funny-parameter')
# ...
print my_funny_parameter.value
```

erikbern:
Just to be clear what I want to achieve is to deprecate `is_global` but we're still using it in `luigi.interface` and I'm not sure how to replace it there. Take a look at `EnvironmentParamsContainer`. It's defined as a Task, but it's clearly not a task, it just happens to be a convenient way to parse config and parameters. So what I want to achieve is some similar mechanism that doesn't require `is_global`.

Could of course just add some new object that behaves like a Task but is called something else (Configuration?) but that seems pretty stupid

erikbern:
Closing this for now #665, let's start with that

PR #622, opened 2015-01-21
Set default temp directory in a Windows friendly way.

### LINKED ISSUE #435: hardcoded unix tyle tmp directory cause failure on Windows
The culprit line is: https://github.com/spotify/luigi/blob/master/luigi/interface.py#L92
```
s_global=True, default='/var/tmp/luigi',
```

needs to be something like:
```
is_global=True, default=tempfile.gettempdir()+os.sep+'luigi',
```
after importing `tempfile` of course.

workaround - manually create directory `\var\tmp\luigi`

themalkolm:
Using tempfile.gettempdir() makes sure that a proper temporary
directory will be picked up with respect to os and even env
variables like TMPDIR, TEMP, TMP etc.

Fix #435

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -27,6 +27,7 @@ import re
 import argparse
 import sys
 import os
+import tempfile
 from task import Register
 
 
@@ -79,7 +80,7 @@ class EnvironmentParamsContainer(task.Task):
         is_global=True, default=False,
         description='Ignore if similar process is already running')
     lock_pid_dir = parameter.Parameter(
-        is_global=True, default='/var/tmp/luigi',
+        is_global=True, default=os.path.join(tempfile.gettempdir(), 'luigi'),
         description='Directory to store the pid file')
     workers = parameter.IntParameter(
         is_global=True, default=1,


luigi/interface.py
```diff
@@ -79,7 +80,7 @@ class EnvironmentParamsContainer(task.Task):
         is_global=True, default=False,
         description='Ignore if similar process is already running')
     lock_pid_dir = parameter.Parameter(
-        is_global=True, default='/var/tmp/luigi',
+        is_global=True, default=os.path.join(tempfile.gettempdir(), 'luigi'),
```
Tarrasch: But this should change behavior right? When I ran `tempfile.gettempdir()` on our production machines I just got `/tmp`!

themalkolm: Indeed, it will choose `/tmp` and only than will try `var/tmp` if `/tmp` is not available. Personally I would expect people to explicitly specify `lock_pid_dir` the configuration if they rely on its location. But sure, behavior changes.  Up to you.

Tarrasch: What? Where's the code logic that changes it to `/var/tmp` if `/tmp` is not available.

> Personally I would expect people to explicitly specify lock_pid_dir the configuration if they rely on its location. 

Yes.

Also, I read up a little bit on /tmp vs /var/tmp [1]. And it seems like `/tmp` is actually what we want here.

http://www.google.se/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&ved=0CCIQFjAA&url=http%3A%2F%2Flinuxers.org%2Farticle%2Fdifferences-between-tmp-and-vartmp&ei=HvPAVInWNMviywOUgIH4CQ&usg=AFQjCNE7JtC--WEDeW9R5D6CaJdcmDhdFg&bvm=bv.83829542,d.bGQ

themalkolm: https://docs.python.org/2/library/tempfile.html#tempfile.tempdir

Tarrasch: Aha, it's inside the logic of the `tempfile` library. I see. Thanks!

PR #623, opened 2015-01-21
Default values should not be mutable.

themalkolm:
It is a kind of popular gotcha in python. Just to avoid incidents
it is a good idea to not use mutable objects as default values.

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -99,8 +99,10 @@ class EnvironmentParamsContainer(task.Task):
     )
 
     @classmethod
-    def env_params(cls, override_defaults={}):
+    def env_params(cls, override_defaults=None):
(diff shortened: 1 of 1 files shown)


Tarrasch:
Can you link to some blog post or something explaining why it's a bad python practice to have mutable objects as default values?

themalkolm:
@Tarrasch, http://stackoverflow.com/questions/1132941/least-astonishment-in-python-the-mutable-default-argument

I tried to find a logic where you expect to share `override_default` default instance across invocations but could not find anything like that. So I assumed it is better fix this way. Am I wrong?

Tarrasch:
Cool!

Tarrasch:
No. I think you're doing the right thing! :)

themalkolm:
okidoki!

PR #624, opened 2015-01-21
Explicitly note that examples work with local_scheduler only.

themalkolm:
Update Interface.run as well to be able to specify local_scheduler
via kwargs.

Not sure but when I started learning luigi this documentation page
was super confusing as I constantly was getting some timeouts..

### DIFF
--- a/doc/command_line.rst
+++ b/doc/command_line.rst
@@ -21,9 +21,9 @@ Any task can be instantiated and run from the command line:
 
 You can run this task from the command line like this::
 
-    $ python my_task.py MyTask --x 123 --y 456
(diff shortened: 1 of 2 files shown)


Tarrasch:
This seems good. :)

themalkolm:
ping

erikbern:
I wonder if we should do something with `**kwargs` instead, but this looks good for now

themalkolm:
Yep, if interface.run & interface.build share same kwargs than everything will be much easier.

erikbern:
ok let's refactor interface.py soon. It's a bit messy. I've been thinking about ways to get rid of `EnvironmentParamsContainer` anyway so maybe let's try to do that first

PR #625, opened 2015-01-22
Add date minute parameter

themalkolm:
Surprisingly luigi could be used to run pipelines with minute precision in configuration. This represents an extra parameter we already have for our project. Why not share, right?

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -309,20 +309,23 @@ class DateHourParameter(Parameter):
     19:00.
     """
 
+    date_format = '%Y-%m-%dT%H'  # ISO 8601 is to use 'T'
+
     def parse(self, s):
         """
         Parses a string to a :py:class:`~datetime.datetime` using the format string ``%Y-%m-%dT%H``.
         """
         # TODO(erikbern): we should probably use an internal class for arbitary
         # time intervals (similar to date_interval). Or what do you think?
-        return datetime.datetime.strptime(s, "%Y-%m-%dT%H")  # ISO 8601 is to use 'T'
+        return datetime.datetime.strptime(s, self.date_format)
 
     def serialize(self, dt):
         """
         Converts the datetime to a string usnig the format string ``%Y-%m-%dT%H``.
         """
-        if dt is None: return str(dt)
-        return dt.strftime('%Y-%m-%dT%H')
+        if dt is None:
+            return str(dt)
+        return dt.strftime(self.date_format)
 
 
 class DateParameter(Parameter):
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -328,6 +328,17 @@ class DateHourParameter(Parameter):
         return dt.strftime(self.date_format)
 
 
+class DateMinuteParameter(DateHourParameter):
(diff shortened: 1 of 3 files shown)


erikbern:
Nice!

Tarrasch:
Cool. You've used this in production?

themalkolm:
@Tarrasch, well, we actually use `%Y-%m-%dT%HH%MM` (note an extra M on the end) and `'%Y-%m-%dT%HH%MM%SS`.

luigi/parameter.py
```diff
@@ -309,20 +309,34 @@ class DateHourParameter(Parameter):
...
+        if dt is None:
+            return str(dt)
+        return dt.strftime(self.date_format)
+
+
+class DateMinuteParameter(DateHourParameter):
+    """Parameter whose value is a :py:class:`~datetime.datetime` specified to the minute.
+
+    A DateMinuteParameter is a `ISO 8601 <http://en.wikipedia.org/wiki/ISO_8601>`_ formatted
+    date and time specified to the minute. For example, ``2013-07-10T19H42`` specifies July 10, 2013 at
```
Tarrasch: Please include also (or only) an example with a leading 0 minute. like 19:07 rather than 19:42. How will it be formatted too?

Tarrasch:
Ah wait a minute then, is it ever ever ever a good idea to not include the 0 in the minute? will files look like

``
8
9
10
11
12
```

or 

``
08
09
10
11
12
```

??

Tarrasch:
> Ah wait a minute then

Pun not intended :p

luigi/parameter.py
themalkolm: To give a hint that minutes will always have a leading zero if below 10?

themalkolm:
@Tarrasch 

https://docs.python.org/2/library/datetime.html
> %M	Minute as a zero-padded decimal number.	00, 01, ..., 59

themalkolm:
So minutes are always zero-padded.

erikbern:
Merging this, we can always add more tests later

Tarrasch:
> So minutes are always zero-padded.

Awesome then! :)

fphilipe:
Quick comment about the format. I think it is incorrect. I've never seen `H` being used as a separator between the hours and the minutes. I believe this should be a colon instead. Please correct me if I'm wrong.

erikbern:
Yes – you are right. ISO 8601 specifies T between date and time, but between hour and minute it should be colon. `2015-07-28T19:19:34Z` etc

Can we fix this? Sorry I glanced at this PR too quickly

fphilipe:
I suppose the incorrect format would need to be deprecated and supported for a while. I'd be happy to work on a PR.

erikbern:
Another problem is also that colons generally don't work in filenames

fphilipe:
@erikbern ISO 8601 states that times don't necessarily need colons, they can simply be omitted. https://en.wikipedia.org/wiki/ISO_8601#Times

erikbern:
Perfect then let's use 2015-07-29T1356

PR #626, opened 2015-01-22
Rename BooleanParameter to BoolParameter.

themalkolm:
It is confusing indeed. Lets fix this nicely put TODO but keep BooleanParameter as an alias for backward compatibility.

Following this logic we should rename kwarg `is_boolean` to `is_bool` in `luigi.Parameter.__init__ `as well. But to keep backward compatibility there would require us to keep two similar kwargs `is_bool` & `is_boolean` in `__init__` for a while. It is confusing so I skipped this.


erikbern:
LGTM I like it. Yeah makes sense to skip the `is_bool` thing

Tarrasch:
You can use `deprecate_kw_arg` for `is_boolean` --> `is_bool` maybe?

https://github.com/spotify/luigi/blob/4cecc7fd6e4d6fb60f6c685ef5d26690f8cb851b/luigi/util.py#L283

themalkolm:
@Tarrasch, should I? Is the desired approach?

Tarrasch:
Well, it does exactly what you want in this case right? I'll let you make the final call but at least try it out I think. Would be bad if you couldn't specify `is_bool` if you rename the class from `BlahBoolean` to `BlahBool`.

themalkolm:
@Tarrasch will give a try.

erikbern:
But let's merge this first  or what do you think @Tarrasch ?

luigi/__init__.py
```diff
@@ -43,7 +43,8 @@
 DateParameter = parameter.DateParameter
 IntParameter = parameter.IntParameter
 FloatParameter = parameter.FloatParameter
-BooleanParameter = parameter.BooleanParameter
+BooleanParameter = parameter.BoolParameter  # backward compatibility
```
Tarrasch: I think it should be `BooleanParameter = parameter.BooleanParameter`

Tarrasch: (see below why)

luigi/parameter.py
```diff
@@ -348,21 +348,18 @@ def parse(self, s):
...
         specifies ``is_boolean=True``.
         """
-        super(BooleanParameter, self).__init__(*args, is_boolean=True, **kwargs)
+        super(BoolParameter, self).__init__(*args, is_boolean=True, **kwargs)
 
     def parse(self, s):
         """Parses a ``boolean`` from the string, matching 'true' or 'false' ignoring case."""
         return {'true': True, 'false': False}[str(s).lower()]
-
+BooleanParameter = BoolParameter
```
Tarrasch: (from above)...

So here I would like to see that you inherit from `BoolParameter` and include a `warning.warning("deprecated blah")` in the `__init__()`. That way more peole are likely to start using the new version if they get spammed with deprecation warnings.

Tarrasch:
I just added a few more comments which I think we should take a look at before merging. Sorry for delaying this. :)

luigi/parameter.py
erikbern: Yeah I think that's a great point. We should do that before we merge this commit. Then we can probably drop `BooleanParameter` sometimes in 6-9 months :)

erikbern:
LGTM

Tarrasch:
LGTM

PR #627, opened 2015-01-22
Parameter cleanup

erikbern:
First half of #620 – this is just cleanup/refactoring that doesn't change any functionality and removes more lines than it adds.


Tarrasch:
LGTM (at a glance)

Tarrasch:
This is what caused #738 (just wanted to create the clickable hyperlink here :))

PR #628, opened 2015-01-22
added chicken param to remove methods, consistent apis

coderfi:
I added the chicken parameter to the HdfsClient remove methods.
By default, this is True, which tells the remove method to raise an exception if the requested remove path is dangerous.
Such paths include the root and single level root paths, i.e. '/' or '/etc'.
Hopefully this prevents embarrassing 'hadoop fs -rmr /'  attempts. :)
In addition, the HdfsClientCdh3 mkdir and remove method was api incompatible with the base class's methods, I went ahead and fixed those.


luigi/hdfs.py
```diff
@@ -105,6 +105,51 @@ def list_path(path):
...
+    >>> is_dangerous_rm_path('/opt/foo/bar')
+    False
+    >>> is_dangerous_rm_path('my_relative')
+    False
+    """
+    if not path:
+        return True
+
+    if '//' in path:
+        raise ValueError('Funny path, double slash: %s' % path)
```
Tarrasch: Why do you care about this at all?

Tarrasch:
You should amend your commits. Other than that this looks good I think.

luigi/hdfs.py
```diff
@@ -105,6 +105,51 @@ def list_path(path):
...
+    ...     pass
+    >>> is_dangerous_rm_path('/opt/')
+    True
+    >>> is_dangerous_rm_path('/opt/foo/bar')
+    False
+    >>> is_dangerous_rm_path('my_relative')
+    False
+    """
+    if not path:
+        return True
```
Tarrasch: What happend if we don't have this? I mean, can't we assume that the user passes in a string? Can something bad happen if we don't guard against this?

luigi/hdfs.py
```diff
@@ -105,6 +105,51 @@ def list_path(path):
...
+    if '//' in path:
+        raise ValueError('Funny path, double slash: %s' % path)
+
+    path = path.strip()
+
+    if path.endswith('/'):
+        path = path[:-1]
+
+    if not path:
+        return True
```
Tarrasch: Again?

Tarrasch: I guess this somehow checks that path isn't the empty string, but can't you guard only against that? The code is kind of unreadable like this I think.

luigi/hdfs.py
```diff
@@ -105,6 +105,51 @@ def list_path(path):
...
+
+    if '//' in path:
+        raise ValueError('Funny path, double slash: %s' % path)
+
+    path = path.strip()
+
+    if path.endswith('/'):
+        path = path[:-1]
```
Tarrasch: Can't you pass `/` as a parameter to strip here?

luigi/hdfs.py
coderfi: You mean the '//' check? Somewhat a kludge to simplify the following code.
Also, since ultimately this path string might be passed to a hadoop fs -rmr process, I would definitely be concerned at the behavior of:
   hadoop fs -rmr //

Not sure if the hadoop CLI would try to delete the root directory, or exit 1.

luigi/hdfs.py
coderfi: The method itself doesn't know that. 

The original impetus of this whole check is that, is to help prevent developers writing Luigi tasks from shooting themselves in the foot and giving some sysadmin guy a very long night restoring from backups (if they exist!).

I've seen (and also caught myself), that through whatever obscure luigi command line mechanism /  programmatic calls, the ultimate path passed in to the remove method can be basically anything -- and when such luigi task is invoked by a cron job or some other automatic process, the anything could be doubly anything. :)

The path could actually be None or a blank string, depending on the calling method and how that method got its parameters (through the luigi.Parameter via some shell command line argument, and whether said parameter had default=None, etc.).

In the case something bad happens... presumedly the caller (remove) method asks if path=None or path='' is dangerous.

If it is not dangerous, the invoked CLI process might look like:
    1) hadoop fs -rmr None   (if the caller did naive string concatenation) 
    or
    2) hadoop fs -rmr 

For #1, best case is we get a file not found error from hadoop cli. Worst case is we deleted someone's file called 'None' :(
For #2, best case is hadoop cli complains that a required command line argument is missing. Worst case is some hadoop distribution (or maybe someone created a hadoop wrapper), assumed something else and inadvertently deletes something.

luigi/hdfs.py
coderfi: Yeah I agree... I couldn't think of a cleaner way, but then again, I didn't think too hard. :) Open to suggestions.
Anyway, I wrote a few doctests that helps validate potential pitfalls (particularly around paths with whitespace). Not sure if your testing system is setup to run doctests, would be happy to convert to unit tests.

luigi/hdfs.py
coderfi: Hmm, I actually never realized strip can take a parameter. Learn something new everyday, and something so basic!
strip('/' would probably then strip the leading '/', thereby making the path appear relative, and by the intention of the function, we allow users to delete relative files with impunity.
I would have to check to see if the path appears relative, but leading spaces may interfere with that check.

coderfi:
Responded to the various comments from @Tarrasch , thanks for the feedback!
One thing I did not think to check is if the path is a full hdfs path, i.e. hdfs:/// or hdfs:// hdfs://master/
All three cases could potentially trigger a disastrous consequence.

I think this is worth looking into, but maybe add checks for that sort of thing in a separate commit?

luigi/hdfs.py
Tarrasch: Yea, your worst case for 2) is *completly* unlikely. You mean default would be to delete `/`? And your worst case for 1) is not that bad. Besides. case 1) will not happen. What will happen is that you get an exception thrown from `path.strip()`.

I don't like this line as it's logic that gives no value, and I get confused reading it since I expect lines of code to be useful. :)

luigi/hdfs.py
Tarrasch: `if path == ''` Is narrower I think.

luigi/hdfs.py
Tarrasch: Good catch! Here's anothing thing for you too learn. [`rstrip`](https://docs.python.org/2/library/stdtypes.html#str.rstrip) :)

Tarrasch:
Yea, wait with the `hdfs:///` stuff please. Meanwhile read through my comments and let me know what you think. :)

luigi/hdfs.py
Tarrasch: I don't think this code is neccessary but it can stay if you're worried about cases like `//`. The code is at least easy to digest here.

coderfi:
Code updated per suggestions, hopefully it is more too your liking :)
I noticed conflicts since my branch was too far behind with your master, so I merged those in as well.

Tarrasch:
The commits are messed up now. The "updates to is_dangerous_rm_path(): " commit is shown twice. Can you properly rebase your commits and force push? :)

coderfi:
Ugh, you might have to help me on that.
I'm an old school svn guy, and it was miraculous for me to even get the
branches working -_-

How's the code aside from the source control mess?

erikbern:
try `git rebase master`

Then you can also try `git rebase -i HEAD~9` to squash the last commits (it will let you cherry pick which ones)

coderfi:
Bah, I think I really messed things up on my branch, will try one more time on a new pull.

PR #630, opened 2015-01-22
Add DateMinuteParameter support to previous()

themalkolm:
Note that we need to check for DateMinuteParameter before we check
for DateHourParameter as they are part of the same hierarchy.


erikbern:
You're on a roll, @themalkolm! Already the 16th most prolific contributor :)

themalkolm:
oatmeal

Tarrasch:
Good job @themalkolm ! :)

PR #631, opened 2015-01-22
Do not open atomic_file for mkdirs result.

themalkolm:
Looks like we open file here only for a side effect that we mkdir
all folders. It is quite silly, lets just mkdir without open/close
invocations.

Lets see what Travis thinks :D

### DIFF
--- a/luigi/file.py
+++ b/luigi/file.py
@@ -81,14 +81,16 @@ class File(FileSystemTarget):
         self.format = format
         self.is_tmp = is_tmp
 
+    def makedirs(self):
+        """Create all parent folders if they do not exist."""
+        normpath = os.path.normpath(self.path)
+        parentfolder = os.path.dirname(normpath)
+        if parentfolder and not os.path.exists(parentfolder):
+            os.makedirs(parentfolder)
+
     def open(self, mode='r'):
         if mode == 'w':
-            # Create folder if it does not exist
-            normpath = os.path.normpath(self.path)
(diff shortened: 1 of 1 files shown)


erikbern:
LGTM assuming it works

luigi/file.py
```diff
@@ -81,14 +81,16 @@ def __init__(self, path=None, format=None, is_tmp=False):
         self.format = format
         self.is_tmp = is_tmp
 
+    def mkdirs(self):
```
Tarrasch: Really short method name. Make it describe what you do. Like `create_parent_directory` or whatever you think suits.

themalkolm: I think mkdirs is enough and it is on the same level of understanding as open,close,copy,move. Surely could rename to `makedirs` mimicking the actual os call `os.makedirs` if it helps.

Tarrasch: I didn't know `os.makedirs` was a thing. But since it it? Why don't you call it `makedirs` then?

themalkolm: Up to you. I talk sed and awk fluently so mkdirs is a least terrible thing I do in my work :D

themalkolm:
TODO: s/mkdirs/makedirs/g

luigi/file.py
themalkolm: Yeah, `mkdirs` came as a child of `mkdir(linux)` & `os.makedirs` :)

themalkolm:
ping

PR #632, opened 2015-01-22
Add LocalGzipTarget.

themalkolm:
It is more Windows friendly as does not require you to have gunzip installed on the target machine.

Sure, I understand that this duplicates luigi.format.Gzip logic. But this works better for us as doesn't require to install gunzip everywhere. Not sure if you want it, though.

So, what do you think?

### DIFF
--- a/luigi/__init__.py
+++ b/luigi/__init__.py
@@ -31,6 +31,7 @@ Target = target.Target
 
 File = file.File  # TODO: remove, should be LocalTarget
 LocalTarget = File
+LocalGzipTarget = file.GzipFile
 Parameter = parameter.Parameter
 RemoteScheduler = rpc.RemoteScheduler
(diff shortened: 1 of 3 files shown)


erikbern:
LGTM but don't we already have a format argument that lets you do gzip?

themalkolm:
@erikbern 

>Sure, I understand that this duplicates luigi.format.Gzip logic. But this works better for us as doesn't require to install gunzip everywhere. Not sure if you want it, though.

Tarrasch:
I don't know. Can you somehow squeeze this into `luigi.contrib`? It's not core functionality and really all non-core stuff should go there, even `luigi.hdfs` I think.

erikbern:
Yeah good point, makes sense.

themalkolm:
Nah, maybe in next life.

PR #634, opened 2015-01-22
Include host in error notifications

jackdwyer:
I've added a slight modification to the subject line of error emails to include the host the worker was running on.
I am just including the already defined self.host

PR #636, opened 2015-01-23
Remove tests that are not run on Travis CI

Tarrasch:
Since they're not run there, they are not useful for anything I think.
They just make the `hdfs_test.py` file harder to navigate and wastes
developer time. Plus they are really old.

Also the `main():` part is not superseded by whatever things we have, in
form of tox and stuff. For instance you can run

    export TOX_ENV=hdp; export PYTHONPATH='';
    JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk-amd64 tox -e $TOX_ENV
    test/hdfs_test.py:AtomicHdfsOutputPipeTests.test_atomicity

To run a single test.


erikbern:
What's the reason they aren't being run?

Btw we should try to add test coverage reports, would be good to watch if test coverage goes down with things like this

Tarrasch:
It takes 4 minutes to run on my laptop. Not sure if that's why they were disabled.

erikbern:
Something weird was introduced in #634 – I'll merge #638 (assuming it builds) and then you can rebase?

Tarrasch:
Hm, only one test was failing and I can fix that. Then we can keep the tests. Not sure if it's good to have these slow tests running, but I guess they can one day detect an actual bug. Should I just enable the tests?

erikbern:
If the tests are covering code already covered by other tests, no need to keep them.

If they cover unique cases then let's keep them

erikbern:
Btw there's now coverage reports automatically generated: https://coveralls.io/jobs/4150321

I think next time you update the PR it will be triggered automatically so we can see if coverage goes down

Tarrasch:
It won't go down as those tests are run (I hypothesize). However, I think I can enable the tests actually. I realize they are can be helpful in fact as they test something I don't think other tests are testing.

Tarrasch:
I close this in favor of #663

PR #638, opened 2015-01-23
Fixed bug introduced in #634

erikbern:
TaskProcess has no attribute host, not sure why nosetests ran


erikbern:
This one resolves some of the current failing tests, but not all. Not sure what happened. We merged a bunch of PR's at the same time so maybe something weird happened

PR #642, opened 2015-01-23
Do not initialize _task_history twice

themalkolm:
This code looks strange. Could it be a case of a merge failure?


erikbern:
Yeah agree. Thanks!

Tarrasch:
Nice catch!

Issue #643, opened 2015-01-24
Update SQLA module


gouthambs:
@erikbern @Tarrasch 
I have been testing the sqla module to incorporate sqlalchemy based task. I have identified some improvements that bring enhancements and fixes, namely:

- Insertions in marker table is NULL in MySQL
- Enhance sqla.CopyToTable to allow table updates apart from insertions

This is a placeholder to bring in these improvements.

PR #644, opened 2015-01-24
Sqlabugs

gouthambs:
Addresses issue #643


luigi/contrib/sqla.py
```diff
@@ -130,9 +130,25 @@ def requires(self):
 import datetime
 import itertools
 import sqlalchemy
-
+from threading import Lock
 logger = logging.getLogger('luigi-interface')
 
+_engine_hash={}
+_mutex = Lock()
+def _get_engine(connection_string, echo):
```
erikbern: would prefer to put this in a classmethod on SQLAlchemyTarget and use class variables instead of global variables

Tarrasch:
LGTM at a glance. But avoid globals as @erikbern  said :)

erikbern:
LGTM except the global variable. Is it easy to fix?

gouthambs:
@erikbern  I will move the global variables out. That's a good suggestion. 

Let me give you some background around the changes I made. I want to get your advice on making the right fix for the problem I observed. 

I saw that even though I had only 2 workers, all the tasks were getting created simultaneously. So if I had, say,  100 sqla.CopyToTable jobs, then that resulted in 100 objects creations and hence 100 connections all at once. After my modifications I was hoping that the two processes would hold the connection throughout the pipeline. But I noticed that the workers keep changing (based on the PID). Is that how it is supposed to work?

Overall I want to modify the create_engine part so that there aren't too many connections being made to the database. Do you have any suggestions for me .

Tarrasch:
That should not be the case I believe. Luigi should start one task per worker and wait for them to finish. Maybe you job finshes without closing the database connection? You maybe (my mistake) do something asynchronous in your `run()`?

erikbern:
If you only have two processes then there shouldn't be more than two simultaneous connections.

Note that Luigi might do some forking under the hood (at least between scheduling and execution) so you might end up with a broken connection in some cases. But I'm guessing sqlalchemy will handle the failure and reconnect

gouthambs:
Let me try to create a small script to replicate this issue. Let us not merge this thing yet.

gouthambs:
@erikbern I have fixed some of the other issues I encountered. I have also removed the globals. Take a look!

gouthambs:
@erikbern I am getting some errors after removing the globals. Let me see if I can fix.

erikbern:
Would be good if you can run all your tests twice: one with `n_workers=1` and one with `n_workers>1` just to make sure it works with more than one process

gouthambs:
@erikbern  Where should I use the `n_workers`? Should it go part of `luigi.build()` in the tests?

erikbern:
sure, sounds good!

gouthambs:
@erikbern I have added tests for the different num workers.

erikbern:
Looks like you're failing pep8 checks, which should be easy to fix!

erikbern:
Still a pep8 issue. You can try it locally using `pep8 --ignore E501 luigi test examples bin`.

luigi/contrib/sqla.py
```diff
@@ -152,8 +150,21 @@ def __init__(self, connection_string, target_table, update_id, echo=False):
         """
         self.target_table = target_table
         self.update_id = update_id
-        self.engine = sqlalchemy.create_engine(connection_string, echo=echo)
+        self.connection_string = connection_string
+        self.echo = echo
         self.marker_table_bound = None
+        self._engine = None
+        self._mutex = Lock()
```
erikbern: Shouldn't this mutex be on a class level rather than object level?

test/contrib/sqla_test.py
```diff
@@ -252,10 +257,66 @@ class ModSQLATask(sqla.CopyToTable):
...
+        # Running only task1, and task2 should fail
+        task1, task2, task3 = ModBaseTask(), ModSQLATask(), UpdateSQLATask()
+        luigi.build([task1, task2, task3], local_scheduler=True, n_workers=self.NUM_WORKERS)
         self._check_entries(self.engine)
+
+
+class TestSQLA2(TestSQLA):
+    """ 2 workers version
+    """
+    NUM_WORKERS = 2
```
erikbern: Great – this is exactly what I meant.

erikbern: You might want to add a test that runs many tasks too. But it's optional (sorry realize this PR has been going on for a long time already)

luigi/contrib/sqla.py
erikbern: I guess it depends on where the race condition is.

erikbern: Actually it's probably not ideal to create on a class level since then it won't work with more than one process.

If you think it's important to re-use the engine you should probably store it in a dict based on `os.getpid()` or something like that.

erikbern: Also, now that I think about it: Luigi doesn't use threads, so I'm not sure if you really need a Mutex.

gouthambs:
@erikbern I like the `os.getpid()` idea. How do I get the dict shared across processors? 

I will fix the pep issues. I know the PR has been long, but lets get it right! I like that the functionality has improved substantially through this PR.

gouthambs:
@erikbern I tried sharing engines using the `multiprocessing.Manager().dict()`. I get `expected string or Unicode object, NoneType found` exception. This I think has to do with the fact that the `sqlalchemy` engines may not be pickleable.

erikbern:
You shouldn't share engines, across processes, you want to make sure they are different.

Note that multiprocessing forks processes so it might be enough to just set `SQLAlchemyTarget._engine = None` and then create it in `def engine`. That way you don't have to mess around with multiprocessing or pids.

gouthambs:
I have this working on my end. Let me fix the conflicts and do some more testing. 

BTW, is there a good place to put a link to a blog post with a tutorial on using the SQLA module?

gouthambs:
@erikbern How do I find out if any tasks failed after I do `luigi.build([]) ` in tests?

erikbern:
I think luigi.build will raise an exception if any task fails?

gouthambs:
I think luigi.build doesn't throw an exception. I had test cases that were marked as passing. I dug into the logs to discover there were failures.

erikbern:
Yeah looks like it returns whether all tasks succeeded.

We could add a flag to let it raise if something fails?

Otherwise you can use luigi.run which should raise if it fails

gouthambs:
@erikbern Thanks for the info.

I have been having issues running the current version of the code in my production environment with MySQL database. I think I have a hunch on the issue I am facing with the current code. I think that when I use static class variables as a means to `create_engine`, it is possible that when it forks, every process doesn't get a connection of its own. However when I used the module level global variable set to None and then checking that to create an engine, things seem to work fine. 

I am not sure if there is a work around. I am not a fan of `globals` but I don't know how to avoid it.

gouthambs:
@erikbern Can you take a look at the latest commit? This removes all the `globals`. 

It is my understanding that the workers are forked after the initial engine creation. So now I check for the `pid` of the engine with that of the process running the job.

luigi/contrib/sqla.py
```diff
@@ -153,22 +155,37 @@ class SQLAlchemyTarget(luigi.Target):
...
         :return:
         """
         self.target_table = target_table
         self.update_id = update_id
-        self.engine = sqlalchemy.create_engine(connection_string, echo=echo)
+        self.connection_string = connection_string
+        self.echo = echo
         self.marker_table_bound = None
 
+    @property
```
gouthambs: The engine instance is checked to make sure it belongs to the same process as that of the worker process.

luigi/contrib/sqla.py
```diff
@@ -153,22 +155,37 @@ class SQLAlchemyTarget(luigi.Target):
...
+    def engine(self):
```
gouthambs: The engine instance is checked to make sure it belongs to the same process as that of the worker process.

erikbern:
This looks great now, thanks!

Do you mind just squashing a bunch of your commits? There are 29 at the moment. I can merge when you have rebased!

Tarrasch:
See #742

PR #646, opened 2015-01-24
Use iteritems(), it is faster as does not create a list.

themalkolm:
Just random findings while reading the code.


luigi/hdfs.py
```diff
@@ -288,7 +288,7 @@ def get_bite(self):
             client_kwargs = dict(filter(lambda (k, v): v is not None and v != '',  {
                 'hadoop_version': self.config.getint("hdfs", "client_version", None),
                 'effective_user': self.config.get("hdfs", "effective_user", None)
-            }.items()))
+            }.iteritems()))
```
erikbern: doubt this matters...

themalkolm: It doesn't. But hey, I'm free to contribute whatever I want, right? It is always up to you to accept or reject it. No hard feelings from me ever if you reject something. Deal?

erikbern: tumblr_masgq6hvlo1qk1j3lo1_500

PR #647, opened 2015-01-24
Add missing call the the parent.

themalkolm:
Random finding when reading the code.


Tarrasch:
It would be nice so say in the commit if anything changes with this patch. For the next patch. :)

PR #648, opened 2015-01-25
Add coveralls badge to README

themalkolm:
* Add coveralls badge to README
* Shielded both badges via http://shields.io to keep the same style

This gives and instant hint to a user where to find the coverage information.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,8 +6,11 @@
 About Luigi
 -----------
 
-.. image:: https://travis-ci.org/spotify/luigi.svg?branch=master
+.. image:: https://img.shields.io/travis/spotify/luigi/master.svg?style=flat
     :target: https://travis-ci.org/spotify/luigi
+
+.. image:: https://img.shields.io/coveralls/spotify/luigi/master.svg?style=flat
+    :target: https://coveralls.io/r/spotify/luigi?branch=master
 
 Luigi is a Python package that helps you build complex pipelines of
 batch jobs. It handles dependency resolution, workflow management,


erikbern:
Cool. Unfortunately it's a bit embarrassing how low the coverage is right now :(

Maybe we should exclude luigi.contrib from the coverage? Or is this to game the system? :)

erikbern:
Hm it looks like my coveralls commit has fucked up Travis – it's now running all tests multiple times. Let me fix

erikbern:
Fixed it in https://github.com/spotify/luigi/commit/9ed55f57ff99e7dec6067688d855a7de2d2aa1da – too lazy to send PR because it's an obvious bug

themalkolm:
77% is low? Ha, ha. It is a quite a number especially if you take in account that some tests are simply disabled e.g. for ssh.

themalkolm:
@erikbern Btw, as you are playing with coveralls you probably should check something like landscape https://landscape.io/github/themalkolm/luigi. Instantly gives some feeling about overall project's "health".

erikbern:
Cool is landscape for static analysis and style checking? 

I set it up: https://landscape.io/github/spotify/luigi

erikbern:
Yeah I guess 77% is acceptable (and there's 415 tests) but it would be great to get it up a bit more. Also some code is run through subprocesses (`mrrunner.py` etc) so it's slightly higher in reality. But we should aim to get it up to 85-90% and I think it's pretty easy

PR #653, opened 2015-01-25
Add pep8 and autopep8 targets to tox

themalkolm:
Add handy tox targets to check against simplest pep8 check and automatically
correct them using autopep8.

Note that both pep8 and autopep8 could be run in more aggressive modes but
for the first time I leave them in the 'easy' mode.

```
$ tox -e pep8 | wc -l
425
$ tox -e autopep8
$ tox -e pep8 | wc -l
45
```

I leave the exercise to run autopep8 and observe changes for core developers.
The best part is you make one step towards a repo with a single code style.
One style - easier to read the code, right?

How cool is that? No more questions about code style ever - you just conform
to pep8.


Tarrasch:
Awesome! LGTM!

erikbern:
How does this work? When I submit a violating line of code, where does it complain?

themalkolm:
No, this won't complain anywhere. This is a standalone targets for tox. So you could play with it for a while.

Later, when repo is pep8 conformant to the desired level you can enable 'tox -e pep8' check in travis to make sure that no pull request is merged that violates pep8. I can help with this later as well.

Is it sensible?

Tarrasch:
It doesn't interact with github at all afaict. It's just locally. You can run tox with `-e pep8` now to get a count on the violations.

Tarrasch:
Damn. You were one second ahead of me! @themalkolm :)

erikbern:
Ok and for now the output of tox will be available if you go to Travis and look at the build logs?

themalkolm:
@erikbern No, pep8 and autopep8 are not listed in `envlist` so they are not run by default.

> No, this won't complain anywhere. This is a standalone targets for tox. So you could play with it for a while.

themalkolm:
I can fix everything, just think if you want to go this way to force code style for every pull request.

Tarrasch:
Yea as this only affects local devlopment I think this makes sense and can be merged. But isn't there a way to get the actual pep8 violations instead of just the count?

themalkolm:
@Tarrasch Sure, note that I *count* them explicitly. Do not count and you get all violations.

> tox -e pep8 | wc -l

Tarrasch:
Ah I'm such an idiot. Look very good then. I'll let @erikbern pull the trigger.

themalkolm:
@Tarrasch nah, here is the updated version. Here we add target, fix all issues and ask travis to always check for pep8 style. Should work.

autopep8 could be either run manually or we can make a git hook to always check for style on commit for changed files.

themalkolm:
I encourage you to read the whole diff.

Tarrasch:
Wow. interesting. I think this might be worth merging. I can live with the merge failures I suppose. :)

I did a quick scroll. Looks good.

themalkolm:
Travis approves!

like-a-boss

themalkolm:
coveralls commenting is annoying though

PR #654, opened 2015-01-25
Remove unused import

themalkolm:
warnings are not used in this module


themalkolm:
https://landscape.io/github/spotify/luigi/16/messages/272

Tarrasch:
Sweet! This is almost a milestone! It shows we don't emit any more damn deprecation warnings since we've removed everything deprecated from `interface.py`.

PR #655, opened 2015-01-25
Do not import postgres twice

themalkolm:
Looks like we import postgres twice while only the second import is used.

https://landscape.io/github/spotify/luigi/16/modules/luigi/contrib/redshift.py#L7


Tarrasch:
I would rather you remove the other import and change the code to use absolute paths. But this is OK too.

themalkolm:
I went for the smallest diff. I agree that absolute import would be better.

PR #656, opened 2015-01-25
Remove unnecessary pass statement

themalkolm:
There is not need for pass here as statement is not empty otherwise.

https://landscape.io/github/spotify/luigi/16/messages/230


Tarrasch:
LGTM

PR #659, opened 2015-01-25
Removed workaround for deprecated get_work RPC call

erikbern:
Deprecated in 2013, I think people have had enough time by now :)


erikbern:
Hm should we really fail builds on pep8 issues?

Tarrasch:
I think we should. It's annoying to get all the issues from other people being shown in my editor each time I save a file.

PR #660, opened 2015-01-25
Improve most common memory footprint

themalkolm:
No need neither to create a list in items() nor to sort the whole
sequence. We need only the max one, right?

If we have two values with the same frequence than we still have
undefined result as before as dict.items() & dict.iteritems()
return values in an undefined order.

Yeah, I just read random code in luigi and see if I can improve
it.


erikbern:
LGTM

Tarrasch:
LGTM, any opinion @ulzha  ?

PR #661, opened 2015-01-26
Rename worker parameter to worker_id

themalkolm:
It is confusing that we use worker & worker_id interchangeable in the code.
Lets be explicit where it is possible (same as with task_id).


themalkolm:
Right, people started using this keyword.

Tarrasch:
Wait. Why did you close?

themalkolm:
@Tarrasch because Travis told me that 'worker' is used as keyword argument in tests. So I guess it is already too late to change to a proper name. Other users must be doing the same.

themalkolm:
It is annoying that after closing an issue we no longer see build status from Travis.

PR #641, opened 2015-01-23
Trying coveralls

erikbern:

PR #664, opened 2015-01-26
Restructure tox.ini and test environments

Tarrasch:
This patch fixes some fundamental flaws:

 * We managed to break the support of running individual tests (again)
   in spotify/luigi#641. This patch fixes this.
 * I don't wanna download a hadoop distribution if I run a non-hdfs
   tests. This patch fixes this.
 * We're running the non-hdfs tests twice, once for each hadoop
   distribution.  I solve this by introducing the nonhdfs environment,
   as a bonus, the tests will run faster.
 * We simplify the CI stuff by removing verify.sh


Tarrasch:
By the way. We should make so that `doc/contributing.rst` only contains how to run the tests, all the other information in there is very outdated and makes luigi look inactive IMO.

Tarrasch:
Sweet! It works. Will merge on +1

erikbern:
LGTM but I guess we should fix the pep8 thing

themalkolm:
It is nice to be able all unit tests w/o hadoop big guns..

erikbern:
Did you see my note on coveralls? I think it gets confused because the coverage reports for the hadoop tests and the non-hadoop tests are being uploaded separately and it doesn't know which one to choose

PR #665, opened 2015-01-26
Expose class parameters on command line

erikbern:
Instead of #620, let's limit the scope a bit and start with this thing first


Tarrasch:
LGTM

erikbern:
I think coveralls is confused by the latest tox refactoring :(

PR #666, opened 2015-01-26
Return 0 even when pep8 fails

Tarrasch:
Currently, all our travis builds are failing, because we don't strictly
comply with pep8. With this patch, we do see what the failure is but we
still get it green.


erikbern:
LGTM

PR #667, opened 2015-01-26
Adds event triggering for various streaming mapreduce init events

freider:


test/hadoop_test.py
```diff
@@ -319,7 +358,3 @@ def test_create_packages_archive_package_subpackage_submodule(self, tar):
         package_subpackage_submodule = __import__("package.subpackage.submodule", None, None, 'dummy')
         luigi.hadoop.create_packages_archive([package_subpackage_submodule], '/dev/null')
         self._assert_package_subpackage(tar.return_value.add)
-
-
-if __name__ == '__main__':
-    HadoopJobTest.test_run_real()
```
Tarrasch: Nice

erikbern:
LGTM

Tarrasch:
I would like to know why you do this. Out of curiosity. LGTM.

freider:
It's nice to have events to hook into for third party modules to be able to interact or extract info from the hadoop running environment when it isn't feasible to create a new base class for people to use.
In this specific case, I was mostly looking for a way for us to change a runtime pyschema configuration setting before a job starts in hadoop, without having that code going into a derivative base class of the HadoopJobTask. In the long run, these sort of handlers will let people all inherit from the recommended/shipped classes and extend them through event handlers instead. Not sure if that's a good explanation, but maybe clear enough? :)

Tarrasch:
I think it's good. Would be awesome to see it in the commit message too.

freider:
I'll fix the failing test... The assertion I made isn't quite true for remotely executing jobs using the minicluster.

freider:
I have to rethink this a bit because of how code distribution works with luigi mapreduce. Closing the PR for now

PR #670, opened 2015-01-26
Removed the RemoteSchedulerResponder class

erikbern:
Since Landscape went nuts about rpc.py and the existance of `RemoteSchedulerResponder` always bothered the hell out of me – decided to kill it once and for all


erikbern:
Haha looks like Landscape is now complaining about server.py instead (for all the "unused" kwargs). Well, at least we removed 40 lines.

The reason we need kwargs there is that otherwise you can't upgrade the workers independently of the server.

Would be good to factor out a proper API at some point. But I don't think RemoteSchedulerResponder was the right approach.

Tarrasch:
LGTM

PR #671, opened 2015-01-27
if the glob is not present at all, a FileNotFoundException is raised. Ca...

plinton:
...tch that exception and continue which should allow execution to continue.

The goal is to have RangeHourly continue and eventually make the glob once it has finished.

### DIFF
--- a/test/range_test.py
+++ b/test/range_test.py
@@ -18,7 +18,6 @@ import luigi
 from luigi.tools.range import RangeEvent, RangeDaily, RangeDailyBase, RangeHourly, RangeHourlyBase, _constrain_glob, _get_filesystems_and_globs
 from luigi.mock import MockFile, MockFileSystem
 import mock
-import time
 import unittest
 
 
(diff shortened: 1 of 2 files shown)


ulzha:
Yeah, limiting to snakebite isn't intended. Checking exists() would be good.

test/range_test.py
```diff
@@ -325,6 +332,7 @@ def test_successfully_inferred(self):
         ])
 
     @mock.patch('luigi.mock.MockFileSystem.listdir', new=mock_listdir)  # fishy to mock the mock, but MockFileSystem doesn't support globs yet
+    @mock.patch('luigi.mock.MockFileSystem.exists', new=mock_listdir)
```
Tarrasch: You mock exists with listdir? Looks fishy :)

Tarrasch:
Can you reword the commit message? It violates 50 character length [1], and it doesn't start with describing that it solves a bug for `tools.range`.

[1]: http://stopwritingramblingcommitmessages.com/

plinton:
@ulzha @Tarrasch @erikbern 
Updated with the exists check, better commit message, tests that actually check the right thing and a little documentation.

erikbern:
LGTM

plinton:
I don't have merge permission so @erikbern, @Tarrasch or @ulzha can you merge please?

PR #672, opened 2015-01-27
Remove global parameters

erikbern:
This is one attempt at removing `is_global=True`. The idea is to support "taskless" parameters by putting them on a separate `luigi.Config` class (that happens to be a Task, but that's another story).

See `luigi.interface` for an example of this, where `EnvironmentParamsContainer` uses `luigi.Config`

This needs some more tests and some cleanup in `luigi.parameter` to be perfect, but let me know what you think of the overall idea!

### DIFF
--- a/luigi/__init__.py
+++ b/luigi/__init__.py
@@ -26,6 +26,7 @@ Event = event.Event
 
 Task = task.Task
 Config = task.Config
+ConfigWithoutSection = task.ConfigWithoutSection
 ExternalTask = task.ExternalTask
 WrapperTask = task.WrapperTask
 Target = target.Target
--- a/luigi/__init__.py
+++ b/luigi/__init__.py
@@ -25,6 +25,7 @@ import event
 Event = event.Event
 
 Task = task.Task
+Config = task.Config
 ExternalTask = task.ExternalTask
 WrapperTask = task.WrapperTask
 Target = target.Target
--- a/luigi/hdfs.py
+++ b/luigi/hdfs.py
@@ -22,10 +22,20 @@ import datetime
 import re
 import warnings
 from luigi.target import FileSystem, FileSystemTarget, FileAlreadyExists
-import configuration
 import logging
 import getpass
 logger = logging.getLogger('luigi-interface')
+
+
+class hdfs(luigi.Config):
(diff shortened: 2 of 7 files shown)


erikbern:
Any thoughts? This one is extra powerful with #682 since it merges global parameters, command line options, and configuration options

ulzha:
Looks good; +1 to shunning "is_global" from object oriented code.

Why introduce is_config to object oriented code? :) It seems to me that that distinction on the parameter itself is unneeded, and neither is include_config knob. Is this to somehow work around collisions? We should fully qualify parameters to solve that I think. I would like to be able to opt to fully qualify with --org.module-xyz.param-name or something and move on, whenever Luigi can't decide which task, configurable core feature, or user defined plugin should consume a given parameter.

Is there a story behind ParamContainer-is-a-Task, other than that this way right now we're moving to consistent options handling with the least friction?

Eventually we'd probably be turning it around, Task-is-or-has-a-ParamContainer? Well I've googled for Python plugin frameworks on this topic, because option parsing and scoping has some strong wheel reinvention flair to it. But haven't had time to evaluate any.

erikbern:
`is_config` is needed to support options like `--n-workers` that aren't of the form (section)-(option). But I realize the name is a bit confusing and it would make more sense to call it something like `ignore_task_name`. I can fix. It would probably be better anyway if the default is to be explicit about the section. What I mean is that if you overload the `Config` option then it *should* include the section in the command line flag by default, unless you specify that it shouldn't.

The story behind ParamContainer-is-a-Task is mainly less friction. I think it's a great idea to flip it around. I can do that in a subsequent PR. I like to separate "feature" PR's from "refactoring" PR's since it's usually easier to verify the latter and make sure that the functionality is equivalent.

erikbern:
Getting an error in Python 2.7 that I can't reproduce locally, messing around with Travis to see how I can repro it

erikbern:
Ok was finally able to repro this locally – turns out something changed in argparse for Python 2.7.9. I'm researching

erikbern:
jeez finally all tests passed...

Tarrasch:
I never thought it would come to use to have both 2.6 and 2.7 on travis. Nice that it's useful :)

PR #678, opened 2015-01-27
Run autopep8 on repo in aggressive mode

themalkolm:
Lets run autopep8 on repo in aggressive mode. It finds various minor
style issues and fixes type detection.

### DIFF
--- a/luigi/date_interval.py
+++ b/luigi/date_interval.py
@@ -71,7 +71,7 @@ class DateInterval(object):
         return hash(repr(self))
 
     def __cmp__(self, other):
-        if type(self) != type(other):
+        if not isinstance(self, type(other)):
             # doing this because it's not well defined if eg. 2012-01-01-2013-01-01 == 2012
             raise TypeError('Date interval type mismatch')
         return cmp((self.date_a, self.date_b), (other.date_a, other.date_b))
--- a/luigi/hdfs.py
+++ b/luigi/hdfs.py
@@ -225,7 +225,7 @@ class HdfsClient(FileSystem):
                    (['-p'] if parents else []) +
                    [path])
             call_check(cmd)
-        except HDFSCliError, ex:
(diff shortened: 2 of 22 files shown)


luigi/date_interval.py
```diff
@@ -71,7 +71,7 @@ def __hash__(self):
         return hash(repr(self))
 
     def __cmp__(self, other):
-        if type(self) != type(other):
+        if not isinstance(self, type(other)):
```
Tarrasch: `autopep8` converted this for you?

luigi/hdfs.py
```diff
@@ -308,7 +308,7 @@ def get_bite(self):
         If Luigi has forked, we have a different PID, and need to reconnect.
         """
         if self.pid != os.getpid() or not self._bite:
-            client_kwargs = dict(filter(lambda (k, v): v is not None and v != '', {
+            client_kwargs = dict(filter(lambda k_v: k_v[1] is not None and k_v[1] != '', {
```
Tarrasch: Is this really prettier?

Tarrasch:
Nice!

luigi/hdfs.py
themalkolm: Nah, personally I think this lambda should be replaced with a simple operator.itemgetter(1) or lambda x: x[1] instead

But I left whatever autopep8 suggested

luigi/date_interval.py
themalkolm: Yep, how cool is that?

luigi/hdfs.py
erikbern: I really don't like treating tuples like arrays and using indices. Much nicer to unpack into variables

themalkolm: Well, at least we all agree that autopep8 went for the weirdest solution.

PR #679, opened 2015-01-27
Fixed various things that Landscape was complaining about

erikbern:


erikbern:
Doesn't really seemed like it helped, still 88% :(

erikbern:
oops not sure why i closed this accidentally

PR #682, opened 2015-01-28
Pull parameter default from config

erikbern:
Depends on #681

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -17,7 +17,32 @@ each controlling a different part of the config. Example
     default-scheduler-host: luigi-host.mycompany.foo
     error-email: foo@bar.baz
 
-Below, we describe each section and the parameters available within it.
+By default, all parameters will be overridden by matching values in the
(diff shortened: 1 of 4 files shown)


erikbern:
Any thoughts?

erikbern:
Can we merge this?

ulzha:
Looks useful. I was thinking about this when arriving at the default values in Range stuff (kind of suited for our arbitrary situation and maybe unsuitable for others'): can have the hardcoded default overridable by a configured default, overridable by the dynamically specified value? Great if this does that; a test with both a hardcoded default and a configured value is lacking.

(Not sure of the usefulness of config_path, wonder why it was invented. Letting user come up with accidental complexity, namely mismatch between config section names and task names - for what purpose?)

erikbern:
`config_path` is an artifact from before, now that they are pulled by default, we should probably deprecate `config_path`.

Let me add a couple of more tests then merge this.

Issue #683, opened 2015-01-28
Proposal for re-evaluating completion of ExternalTasks


boosh:
Currently because ExternalTasks don't implement `run()`, they are marked as not runnable when the task is added. That means that they aren't assigned to a worker if they haven't completed. As a result, the scheduler will never return a PENDING ExternalTask to a worker in `get_work`. So if an ExternalTask completes while luigi is running, it will never be detected and so dependent Tasks will not run on that invocation. I want to change that.

I suggest implementing logic to retry ExternalTasks as follows:

1. Set `runnable=True` for tasks that don't implement `run()`.
2. Create a new multiprocessing.Process subclass that calls `task.complete()` if `run()` is not implemented.

What are your thoughts? I don't want to do this work if it won't be merged in since we don't want to maintain our own fork.

boosh:
Actually I've created something that seems to work on initial testing. I'll do more testing and submit a PR if it is OK.

erikbern:
Ok, sounds interesting. I think the easiest thing is to just re-schedule the graph after it finishes (meaning re-evaluate all external tasks)

boosh:
See PR #684

boosh:
Resolved in #684.

PR #684, opened 2015-01-29
Recheck dependencies (relates to issue #683)

boosh:
This PR allows incomplete external dependencies to be rechecked while Luigi is running. It can be tested with the following config:

```
# client.cfg
[core]
max-reschedules: 3
worker-keep-alive: true
retry-external-tasks: true

[scheduler]
retry-delay: 10
disable-num-failures: 5
```

...and simple workflow:

```
#!/usr/bin/python
# -*- coding: utf-8 -*-
"""
Flow for testing recomputation of dependencies
"""
import logging
from luigi.file import File
import time

import luigi


logging.basicConfig(level=logging.INFO)
logger = logging.getLogger(__name__)


class ExternalFileTask(luigi.ExternalTask):
    path = luigi.Parameter()
    def output(self):
        return File(path=self.path)


class SlowTask(luigi.Task):
    """
    Requires a single file dependency
    """
    path = luigi.Parameter()

    def requires(self):
        yield ExternalFileTask(path="/tmp/luigi/%s.dep" % self.path)

    def output(self):
        return File(
            path="/tmp/luigi/%s.output" % self.path)

    def run(self):
        logger.info("Sleeping for a long time...")
        time.sleep(20)

        # touch file
        open("/tmp/luigi/%s.output" % self.path, 'a').close()


class LastTask(luigi.Task):
    """
    Final task with multiple dependencies
    """
    def requires(self):
        """
        Loads the config file and yields tasks for each one.
        """
        for p in range(1, 3):
            yield SlowTask(path="file-%d" % p)

    def output(self):
        return File(
            path="/tmp/luigi/final.output")

    def run(self):
        open("/tmp/luigi/final.output", 'a').close()


if __name__ == '__main__':
    luigi.run(main_task_cls=LastTask)
```

Run the above, then `touch /tmp/luigi/file-1.dep`. Luigi should now detect that this dependency is met and then run `SlowTask(path="file-1")`. While this is running (or afterwards if config has `worker-keep-alive: true`), `touch /tmp/luigi/file-2.dep`. Luigi should then detect that the dependency for `SlowTask(path="file-2")` is satisified, and run it. Finally `LastTask` should run.

The worker returns a state of FAILED if an ExternalTask is incomplete. This allows us to reuse the current retry logic. However, it'd be better if a distinction was made between "incomplete and will try later" vs "failed due to too many retries". This last case is covered by DISABLED if `disable-num-failures` is set though, so I thought it was just simpler to use FAILED as I've done. We can always introduce a WILL_RETRY state later if necessary to make a distinction in the UI between queued tasks vs disabled/failed ones (I'd like per-task configuration of retries, but not sure if we'll be going down that route internally at the moment. If we do, I might address it then).

Note: This feature is beneficial for my needs (polling an FTP server for existence of a file while other long-running computations are taking place). However, handling failure is still an issue. I think failures should cascade so that if a task is disabled by the scheduler (due to the `disable-num-failures` setting), tasks dependent on the failed task should be marked as disabled or "failed due to an unmet dependency". This would allow the workflow to terminate even if `worker-keep-alive` is `true`. Currently the workflow will not terminate until all pending tasks have completed or failed. I think this should be addressed in a separate issue though.


boosh:
BTW if this looks good I'll create an automated test.

luigi/worker.py
```diff
@@ -490,9 +556,16 @@ def _get_work(self):
...
     def _run_task(self, task_id):
         task = self._scheduled_tasks[task_id]
-        p = TaskProcess(task, self._id, self._task_result_queue,
-                        random_seed=bool(self.worker_processes > 1),
-                        worker_timeout=self.__worker_timeout)
+
+        if task.run == NotImplemented:
+            p = ExternalTaskProcess(task, self._id, self._task_result_queue,
+                                    random_seed=bool(self.worker_processes > 1),
+                                    worker_timeout=self.__worker_timeout)
```
erikbern: Trying to understand your code. Doesn't this mean it will only be run once, possibly at the beginning of the pipeline? What you really want to do is re-check the existence later.

luigi/worker.py
```diff
@@ -412,7 +478,7 @@ def _add(self, task, is_complete):
         elif task.run == NotImplemented:
             deps = None
             status = PENDING
-            runnable = False
+            runnable = configuration.get_config().getboolean('core', 'retry-external-tasks', False)
```
boosh: This line means that the task will be given to a worker to run (depending on a config value).

luigi/worker.py
boosh: This code makes a worker use an ExternalTaskProcess to execute a task that the scheduler has marked as `runnable`, but which has no implemented `run` method (and therefore is assumed to be an ExternalTask).

An ExternalTaskProcess is very similiar to a TaskProcess, but instead of calling the task's `run` method, it checks its `complete` method and returns a status based on that. If the task's `complete` method returns `False`, the task is marked as FAILED. The scheduler will then reschedule it later, subject to the normal retry logic (`disable-num-failures` especially applies here). 

So no, that code doesn't mean it will only be run once at the start of the pipeline (perhaps you're thinking of `check_complete` in worker.py?). The ExternalTaskProcess will be re-evaluated every time the task is given to a worker to execute.

luigi/worker.py
```diff
@@ -65,6 +66,69 @@ def __init__(self, task, worker_id, result_queue, random_seed=False, worker_time
...
+            random.seed((os.getpid(), time.time()))
+
+        status = FAILED
+        error_message = ''
+        try:
+            self.task.trigger_event(Event.START, self.task)
+            t0 = time.time()
+            status = None
+            try:
+                status = DONE if self.task.complete() else FAILED
```
boosh: This class is pretty much the same as TaskProcess, but it checks the task's `complete` method instead of calling `run`.

luigi/worker.py
erikbern: That makes a lot of sense, thanks for the clarification!

erikbern:
I think I understand your code now. One thing remaining that I don't understand: isn't some kind of modification needed in the scheduler in order to return external task dependencies to the worker in `get_work`? Or are missing external dependencies always marked as `FAILED`?

boosh:
That's done by marking a task as `runnable` with a status of PENDING (that's what it was before as well). Only `runnable` tasks are returned to workers to evaluate.

boosh:
Tests added. Let me know if you want anything else.

Cheers

erikbern:
Looks like it's good to go except the pep8 issues. If you fix, I'll merge

boosh:
All done. Thanks.

Issue #331, opened 2014-05-18
A dependency failed does not cause job fail cascade


BusyJay:
For example, job A require job B, job B require job C. If job C fail, job B fail of course. However, job A will not fail. In fact, it won't be run. And it outputs:

> There are 1 pending tasks possibly being run by other workers

This behavior makes it hard to write a job with fault tolerance because A is never be run or triggered any event.

A quick work around is to change a line of code in function get_work, which is in scheduler.py, from

```
elif self._tasks[dep].status != DONE:
```

to

```
elif self._tasks[dep].status not in (DONE, FAILED):
```

Tarrasch:
I'm closing this issue. 

- [x] It has been inactive for +4 months.
- [ ] It's not about luigi core, so not as many users are affected about this.
- [ ] The change seems quite big, it's unlikely to be sporadically picked up.
- [ ] The owner haven't responded or disappeared.
- [ ] I don't understand what this issue is about.
- [ ] There exists a reasonable workaround for this.
- [ ] We need to check if this hasn't been fixed by now (for old issues).
- [ ] This is kind of by design and not a bug.
- [ ] Resolving this would probably add a lot of complexity.


Every open issue adds some clutter, and we try to make the issues fewer and make it easier for new collaborators to find. Currently we try to close any issue that meets the first checkbox + one other.

Feel free to reopen this issue at any point if you have the intent to continue to work this. :)

PR #686, opened 2015-01-30
Cascade the DISABLED state if all dependencies are either DISABLED/DONE ...

boosh:
...and at least one is DISABLED.

This PR will set a task to be DISABLED if any of its dependencies are DISABLED, and the rest are either DISABLED or DONE.

This allows workers to terminate when something is DISABLED, even if worker keep-alives are enabled. WIthout this patch, keep-alive workers never terminate even if tasks are DISABLED, effectively causing the workflow to hang.

The approach taken here means that as much work as possible is done, and only the critical path is disabled (i.e. we "fail slow" instead of immediately bailing out if anything gets DISABLED).

This sort of relates to #331 but not exactly, but does address the note at the end of #684.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -234,6 +234,27 @@ class SimpleTaskState(object):
         else:
             logger.info("No prior state file exists at %s. Starting with clean slate", self._state_path)
 
+    def cascade_disabled_state(self, tasks):
+
+        """ Iterate through all pending tasks, setting the status to DISABLED
+        if at least one of it's dependencies is DISABLED and the rest are
+        either DONE or DISABLED. """
+
+        for task in tasks:
+            if task.status != PENDING:
+                continue
+
+            any_disabled = False
+            should_disable = True
+            for dep in task.deps:
+                t = self.get_task(dep)
+                if t.status == DISABLED:
+                    any_disabled = True
+                should_disable &= (t.status == DISABLED or t.status == DONE)
+
+            if any_disabled and should_disable:
+                self.set_status(task, DISABLED)
+
     def get_active_tasks(self, status=None):
         if status:
             for task in self._status_tasks[status].itervalues():
@@ -591,6 +612,7 @@ class CentralPlannerScheduler(Scheduler):
                               for worker in self._state.get_active_workers())
 
         tasks = list(self._state.get_pending_tasks())
+        self._state.cascade_disabled_state(tasks)
         tasks.sort(key=self._rank(), reverse=True)
 
         for task in tasks:


Tarrasch:
Tests are usually nice :)

boosh:
I'll do the tests if it gets the thumbs up for a merge. Don't want to do unnecessary work if this is somehow against the direction the maintainers want to go in :-)

boosh:
I'll do the tests if it gets the thumbs up for a merge. Don't want to do unnecessary work if this is somehow against the direction the maintainers want to go in :-)

Tarrasch:
Ok, makes sense @boosh !

@erikbern ? I can't say I know enough about the scheduler to sya if this is a good idea.

erikbern:
OK lemme take a look

luigi/scheduler.py
```diff
@@ -234,6 +234,27 @@ def load(self):
         else:
             logger.info("No prior state file exists at %s. Starting with clean slate", self._state_path)
 
+    def cascade_disabled_state(self, tasks):
+
+        """ Iterate through all pending tasks, setting the status to DISABLED
+        if at least one of it's dependencies is DISABLED and the rest are
```
erikbern: minor typo: *its* dependencies

luigi/scheduler.py
```diff
@@ -234,6 +234,27 @@ def load(self):
...
+        either DONE or DISABLED. """
+
+        for task in tasks:
```
erikbern: I don't see how this will always work? If `tasks` isn't topologically sorted then you might not disable all tasks

luigi/scheduler.py
```diff
@@ -234,6 +234,27 @@ def load(self):
...
+            if task.status != PENDING:
+                continue
+
+            any_disabled = False
+            should_disable = True
+            for dep in task.deps:
+                t = self.get_task(dep)
+                if t.status == DISABLED:
+                    any_disabled = True
+                should_disable &= (t.status == DISABLED or t.status == DONE)
```
erikbern: why do you disable tasks where the dependencies are done?

erikbern: i guess i'm confused what's the difference between `any_disabled` and `should_disable`

luigi/scheduler.py
```diff
@@ -591,6 +612,7 @@ def get_work(self, worker, host=None, **kwargs):
                               for worker in self._state.get_active_workers())
 
         tasks = list(self._state.get_pending_tasks())
+        self._state.cascade_disabled_state(tasks)
```
erikbern: (1) I don't really understand why you need to do this. We won't run a disabled task anyway if any of its dependencies are disabled right?
(2) Would be much better if `cascade_disabled_state`  was an immutable method that returns all disabled tasks. Otherwise consider the problem when a top task is "undisabled" – now you need to "undisable" all the dependencies that you just marked as disabled in its transitive closure

luigi/scheduler.py
boosh: A task should be disabled if both of the following are true:

1. At least one of its dependencies are DISABLED (this corresponds to `any_disabled), and
2. All of its dependencies have finished processing. This means all of its dependencies must either be DONE or DISABLED (because FAILED tasks will be rescheduled until they have been disabled by the scheduler). This corresponds to `should_disable`.

So this code sets `any_disabled` to True if any of the dependencies of a task have been DISABLED, and then sets `should_disable` to True only if every dependency is either `DISABLED` or `DONE`

luigi/scheduler.py
boosh: 1. The problem is that workers that are kept-alive never terminate if any tasks become DISABLED. So although currently luigi won't run a task if its dependencies are DISABLED, it also won't terminate, since it'll never try to run a disabled task, so therefore will never fail, so therefore will never be able to apply retry logic. That's why we need to cascade the DISABLED state - to make kept-alive workers terminate.
2. I don't think we'd need to "undisable" anything. cascade_disabled_state goes through all tasks, and for each one it might update the state depending on the states of its dependencies. Or are you saying there's some way that luigi itself can "undisable" tasks? I've seen things go from FAILED -> PENDING, but never from DISABLED -> PENDING...

luigi/scheduler.py
boosh: We don't need to worry about that because this method gets evaluated in every loop of the scheduler. So we may not disable all tasks on the first pass, but we will eventually (within a couple of passes).

boosh:
@erikbern Where did we get to with this? If I write tests, are you happy to merge in?

luigi/scheduler.py
daveFNbuck: Tasks go from disabled to pending after a time delay if you have automatic disabling enabled for repeated failures. Tasks also go from disabled to pending if you schedule a task as disabled and later schedule it as pending. It seems that the logic you'd actually want for this is to not count upstream disabled jobs as pending for the purposes of keep alive.

boosh: Hmm I haven't seen tasks go from disabled back to pending again, but then maybe I wasn't looking for it. Is there a config setting that controls the number of times this can happen, and the delay for making them pending again? I assume after it's flapped a couple of times it's just set to failed?

> Tasks also go from disabled to pending if you schedule a task as disabled and later schedule it as pending

How can you schedule a task as disabled and later schedule it as pending?

> It seems that the logic you'd actually want for this is to not count upstream disabled jobs as pending for the purposes of keep alive.

Yes, that's what my patch does. But it then propagates up through the graph so that the root nodes will be set to disabled if any of their children are disabled and everything else isn't pending.

Do you have an alternative suggestion for how to implement this? Perhaps adding a config setting to enable this behaviour (and default it to disabled) would be good so it doesn't affect people with current deployments.

daveFNbuck: > Is there a config setting that controls the number of times this can happen, and the delay for making them pending again?

No, the intention is actually that if a job flaps between failed and pending too often that it'll get disabled and persist in the disabled state for much longer. This can prevent retries from getting too nasty overnight. See http://luigi.readthedocs.org/en/latest/configuration.html#scheduler for the related parameters.

> How can you schedule a task as disabled and later schedule it as pending?

This happens if you set `disabled = True`, schedule a task, then set `disabled = False` and schedule it again. This is a common pattern if something causes a job to break and I need to disable it until I'm able to fix the conditions that are causing it to break. Once I set `disabled = False` again, I want the job and all its dependencies to be able to run again. The problem with your patch is that temporarily disabling a job would disable all of its downstream dependents until they're rescheduled, whether the disable was triggered within the scheduler or by scheduling it as disabled.

> Do you have an alternative suggestion for how to implement this?

I think you shouldn't disable tasks to prevent workers from being kept alive. If we want to let workers die when all of their tasks are disabled or upstream disabled, just cache the upstream disabled state and use that rather than marking them as disabled.

erikbern:
I forgot this. Do you still want to merge it? Let me take a look

daveFNbuck:
This still doesn't handle re-enables, so it'll take a bit of re-writing.

On Sun, Mar 8, 2015 at 12:29 PM, Erik Bernhardsson <notifications@github.com

boosh:
I can work around this for now by running a cron job that checks that the final output of a job was created by a certain time, so if other people think this may complicate how luigi currently works, we should leave it for now.

daveFNbuck:
It should be fairly simple to get the effect you want. Just keep track of
upstream status and don't count upstream disabled jobs toward keeping
workers running. The only issue I have is that marking them as disabled
rather than upstream disabled will have other negative effects.

erikbern:
should we close this then?

boosh:
I'll close this because I don't have the time to look at it right now. If I need this in future I'll make the changes suggested by @daveFNbuck and reopen the PR.

PR #689, opened 2015-02-02
Performance fix in argparse interface

erikbern:
Instead of adding all global parameters to all subparsers (leading to a O(n^2) complexity where n is the number of tasks), just figure out the subparser that's needed and add it to that one

Running nosetests test/parameter_test.py:

Before: Ran 62 tests in 5.248s
After:  Ran 62 tests in 1.381s

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -250,6 +250,8 @@ class ArgParseInterface(Interface):
 
         add_global_parameters(parser)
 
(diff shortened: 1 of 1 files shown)


erikbern:
We should probably merge this pretty soon – I'm a bit concerned that #665 introduced a pretty huge performance penalty. If people have Luigi workflows with 1000s of tasks it could get really slow.

PR #692, opened 2015-02-03
Fixed a bug in argparse for Python 2.7.9 + added some new tests

erikbern:


erikbern:
Just wasted five hours of my live on this... this is the best workaround that I could think of.

Note that the behavior was completely messed up on Python 2.7.9 (which uses the same argparse as Python 3) but we didn't have any tests that caught it.

Tarrasch:
Sweet! :)

erikbern:
This could probably have been solved a little cleaner, but me and argparse aren't best friends.

Another benefit is the non-HDFS tests are running much faster now

PR #694, opened 2015-02-03
Improvement on ftp support

interskh:
Added support of:
1) custom server port
2) exists() optionally check modified time
3) tls support

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -7,6 +7,7 @@ You can also find unittest for each class.
 
 Be aware that normal ftp do not provide secure communication.
 """
+import datetime
 import os
 import random
 import ftplib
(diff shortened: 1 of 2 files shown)


Tarrasch:
LGTM

interskh:
@themalkolm If mtime is provided, then exists() would check against modified time also - it returns False if the modified time is older than mtime. exists() in luigi.target doesn't necessarily mean the file exists but also means it meets some conditions. And because of that, luigi.target.exists() is called implicitly in a lot of places, like in luigi.task.complete().

themalkolm:
I understand why it is the way it is. But IMHO mtime in exists overloads the definition of exists and kind of goes the wrong way. It is the beauty of Target that it has only one concept and it is a boolean property exists/does not exist.

erikbern:
Didn't see the mtime thing – it's actually a pretty interesting (ab)use of the API. My gut reaction is that it will probably lead to bad things, but maybe it's actually a good pattern. So I'm fine keeping it in `luigi.contrib.ftp` and see how well it works out

themalkolm:
@erikbern well, it is kind of OK if you remember how Makefile works. So nothing new about this idea. My problems is I think the more luigi-way is to encode time information in the file's name. Using mtime, though, allows you to encode time information in mtime. Two ways to do the same thing.

But OK, lets say it is OK. Using extra mtime here makes ftp target not interchangeable with other targets and this is the problem. Like you will need to implement mtime for other targets as well.

erikbern:
Trying to understand your Makefile analogy – could this potentially be used more generally to pick up which files have changed and re-run dependencies?

It doesn't break the API contract right? The only method a `Target` needs to implement is `exists()`

themalkolm:
It totally could be used to re-run and stuff. Makefile proved that this is a very good idea. I just think `exists()` answers a simple question if target exists or not and has nothing to do with mtime. This easily could be my personal misconception.

Adding mtime-like parameters changes the way people can start using luigi. It could be both good and bad. Right now to make luigi run pipeline again you need to make it output e.g. file targets with different names e.g. `foo-HH-MM.json` for every run. With mtime-like parameters you can easily re-run the whole pipeline on *exactly* the same file outputs rewriting them when needed.

interskh:
Here is my opinion on this. A target can be anything that meets some criteria - it could be a `FtpRecentRemoteTarget` with extra mtime check, or a `NonEmptyFileTarget` that checks if files exists AND non-empty or `MysqlTableRowTarget` that checks against mysql table for something.

Agreed some of what @thelinmichael said, the pipeline needs to be re-run. That's why `mtime` I pass in is the absolute datetime not a relative number of seconds to be calculated against now.

Also agreed that `foo-MM-dd.json` is optimal output format. But in some cases, especially in ftp cases, there are times that you are dealing with a 3rd-party vendor and you have no control of file name. Hence the change.

PR #695, opened 2015-02-03
Add Task for running Pig jobs.

jeremykarn:


erikbern:
Nice! Is there any way this can be tested? Either by mocking or using the minicluster.

I think testing is optional for `luigi.contrib` but strongly encouraged :)

Tarrasch:
Yep, tests would be nice. Also the last two commits should be amended . :)

Thanks for the patch!

erikbern:
This is great! Minor thing is that the test should go into `test/contrib`. But let me merge this pull request (assuming the Travis build passes) and I will move the file using a direct commit

jeremykarn:
Thanks!

PR #698, opened 2015-02-03
Add docsapi target to tox.

themalkolm:
Maybe `doc/README.rst` should be updated as well as now everything is much simplier:

```
tox -e docsapi
tox -e docs
```
Configuration is updated.


themalkolm:
Maybe we should embed docsapi invocation into docs so docs always run against the updated doc/api folder. But eh, up to you.

PR #703, opened 2015-02-04
refactored email tests to get test coverage a bit higher

erikbern:


Tarrasch:
Did we remove coveralls?

luigi/notifications.py
```diff
@@ -113,6 +113,8 @@ def send_email_sendgrid(config, sender, subject, message, recipients, image_png)
 
 
 def send_email(subject, message, sender, recipients, image_png=None):
+    config = configuration.get_config()
```
Tarrasch: Looks like you're also changing logic and not only tests.

Tarrasch:
LGTM, but please explain the non-test changes :)

erikbern:
The non test changes were because we wouldn't send email id either (a) you
were running in debug mode (b) running in a terminal. I added a config that
lets you override that check and send email anyway. Otherwise you can't
test it

erikbern:
Coveralls is still running but I disabled the comments on PR's since I thought they were too spammy

See here: https://coveralls.io/r/spotify/luigi

Coverage went up from about 77.5% to 78.3% with this PR. Nothing crazy. But the code was bothering me and hopefully the PR demonstrates how you can make everything testable if you want to,

luigi/notifications.py
```diff
@@ -124,12 +126,10 @@ def send_email(subject, message, sender, recipients, image_png=None):
                  "-------------", recipients, sender, subject, message)
     if not recipients or recipients == (None,):
         return
-    if sys.stdout.isatty() or DEBUG:
+    if (sys.stdout.isatty() or DEBUG) and (not config.getboolean('email', 'force-send', False)):
```
themalkolm: Update docs too maybe?

erikbern: yeah i'm planning to change all config to use the new `Config` class anyway so I'll do it later

PR #710, opened 2015-02-04
usage of string format as logging function parameters

steenzout:
fixes 3 landscape errors on luigi.contrib.redshift


steenzout:
for reference: https://landscape.io/github/spotify/luigi/150/diff/fixed

PR #715, opened 2015-02-05
More documentation to RangeDaily/RangeHourly

ulzha:


doc/luigi_patterns.rst
```diff
@@ -19,21 +19,94 @@ have some kind of periodical garbage collection clean it up.
 Triggering Many Tasks
```
ulzha: I separated the range aspect out of this pattern's original content; the wrapping pattern still kind of can stand on its own.

Tarrasch:
Didn't read too carefully but LGTM

erikbern:
Would have a minor preference for putting most of the documentation on the class/module and then linking to it, but LGTM

PR #716, opened 2015-02-05
Fixed #713.

### LINKED ISSUE #713: hadoop --config cannot be set in client.cfg
We use a config option for hadoop command, so I put the following line to client.cfg:
command: /usr/bin/hadoop --config /etc/rutarget
This results in a "no such file" error raised by Popen.

Here is a simple [fix](https://github.com/kalaidin/luigi/commit/f9db4c189e5171812f959a19d516ca51850bc5fa) I can provide, not sure if this is a right way to go.

kalaidin:
Fixed #713.
Also fixed a warning about deprecated parameter.

### DIFF
--- a/luigi/hdfs.py
+++ b/luigi/hdfs.py
@@ -52,7 +52,7 @@ def call_check(command):
 
 
 def load_hadoop_cmd():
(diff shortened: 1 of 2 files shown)


Tarrasch:
LGTM

PR #720, opened 2015-02-06
Some small changes to the Hive support

erikbern:


Tarrasch:
Ok, so you moved to contrib. Shouldn't the content of the test file change as well (I mean the import should change right?)

erikbern:
I don't really get coveralls... it used to say that hive.py had really bad coverage, so I put together this test. Now it doesn't seem to even generate coverage reports for hive :(

PR #721, opened 2015-02-06
enhanced examples documentation

steenzout:
* added coding statement where it was missing
* added license where it was missing
* added docstring documentation with details that should make it easy to decipher what is going on

I used my current understanding to add the details and finally managed to understand some of the "magic" under the hood but it's not perfect so there may be a few mistakes.

I think that some examples would be easier to understand if there were some diagrams.
I think this would mean adding them to the sphinx docs.
I can help with this if it's worthwhile or needed.

### DIFF
--- a/examples/__init__.py
+++ b/examples/__init__.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2012-2015 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


themalkolm:
So I have no clue and want to ask a question. You just granted Spotify copyrights for all files by default. Are you sure all of them belong to Spotify? From my understating (I'm not a lawyer, though)  something in this repo could be contributed to project by other peoples under Apache License but this does not implies copyright to Spotify.

erikbern:
I tend to agree that some of the docs is taking it a bit too far – we don't need to autogenerate html docs from examples, the point is to read them as they are. I might remove some of the more excecessive stuff

erikbern:
afaik, you need to explicitly assign copyright to something, then grant permission for other people to use it

if you don't assign copyright, then the default assumption is anyone else is not permitted to use it

steenzout:
@themalkolm I've raised this question before regarding the copyright notice.

I've finished my research and from what I'm seeing a project should have a Contributor License Agreement and a Corporate Contributor License Agreement.

a short article : http://oss-watch.ac.uk/resources/cla

a website to help understand CLAs more in-depth and has templates : http://harmonyagreements.org
(Canonical and others used this to generate their CLAs)

a beta tool that integrates with GitHub: https://www.clahub.com
I've tested this with an organization I own and it seems to work ok.

PR #723, opened 2015-02-06
Activated all tests in test.contrib

erikbern:
Realized they weren't running.

Before: Ran 407 tests in 40.602s
After:  Ran 426 tests in 42.342s


erikbern:
This is also why coveralls doesn't display coverage for hive.py in #720

PR #725, opened 2015-02-07
Use new-style config for server/scheduler

erikbern:
Taking advantage of the new mechanism where config & params are the same thing

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -154,7 +154,7 @@ class Parameter(object):
         if self.__config:
             section, name = self.__config['section'], self.__config['name']
         elif task_name is not None and param_name is not None:
-            section, name = task_name, param_name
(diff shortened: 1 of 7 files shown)


erikbern:
I'm also considering making `config_path` a deprecated parameter and always pull from the default location as well

erikbern:
I don't know what's up with these PEP8 changes, but they were definitely not introduced by me:

```
luigi/__init__.py:61:1: E402 module level import not at top of file
luigi/hadoop.py:266:25: E731 do not assign a lambda expression, use a def
luigi/hadoop.py:301:9: E731 do not assign a lambda expression, use a def
test/clone_test.py:41:5: E731 do not assign a lambda expression, use a def
```

I might just turn off E402 and E731 for now

PR #727, opened 2015-02-08
Use new-style config mechanism for the scheduler

erikbern:
Also removed _create_scheduler

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -24,9 +24,12 @@ import logging
 import os
 import time
 
+import configuration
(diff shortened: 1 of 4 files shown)


erikbern:
It's pretty annoying how inconsistent the configuration options are. Some use dashes, some use underscores.

I think we should move to use underscores consistently. With #726 we will throw a deprecation warning if `config_path` is actually being used. We can probably remove `config_path` sometimes next year :)

Tarrasch:
LGTM

PR #728, opened 2015-02-09
enable range use with luigi.rdbms.CopyToTable

ulzha:


erikbern:
LGTM. Btw `luigi.postgres` has really bad test coverage, would be great to bump it up a bit

Tarrasch:
LGTM

ulzha:
Yeah, I was about to add more but gave up after figuring I lack knowledge of the best tools to test a database interface. One generic way would be just to use plain MagicMock and assert on the entire mock_calls list, but one needs to specify fetchone return values. I would hate to laboriously implement all sorts of mock cursors. There should be a quite trivial generic way to script all the important `side_effect`s inline with the definition of the expected calls list itself. But I couldn't google the magic quickly. Wonder if I should invent it?

erikbern:
Yeah using MagicMocks is a bit messy.

I was also thinking about maybe just spin up a Postgres cluster locally or something, but it might be overkill

PR #732, opened 2015-02-10
Pass sys.argv to daemon main function.

splusminusx:


erikbern:
Does this change any functionality or fix any bug?

splusminusx:
When I trying to run daemon it says:

```bash
root@mte:/# luigid --background --pidfile /run/luigi.pid
  File "/usr/local/bin/luigid", line 42, in <module>
    main()
TypeError: main() takes exactly 1 argument (0 given)
```

erikbern:
Crap, you are right. I will fix. Let me see if we can add unit tests too

PR #733, opened 2015-02-10
Remove hdp from tox

erikbern:
Hadoop builds are super slow and Travis doesn't fully parallelize the builds. This should cut down build times by a third at least. CDH and HDP should be almost entirely correlated (in fact we don't have a single check in the code for it)


Tarrasch:
Sure, if somebody wants `hdp` they can run it manually

PR #737, opened 2015-02-11
Fixes for some issues with task history

erikbern:
See https://github.com/spotify/luigi/commit/7707525a4f929226acfb0cd10e5667ddee8ec9cd

@jbalogh do you think this resolves the issue?

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -412,19 +412,18 @@ class CentralPlannerScheduler(Scheduler):
     Can be run locally or on a server (using RemoteScheduler + server.Server).
     """
 
-    def __init__(self, config=None, resources=None, task_history=None, **kwargs):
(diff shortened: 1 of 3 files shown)


jbalogh:
It certainly looks like it. Do you think there was anything relying on the default `scheduler` argument always being the same object?

erikbern:
No, the `Worker` object shouldn't be instantiated anywhere else than from interface.py, so it should be fine

jbalogh:
Thanks for fixing this so fast!

erikbern:
Cool can you let me know if the problem persists?

jbalogh:
I pulled the latest and I'm not seeing any issues.

erikbern:
Great – I'll make another release tomorrow morning, hopefully the last one for a while (there was a rapid succession of bugfixes)

PR #739, opened 2015-02-11
Resolve #738 by requiring positional params to be significant

### LINKED ISSUE #738: Latest Luigi version breaks my hadoop jobs
I used to be able to initialize a class like

```
class MyStreamingJob(luigi.hadoop.JobTask):
    param = luigi.Parameter()
```

with code like

```
MyStreamingJob('param_value')
```

but now this results in an error

```
luigi.parameter.MissingParameterException: MyStreamingJob[args=('param_value',), kwargs={}]: requires the 'param' parameter to be set
```

I believe it's now interpreting the first argument as pool, where it never did that before.

erikbern:


erikbern:
@daveFNbuck this should resolve it!

erikbern:
I don't know if this is an amazing solution – it's probably better to start enforcing that you don't have "config-like" parameters on a `Task` itself – this should go on a config object. On the other hand sometimes you might want to be able to override config on a per-instance basis. I think for now it's better to (ab)use the existing `significant` functionality.

daveFNbuck:
Thanks! This at least seems less nasty than the solution I was considering of changing the parameter sorting.

PR #741, opened 2015-02-11
Allow disabling Hadoop atomic output and S3 input/output in JobRunner

mikepii:
For working with data in S3 with `luigi.hadoop.JobTask`, this PR allows inputs of type `S3Target` and output of type `S3FlagTarget`. Additionally, it allows disabling of atomic output, which is not possible in S3. It requires that atomic output be disabled for S3 output.

```python
class MyHadoopJob(luigi.hadoop.JobTask):
    # . . .
    def job_runner(self):
        return luigi.hadoop.HadoopJobRunner('hadoop-streaming.jar', atomic_output=False)
```

This is also an important step toward supporting the Elastic MapReduce API. (We just use the regular Hadoop interface on an Elastic MapReduce cluster -- Luigi already supports Elastic MapReduce in that sense.)

### DIFF
--- a/luigi/hadoop.py
+++ b/luigi/hadoop.py
@@ -38,6 +38,7 @@ from itertools import groupby
 import configuration
 import luigi
 import luigi.hdfs
+import luigi.s3
 import mrrunner
 
 logger = logging.getLogger('luigi-interface')
@@ -352,7 +353,10 @@ class HadoopJobRunner(JobRunner):
     TODO: add code to support Elastic Mapreduce (using boto) and local execution.
     """
 
-    def __init__(self, streaming_jar, modules=None, streaming_args=None, libjars=None, libjars_in_hdfs=None, jobconfs=None, input_format=None, output_format=None):
+    def __init__(self, streaming_jar, modules=None, streaming_args=None,
+                 libjars=None, libjars_in_hdfs=None, jobconfs=None,
+                 input_format=None, output_format=None,
+                 end_job_with_atomic_move_dir=True):
         def get(x, default):
             return x is not None and x or default
         self.streaming_jar = streaming_jar
@@ -363,6 +367,7 @@ class HadoopJobRunner(JobRunner):
         self.jobconfs = get(jobconfs, {})
         self.input_format = input_format
         self.output_format = output_format
+        self.end_job_with_atomic_move_dir = end_job_with_atomic_move_dir
         self.tmp_dir = False
 
     def run_job(self, job):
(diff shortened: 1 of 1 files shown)


Tarrasch:
Wait a second. The reason we implemented the atomic `move_dir` in luigi was because hadoop-streaming *wasn't* atomic. I don't understand the meaning of the `atomic_output` kwarg.

mikepii:
The functionality of the `atomic_output` arg is that if it is `True` (the default), the atomic `move_dir` should be used (the previous behavior). If it is `False`, then the atomic `move_dir` should be skipped and the hadoop-streaming job should instead output directly to the final directory. Having the option to disable that is necessary when allowing S3 output because there is no atomic move operation in S3 like there is in HDFS.

erikbern:
Maybe atomic output should be a property on the target then... anyway it's probably fine for now

Tarrasch:
Can you rename the kwarg `atomic_output` to something explicit? It's a bad sign that there are two interpretations of a boolean parameter. I'm thinking of `end_job_with_atomic_move_dir`. Yes it's really long, but has only one interpretation. Would that be ok? :)

(If you come up with a better/shorter unambigious name. Go for it! :))

mikepii:
Sounds good, I made the change

Tarrasch:
Thanks @mikepii !! I can merge this. But would you mind just reword your commit message? It's not so useful for to mention that it's "(attempt 2)" :)

mikepii:
There you go, I changed it to say "(attempt 3)". Just kidding

Tarrasch:
Thanks!

luigi/hadoop.py
```diff
@@ -398,10 +403,17 @@ def run_job(self, job):
...
         red_cmd = '{0} mrrunner.py reduce'.format(python_executable)
 
-        # replace output with a temporary work directory
         output_final = job.output().path
-        output_tmp_fn = output_final + '-temp-' + datetime.datetime.now().isoformat().replace(':', '-')
-        tmp_target = luigi.hdfs.HdfsTarget(output_tmp_fn)
+        # atomic output: replace output with a temporary work directory
+        if self.end_job_with_atomic_move_dir:
+            if isinstance(job.output(), luigi.s3.S3FlagTarget):
+                raise TypeError("end_job_with_atomic_move_dir is not supported"
```
geowurster: @mikepii I realize this PR is over a year old and you may not be able to answer my question, but do you remember why `end_job_with_atomic_move_dir=False` is required for S3?  I'm working with Hadoop on Google Cloud Platform and also had to set that flag to `False` several months ago.  I don't fully recall the reason, but i believe the output log files contained URL's that `$ hadoop fs` could not handle.  Something like with an escaped `gs://` like: `'/path/to/output/_log/something_gs%3A//'`.  The path was valid for whatever process wrote it, but not for the copy operation.  Ultimately we found the copy was too slow and relying on the `_SUCCESS` flag was fine, but is this similar to what you were seeing on AWS?

I'm working on allowing [`GCSTarget()`'s as input and output for Hadoop streaming](https://github.com/geowurster/luigi/commit/2274d905da0a33d2af4d109d5c8af01781bc0523) tasks as they are valid paths when working on the Google Cloud Platform, found this PR when doing some research, and am wondering if I should raise the same exception for `GCSFlagTarget()`'s.

geowurster: Ah - https://github.com/spotify/luigi/issues/1092 details the issue.  It's not a blocker, more a question of responsibly blocking something that is broken.

mikepii: Yeah, I don't remember whether there were other issues, but one of them was that it's too slow, and another is that it's not worth it because there's no atomic move operation on S3 like there is on HDFS.

PR #747, opened 2015-02-15
Py3k unit test

gpoulin:
some test fix for python3. I suggest that we merge this only when all non-hdfs test pass for py3k.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,9 @@ env:
 matrix:
   allow_failures:
     - python: "3.4"
+      env: TOX_ENV=cdh
     - python: "3.3"
+      env: TOX_ENV=cdh
 sudo: false
 
 cache:
--- a/bin/deps.py
+++ b/bin/deps.py
@@ -92,16 +92,16 @@ if __name__ == '__main__':
     deps = find_deps_cli()
     for d in deps:
         task_name = d
-        task_output = "n/a"
+        task_output = u"n/a"
         if isinstance(d.output(), RemoteTarget):
-            task_output = "[SSH] {0}:{1}".format(d.output()._fs.remote_context.host, d.output().path)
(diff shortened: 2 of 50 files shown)


Tarrasch:
I like your suggestion! :)

gpoulin:
I'll try to find sometime today to fix the module in contrib that conflict with the docs building.

Any idea on why Travis crashes on python 3.4? On my computer everything run smoothly. Only issue that I have are with test that required snakebite. I looked at the code and really don't understand how this traceback can occur.

gpoulin:
The issue with the python3.4 seems to be due to tox environment. It run fine when running directly nosetest from my system environment but I get the same issue as Travis when running from tox. I'm not so familiar with tox but I'll try to figure out the issue. Is there a reason why python-daemon not part of the tox environment?

erikbern:
Looking at the Travis output it seems like it's an issue in `luigi.contrib.ftp`

https://travis-ci.org/spotify/luigi/jobs/50870719

I guess we aren't running any unit tests for it

```
Warning, treated as error:
/home/travis/build/spotify/luigi/doc/api/luigi.contrib.rst:18: WARNING: autodoc: failed to import module 'luigi.contrib.ftp'; the following exception was raised:
  File "/home/travis/build/spotify/luigi/.tox/docs/lib/python3.3/site-packages/sphinx/ext/autodoc.py", line 378, in import_object
    __import__(self.modname)
  File "/home/travis/build/spotify/luigi/luigi/contrib/ftp.py", line 180, in <module>
    class AtomicFtpfile(file):
NameError: name 'file' is not defined
```

gpoulin:
The docs is failing because of the `contrib.ftp` module, I didn't have time to look at it yet. What I was more concerned about is Travis crashing when starting tornado server on python3.4. I turn out to be a bug in a dependency of spulec/moto#308. The nonhdfs test should pass now. Will fix the docs this week.

Tarrasch:
Should I merge?

PR #748, opened 2015-02-15
Py3k

gpoulin:
More fixes for python3. With those fixes all  nonhdfs test succeed (see #747). I think there is still some work to make sure the IO interface is consistent (bytes vs unicode) #745. I'll also try to make the docs build on python3. For hdfs, it rely on snakebite which doesn't support python3 at the moment spotify/snakebite#62.

### DIFF
--- a/luigi/contrib/hive.py
+++ b/luigi/contrib/hive.py
@@ -29,6 +29,9 @@ import luigi.hadoop
 from luigi.target import FileAlreadyExists, FileSystemTarget
 from luigi.task import flatten
 
+if six.PY3:
+    unicode = str
+
 logger = logging.getLogger('luigi-interface')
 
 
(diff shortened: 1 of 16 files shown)


Tarrasch:
Can you have better names on your PRs? I have no idea what's the difference of "Py3k" and "Py3k test"...

Anyway, this looks good, how did you deal with the fact of snakebite running python 2?

gpoulin:
@Tarrasch The difference between "py3k" and "py3k test" is that "py3k" contains only to Luigi and "py3k test" contains modification to the Luigi and the unit tests. So "py3k" is a subset of "py3k test". If you want I can whether close this PR and keep "py3k test", come with better names or both.

For snakebite, I currently do nothing. Only the nonhdfs test are passing at the moment. I'll look if snakebite people need help for the support of py3k.

PR #752, opened 2015-02-17
Renames inverse_dependencies to inverse_dep_graph to match the api call

daveFNbuck:
This api call was broken by the recent rpc refactoring.


Tarrasch:
Hi @daveFNbuck , what error messages did you get before this patch? Was it `UrlError()` something?

daveFNbuck:
I just saw 404s in my browser and the server logs. I didn't dig deeper than that before finding the cause.

PR #753, opened 2015-02-18
fix failure when ftp._rm_recursive() is called on a file

interskh:
Let's say if the ftp folder is like this:
```
- dirA
  |- fileB
```

`_rm_recursive()` fails on `/dirA/fileB` because `ftplib.nlst("/dirA/fileB")` returns "/dirA/fileB/fileB" in our environment.

It is a problem because `ftp._rm_recursive()` is called in `ftp.RemoteTarget().remove()`


Tarrasch:
LGTM. Let me know if you want more reviews or just merge

PR #756, opened 2015-02-18
Mkae python-daemon a core requirement

Tarrasch:
Previously, it was only required to run the tests. Though given that
luigi is useless without it, I move it as a requirement for the whole
package (like it it in debian).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ install_requires = [
     'pyparsing',
     'tornado',
     'snakebite>=2.5.0',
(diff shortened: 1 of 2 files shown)


gpoulin:
@Tarrasch can you remove ipdb from `requirement.txt` at same time? I put it there for debugging unit test, I did meant to commit that.

Sorry, for the `six` library, I need to sync more often with master.

Tarrasch:
That's cool. We all make mistakes! :)

PR #758, opened 2015-02-18
Remove ipdb

Tarrasch:
As was discussed in #756

### DIFF
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -4,4 +4,3 @@ nose
 psycopg2
 boto
 sqlalchemy
-ipdb


Tarrasch:
Ah, I caught myseflt in that the commit message sucks here. Though since it's a solely red commit I guess people won't find this commit in their blame history anyway.s

gpoulin:
@Tarrasch I thought you would "subtly" include it with #746.

Tarrasch:
I guess you mean #756. Nah, that's more work I think. The online editor is much easier to use than to append commits through CLI. :)

Issue #382, opened 2014-07-21
Adding Python 3 support


konklone:
I've moved all of my Python work over to Python 3 over the last 6 months, and am not eager to look back. This topic also [came up on the list](https://groups.google.com/forum/#!topic/luigi-user/nhZfyjSjKaA) back in April.

Transitioning to Python 3 has been easy so far for me, but I don't have any experience at managing a project that supports both Python 2 *and* 3. Does anyone here?

gpoulin:
I started some work in PR #745. It's still not working but the Top10Artist with a local scheduler is now working.

gpoulin:
@erikbern You can close this one too.

PR #760, opened 2015-02-18
Force hadoop client to hadoopcli on python3

gpoulin:
Also disable snakebite test for python3

### DIFF
--- a/luigi/hdfs.py
+++ b/luigi/hdfs.py
@@ -616,6 +616,14 @@ def get_configured_hdfs_client(show_warnings=True):
     """
     config = hdfs()
     custom = config.client
+    if six.PY3 and (custom == "snakebite" or config.use_snakebite):
+        if show_warnings:
+            warnings.warn(
+                "snakebite client not compatible with python3 at the moment"
+                "falling back on hadoopcli",
+                stacklevle=2
+            )
+        return "hadoopcli"
     if custom:
         # Eventually this should be the only valid code path
         return custom
--- a/luigi/mock.py
+++ b/luigi/mock.py
@@ -24,7 +24,6 @@ except ImportError:
     from io import StringIO
 import sys
 
-from luigi import six
 
 import luigi.util
 from luigi import target
@@ -164,12 +163,3 @@ class MockFile(target.FileSystemTarget):
             return atomic_type()
         else:
             return atomic_type(self.fs.get_all_data()[fn])
-
-
-def skip(func):
-    """
-    Sort of a substitute for unittest.skip*, which is 2.7+.
-    """
-    def wrapper():
-        pass
-    return wrapper
--- a/test/_s3_test.py
+++ b/test/_s3_test.py
@@ -27,26 +27,13 @@ from luigi import six
 import luigi.format
 from boto.exception import S3ResponseError
 from boto.s3 import key
+from moto import mock_s3
 from luigi import configuration
 from luigi.s3 import FileNotFoundException, InvalidDeleteException, S3Client, S3Target
 
-# moto does not yet work with
-# python 2.6. Until it does,
-# disable these tests in python2.6
-try:
-    from moto import mock_s3
-except ImportError:
-    # https://github.com/spulec/moto/issues/29
-    print('Skipping %s because moto does not install properly before '
-          'python2.7' % __file__)
-    from luigi.mock import skip
-    mock_s3 = skip
-
-
 if sys.version_info[:2] == (3, 4):
-    # moto break stuff under python3.4
-    from luigi.mock import skip
-    mock_s3 = skip
+    # spulec/moto#308
+    mock_s3 = unittest.skip('moto mock doesn\'t work with python3.4')
 
 
 AWS_ACCESS_KEY = "XXXXXXXXXXXXXXXXXXXX"
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,6 @@ env:
     - TOX_ENV=nonhdfs
     - TOX_ENV=docs
 
-matrix:
-  allow_failures:
-    - python: "3.4"
-      env: TOX_ENV=cdh
-    - python: "3.3"
-      env: TOX_ENV=cdh
 sudo: false
 
 cache:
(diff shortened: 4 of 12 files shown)


luigi/mock.py
```diff
@@ -170,6 +171,22 @@ def skip(func):
...
+def skipIf(condition):
+    """
+
+    """
+    if condition:
+        return skip
+    return lambda x: x
+
+
+skipIfPy3k = skipIf(six.PY3)
```
erikbern: Why did you put it in mock.py?

gpoulin: `skip` decorator was already there. It makes to keep those 2 together, no? Maybe `skip`shouldn't be there at the first place. You prefer that I move those `skip`, `skipIf` and `skipIfPy3k` somewhere in /test

erikbern: Yeah I think that makes sense. But if you are busy we can merge now and fix it later.

gpoulin:
@erikbern I moved the `skip` to test/helpers.py. It's not used a lot so it was better the fix now.

erikbern:
Awesome! Any idea why the Hadoop tests are failing? I wonder if we need to bundle some module with the hadoop jobs.

gpoulin:
The issue was really strange. The only fact of importing helpers was making the test fail. I think it was some issue with pickling and subprocess (helpers was not importable), but I'm not sure. However adding a simple __init__.py to the test folder seems to solve the issue.

gpoulin:
Adding an __init__.py to the test folder cause other test to fail. I'm giving up, I put back the skip family in `luigi.mock` adding a `_` to those decorator. I prefer to spend time fixing test rather than skipping them.

luigi/hdfs.py
```diff
@@ -616,6 +616,13 @@ def get_configured_hdfs_client(show_warnings=True):
     """
     config = hdfs()
     custom = config.client
+    if six.PY3 and (custom == "snakebite" or config.use_snakebite):
+        if show_warnings:
+            warnings.warn(
+                "snakebite client not compatible with python3 at the moment"
+                "falling back on hadoopcli"
```
berkerpeksag: needs `stacklevel=2`

test/_s3_test.py
```diff
@@ -39,20 +42,18 @@
...
-    mock_s3 = skip
+# moto break stuff under python3.4
+skip_tests = skip_tests or sys.version_info[:2] == (3, 4)
 
 
 AWS_ACCESS_KEY = "XXXXXXXXXXXXXXXXXXXX"
 AWS_SECRET_KEY = "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
 
 
+@_skipIf(skip_tests)
```
berkerpeksag: I don't understand this function. Can't you use a `unittest.skipIf` decorator instead?

```py
@unittest.skipIf(sys.version_info[:2] == (3, 4))
class TestS3Target(unittest.TestCase):
    # ...
```

test/_s3_test.py
```diff
@@ -39,20 +42,18 @@
     # https://github.com/spulec/moto/issues/29
     print('Skipping %s because moto does not install properly before '
           'python2.7' % __file__)
-    from luigi.mock import skip
-    mock_s3 = skip
+    skip_tests = True
```
berkerpeksag: Looks like https://github.com/spulec/moto/issues/29 has been already fixed.

test/_s3_test.py
gpoulin: `unittest.skipIf` is not documented for python 2.6. I assume it's not compatible.

erikbern:
That's fine, you can leave them in luigi.mock. I can try to factor it out

test/_s3_test.py
berkerpeksag: Ah, I see. Thanks for the clarificaiton. You can import unittest2 conditionally. unittest2 will be installed if you're on Python 2.6: https://github.com/spotify/luigi/blob/master/tox.ini#L9

gpoulin: Thanks for pointing out. I'll modify appropriately.

erikbern:
This is great except the merge conflict now :(

gpoulin:
@erikbern I fixed the conflict. I also incorporated @berkerpeksag's comments (no more `luigi.mock.skip`). Also, #763 was skipping all hadoop test under python3. With this PR we only skip the test that actually require snakebite.

gpoulin:
@berkerpeksag unittest2 doesn't seem to be available for python 2.6. The hadoop_test are failing with 
```pytb
@unittest.skipIf(six.PY3, "snakebite minicluster is not supported in python3 yet.")
AttributeError: 'module' object has no attribute 'skipIf'
```

even thought I put the following at the beginning of the file:

```python
try:
    import unittest2 as unittest
except ImportError:
    import unittest
```

gpoulin:
I'm not familiar with tox, but the my understanding is that in the following configuration `py26` doesn't refer to the python version but to the tox environment. Since we are using travis to manage python version, there isn't any tox environment matching py26.

```ini
[testenv]
deps=
  py26: unittest2
```

To fix this issue, I added unittest2 to requirement and use unitest2 only when the corresponding function doesn't exist in unittest

```python
try:
    from unittest import skipIf
except ImportError:
    from unittest2 import skipIf
```

gpoulin:
@erikbern With this last commit all tests that doesn't require Minicluster or snakebite are working on python3. I think we can close #382.

`test_multiple_workers` fails when it's run with all other tests, but it passes when run individually.

Tarrasch:
Please also remove the "allowed failures" as they are not allowed anymore. https://travis-ci.org/spotify/luigi/builds/51701754

Tarrasch:
Please also remove the "allowed failures" as they are not allowed anymore. https://travis-ci.org/spotify/luigi/builds/51701754

Very good job on this @gpoulin !! I guess I could merge this now. Right?

Tarrasch:
Hmm, why is there only one test running? Is there only one non-snakebite&non-minicluster test for our hadoop stuff?

https://travis-ci.org/spotify/luigi/jobs/51701764

gpoulin:
@Tarrasch There is indeed only one hdfs test that doesn't require Minicluster. The other "hadoop" tests are local test and are run as part of the nonhdfs environment. I'll try to see if I can make minicluster run under python3.

It's good to be merged.

erikbern:
Awesome

PR #762, opened 2015-02-18
Override default encoding to utf8 in test that assumes unicode compatibility

freider:
Previously the `test_unicode` test would fail on machines with ascii (or similar) as the default locale encoding

### DIFF
--- a/test/file_test.py
+++ b/test/file_test.py
@@ -23,6 +23,7 @@ import os
 import random
 import shutil
 import unittest
+import mock
 
 import luigi.format
 from luigi import File
@@ -165,7 +166,8 @@ class FileTest(unittest.TestCase):
         self.assertFalse(os.path.exists(self.path))
         self.assertTrue(os.path.exists(self.copy))
 
-    def test_unicode(self):
+    @mock.patch("locale.getpreferredencoding", return_value="utf8")
+    def test_unicode(self, mock_encoding):
         t = File(self.path)
         a = u'我éçф'
         f = t.open('wt')


freider:
@gpoulin fyi :point_up: 
This is a quick fix for a problem we've been seeing in our build host. I guess we should probably add a way for the user to specify the encoding in the target specification. Personally I think using the existing "format" concept is a better fit for setting text mode/encoding than using open flags similarly to the python file `open`. Would be nice if open() pretty much only set if a file is opened for reading or writing and the rest of all format specification is done on the target level of the abstraction. This was at least part of the vision for the Target abstraction when we came up with it, that you wouldn't need to know a whole lot about how the content was stored when you consume a target in a subsequent task.

What do you think @erikbern ?

erikbern:
I agree with flags on the Target object.

There's also a unicode test failing in Hadoop (on the minicluster) – maybe the same reason? See hadoop_test.py

freider:
@erikbern I think that's a separate issue since that job doesn't use any non-ascii characters afaict

gpoulin:
@freider I thought about using `**kwargs` on target to be able to set those kind of thing but I had a second thought and decided to don't because it might be more complex to set for other target. The issue that I had (#745) was that Luigi doesn't make the difference between string stream and bytes stream which make it harder to work with in python3 (it also probably cause with binary file on system that don't use `\n` has newline).

My idea was to have the target support both bytes and string mode. But yeah, I think your approach is better. It's probably easy to implement, simply add a formatter that wrap the target in a `io.TextIOWrapper`. Currently is it possible to use formatter chaining? If not it could be use, for example write unicode text in a gzip.

There is only 2 issues that I see with that:
  - All current test/Task will need to use this formatter with unicode literal or don't use the formatter and use bytes literal (it's not a real issue, it's just gonna take time)
  - To work as expected, the target need to operate in binary mode (don't modify the bytes stream that it receives). Currently, the target that write a temporary open the file in `mode='w'` which convert the `\n` to the system specific newline (in python3 it also encode to system specific). So this mode should be change to `mode='wb'` which would break newline for user on system that doesn't use `\n` as newline. The solution (and probably recommended way) would be to use the text formatter, but it would break the current working code.

gpoulin:
@freider I created a PR (#764) which implement the format and replace the unicode test with new ones that use this format (should be platform agnostic).

I also had few questions about the intended use of the format:
  - Is there reason why `pipe_reader` and `pipe_writer` are `@classmethod`? I see plenty of case are passing extra arguments to the format using the `__init__` could be useful (like specifying the encoding).
  - Doesn't `hdfs_reader` and `hdfs_writer` go again the abstraction of the storage backend? I feel like it should be somewhere in hdfs module and a completely different concept from format

freider:
@gpoulin To answer your first comment. Yeah, format chaining is definitely something I've been meaning to implement (that's a large reason for the pipe_* methods). I've been thinking about implementing a pipe operator on Format or something like that: `File("some/path", format=Text | Gzip)` (just brain storming here).
Using targets with non-default r/w modes was added quite recently to luigi as far as I can remember, so hopefully not too many people are opening targets in text mode, but it would definitely be a slightly backwards breaking change. If I'm not mistaken 'wb'/'rb' is the same as 'w'/'b' on *nix platforms at least.
What we could do in an interim to ease a transition to using formats for specifying text mode is to add a DeprecationWarning and add a Text format wrapper if you open a file in 't*' mode.

To answer your second comment:
* The fact that they are class methods is actually just backwards compatibility that haven't been touched in forever. We had some format specifiers as classes when we started prototyping, and then started introducing some formats as instances instead, but to avoid breaking places where we used the classes as formats (e.g. format=Gzip as opposed to format=Gzip()) I think I added the classmethod decorator as a quick hack to allow both using formats as classes and as instances. When defining new formats that could potentially take parameters (like encoding for a Text format) I would definitely encourage using instance methods instead and rely on duck typing. We might want to get rid of the class type formats as well, but I don't see it as urgent (mostly would make it a bit clearer that they can be used as instances instead).
* Yes, it's not pretty. If I remember correctly I made specialised methods for HDFS as an optimisation to since reading from HDFS in various formats can often be done using single command instead of piping through the normal commands (and for some formats we couldn't even do it using piping). I agree there is probably a cleaner way to do this - at the very least define the hdfs_reader/writers in the hdfs module, but I'm not sure we can cleanly decouple it from format. Any ideas?

Tarrasch:
I merged without carefully reading your conversations, but it's important that organizations can run the tests on their own environments which this PR helps with.

erikbern:
I think it's going to be messy no matter what – you have m underlying data storages (local, hdfs, etc) and n different formats (plain, gzip, etc). Both on the storage and format side you have specific code, but then on top of that you might have specific implementations for any of the m*n combinations. This is something that's awkward in most languages because do you put the m*n combinations on the m-side or the n-side?

gpoulin:
@freider I also thought about using `|` as a pipe operator but finally went for `>>`. The rational being that pipe chaining is not commutative. Something like `open('path', Text | Gzip)` might look like C flags which give the impression that the order doesn't mater. On the other hand `>>` doesn't look commutative, look like the C++ stream and `print >> f, 'text'`.

For the rt/rb/wt/wb, it's me who added it less than a week ago to be able to deal more easily with bytes vs unicode. The distinction is pretty important in python3. Since it is there for a really short moment, it's not documented and not part of any release, I think is relatively save to remove it. I'm coming with a fix for that (3e69c4fa818be3313e5cbd8d9ac69659ceb78740).

For the HDFS format, I think it should be treated the same way as any format. By default, the HdsfTarget would use a Hdfs format. If a user want to use gzip in combination of hdfs, he would have to explicitly pipe the two format `Gzip >> HDFS`. It should be possible to implement this and keeping the current code working?

In fact, I would go even further and generalize the concept of Target and Format. A Target would have to document the "type" of input/output that it expecting (the majority of the target would have bytes stream as expected "type"). The Format would take a input of type A and give output of type B (in case of Gzip A == B == bytes stream). The job of the user would be to create a valid pipe. So for example it would be possible to do something like this `JSON >> Text >> Gzip >> HDFS`. This expect Json serializable python objects, convert the objects to Json Text (one object per line), the produced text is converted to bytes stream, which is compress with gzip and then store to hdfs. It could be used with something like this

```python
class CopyTask(luigi.Task):

    def requires(self):
        return SqlReadOnlyTarget(some_fancy_sql_query, format=SqlDictFormat)

    def output(self):
        return HdfsTarget('some/path', format=JSON >> Text >> Gzip >> HDFS)

    def run(self):
        with self.input().open('r') as input:
            with self.output().open('w') as output:
                for obj in input:
                    output.write(obj)
```
That just an example, it would be terrible in term of performance to get one record at the time. To overcome this issue, create a format that retrieve the records in chunks. The `run` function is completely general.

Advantage:
  - Could work with targets for which bytes stream doesn't make sense (DB, elasticsearch, ...)
  - Flexible
  - Make code reusable (convert from one back end to another is simply a question of pipe)
Disadvantage:
  - Require the user to know which Format can be piped with which one
  - Knowing that an object is a target doesn't give you information of the expected output and input (knowing that the target return bytes stream is not that instructive neither)

Any thought? It's just an idea I had.

PR #763, opened 2015-02-18
Skip snakebite tests on Python 3.

berkerpeksag:


test/minicluster.py
```diff
@@ -17,13 +17,20 @@
...
+from luigi import hadoop, hdfs, six
 from nose.plugins.attrib import attr
 
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest
+
+if six.PY3:
+    raise unittest.SkipTest("snakebite doesn't work on Python yet.")
```
gpoulin: By doing this we also skip the definition of `class MiniClusterHadoopJobRunner` which is needed by hadoop_test

erikbern:
I wonder if we should just convince @wouterdebie to move the minicluster functionality to a separate repo and make it compatible with Python 3 :)

PR #766, opened 2015-02-18
[LUIGI-765] S3FlagTarget now uses == to check slash

### LINKED ISSUE #765: `S3FlagTarget` checks for slash terminated path with `is` instead of `==`
See https://github.com/spotify/luigi/blob/master/luigi/s3.py#L510

This screws up situations where the provided data may be a unicode string.  Is this deliberate?

laserson:
Fixes #765.


erikbern:
Good catch. I wish we had more tests for S3. LGTM

PR #767, opened 2015-02-19
Text format

gpoulin:
Remove text vs binary mode (`rt` vs `rb`) for the target in favor of Format.

In detail:
  - Remove binary and text mode for target that was introduce by #745 (functionality replaced by Nop and Text format)
  - Add a default format to give expected behavior on py3k and retro compatibility on system not using `\n` as newline (https://github.com/gpoulin/luigi/blob/text_format/luigi/format.py#L447)
  - Create a base AtomicLocalFile to be use by target that first write to the local filesystem before moving to the final destination (https://github.com/gpoulin/luigi/blob/text_format/luigi/target.py#L199)
  - Modify the targets to take advantage of those changes
  - small changes to the tests to compensate for the difference between bytes and string (only affect py3k which is not fully supported yet)

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -257,7 +257,7 @@ class RemoteTarget(luigi.target.FileSystemTarget):
             self._fs.get(self.path, self.__tmp_path)
 
             return self.format.pipe_reader(
-                FileWrapper(io.BufferedReader(io.FileIO(self.path, 'r')))
+                FileWrapper(io.BufferedReader(io.FileIO(self.__tmp_path, 'r')))
             )
         else:
             raise Exception('mode must be r/w')
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -36,7 +36,6 @@ import luigi.file
 import luigi.format
 import luigi.target
 from luigi.format import FileWrapper
-from luigi import six
 
 
 class RemoteFileSystem(luigi.target.FileSystem):
@@ -189,7 +188,7 @@ class RemoteFileSystem(luigi.target.FileSystem):
         os.rename(tmp_local_path, local_path)
 
 
-class AbstractAtomicFtpFile(object):
+class AtomicFtpFile(luigi.target.AtomicLocalFile):
     """
     Simple class that writes to a temp file and upload to ftp on close().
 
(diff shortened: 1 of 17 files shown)


erikbern:
Wow this is a big one. Looks good. Is it backwards compatible?

gpoulin:
@erikbern It's indeed a big PR with a short description (I needed some sleep after all this work). I'll edit the PR description to had more details.

And yeah, it should be backward compatible. There is only one edge case that I see could happen only for user on system that doesn't use `\n` as newline (aka windows). The file are now open the file with `io.BufferedWriter(io.FileIO('w'))` instead of `open('w')` which mean that the file doesn't convert `\n` to the system specific newline. To overcome this issue I added a default format, the targets use this default format if not otherwise specified. On system that doesn't use `\n` as newline, the default format is SysNewLine which replicate the newline conversion behavior. A breakage would happen if a user use a custom format on those platform. Even then, the format need to be relying on python to convert `\n` to the system specific newline. It's not the case for the format profited in Luigi Gzip and Bzip2. In fact, I think (didn't tested) those format are broken on those platforms. In addition, the fix for this issue is simple, simply chain the custom format with the SysNewLine format.

erikbern:
Just looking at the code it seems like you are ignoring the 'b' in mode, so how do you make sure it's backwards compatible? Trying to understand the code but it's a lot :) https://github.com/spotify/luigi/pull/767/files#diff-92c74f75f768b714084b20caa61aafc9L128

gpoulin:
@erikbern Yeah it's not backward compatible with 'wb'/'wt'/'rt'/'rb' file mode. However those were added 5 days ago with #745, are not documented anywhere and not part of any release. If you think it's still important to stay compatible with those I can add some work around for this. I just feel it would be a little messy to keep a functionality that probably nobody rely on.

erikbern:
Ok, cool, I thought they had been there for a long time. In that case no biggie, don't worry about backwards compatibility. LGTM (I'll let @freider or @Tarrasch go ahead and push the button)

Tarrasch:
Should we close this in favor of #769 ?

gpoulin2:
@Tarrasch I was thinking that this one was already big I didn't want to make it bigger, but yeah we can close in favor of #769.

Tarrasch:
Yeah sorry I should have merged this before you created to other PR, oh well. :)

PR #768, opened 2015-02-20
Update index.rst

aadu:
Fixed the relative path of the sphinx include directive so that it can find README.rst.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -3,7 +3,7 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-.. include:: README.rst
+.. include:: ../README.rst
 
 Table of Contents
 -----------------


Tarrasch:
Seems like the doc build failed because of this ...

berkerpeksag:
The patch LGTM in my opinion. Currently, the index page at http://luigi.readthedocs.org/en/latest/ is broken (relevant commit is https://github.com/spotify/luigi/commit/7d0b4ec1215059d340cd9711a584abaca44fd49c)

`tox -e docs` command can be simplified. Sphinx already provides some useful make rules. `make -C doc linkcheck` for example. I'll open a PR to simplify it.

Tarrasch:
Thanks @berkerpeksag, you're like an guardian angel for our continuous integration! :)

erikbern:
Currently some build issue but I think it's because this has to be rebased. I'll fix it if it causes build issues in master

PR #769, opened 2015-02-22
Hdfs support for python3

gpoulin:
This PR add a way for the format to explicitly declare the type of there input and output. The chaining of Format do a basic to ensure that the chaining is correct based on the types.

This PR also deprecate `luigi.format.Format.hdfs_reader` and `luigi.format.Format.hdfs_writer` in favor of Format chaining and explicit declaration of output type. `HdfsTarget` still verify for `hdfs_reader` and `hdfs_writer` and use them but would raise a deprecation warning.

This PR also include other minor fixes for python3.

This PR is based on #767 only the 2 last commits are different.

With this PR and spotify/snakebite#138, all hdfs test that doesn't necessitate SnakebiteClient are passing on python3.

### DIFF
--- a/luigi/format.py
+++ b/luigi/format.py
@@ -18,6 +18,11 @@
 import signal
 import subprocess
 import io
+import os
+import re
+import locale
+
+from luigi import six
 
 
 class FileWrapper(object):
@@ -211,6 +216,86 @@ class OutputPipeProcessWrapper(object):
         return False
 
 
+class BaseWrapper(object):
+
+    def __init__(self, stream, *args, **kwargs):
+        self._stream = stream
+        try:
+            super(BaseWrapper, self).__init__(stream, *args, **kwargs)
+        except TypeError:
+            pass
+
+    def __getattr__(self, name):
+        if name == '_stream':
+            raise AttributeError(name)
+        return getattr(self._stream, name)
+
+    def __enter__(self):
+        self._stream.__enter__()
+        return self
+
+    def __exit__(self, *args):
+        self._stream.__exit__(*args)
+
+
+class NewlineWrapper(BaseWrapper):
+
+    def __init__(self, stream, newline=None):
+        if newline is None:
+            self.newline = newline
+        else:
+            self.newline = newline.encode('ascii')
+
+        if self.newline not in (b'', b'\r\n', b'\n', b'\r', None):
+            raise ValueError("newline need to be one of {b'', b'\r\n', b'\n', b'\r', None}")
+        super(NewlineWrapper, self).__init__(stream)
+
+    def read(self, n=-1):
+        b = self._stream.read(n)
+
+        if self.newline == b'':
+            return b
+
+        if self.newline is None:
+            newline = b'\n'
+
+        return re.sub(b'(\n|\r\n|\r)', newline, b)
+
+    def writelines(self, lines):
+        if self.newline is None or self.newline == '':
+            newline = os.linesep.encode('ascii')
+        else:
+            newline = self.newline
+
+        self._stream.writelines(
+            (re.sub(b'(\n|\r\n|\r)', newline, line) for line in lines)
+        )
+
+    def write(self, b):
+        if self.newline is None or self.newline == '':
+            newline = os.linesep.encode('ascii')
+        else:
+            newline = self.newline
+
+        self._stream.write(re.sub(b'(\n|\r\n|\r)', newline, b))
+
+
+class MixedUnicodeBytesWrapper(BaseWrapper):
+    """
+    """
+
+    def write(self, b):
+        if isinstance(b, unicode):
+            b = b.encode(locale.getpreferredencoding())
+        self._stream.write(b)
+
+    def writelines(self, lines):
+        for x in range(len(lines)):
+            if isinstance(lines[x]):
+                lines[x] = lines[x].encode(locale.getpreferredencoding())
+        self._stream.writelines(lines)
+
+
 class Format(object):
     """
     Interface for format specifications.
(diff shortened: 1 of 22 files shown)


gpoulin:
I added some fixes for hadoop. With those, MapReduce is working on python3.

erikbern:
lots of changes, but looks good superficially.

luigi/format.py
```diff
@@ -18,6 +18,14 @@
...
 import subprocess
 import io
+import os
+import re
+import tempfile
+
+from luigi import six
+
+if six.PY3:
+    unicode = str
```
berkerpeksag: this looks unused (you can also use `six.text_type`)

test/file_test.py
```diff
@@ -169,36 +169,64 @@ def test_move(self):
...
+        with open(self.path, 'rb') as f:
+            c = f.read()
+
+        self.assertEqual(b'a\nb\nc\nd', b)
+        self.assertEqual(b'a\r\nb\r\nc\r\nd', c)
+
+    def test_del_with_Text(self):
+        t = File(self.path, luigi.format.UTF8)
+        p = t.open('w')
+        print(u'test', file=p)
```
berkerpeksag: debug print?

test/hdfs_test.py
```diff
@@ -24,9 +24,13 @@
 import luigi
 import mock
 from luigi import hdfs
+from luigi import six
 from minicluster import MiniClusterTestCase
 from nose.plugins.attrib import attr
 
+if six.PY3:
+    basestring = (str, bytes)
```
berkerpeksag: or `six.string_types`

luigi/format.py
```diff
@@ -211,37 +237,120 @@ def seekable(self):
...
+
+    def __enter__(self):
+        self._stream.__enter__()
+        return self
+
+    def __exit__(self, *args):
+        self._stream.__exit__(*args)
+
+    def __iter__(self):
+        for line in self._stream:
```
berkerpeksag: to make the code safer, you may want to wrap this with a `try..except`:

```py
try:
    # yielding
finally:
    self.close()
```

luigi/format.py
```diff
@@ -211,37 +237,120 @@ def seekable(self):
...
+        return re.sub(b'(\n|\r\n|\r)', newline, b)
+
+    def writelines(self, lines):
+        if self.newline is None or self.newline == '':
+            newline = os.linesep.encode('ascii')
+        else:
+            newline = self.newline
+
+        self._stream.writelines(
+            [re.sub(b'(\n|\r\n|\r)', newline, line) for line in lines]
```
berkerpeksag: can this be a generator expression instead list comprehension?

luigi/format.py
```diff
@@ -211,37 +237,120 @@ def seekable(self):
...
         self.args = args
+        try:
+            self.input = args[0].input
+        except AttributeError:
+            pass
+        try:
+            self.output = args[-1].output
+        except AttributeError:
+            pass
+        if 'check_consistency' in kwargs and not kwargs['check_consistency']:
```
berkerpeksag: or `if not kwargs.get('check_consistency', True):`

luigi/format.py
```diff
@@ -254,20 +363,77 @@ def pipe_writer(self, output_pipe):
...
+        return input_pipe
+
+    def pipe_writer(self, output_pipe):
+        return output_pipe
+
+
+class WrappedFormat(Format):
+
+    def __init__(self, wrapper_cls, *args, **kwargs):
+        self.wrapper_cls = wrapper_cls
```
berkerpeksag: `wrapper_cls` can be a class attribute too. that way you can do

```py
class TextFormat(WrappedFormat):
    input = 'unicode'
    output = 'bytes'
    wrapper_cls = TextWrapper
```

luigi/hdfs.py
```diff
@@ -621,7 +621,7 @@ def get_configured_hdfs_client(show_warnings=True):
             warnings.warn(
                 "snakebite client not compatible with python3 at the moment"
                 "falling back on hadoopcli",
-                stacklevle=2
+                stacklevel=2
```
Tarrasch: Nice catch! :)

Tarrasch:
LGTM. So this is all backwards comptible, right?

Can you fix all the comments from @berkerpeksag as well first?

gpoulin:
@Tarrasch for the backward compatibility is the same as #767, so yeah apart from really edge case (https://github.com/spotify/luigi/pull/767#issuecomment-75241916) and really recent change (<7days, 'wb'/'wt'/'rt'/'rb' to open targets). It also deprecates `luigi.format.Format.hdfs_reader` and `luigi.format.Format.hdfs_writer`, but those are still working (just "raise" a deprecation warning).

I'll try to include @berkerpeksag by the end of the day.

test/file_test.py
gpoulin: No, it's that the io.TextIOWrapper try to close the file on `del` which we don't want since in our case close => commit

test/hdfs_test.py
gpoulin: I wanted to capture both `str` and `bytes`, but it this case `str` is enough.

luigi/format.py
gpoulin: I tested, it can. It rely on the fact that the underlying stream will be able to handle. I'll change it.

gpoulin:
@Tarrasch I incorporated the comment of @berkerpeksag. So apart if somebody else have comment I won't touch it. (I won't open a bigger PR... I swear).

Tarrasch:
I would like to press merge. But there's a conflict. Can you resolve it?

gpoulin:
@Tarrasch I was indeed conflicting with myself. It solved now.

Tarrasch:
I'll merge once tests passes :)

PR #771, opened 2015-02-23
make filesystem inference work for RangeDaily

ulzha:
This brings RangeDaily feature set on par with RangeHourly, by generalizing infer_bulk_complete_from_fs implementation.
    
Previously it just was not implemented (as the removed FIXME in RangeDaily docstring says). It was only implemented for the hourly case. Scheduling a daily job outputting to e.g. HdfsTarget raised NotImplementedError.

### DIFF
--- a/luigi/tools/range.py
+++ b/luigi/tools/range.py
@@ -374,28 +374,29 @@ def _get_per_location_glob(tasks, outputs, regexes):
         if m is None:
             raise NotImplementedError("Couldn't deduce datehour representation in output path %r of task %s" % (p, t))
 
-    positions = [most_common((m.start(i), m.end(i)) for m in matches)[0] for i in range(1, 5)]  # the most common position of every group is likely to be conclusive hit or miss
+    n_groups = len(matches[0].groups())
+    positions = [most_common((m.start(i), m.end(i)) for m in matches)[0] for i in range(1, n_groups + 1)]  # the most common position of every group is likely to be conclusive hit or miss
 
-    glob = list(paths[0])  # TODO sanity check that it's the same for all paths?
+    glob = list(paths[0])  # FIXME sanity check that it's the same for all paths
     for start, end in positions:
         glob = glob[:start] + ['[0-9]'] * (end - start) + glob[end:]
     return ''.join(glob).rsplit('/', 1)[0]  # chop off the last path item (wouldn't need to if `hadoop fs -ls -d` equivalent were available)
 
 
-def _get_filesystems_and_globs(task_cls):
+def _get_filesystems_and_globs(datetime_to_task, datetime_to_re):
     """
-    Yields a (filesystem, glob) tuple per every output location of task_cls.
+    Yields a (filesystem, glob) tuple per every output location of task.
 
-    task_cls can have one or several FileSystemTarget outputs.
+    The task can have one or several FileSystemTarget outputs.
 
-    For convenience, task_cls can be a wrapper task,
+    For convenience, the task can be a luigi.WrapperTask,
     in which case outputs of all its dependencies are considered.
     """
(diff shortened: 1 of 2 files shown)


luigi/tools/range.py
```diff
@@ -374,28 +374,29 @@ def _get_per_location_glob(tasks, outputs, regexes):
         if m is None:
             raise NotImplementedError("Couldn't deduce datehour representation in output path %r of task %s" % (p, t))
 
-    positions = [most_common((m.start(i), m.end(i)) for m in matches)[0] for i in range(1, 5)]  # the most common position of every group is likely to be conclusive hit or miss
+    n_groups = len(matches[0].groups())
+    positions = [most_common((m.start(i), m.end(i)) for m in matches)[0] for i in range(1, n_groups + 1)]  # the most common position of every group is likely to be conclusive hit or miss
 
-    glob = list(paths[0])  # TODO sanity check that it's the same for all paths?
+    glob = list(paths[0])  # FIXME sanity check that it's the same for all paths
```
Tarrasch: Mind sharing your take on TODO vs FIXME? :)

test/range_test.py
```diff
@@ -455,7 +458,78 @@ def test_start_long_before_long_hours_back_and_with_long_hours_forward(self):
...
+
+    def test_wrapped_inconsistent_datehour_globs_not_inferred(self):
+        class InconsistentlyParameterizedWrapperTask(luigi.WrapperTask):
+            dh = luigi.DateHourParameter()
+
+            def requires(self):
+                yield TaskA(dh=self.dh - datetime.timedelta(days=1))
+                yield TaskB(dh=self.dh, complicator='no/worries')
+
+        with self.assertRaises(NotImplementedError):
```
Tarrasch: Ah, didn't know this could be a context manager :)

luigi/tools/range.py
```diff
@@ -498,4 +503,7 @@ def missing_datetimes(self, task_cls, finite_datetimes):
         try:
             return set(finite_datetimes) - set(map(self.parameter_to_datetime, task_cls.bulk_complete(list(map(self.datetime_to_parameter, finite_datetimes)))))
         except NotImplementedError:
-            return infer_bulk_complete_from_fs(task_cls, finite_datetimes)
+            return infer_bulk_complete_from_fs(
+                finite_datetimes,
+                lambda d: task_cls(self.datetime_to_parameter(d)),
+                lambda d: '(%04d).*(%02d).*(%02d).*(%02d)' % (d.year, d.month, d.day, d.hour))
```
Tarrasch: Would it work to use `d.format()` perhaps?

Tarrasch:
Can you just briefly say what the end effect will be? Will there be less round-trips to the namenode after this patch when using RangeDaily? The logical is kind of non-trivial and all over the place (which I understand, what you're doing here will require some tricky logic). So it could be nice with a little bit of extra description. :)

Tarrasch:
But overall LGTM

luigi/tools/range.py
ulzha: In my mind TODO is for something missing but not astonishingly so, while FIXME is for when a function or UX is broken or quite inevitably going to break astonishingly.

In this case I judged astonishment by my own inability to recall what I expected from this code :) Today I expected path formatting like `MockFile(self.dh.strftime(('/even' if self.dh.hour % 2 == 0 else '/odd') + '/%Y%m%d%H'))` to trigger an exception from this inference code but it didn't.

luigi/tools/range.py
ulzha: Actually strftime would work...

test/range_test.py
gpoulin: Not under 2.6 https://travis-ci.org/spotify/luigi/jobs/51823860#L775

I will add fix in #777

PR #772, opened 2015-02-23
Fix parameter passing

Tarrasch:
Copy pasting the commit message from the second commit, as that is what's essential here:

-----

At the moment, we think the behavioral change introduced in a78338c was
a mistake. There were no test cases that covered the behavior of passing
global parameters positionally, so this commits adds that.

The major issue was partially solved in f5c3316, but I think that was
kind of an arbitrary fix that still breaks backward compatibility.
Please see the discussion in spotify/luigi#738.  Hopefully, nobody
started to rely on this intermediete behavior that would have existed
for about 2 weeks.

Eventually, we'll have to change the condition

        positional_params = [(n, p) for n, p in params if not p.is_global]

to just

        positional_params = [(n, p) for n, p in params]

just like we did in a78338c. Though that should be done after global
parameters have been deprecated for half a year or something at least.
People need time to change their cron lines from `--pool development` to
`--BaseHadoopJobTask-pool development`.  Then we can change it to

        positional_params = [(n, p) for n, p in params]

AND remove is_global (you can't just do one of them like we did in
a78338c)

----

### DIFF
--- a/test/parameter_test.py
+++ b/test/parameter_test.py
@@ -341,22 +341,28 @@ class TestRemoveGlobalParameters(unittest.TestCase):
         MyConfigWithoutSection.mc_r.reset_global()
         MyConfigWithoutSection.mc_s.reset_global()
 
+    def run_and_check(self, args):
+        run_exit_status = luigi.run(['--local-scheduler', '--no-lock'] + args)
+        self.assertTrue(run_exit_status)
+        return run_exit_status
+
+
     def test_use_config_class_1(self):
-        luigi.run(['--local-scheduler', '--no-lock', '--MyConfig-mc-p', '99', '--mc-r', '55', 'NoopTask'])
+        self.run_and_check(['--MyConfig-mc-p', '99', '--mc-r', '55', 'NoopTask'])
         self.assertEqual(MyConfig().mc_p, 99)
         self.assertEqual(MyConfig().mc_q, 73)
         self.assertEqual(MyConfigWithoutSection().mc_r, 55)
         self.assertEqual(MyConfigWithoutSection().mc_s, 99)
 
     def test_use_config_class_2(self):
-        luigi.run(['--local-scheduler', '--no-lock', 'NoopTask', '--MyConfig-mc-p', '99', '--mc-r', '55'])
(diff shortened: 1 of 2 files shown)


test/parameter_test.py
```diff
@@ -385,14 +411,71 @@ class Dogs(luigi.Config):
...
+            # This could typically be "--num-threads=True"
+            x_g3 = luigi.Parameter(default='y', is_global=True)
+            local_param = luigi.Parameter()
+
+        MyTask('setting_local_param')
+
+    def test_contrived_positional_parameter(self):
+        """ The original minimal test case we found for #738 looked like this,
+        though it is in fact pretty contrived. Though it's good to have a test
+        case that is more similar to the real-world use case """
```
ulzha: This kind of history documentation belongs better in commit message than in code, where it will confuse people as to what is true now and what was true back then... especially when code evolves and diverges from comment contents as it tends to do.

E.g. "Now they need to be insignificant" is not true anymore, right? It reads very confusingly. One can't be expected to notice all the "looked like this" layers and keep them in mind.

test/parameter_test.py
```diff
@@ -385,14 +411,71 @@ class Dogs(luigi.Config):
...
+            """Subclass the JobTask to mess with positional parameters"""
+            # Before commit https://github.com/spotify/luigi/commit/a78338c77d8d79497a7f89bbfde78ba9877acb36
+            # it was possible to use positional parameters and subclasses if the
+            # superclass parameters where is_global
+            # Now they need to be insignificant:
+            # https://github.com/spotify/luigi/commit/f5c33160573a87486c5f0b7423245a89a6d02e0b
+            contrived_global = luigi.Parameter(default='y', is_global=True)
+
+        class TaskWithOneParam(TaskWithOneGlobalParam):
+            the_one_and_only_param = luigi.Parameter()
```
ulzha: the_one_and_only_positional_param? The task actually does have more than one param, right?

Tarrasch: You're right. I didn't dare change code I copy pasted, though of course I should dare too. :)

Will fix.

PR #776, opened 2015-02-23
Remove coverage and coveralls for now

Tarrasch:
This is about fixing the SUPER-SCARY false positives we get from Travis
CI service. See #759. I see this as urgent, so if the tests fail for
this PR, that's good.

While we haven't had any incident yet, we were about to have two on
Spotify's side and apparently @daveFNbuck have had one already (#752).

Just now, I got some 2.6 error which is still green.
https://travis-ci.org/spotify/luigi/jobs/51362963

Issue #759, opened 2015-02-18
[Critical] Travis gives green on failures


Tarrasch:
Hey! I found this super serious issue. Ping @erikbern @freider @daveFNbuck @berkerpeksag. Our travis says green *regardless* if there are failures or not. See for example https://travis-ci.org/spotify/luigi/jobs/51103701 from #751. I guess this is what @daveFNbuck fixed in #752.

I'll fix this today after lunch unless anybode else wants to do this. :)

erikbern:
that's weird!

berkerpeksag:
Interesting, here is an example with non zero exit code: https://travis-ci.org/spotify/luigi/jobs/49392611#L371 The code looks a bit different since I was worked on it, I need to take a look first.

Just a wild guess: If you see this behavior only in "nonhdfs" tests, you can try to change NOSE_ATTR=!minicluster to NOSE_ATTR="!minicluster" in https://github.com/spotify/luigi/commit/af8f3e475a3c484b12b4bcd5a700cb8248d80253#diff-b91f3d5bd63fcd17221b267e851608e8R18

Tarrasch:
Again I think it's a nose bug or something. It does in fact report errors when there are error. It's just that `bin_test.py` is magical or something.

It seems to behave correctly when I run the individual tests but not both (I trimmed away other tests so I can iterate faster, though the following log scales to the clean repository)

```
~/spotify/repos/luigis/github_luigi fix-travis-false-positives*  11s
❯ export PYTHONPATH=''; tox -e nonhdfs test/bin_test.pyc; echo $?  
nonhdfs develop-inst-noop: /home/arash/spotify/repos/luigis/github_luigi
nonhdfs runtests: PYTHONHASHSEED='321437419'
nonhdfs runtests: commands[0] | coverage run test/runtests.py -v test/bin_test.pyc
test_404 (bin_test.LuigidTest) ... Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/arash/spotify/repos/luigis/github_luigi/.tox/nonhdfs/local/lib/python2.7/site-packages/mock.py", line 1201, in patched
    return func(*args, **keywargs)
[...]

PR #777, opened 2015-02-23
Fix tests that were failing unnoticed

gpoulin:
Some minor fixes to tests. Some failing test were introduced due to Travis issue #759 #776.

### DIFF
--- a/test/range_test.py
+++ b/test/range_test.py
@@ -509,10 +509,12 @@ class FilesystemInferenceTest(unittest.TestCase):
                         'spl': MockFile(base + '/something.spl'),
                     }
 
-        with self.assertRaises(NotImplementedError):
+        def test_raise_not_implemented():
             list(_get_filesystems_and_globs(
                 lambda d: InconsistentlyOutputtingDateHourTask(d),
                 lambda d: d.strftime('(%Y).*(%m).*(%d).*(%H)')))
+
+        self.assertRaises(NotImplementedError, test_raise_not_implemented)
 
     def test_wrapped_inconsistent_datehour_globs_not_inferred(self):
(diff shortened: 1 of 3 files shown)


gpoulin:
I added some fixes for https://travis-ci.org/spotify/luigi/jobs/51823860#L775

test/range_test.py
```diff
@@ -509,11 +509,13 @@ def output(self):
                         'spl': MockFile(base + '/something.spl'),
                     }
 
-        with self.assertRaises(NotImplementedError):
+        def test_raise_not_implemented():
```
berkerpeksag: I'd suggest to add the following pattern to test/helpers.py:

```py
try:  # or ``if py_version < 2.7, import unittest2``
    import unittest2 as unittest
except ImportError:
    import unittest
```

then import `from helpers import unittest` in all test files.

gpoulin:
There is some strange test failing under python3. `MultipleWorkersTest.test_multiple_workers` failed when run with:

```sh
. .tox/nonhdfs/bin/activate
coverage run test/runtests.py -v test
```

but won't if run with

```sh
. .tox/nonhdfs/bin/activate
coverage run test/runtests.py -v test/worker_test.py
```

or

```sh
. .tox/nonhdfs/bin/activate
python test/runtests.py -v test
```

test/range_test.py
gpoulin: In theory, I agree. However some tests fail when importing helpers even when helpers is an empty file (https://github.com/spotify/luigi/pull/760#issuecomment-75191878). It can be fixed by adding a `__init__.py` file in the test folder but that makes fail other tests.

Tarrasch:
Thank you so much! Again, you really made my day so much easier! :)

PR #779, opened 2015-02-23
Fixes some outstanding Python 2/3 compatability issues.

gwax:


luigi/postgres.py
```diff
@@ -25,6 +25,7 @@
 import tempfile
 
 from luigi import six
+from luigi.six.moves import xrange
```
berkerpeksag: or you can just change line 324 from `xrange(2)` to `range(2)`. 2 is not an important value to use xrange in my opinion :)

gwax: Done.

Tarrasch:
LGTM.

Tarrasch:
Maybe you could squash the two commits though :)

gwax:
Squashed.

tym-oao:
@gwax Could you take a look my Google group post here: https://groups.google.com/d/msg/luigi-user/2UyGTJHFx_8/8n4TcGP9860J - the encode('utf-8') in line 315 seems to be causing the tmp_file write to fail for me. Just wondering if you could help. Thanks!

gwax:
It looks to me like the problem is that, in some cases, we are doing an encode('utf-8') on lines 262 AND 315. I would suggest removing the .encode('utf-8') from line 262 and submitting a pull request.

tym-oao:
Thanks! I tried that, but then I still get the error from line 315. Removing the .encode('uff-8') from that line is the only thing that works for me. I imagine that change would break other use cases, however. I'm afraid I don't understand enough about this commit to understand what other sorts of data I could try passing through it to find a solution that works for more than my particular file. :/

PR #781, opened 2015-02-24
Render documentation on separate pages

erikbern:
I hope this works – last time there were discrepancies between my local environment and readthedocs


erikbern:
Also nice to remove 400 lines :)

Tarrasch:
What will the difference be?

erikbern:
Right now http://luigi.readthedocs.org/en/latest/api/luigi.html is one big page. This breaks it down into one page per submodule

erikbern:
I'll try merging and if it doesn't work I'll revert

erikbern:
Crap, readthedocs still does something weird. No idea what. I'll revert for now. Will see if there's a version mismatch

berkerpeksag:
The reason is that readthedocs is not invoke `tox -e docs`. It invokes `sphinx-build` instead: http://docs.readthedocs.org/en/latest/builds.html#understanding-what-s-going-on

PR #782, opened 2015-02-24
test if travis fail with process instead of thread

gpoulin:
On my local machine starting the server in a process instead of a thread seems to make the tests fail as expected.

### DIFF
--- a/test/server_test.py
+++ b/test/server_test.py
@@ -15,10 +15,9 @@
 # limitations under the License.
 #
 
-import mock
+import multiprocessing
 import random
 import time
-import threading
 import unittest
 try:
     from urllib2 import Request, urlopen, HTTPError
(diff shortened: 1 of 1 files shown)


gpoulin:
Don't merge this, it's really draft. If it works I'll create a PR with only modification in `server_test.py`.

gpoulin:
It seems to work I'll do one more test.

gpoulin:
I kept the two commits to keep reference of the two Travis build. When everything seems OK and if you want to merge let me know I'll squash the commit and give a better description.

@Tarrasch this should fix #759.

gpoulin:
https://travis-ci.org/spotify/luigi/jobs/51936789#L388 ??? That shouldn't have passed and it fail on my local machine. Anyway I would say merge and try to see why it passes later. I'll also try to see why the redshift tests are failing on python3.3.

Tarrasch:
This is an awesome catch. Please squash and give description and I'll merge! :)

gpoulin:
@Tarrasch Done! Travis should be red because of an issue in the redshift test on python3. I'll try to fix later. There is also some error at the shutdown of the test https://travis-ci.org/spotify/luigi/jobs/51953774#L736. It seems to be due to worker that stay alive after the tests and try to acquire a lock that is deleted before the worker. Not sure thou.

Tarrasch:
Thank you so much!!!! :)

PR #784, opened 2015-02-24
Revert "Render documentation on separate pages"

erikbern:
Reverts spotify/luigi#781

### DIFF
--- a/doc/api/luigi.contrib.rst
+++ b/doc/api/luigi.contrib.rst
@@ -1,0 +1,126 @@
+luigi.contrib package
+=====================
+
+Submodules
+----------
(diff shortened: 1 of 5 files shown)


erikbern:
This is so annoying

erikbern:
Btw this build is going to fail because there's a 404 on readthedocs right now (all builds will fail until this one is merged and built).

I think I realized what's going on. I've been testing my stuff using `tox -e docs` but I guess readthedocs builds the docs in a different way. Will figure out how to configure readthedocs

Tarrasch:
Ah, nice catch. :)

PR #785, opened 2015-02-24
Fix redshift and s3 for python3

gpoulin:
Also reactivate s3 unittest but skip broken tests

PR #787, opened 2015-02-25
Accept to write unicode in MockFile for compat

### LINKED ISSUE #786: Did python 3 refactoring break backwards compat?
Spotify just noticed a type error like this.

```
  File " ... ", line 433, in test_nulls
    result = job.test_run(records)
  File "/usr/lib/python2.6/dist-packages/spotify/luigi/hadoop.py", line 421, in test_run
    print >> f, l
  File "/usr/lib/python2.6/dist-packages/luigi/mock.py", line 143, in write
    BytesIO.write(self2, data)
TypeError: expecting a bytes object, got unicode
```

@gpoulin, you did a lot of nice improvements and refactorings regarding the way FileTargets, can you immediately see if a stack trace like this can be related?

gpoulin:
Only accept on python2
Fix #786

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -28,6 +28,7 @@ Be aware that normal ftp do not provide secure communication.
 import datetime
 import ftplib
 import os
+import sys
 import random
 import io
 
@@ -35,7 +36,7 @@ import luigi
 import luigi.file
 import luigi.format
 import luigi.target
-from luigi.format import FileWrapper
+from luigi.format import FileWrapper, MixedUnicodeBytes
 
 
 class RemoteFileSystem(luigi.target.FileSystem):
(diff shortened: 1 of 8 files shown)


gpoulin:
@Tarrasch The reds on Travis are expected. Those are fixed by #785.

luigi/contrib/ftp.py
```diff
@@ -226,6 +227,11 @@ def __init__(
     ):
         if format is None:
             format = luigi.format.get_default_format()
+
+        # Allow to write unicode in file for retrocompatibility
+        if sys.version_info[:2] <= (2, 6):
```
berkerpeksag: Shouldn't 2.6 be 2.7 here?

gpoulin: No it's really 2.6. The io.BufferedWriter accept unicode on python2.7. `io.BufferedWriter(io.FileIO('test.txt', 'w')).write(u'test')` works fine on 2.7. In fact, that's the reason that I didn't put the MixedUnicodeBytes format at the first place, I didn't think it was needed.

PR #794, opened 2015-02-26
add configurable per-invocation task-limit

ulzha:
Prevents incidents due to spamming of the central scheduler, usually accidental.

The default limit is None, so nothing existing should be affected.

### DIFF
--- a/test/cmdline_test.py
+++ b/test/cmdline_test.py
@@ -140,7 +140,7 @@ class CmdlineTest(unittest.TestCase):
 
         with mock.patch("luigi.configuration.get_config") as getconf:
             getconf.return_value.get.side_effect = ConfigParser.NoOptionError(section='foo', option='bar')
-            getconf.return_value.get_boolean.return_value = True
+            getconf.return_value.getint.return_value = 0
 
             luigi.interface.setup_interface_logging.call_args_list = []
             luigi.run(['SomeTask', '--n', '42', '--local-scheduler', '--no-lock'])
--- a/test/worker_test.py
+++ b/test/worker_test.py
@@ -31,6 +31,7 @@ import luigi.worker
 import mock
 from helpers import with_config
 from luigi import ExternalTask, RemoteScheduler, Task
+from luigi.mock import MockFile, MockFileSystem
 from luigi.scheduler import CentralPlannerScheduler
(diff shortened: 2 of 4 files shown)


freider:
lgtm

test/worker_test.py
```diff
@@ -838,5 +839,53 @@ def test_purge_hung_worker_override_timeout_time(self, mock_time):
...
+    def output(self):
+        return MockFile('.'.join(map(str, self.p)))
+
+    def run(self):
+        with self.output().open('w') as f:
+            f.write('Done!')
+
+    def requires(self):
+        if len(self.p) < self.depth:
+            for i in xrange(self.breadth):
```
gpoulin: As my previous comment, can you change this to range or six.moves.range(self.breadth). Otherwise, the test won't pass under python3.

ulzha:
Aww https://travis-ci.org/spotify/luigi/jobs/52282377 how did mocking go this wrong on me? :/

    ERROR: test_cmdline_logger (cmdline_test.CmdlineTest)
    ...
      File "/home/travis/build/spotify/luigi/luigi/worker.py", line 466, in add
	for next in self._add(item, is_complete):
      File "/home/travis/build/spotify/luigi/luigi/worker.py", line 483, in _add
	if self.__task_limit is not None and len(self._scheduled_tasks) >= self.__task_limit:
    TypeError: unorderable types: int() >= MagicMock()

ulzha:
Wait... I'll fix the cmdline test, it looks like outdated and relying on configuration.py too much.

test/cmdline_test.py
```diff
@@ -140,7 +140,7 @@ def test_cmdline_logger(self, setup_mock, warn):
 
         with mock.patch("luigi.configuration.get_config") as getconf:
             getconf.return_value.get.side_effect = ConfigParser.NoOptionError(section='foo', option='bar')
-            getconf.return_value.get_boolean.return_value = True
+            getconf.return_value.getint.return_value = 0
```
ulzha: @gf-atebbe may clarify why the True was relevant, apparently it has become irrelevant (get_boolean doesn't exist anymore), yet the MagicMock existence has become relevant with this patch...

PR #613, opened 2015-01-20
Fix test @attr-ibutes in hadoop_test.py

Tarrasch:
Apparently the fixtures came and bit us in the end after all. As was
feared from spotify/luigi#601. The attributes feature of nosetests is
completely borked when used together with mixins. It's a total mess and
nothing works. So I just removed all meta programming from these tests.

After this patch:

    ~/spotify/repos/luigis/github_luigi fix_hadoop_test_attributes*
    ❯ NOSE_ATTR='minicluster' nosetests --collect-only test/hadoop_test.py
    ....
    ----------------------------------------------------------------------
    Ran 4 tests in 0.000s

    OK

    ~/spotify/repos/luigis/github_luigi fix_hadoop_test_attributes*
    ❯ NOSE_ATTR='!minicluster' nosetests --collect-only test/hadoop_test.py
    ............
    ----------------------------------------------------------------------
    Ran 12 tests in 0.000s

    OK

Before this patch:

    ~/spotify/repos/luigis/github_luigi master
    ❯ NOSE_ATTR='minicluster' nosetests --collect-only test/hadoop_test.py
    .....
    ----------------------------------------------------------------------
    Ran 5 tests in 0.000s

    OK

    ~/spotify/repos/luigis/github_luigi master
    ❯ NOSE_ATTR='!minicluster' nosetests --collect-only test/hadoop_test.py
    .................
    ----------------------------------------------------------------------
    Ran 17 tests in 0.000s

    OK

The reason it shows 17 before was that it had an empty test.

PR #797, opened 2015-03-01
Create BaseTargetTestCase

gpoulin:
Also fix some s3 bugs catched by those tests

### DIFF
--- a/test/hdfs_test.py
+++ b/test/hdfs_test.py
@@ -29,6 +29,8 @@ from luigi import six
 from minicluster import MiniClusterTestCase
 from nose.plugins.attrib import attr
 
+from target_test import FileSystemTargetTestMixin
+
 
 class ComplexOldFormat(luigi.format.Format):
     """Should take unicode but output bytes
@@ -272,7 +274,13 @@ class ComplexOldFormatTest(MiniClusterTestCase):
 
 
 @attr('minicluster')
-class HdfsTargetTests(MiniClusterTestCase):
+class HdfsTargetTests(MiniClusterTestCase, FileSystemTargetTestMixin):
+
+    def create_target(self, format=None):
+        target = hdfs.HdfsTarget(self._test_file(), format=format)
+        if target.exists():
+            target.remove(skip_trash=True)
+        return target
 
     def test_slow_exists(self):
         target = hdfs.HdfsTarget(self._test_file())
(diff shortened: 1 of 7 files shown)


Tarrasch:
It's not clear to my why you did the re-factoring. Since there's so much code, I think you could explain it in the commit message. Also squash the typo commit.

test/hdfs_test.py
```diff
@@ -272,7 +274,13 @@ def test_with_write_success(self):
 
 
 @attr('minicluster')
-class HdfsTargetTests(MiniClusterTestCase):
+class HdfsTargetTests(MiniClusterTestCase, BaseTargetTestCase):
```
Tarrasch: I think this won't work actually (because nosetest is buggy). Can you veryify that the attribute is still honored? I mean, does this still *only* run when running tox with either the `cdh` or `hdp` environment and not `nonhdfs`?

Tarrasch:
Aha, I think I see now, but at lease add a read reference to the `BaseTargetTestCase` class documentation in the commit message.

Tarrasch:
Ah, I understand even more now. The "Typo fix" is not amendable. But can you make your commit messages more detailed in general? The code is great but it becomes even easier to read for others of the commit messages are as great too. :)

gpoulin:
@Tarrasch The idea wasn't to reduce the code but to ensure consistency between the Targets. Having the Target pass the `BaseTargetTestCase` ensure a minimal interface. It would be handy also if someone want to create another Target.

I will reword my commits and try to be more descriptive in the future.

erikbern:
LGTM. Does this increase coverage?

To be strict, the `Target` class only requires one method to be implemented – `exists()`. The refactoring really just touches a subclass of `Target` which is `FileSystemTarget` iirc. So I would have a slight preference for calling it `BaseFileSystemTarget` or something like that. Anyway it's really no big deal

gpoulin:
@erikbern It should increase a little at least for s3. However, my main goal was to ensure that the Target handle correctly `str`, `unicode` and `bytes` on all python version.

I'll will change the name of the base class. It totally makes sense.

erikbern:
You don't have to change it, just wanted to point it out. I've seen some confusion about what's a `Target`, just wanted to over-emphasize for maximum clarity

erikbern:
On a separate note, I really appreciate your hard work with Luigi, @gpoulin – excited to see more PR's!

Tarrasch:
Yep thanks @gpoulin. Yea, Please rename the class name now or I'm afraid it's a thing that will be forgotten! :)

Again, you're totally awesome @gpoulin !! :)

test/target_test.py
```diff
@@ -47,3 +53,193 @@ def open(self, mode):
                 return None
 
         GoodTarget()
+
+
+class BaseTargetTestCase():
```
berkerpeksag: () -> (object)

Tarrasch: Good catch!

test/hdfs_test.py
Tarrasch: To clarify what I mean. Please see https://github.com/spotify/luigi/pull/613 :)

Tarrasch: *Tarrasch stares at the ID*. Ok, seems like we've had almost 200 PRs in 50 days. Nice :)

test/target_test.py
gpoulin: I already started to deprecate 2.x... :P. Seriously, thanks @berkerpeksag, I'll fix.

test/hdfs_test.py
gpoulin: I think the nonhdfs travis environment would have failed since we don't install hadoop. But I'll double check to be sure.

gpoulin:
@Tarrasch Added more description to the commit message and renamed `BaseTargetTestCase` to `FileSystemTargetTestMixin`. I also confirm that `tox -e nonhdfs` doesn't run `HdfsTargetTests`.

Tarrasch:
Sounds good. Thanks!

PR #804, opened 2015-03-02
Remove luigi namespace from postgres_test

gpoulin:
Was making all successive test fail if
postgresql was not accessible

### DIFF
--- a/test/postgres_with_server_test.py
+++ b/test/postgres_with_server_test.py
@@ -27,9 +27,6 @@ import luigi
 import luigi.notifications
 from luigi import postgres
 
-luigi.notifications.DEBUG = True
-luigi.namespace('postgres_test')
-
 """
 Typical use cases that should be tested:
 
@@ -173,5 +170,3 @@ class TestPostgresImportTask(TestCase):
         cursor = conn.cursor()
         cursor.execute('select count(*) from {table}'.format(table=clearer.table))
         self.assertEqual(tuple(cursor), ((3,),))
-
-luigi.namespace()


erikbern:
LGTM

Tarrasch:
LGTM, will merge on green

gpoulin:
@erikbern Could task namespace be a decorator instead? I feel the current approach is pretty error prone, it's easy to forget to revert the namespace. I'm not so familiar with metaclass and not sure how easy it would be. The other option would be context_manager but that looks strange to me

```python
with luigi.namespace('fancy_namespace'):
    class MyTask(luigi.Task):
         def run(self):
              fancy_algorithm()
```

Tarrasch:
Decorator makes sense though it will make a lot of indentation. Probably still better than what we have now

erikbern:
Decorators wouldn't require any indentation. Context managers would (so I don't like that approach).

gpoulin:
Context manager seems more appropriate, but I have the same feeling as @erikbern about the extra indentation. Decorator are a little verbose but manageable. Before discovering Luigi, I built a little workflow framework (far simpler than Luigi) and since I didn't want to mess with metaclass I was using decorator to register tasks. I feel doing something like this should be totally reasonable.

```python
@luigi.namespace('fancy_namespace')
class MyTask(luigi.Task):
    def run(self):
        fancy_algorithm()
```

I'll do some tests on this.

erikbern:
Minor remark is that the first version of Luigi (late 2011) used class decorators to register classes. I think those decorators were still around until a few months ago.

erikbern:
The other option is to use a class attribute:

```python
class MyTask(luigi.Task):
    namespace = 'blah'
```

gpoulin:
@erikbern That was my third option and @freider already implemented it. see #805

erikbern:
haha yeah it felt kind of familiar.

PR #805, opened 2015-03-02
Add support for namespace decorator

gpoulin:
luigi.namespace can be use as a decorator to
change the namespace of a single task

Feel free to say that this functionality is redundant with `task_namespace` property. In that case I'll just keep the documentation of this property. (In fact I discover `task_namespace` when doing this change)

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -45,6 +45,15 @@ def namespace(namespace=None):
     is reset, which is recommended to do at the end of any file where the
     namespace is set to avoid unintentionally setting namespace on tasks outside
     of the scope of the current file.
+
+    The namespace of a Task can also be changed by specifying the property
+    ``task_namespace``. This solution has the advantage that the namespace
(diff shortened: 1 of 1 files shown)


erikbern:
In the spirit of Python I would actually prefer if there's one way to do it and all other ways are deprecated

https://www.python.org/dev/peps/pep-0020/

gpoulin:
@erikbern I agree. I just did this PR because it found the `task_namespace` only when I finished to implement decorator. I'll remove the decorator and keep the documentation for `task_namespace`.

erikbern:
LGTM. To be clear I don't have any strong feelings regarding class attribute vs class decorator. Just want one of them to be deprecated and the other one to be documented.

gpoulin:
Class attribute are there since August 2012. Decorator being subjectively nicer it's not a good enough reason to remove class attribute and potentially break stuff.

PR #806, opened 2015-03-02
Add spark-master configuration & PySpark1xJob support for --jars

jthi3rry:


Tarrasch:
LGTM!

luigi/contrib/spark.py
```diff
@@ -378,6 +378,9 @@ def run(self):
         py_files = self.py_files()
         if py_files != []:
             options += ['--py-files', ','.join(py_files)]
+        dependency_jars = self.dependency_jars()
+        if dependency_jars != []:
```
berkerpeksag: Nitpick: I see that `xxx != []` is the current convention in this file (see line 379 for example), but I'd prefer `if dependency_jars:`.

jthi3rry: I agree @berkerpeksag.

There's also room for a bit more refactoring as ATM the run methods in PySpark1xJob and Spark1xJob are mostly duplicated.

I can have a go at it in separate PR?

Tarrasch:
Thanks @jthi3rry! Can you please also fix @berkerpeksag's comment in a new PR? :)

PR #810, opened 2015-03-03
Adds assistant argument to RemoteScheduler.get_work

daveFNbuck:
Workers can't use the remote scheduler at all with the bug this fixes.

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -127,10 +127,10 @@ class RemoteScheduler(Scheduler):
             'params': params,
         })
 
-    def get_work(self, worker, host=None):
+    def get_work(self, worker, assistant=False, host=None):
         return self._request(
             '/api/get_work',
-            {'worker': worker, 'host': host},
+            {'worker': worker, 'host': host, 'assistant': assistant},
             log_exceptions=False,
             attempts=1)
 


erikbern:
Thanks!

daveFNbuck:
No problem. Luigi works with this change. I'm working on making it possible to actually run an assistant now.

erikbern:
Wait so Luigi was broken until this fix? That's not good.

daveFNbuck:
Yeah, there aren't any tests to verify the RemoteScheduler interface. So the tests passed even though get_work was giving an invalid named argument.

erikbern:
Ok – we should add some tests for that. I think the tests might have been deactivated because of some issues with Tornado – remember seeing a comment about that. Might be worth turning them back on

gpoulin:
@erikbern If you are talking with the issue with Travis, we didn't deactivated the test, they just run in a subprocess now. However, coverage don't capture them. There is way around it thou: http://nedbatchelder.com/code/coverage/subprocess.html http://blog.fizyk.net.pl/blog/gathering-tests-coverage-for-subprocesses-in-python.html

For the RemoteScheduler, there is some tests but is pretty limited https://github.com/spotify/luigi/blob/master/test/rpc_test.py. The best will probably to subclass https://github.com/spotify/luigi/blob/6c342f3b7c380d63cbdb53578a8ebff8da0adf32/test/central_planner_test.py#L28

erikbern:
Great thanks. I realized I think we test running a server but we don't ever try to query it for work

PR #811, opened 2015-03-04
Add test for RemoteScheduler

gpoulin:
Test are based on CentralScheduler tests
Add some method to remote scheduler to support more features
Some tests are still failing and should be fixed

This is to address an issue found with #810.

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -23,14 +23,12 @@ See :doc:`/central_scheduler` for more info.
 import json
 import logging
 import time
-try:
-    from urllib import urlencode
-    from urllib2 import urlopen, URLError, Request
-except ImportError:
-    from urllib.parse import urlencode
-    from urllib.request import Request, urlopen
-    from urllib.error import URLError
+from socket import gaierror
 
+from tornado.simple_httpclient import SimpleAsyncHTTPClient
+from tornado.httpclient import HTTPError
+
+from luigi import six
 from luigi import configuration
 from luigi.scheduler import PENDING, Scheduler
 
@@ -58,16 +56,22 @@ class RemoteScheduler(Scheduler):
         if connect_timeout is None:
             connect_timeout = config.getfloat('core', 'rpc-connect-timeout', 10.0)
         self._connect_timeout = connect_timeout
+        self._client = SimpleAsyncHTTPClient()
 
     def _wait(self):
         time.sleep(30)
 
-    def _post(self, url, body):
-        url = 'http://%s:%d%s' % (self._host, self._port, url)
-        return Request(url, body)
+    def _async_fetch(self, full_url, body):
+        return self._client.fetch(
+            full_url,
+            method="POST",
+            body=body,
+            connect_timeout=self._connect_timeout,
+        )
 
     def _fetch(self, url, body, log_exceptions=True, attempts=3):
-        req = self._post(url, body)
+
+        full_url = 'http://%s:%d%s' % (self._host, self._port, url)
         last_exception = None
         attempt = 0
         while attempt < attempts:
@@ -76,9 +80,11 @@ class RemoteScheduler(Scheduler):
                 logger.info("Retrying...")
                 self._wait()  # wait for a bit and retry
             try:
-                response = urlopen(req, None, self._connect_timeout)
+                response = self._client.io_loop.run_sync(
+                    lambda: self._async_fetch(full_url, body)
+                )
                 break
-            except URLError as e:
+            except (HTTPError, gaierror) as e:
                 last_exception = e
                 if log_exceptions:
                     logger.exception("Failed connecting to remote scheduler %r", self._host)
@@ -89,11 +95,11 @@ class RemoteScheduler(Scheduler):
                 (attempts, self._host),
                 last_exception
             )
-        return response.read().decode('utf-8')
+        return response.body.decode('utf-8')
 
     def _request(self, url, data, log_exceptions=True, attempts=3):
         data = {'data': json.dumps(data)}
-        body = urlencode(data).encode('utf-8')
+        body = six.moves.urllib.parse.urlencode(data).encode('utf-8')
 
         page = self._fetch(url, body, log_exceptions, attempts)
         result = json.loads(page)
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -120,7 +120,7 @@ class RemoteScheduler(Scheduler):
             'params': params,
         })
 
-    def get_work(self, worker, assistant=False, host=None):
+    def get_work(self, worker, host=None, assistant=False):
         return self._request(
             '/api/get_work',
             {'worker': worker, 'host': host, 'assistant': assistant},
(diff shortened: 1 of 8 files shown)


luigi/rpc.py
```diff
@@ -110,9 +110,9 @@ def ping(self, worker):
         # just one attemtps, keep-alive thread will keep trying anyway
         self._request('/api/ping', {'worker': worker}, attempts=1)
 
-    def add_task(self, worker, task_id, status=PENDING, runnable=False,
-                 deps=None, new_deps=None, expl=None, resources={}, priority=0,
-                 family='', params={}):
+    def add_task(self, worker, task_id, status=PENDING, runnable=True,
+                 deps=None, new_deps=None, expl=None, resources=None, priority=0,
+                 family='', params=None):
```
gpoulin: I changed the default value here to respect the same values as CentralScheduler. I'm not sure if the two schedulers didn't have the same signature.

Tarrasch:
What does this mean that the tests are still failing with this?

erikbern:
This is great. I really like the approach. We should consider doing something similar with `worker_test.py` as well.

1. I hope this doesn't cause the problem with the unit tests we saw before (I never understood what the issue was)
1b. I remember seeing something about tornado's built-in support for multithreaded unit-tests, should we switch to that? or is it not necessary?
2. Why are the tests failing? Sounds pretty dangerous.

luigi/rpc.py
erikbern: sounds good to me

test/rpc_test.py
```diff
@@ -15,30 +15,74 @@
 # limitations under the License.
 #
 
+import time
 from helpers import unittest
 
 import luigi.rpc
-import server_test
```
erikbern: what's the reason you're not using `server_test` here?

test/central_planner_test.py
```diff
@@ -23,6 +23,7 @@
 
 luigi.notifications.DEBUG = True
 WORKER = 'myworker'
+HOST = 'localhost'
```
erikbern: what's the reason you had to add the `host` argument to all `get_work` calls?

gpoulin:
@erikbern 
1. I think the issue was with the IOloop of tornado. I didn't touch this in this commit so it shouldn't be an issue.

2. I was also unsatisfy with the current. I look at tornado testing and I'm 95% there to use the "native" torando tests. I'll amend this PR

@Tarrasch 
3. [It's](https://github.com/spotify/luigi/pull/811/files#diff-5f5b4f174447b514f6689c13b0ce17cbR68) what I mean by failing test. Some tests are working fine with when using CentralScheduler but fail with RemoteScheduler. So if there is something broken, it's already broken. This PR just find out that it's broken, don't break it.

test/rpc_test.py
gpoulin: I had some difficulties with the sever_test. I'll amend this PR as per my [comment](https://github.com/spotify/luigi/pull/811#issuecomment-77155803).

test/central_planner_test.py
gpoulin: Sincerely... I'm not sure. The RemoteScheduler doesn't work properly without it. At first, I thought it was needed by the RemoteScheduler by design, but looking back at the code, it doesn't seem. I didn't have time to investigate the source of the issue.

Tarrasch:
Good. LGTM

erikbern:
Yes this is excellent. Once you merge this I will extend it and run it also for a couple of other test cases. Essentially every case where we're running local scheduler we might as well run it over RPC as well

Tarrasch:
@gpoulin, so we're waiting for you amend to this PR then. Right?

gpoulin:
Yeah, should be there within an hour or 2

gpoulin:
I forgot to mention in my commit that I removed the fall back to `GET` method for the RemoteScheduler request. There was a comment `TODO: rmove in several weeks` that dated from August. I'll added to my commit message later.

berkerpeksag:
> I was also unsatisfy with the current. I look at tornado testing and I'm 95% there to use the "native" torando tests. I'll amend this PR

Feel free to use my WIP branch to use tornado.testing.AsyncHTTPTestCase in server_test: https://github.com/spotify/luigi/compare/master...berkerpeksag:server-tests

gpoulin:
@berkerpeksag Thanks. However I think I already finish my implementation which doesn't require starting a new thread.

For the test that were failing and the extra HOST argument, it was due to the wrong argument order of get_worker in the RemoteScheduler. #810 Added the assistant argument but not in the same order as the CentralScheduler.

Note that to support `tornado.testing.AsyncHTTPTestCase`, I need to [bypass](https://github.com/spotify/luigi/pull/811/files#diff-9b595b54a300214f2172cde2e9c1eec4R33) luigi.server.run() and to [monkey path](https://github.com/spotify/luigi/pull/811/files#diff-9b595b54a300214f2172cde2e9c1eec4R48) the http request of the remote scheduler. I created extra [tests](https://github.com/spotify/luigi/pull/811/files#diff-9b595b54a300214f2172cde2e9c1eec4R78) to make sure that those are still cover

@Tarrasch @erikbern Travis seems happy. If you are too, it's ready for merging.

FYI: Using AsyncHTTPTestCase instead of subprocess allowed me to easily plug a debugger inside the server which I used to find the bug. It also allows to support CentralPlanner tests that [monkey patch](https://github.com/spotify/luigi/blob/master/test/central_planner_test.py#L44) `time.time`.

test/server_test.py
```diff
@@ -14,55 +14,54 @@
...
+
+        def _fetch(obj, url, body, *args, **kwargs):
+            response = self.fetch(url, body=body, method='POST')
+            if response.code >= 400:
+                raise luigi.rpc.RPCError(
+                    'Errror when connecting to remote scheduler'
+                )
+            return response.body.decode('utf-8')
+
+        luigi.rpc.RemoteScheduler._fetch = _fetch
```
erikbern: why do you need to patch `_fetch`?

erikbern:
Great catch with the failing tests. Not sure how I ended up screwing that up. Hopefully going forward tests will break if it happens

erikbern:
LGTM

test/server_test.py
gpoulin: `tornado.testing.AsyncHTTPTestCase` doesn't start a new thread but use the tornado IOLoop to switch context between the client and the server. When calling `self.fetch`, the client is waiting for IO and give away the execution context to the server which can then handle the request and response. `urllib` doesn't work with the tornado IOLoop. So when making a request, it blocks and never give the execution context to the server which cannot reply (we are stuck in a pseudo dead-lock).

erikbern: ok – makes sense. I wonder if using `tornado.httpclient` would resolve it. Anyway doesn't matter.

gpoulin: Using the AsyncHTTPClient should work. It's what do `self.fecth`. It just requires an IOLoop running. The client automatically start one if there isn't. I thought about it, but I thought it would be relatively complex, but it seems simple. I can try to modify RemoteScheduler to use AsyncHTTPClient.

Tarrasch:
@gpoulin  Let me know how your experimentation with AsyncHTTPClient goes, I'm in overall happy to merge this whenever you feel ready! :)

gpoulin:
@Tarrasch It's working!!! Apart if you have comments I won't touch this PR.

erikbern:
this looks awesome, let's merge

PR #812, opened 2015-03-04
Added SparkSubmitTask and deprecated SparkJob, Spark1xJob and PySpark1xJob

jthi3rry:
@Tarrasch, @berkerpeksag, this is my follow up PR from comments in https://github.com/spotify/luigi/pull/806.

I deprecated both Spark1xJob and PySpark1xJob and created a single SparkSubmitTask that follows the command usage and can handle java, scala and python apps using spark standalone, yarn or mesos. I'm wondering whether SparkJob should be deprecated too.

Happy to discuss and make any changes.

Thanks

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -446,7 +446,78 @@ worker-disconnect-delay
 [spark]
 -------
 
-Parameters controlling the running of Spark jobs
+Parameters controlling the default execution of ``SparkSubmitTask``:
+
+.. deprecated:: 1.1.1
+   ``SparkJob``, ``Spark1xJob`` and ``PySpark1xJob`` are deprecated. Please use ``SparkSubmitTask``.
+
+spark-submit
+  Command to run in order to submit spark jobs. Default: spark-submit
+
+master
+  Master url to use for spark-submit. Example: local[*]. Default: Spark default (Prior to 1.1.0: yarn-client)
+
+deploy-mode
+    Whether to launch the driver programs locally ("client") or on one of the worker machines inside the cluster ("cluster"). Default: Spark default
+
+jars
+    Comma-separated list of local jars to include on the driver and executor classpaths. Default: Spark default
+
+py-files
+    Comma-separated list of .zip, .egg, or .py files to place on the PYTHONPATH for Python apps. Default: Spark default
+
+files
+    Comma-separated list of files to be placed in the working directory of each executor. Default: Spark default
+
+conf:
+    Arbitrary Spark configuration property in the form Prop=Value|Prop2=Value2. Default: Spark default
+
+properties-file
+    Path to a file from which to load extra properties. Default: Spark default
+
+driver-memory
+    Memory for driver (e.g. 1000M, 2G). Default: Spark default
+
+driver-java-options
+    Extra Java options to pass to the driver. Default: Spark default
+
+driver-library-path
+    Extra library path entries to pass to the driver. Default: Spark default
+
+driver-class-path
+    Extra class path entries to pass to the driver. Default: Spark default
+
+executor-memory
+    Memory per executor (e.g. 1000M, 2G). Default: Spark default
+
+*Configuration for Spark submit jobs on Spark standalone with cluster deploy mode only:*
+
+driver-cores
+    Cores for driver. Default: Spark default
+
+supervise
+    If given, restarts the driver on failure. Default: Spark default
+
+*Configuration for Spark submit jobs on Spark standalone and Mesos only:*
+
+total-executor-cores
+    Total cores for all executors. Default: Spark default
+
+*Configuration for Spark submit jobs on YARN only:*
+
+executor-cores
+    Number of cores per executor. Default: Spark default
+
+queue
+    The YARN queue to submit to. Default: Spark default
+
+num-executors
+    Number of executors to launch. Default: Spark default
+
+archives
+    Comma separated list of archives to be extracted into the working directory of each executor. Default: Spark default
+
+*Parameters controlling the execution of SparkJob jobs (deprecated):*
 
 spark-jar
   Location of the spark jar. Sets SPARK_JAR environment variable when
@@ -459,12 +530,6 @@ hadoop-conf-dir
 
 spark-class
   Location of script to invoke. Example: /usr/share/spark/spark-class
-
-spark-submit
-  Command to run in order to submit spark jobs. Default: spark-submit
-
-spark-master
-  Master url to use for spark-submit. Example: local[*]. Default: yarn-client
 
 
 [task_history]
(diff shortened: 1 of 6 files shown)


luigi/contrib/spark.py
```diff
@@ -226,76 +215,160 @@ def track_progress(self, proc):
...
-class Spark1xJob(luigi.Task):
-
-    num_executors = None
+class SparkSubmitTask(luigi.Task):
+    """
+    Template task for running a spark job
+
+    Strictly follows spark-submit usage::
+
+        Usage: spark-submit [options] <app jar | python file> [app options]
```
berkerpeksag: It would be better to just say something like "See ``spark-submit -h`` for more information.". Or just add documentation link of the spark-submit script.

luigi/contrib/spark.py
```diff
@@ -347,53 +419,81 @@ def track_progress(self, proc):
...
+    @property
+    def entry_class(self):
+        return self.job_class()
+
+    @property
+    def app(self):
+        return self.jar()
+
+    def run(self):
+        warnings.warn("The use of Spark1xJob is deprecated. Please use SparkSubmitTask.")
```
berkerpeksag: needs `stacklevel=2`

luigi/contrib/spark.py
```diff
@@ -347,53 +419,81 @@ def track_progress(self, proc):
...
-                                stderr=temp_stderr, env=env, close_fds=True)
-        return_code, final_state, app_id = self.track_progress(proc)
-        if final_state == 'FAILED':
-            raise SparkJobError('Spark job failed: see yarn logs for %s', app_id)
-        elif return_code != 0:
-            temp_stderr.seek(0)
-            errors = "".join((x.decode('utf8') for x in temp_stderr.readlines()))
-            logger.error(errors)
-            raise SparkJobError('Spark job failed', err=errors)
+        warnings.warn("The use of PySpark1xJob is deprecated. Please use SparkSubmitTask.")
```
berkerpeksag: needs `stacklevel=2`

test/contrib/spark_test.py
```diff
@@ -350,3 +347,124 @@ def communicate(self):
...
+        return ["arg1", "arg2"]
+
+    def output(self):
+        return luigi.LocalTarget('output')
+
+
+class SparkSubmitTest(unittest.TestCase):
+    ss = 'ss-stub'
+
+    def setUp(self):
```
berkerpeksag: Since you subclass `unittest.TestCase`, there is no need to add this code. Same thing for tearDown, too.

doc/configuration.rst
```diff
@@ -464,7 +464,9 @@ spark-submit
   Command to run in order to submit spark jobs. Default: spark-submit
 
 spark-master
-  Master url to use for spark-submit. Example: local[*]. Default: yarn-client
+  Master url to use for spark-submit. Example: local[*]. Default: See your Spark configuration
+
+  The deprecated Spark1xJob and PySparl1xJob use yarn-client by default
```
berkerpeksag: Using a deprecated directive would be good here:

```rst
.. deprecated:: 1.0.25
   ``Spark1xJob`` and ``PySparl1xJob`` classes were deprecated. [...]
```

(not sure about the next version number. @erikbern or @Tarrasch know better than me :))

Tarrasch: Next version will be 1.1.0 I think

erikbern:
Looks great but on a more general note I think it would be awesome if we can support inline Python code like in `luigi.hadoop`.

jthi3rry:
Thanks for the review. Updated PR contains:

* Incorporation of @berkerpeksag's comments and more global configuration options for the class and in the docs.
* Deprecation warnings using version 1.1.0. @Tarrasch, let me know if that changes and I'll update.
* Deprecation of ``SparkJob`` too as it seemed very YARN specific. Does that sound alright?

@erikbern, that would be awesome as an additional feature. A simple support for spark submit is necessary though to be able to use existing python drivers and java or scala drivers.

That could be introduced as a subclass (``PySparkTask``?) with a ``main($sc)`` abstract method where $sc is a given spark context (like when using the pyspark shell). I'll have a look at ``luigi.hadoop`` to see how it's done. That's a good one for another PR :-)

jthi3rry:
@erikbern, added ``PySparkTask`` to run spark jobs inline.

Usage looks like this:
```
class MyJob(PySparkTask):
    param = IntParameter(default=2)

    def output(self):
        # return a target accessible from the spark cluster (HDFS, S3, ...)

    def main(self, sc, *args):
        exponent = self.param
        output = self.ouput().path
        sc.parallelize([1, 2, 3]).map(lambda x: x ** exponent).saveAsTextFile(output)

```

erikbern:
That's really cool – can you map over an input source as well?

erikbern:
If you want to, feel free to leave out the last commit for a separate PR. That way we can merge what we have so far

jthi3rry:
I removed the last commit and squashed the existing ones together.

Yes, the input path can be used as a source by doing something like ``sc.textFile(self.input().path)``

luigi/contrib/spark.py
```diff
@@ -99,7 +83,247 @@ def __str__(self):
...
+        if not self.app:
+            raise NotImplementedError("subclass should define an app (.jar or .py file)")
+        return [self.app] + self.app_options()
+
+    def run(self):
+        args = map(str, self.spark_command() + self.app_command())
+        env = os.environ.copy()
+        temp_stderr = tempfile.TemporaryFile()
+        logger.info('Running: %s', repr(args))
+        proc = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=temp_stderr, env=env, close_fds=True)
```
gpoulin2: If stdout and stderr will always be text, you can open the process with `universal_newline=True`. The process will accept unicode as stdin and return unicode.

Also, any reason for `env=env`. According to my understanding of the doc, it's the default behavior.

jthi3rry: Added ``universal_newline=True`` to https://github.com/spotify/luigi/pull/837.

https://github.com/spotify/luigi/pull/837 adds a get_environment() that alters the env before passing it to ``Popen``.

erikbern:
Something seems weird with the delta now. Can you rebase on master?

jthi3rry:
Done @erikbern, does that fix it?

erikbern:
Yes looks good. Is this ready to be merged? How backwards-incompatible is it? Happy to merge just want to make sure there's no risk

jthi3rry:
It should be fully backwards compatible, all existing unit tests pass, SparkJob is unchanged, (Py)Spark1xJob's have more features now as they extend SparkSubmitTask but their behaviour is unchanged.

PR #813, opened 2015-03-04
Read record_task_history parameter in scheduler.scheduler

neilisaac:
We're trying to do something like this:
```
class MyJob(luigi.task):
    def complete(self):
        scheduler = luigi.scheduler.CentralPlannerScheduler()

        history = scheduler.task_history
        if isinstance(history, luigi.task_history.NopHistory):
            return False

        params = {k: str(getattr(self, k)) for k, _ in self.get_params()}
        tasks = history.find_all_by_parameters(self.task_family, **params)
        return any(any(e.event_name == "DONE" for e in t.events) for t in tasks)
```

When we run `luigi.scheduler.CentralPlannerScheduler()` from ipython, it uses `DbTaskHistory` however when complete() is invoked via the worker, it uses `NopHistory`.

Perhaps there's a nice way to build in this sort of functionality too.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -99,7 +99,8 @@ class scheduler(Config):
     max_shown_tasks = parameter.IntParameter(default=100000,
                                              config_path=dict(section='scheduler', name='max-shown-task'))
 
-    record_task_history = parameter.BoolParameter(default=False)
+    record_task_history = parameter.BoolParameter(default=False,
+                                                  config_path=dict(section='scheduler', name='record_task_history'))
 
 
 def fix_time(x):


erikbern:
Commit looks good to me.

I'm not sure about your code... why can't you just use a dummy file on disk instead?

neilisaac:
It mimics one of the patterns we commonly used before we started porting jobs to luigi.  This came about for jobs which don't have file outputs (ex. batch updating model parameters in couchbase.)  It feels cleaner to use the existing history data, avoiding the need to set up new folders on a distributed filesystem for each job and permutation of job arguments.

erikbern:
Ok. I guess generalizing that a bit would be to factor it out into a separate `Target` class that takes a `Task` as an argument, but it doesn't really matter too much.

erikbern:
`config_path` is deprecated. But I can fix in a separate commit

PR #817, opened 2015-03-05
Makes assistants reasonably runnable

daveFNbuck:
This contains commits that allow the assistant to be run on the command-line, stay alive while waiting for new jobs to be scheduled, and have the scheduler properly recognize that the assistant is running jobs rather than marking them as failed. This provides basic functionality to run the assistant, but it doesn't tend to last long for me before it gets a job type it doesn't know about.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -623,7 +623,6 @@ class CentralPlannerScheduler(Scheduler):
         # Return remaining tasks that have no FAILED descendents
         self.update(worker, {'host': host})
         best_task = None
-        best_task_id = None
         locally_pending_tasks = 0
         running_tasks = []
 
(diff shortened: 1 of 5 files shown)


Tarrasch:
Cool stuff! Excited to hear reviews from the community. I commited in your individual commits. :)

erikbern:
This looks great!

Note that `interface.load_task` can load modules dynamically. If we keep track of the module of all tasks, and return it to the worker/assistant, we should be able to load modules dynamically and execute anything that's available to the assistant

daveFNbuck:
It would be great if we added the module tracking, as my assistant currently fails fairly quickly because it doesn't load everything thoroughly enough. I'm working on making that report a failure (which is still necessary) but that would be a better fix for my actual problem. I'll dig into that when I'm done with the failures.

luigi/scheduler.py
```diff
@@ -224,6 +224,9 @@ def prune(self, config):
         if self.last_active + config.worker_disconnect_delay < time.time():
             return True
 
+    def assistant(self):
```
berkerpeksag: This could be a property. `worker.assistant` instead of `worker.assistant()` looks better to me :)

erikbern:
LGTM!

luigi/scheduler.py
Tarrasch: I still haven't learned about the best practive for when to use properties in python or not. Hmm ...

daveFNbuck: It seems reasonable here, as that gives us the option to make assistant a member variable in the future.

PR #818, opened 2015-03-05
Fails assistant when it can't determine the task type

daveFNbuck:
When the assistant can't create a task for a given family, it should not just
die. This catches the errors caused by unknown task families and logs messaeges
rather than just dying. The scheduler is informed of the failure and the
assistant can continue requesting other jobs.

As a future improvement, the scheduler could also use the failure report to
avoid giving the assistant these types of jobs.


erikbern:
LGTM!

Tarrasch:
LGTM

PR #819, opened 2015-03-06
Passes module information in get_work response for assistants

daveFNbuck:
This allows assistants to run jobs that they have the code for but have not
imported the module. This will unfortunately not work with jobs that were in
the __main__ module of the worker that first added them.

### DIFF
--- a/dummy_test_module/not_imported.py
+++ b/dummy_test_module/not_imported.py
@@ -1,0 +1,6 @@
+import luigi
+
+
+class UnimportedTask(luigi.Task):
+    def complete(self):
+        return False
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -111,7 +111,7 @@ class RemoteScheduler(Scheduler):
 
     def add_task(self, worker, task_id, status=PENDING, runnable=True,
                  deps=None, new_deps=None, expl=None, resources=None, priority=0,
-                 family='', params=None):
+                 family='', module=None, params=None):
         self._request('/api/add_task', {
             'task_id': task_id,
             'worker': worker,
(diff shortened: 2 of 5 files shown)


daveFNbuck:
I'm having a little trouble getting this test to work, as I need a module that hasn't been imported. I tried throwing a file into test, but because test isn't a module, that doesn't work with the way travis runs the tests. The only other option I can think of is to write to a new file in the luigi module during the test, but that seems super sketchy. Better ideas? Maybe this should not have a test? Or should we just test that the right arguments get passed around?

gpoulin:
It's also an issue for other tests. I tried before to make `test` a package by adding a `__init__.py` file, however, it causes issue with task name resolution. I would suggest to whether solve the tasks ambiguity in the test folder and make this folder a package or to create a dummy package eg.: `luigi_test_utils`

erikbern:
That's a good question actually. What if you run with multiple workers and you require tasks that are imported in the local scope?

daveFNbuck:
@erikbern That sounds like a good alternative to how I implemented the test. It still leaves the question of where does the task that my worker hasn't imported live? @gpoulin's suggestion of luigi_test_utils sounds like a reasonable solution.

erikbern:
yes that makes sense. I realized I actually built one alraedy: https://github.com/spotify/luigi/blob/master/test/foo_module.py

daveFNbuck:
The problem here is that the test runner doesn't see foo_module because it's nested under test, which does not have an __init__.py.

erikbern:
That's true. Feel free to move it somewhere else and/or create an ``__init__.py`` and/or create a new dummy module

daveFNbuck:
Got it working with a dummy module and updated the test to exercise more of the code by adding a modified task. Should be good to merge now.

PR #823, opened 2015-03-06
Make sure to kill tornado process at the end of a test

gpoulin:
I saw Travis hanging on bin_test.py. Hopefully that would make sure it doesn't happen again. Might need better way to kill the daemon, thou.


erikbern:
Looks great, thanks

PR #827, opened 2015-03-07
Enable coverage for subprocess and shutdown sever gently

gpoulin:
Test (don't merge yet) to see if we get a better code coverage. Currently some part of the code is reported as uncover because it is run within a sub process.

### DIFF
--- a/luigi/server.py
+++ b/luigi/server.py
@@ -27,6 +27,7 @@ import mimetypes
 import os
 import posixpath
 import signal
+import sys
 
 import pkg_resources
 import tornado.httpclient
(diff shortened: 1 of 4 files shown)


gpoulin:
Those modification allow to have the right coverage from `coverage.py`, even for luigi/mrrunnner.py that is run on the minicluster. However, there seems to be some using with the way coveralls-python export the coverage.

gpoulin:
With this PR the coverage should now be properly reported in coveralls.

luigi/server.py
```diff
@@ -168,10 +169,11 @@ def run(api_port=8082, address=None, scheduler=None, responder=None):
     pruner = tornado.ioloop.PeriodicCallback(scheduler.prune, 60000)
     pruner.start()
 
-    def shutdown_handler(foo=None, bar=None):
+    def shutdown_handler(sig_num=None, trace=None):
         logger.info("Scheduler instance shutting down")
         scheduler.dump()
-        os._exit(0)
+        stop()
+        sys.exit(0)
```
gpoulin: I'm sure if there was a reason to use `os._exit(0)` instead of the more gently `sys.exit(0)`, but it prevents coverage to write coverage file.

gpoulin:
@erikbern it might interest you:

The way the sever was shutdown was an issue to get a proper coverage. But I think it was also the source of the false positive in Travis. The `shutdown_handler` was registered with the `atexit` and `shutdown_handler` was calling `os._exit(0)`. So nose was always exiting with `0`.

erikbern:
Oh that's dangerous. Makes sense. So the server thread shut down the entire process?

Is this ready to be merged?

gpoulin:
The thread shouldn't shutdown the entire process because `shutdown_handler` will not be called within the thread. However, `shutdown_handler` was call at the end of the process during the cleanup part. By calling `os._exit(0)`, it prevents any further clean up (close socket, flush buffer...) and force the return code to be 0 even if there was an error.

This PR is ready to be merged.

erikbern:
84% coverage now, yeah!

Issue #822, opened 2015-03-06
How is Luigi Parameter name decided?


soxofaan:
Take this POC code, which defines a single parameter in a base class, to be reused in various subclasses:
```python
class Thing(luigi.Task):

    color = luigi.Parameter(default='fluffypink')

    has_run = False
    def run(self):
        print self.__class__.__name__, self.color
        self.has_run = True

    def complete(self):
        return self.has_run

class AliveThing(Thing): pass

class Animal(AliveThing): pass
 
class Mammal(Animal): pass

class Horse(Mammal): pass

class Cow(Mammal): pass
```

How is the parameter name for the color parameter decided?
If I run this with v1.0.22, --help tells me this:
```
--Animal-color ANIMAL_COLOR
                        Animal.color [default: fluffypink]
```

which seems a pretty arbitrary decision .
I think  `--Thing-color` would be the most sensible

erikbern:
Interesting – it just picks an arbitrary one? It should either be all of them, or the base class

soxofaan:
Great, thanks.

But now I'm even more confused.
Apparently the command line option `--Thing-color` doesn't work (its value is not passed as parameter value to the task)
If I use `--color`, then it works, but that option is not documented under `--help`

(I tried this with latest master at 67912aba66c2e890b5f524832853ac4f7758f36c)

How is the command line interface supposed to work?
We're considering upgrading from 1.0.19, but are held back by the (unclarity of the ) changes of the command line interface.
I tried to dive in to the command line parsing code, but that part of the code flows in pretty complex ways :)

erikbern:
Let me try to reproduce.

`--MyClass-my-param` style parameters weren't present in 1.0.19 though so I'm not sure what kind of compatibility issues that would present for you. They were introduced in `1.0.21` IIRC.

soxofaan:
To make my previous comment a bit more concrete:

If I have this:
```python
class Foo(luigi.Task):
    color = luigi.Parameter(default='fluffypink')
```

Then both of these commands trigger the `Foo(color=blue)` task:
```
script.py Foo --Foo-color blue
script.py Foo --color blue
```

If I ad just one subclass, like this:
```python
class Foo(luigi.Task):
    color = luigi.Parameter(default='fluffypink')

class Bar(Foo):
    pass
```
Then `script.py Foo --Foo-color blue` Triggers `Foo(color=fluffypink)`
and `script.py Foo --color blue` triggers `Foo(color-blue)`

[...]

PR #828, opened 2015-03-07
Fix for #822

erikbern:
The solution is to let `--MyClass-my-param` only be exposed for params set on the base class. If any class inherits from MyClass, the param name will not change (or be duplicated). Eg. if there will be no `--MySubClass-my-param` option if MySubClass inherits from MyClass

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -29,6 +29,11 @@ import logging
 import traceback
 import warnings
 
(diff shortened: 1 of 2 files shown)


erikbern:
Tests are flaky on Travis but not locally. I'll see if I can make my implementation more watertight

PR #830, opened 2015-03-08
Activate py3k test that are now fixed

gpoulin:
Those succeed now beacuse:
 - release of python 3.4.3
 - snakebite 2.5.2
 - new way of testing webserver

### DIFF
--- a/test/redshift_test.py
+++ b/test/redshift_test.py
@@ -29,7 +29,7 @@ from boto.s3.key import Key
 from luigi.s3 import S3Client
 
 
-if sys.version_info[:2] == (3, 4):
+if (3, 4, 0) <= sys.version_info[:3] < (3, 4, 3):
     # spulec/moto#308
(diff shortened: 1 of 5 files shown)


test/redshift_test.py
```diff
@@ -29,7 +29,7 @@
 from luigi.s3 import S3Client
 
 
-if sys.version_info[:2] == (3, 4):
+if (3, 4, 0) <= sys.version_info[:3] < (3, 4, 3):
```
berkerpeksag: Out of curiosity, why it doesn't work on < Python 3.4.3?

gpoulin: Backward incompatibility in python 3.4 that was fixed in 3.4.3 http://bugs.python.org/issue20386.

berkerpeksag: Oh, I remember that one. Thanks for the link!

PR #831, opened 2015-03-08
Run minicluster setup once in the primary nose process

erikbern:
Trying this to see if it speeds up Travis builds


erikbern:
CDH tests are still ridiculously slow, abandoning this

PR #836, opened 2015-03-09
Fixed additional issues popping up in #822

erikbern:


erikbern:
There's a failing test that I commented out that checks the output of `--help`. I'll see what I can do about it later.

Issue #838, opened 2015-03-09
Improve packaging of "entrypoints"


soxofaan:
Hi, we use buildout to assemble our python environment and have luigi as a dependency. Currently there is no clean way I know of to get the `luigid` "script" installed this way.

Apparently this is because the scripts like `luigid` are specified in setup.py like:
```
setup(
    ...
    scripts=[
        'bin/luigid',
        'bin/luigi'
    ],
```

instead of 
```
setup(
    ...
    entry_points={
        'console_scripts': [
            'luigi = luigi.interface:luigi_main',
            'luigid = luigi.server:luigid_main',
    }
```

soxofaan:
#839 has been merged

PR #839, opened 2015-03-09
setup.py: improve packaging of 'bin' scripts with 'entry_points/console_scripts'

soxofaan:
see #838 

with this setup.py change, the "scripts" under `bin` are automatically generated/managed and cross platform and more package friendly way

I tried to move the original logic from the `bin` folder to sensible places under the `luigi` source folder.
I'd welcome other suggestions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ cache:
   - $HOME/.pip-cache
 
 install:
+  - pip install .
   - pip install tox
 
 before_script:
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,3 @@
-include bin/luigid
 include README.rst
 include LICENSE
 include examples/*.py
--- a/bin/luigi
+++ b/bin/luigi
@@ -1,6 +1,14 @@
 #!/usr/bin/env python
 
-import luigi
+import sys
+import warnings
+import luigi.cmdline
+
+
+def main(argv):
+    warnings.warn("'bin/luigi' has moved to console script 'luigi'", DeprecationWarning)
+    luigi.cmdline.luigi_run(argv)
+
 
 if __name__ == '__main__':
-    luigi.run(use_dynamic_argparse=True)
+    main(sys.argv[1:])
--- a/bin/luigid
+++ b/bin/luigid
@@ -1,40 +1,14 @@
 #!/usr/bin/env python
 
-import luigi.server
-import luigi.process
-import os
-import luigi.configuration
-import optparse
-import logging
 import sys
+import warnings
+import luigi.cmdline
+
 
 def main(argv):
-    parser = optparse.OptionParser()
-    parser.add_option(u'--background', help=u'Run in background mode', action='store_true')
-    parser.add_option(u'--pidfile', help=u'Write pidfile')
-    parser.add_option(u'--logdir', help=u'log directory')
-    parser.add_option(u'--state-path', help=u'Pickled state file')
-    parser.add_option(u'--address', help=u'Listening interface')
-    parser.add_option(u'--port', default=8082, help=u'Listening port')
-        
-    opts, args = parser.parse_args(argv)
+    warnings.warn("'bin/luigid' has moved to console script 'luigid'", DeprecationWarning)
+    luigi.cmdline.luigid(argv)
 
-    if opts.state_path:
-        config = luigi.configuration.get_config()
-        config.set('scheduler', 'state-path', opts.state_path)
-
-    if opts.background:
-        # daemonize sets up logging to spooled log files
-        logging.getLogger().setLevel(logging.INFO)
-        luigi.process.daemonize(luigi.server.run, api_port=opts.port,
-                                address=opts.address, pidfile=opts.pidfile,
-                                logdir=opts.logdir)
-    else:
-        if opts.logdir:
-            logging.basicConfig(level=logging.INFO, format=luigi.process.get_log_format(), filename=os.path.join(opts.logdir, "luigi-server.log"))
-        else:
-            logging.basicConfig(level=logging.INFO, format=luigi.process.get_log_format())
-        luigi.server.run(api_port=opts.port, address=opts.address)
 
 if __name__ == '__main__':
     main(sys.argv[1:])
--- a/doc/central_scheduler.rst
+++ b/doc/central_scheduler.rst
@@ -24,7 +24,7 @@ To run the server as a daemon run:
 
 ::
 
-    PYTHONPATH=. python bin/luigid --background --pidfile <PATH_TO_PIDFILE> --logdir <PATH_TO_LOGDIR> --state-path <PATH_TO_STATEFILE>
+    luigid --background --pidfile <PATH_TO_PIDFILE> --logdir <PATH_TO_LOGDIR> --state-path <PATH_TO_STATEFILE>
 
 Note that this requires ``python-daemon``.
 By default, the server starts on port ``8082``
--- a/doc/example_top_artists.rst
+++ b/doc/example_top_artists.rst
@@ -238,7 +238,7 @@ If you run
 
 ::
 
-    PYTHONPATH=. python bin/luigid
+    luigid
 
 in the background and then run
 
--- a/luigi/cmdline.py
+++ b/luigi/cmdline.py
@@ -1,0 +1,43 @@
+import os
+import optparse
+import logging
+import sys
+
+import luigi.server
+import luigi.process
+import luigi.configuration
+import luigi.interface
+
+
+def luigi_run(argv=sys.argv[1:]):
+    luigi.interface.run(argv, use_dynamic_argparse=True)
+
+
+def luigid(argv=sys.argv[1:]):
+    parser = optparse.OptionParser()
+    parser.add_option(u'--background', help=u'Run in background mode', action='store_true')
+    parser.add_option(u'--pidfile', help=u'Write pidfile')
+    parser.add_option(u'--logdir', help=u'log directory')
+    parser.add_option(u'--state-path', help=u'Pickled state file')
+    parser.add_option(u'--address', help=u'Listening interface')
+    parser.add_option(u'--port', default=8082, help=u'Listening port')
+
+    opts, args = parser.parse_args(argv)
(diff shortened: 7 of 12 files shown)


erikbern:
I'm guessing unit tests will fail since ./bin/luigi doesn't exist

erikbern:
I think this looks great (despite my comments sounding a bit grumpy).

I hope the entry points work on Python 2.6. If we can get the unit tests working (and preserve backwards compatibility) I'm all for merging this

berkerpeksag:
That's great, but I'd prefer all entry points code in a separate module(e.g. luigi/cmdline.py). See https://github.com/hylang/hy/blob/master/hy/cmdline.py for example.

> I hope the entry points work on Python 2.6.

It will work, but you'll need to install setuptools first.

erikbern:
Starting to look good. The unit tests that are failing now are the ones that invoke `bin/luigi`, `/bin/luigid` etc. For backwards compatibility, I suggest you keep those entry points, but remove the code in them and just do a dummy `from luigi.tools.cmdline import luigid` or something similar. Ideally you should also make sure the unit tests run with the new executables.

Hope this makes sense!

gpoulin:
I wouldn't deprecate the `bin` folder. In fact I would replace the script within it with something like this. 

```python
import os
import sys

luigi_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..')
sys.path.append(luigi_dir)

from luigi.tools.cmdline import luigid

luigid.main()
```

I find it neat to be able to do something like this
```bash
git clone git@github.com:spotify/luigi.git
luigi/bin/luigid
```

erikbern:
Thanks @gpoulin for the code sample – that's exactly what I meant.

I think it might be worth considering deprecating `bin/luigid` though since it creates two entry points to the same thing – we could just make sure that `luigi.tools.luigid` is runnable. But it's just a minor preference

berkerpeksag:
From the user point of view, there will be no difference. They still can run `luigi` and `luigid` directly. This is actually an implementation change. I wouldn't bother to deprecate `bin/*` since it's an implementation detail of the old distutils machinery and users should not call `python bin/luigid.py` directly.

gpoulin:
I usually like the possibility to launch python program without having to "install" them. I'm not sure to see what's wrong with

```
export PYTHONPATH=/some/path/to/testing/luigi
bin/luigid
```

berkerpeksag:
Well, most of time that command won't work. For example, you need to install dependencies and probably need to create a virtualenv to avoid messing with PYTHONPATH etc.

The more Pythonic and reliable way to launch a Python program without installing it is:

```sh
$ python3.4 -m venv venv
$ . venv/bin/activate
$ pip install -e .
$ luigid
```

gpoulin:
I agree that virtualenv is more reliable. However, my technique work 99% of the time and is fast for small test. I wouldn't use it for prod environment, but I wouldn't be surprised if there is some that do it. In addition, pyvenv doesn't install pip correctly when using `--system-site-packages`. And, no I don't want to rebuild numpy/scipy/pandas/scikit-learn ... just for the fun of it.

soxofaan:
Hi all, thanks for the feedback

- I moved bin/luigi and bin/luigid to luigi.cmdline as suggested. Note that I had to use `luigi_run()` as function name to not overwrite `luigi`, the imported module variable (which might be an indication that putting all entry points in a dedicated module is not an ideal solution either)
- I left luigi-grep and deps under luigi.tools, I'm not really sure how "official" or polished these are
- I've re-added small (deprecated) wrappers at bin/luigi and bin/luigi. But without the sys.path fiddling: it's not very clean and experience taught me you're only shooting yourself in the foot in the long run. I agree with @berkerpeksag that just doing git clone + running bin/luigid is not "the right thing" as it does not cover dependency management. Moreover, if I'm not mistaken, you don't necessarily need pip either, just `python setup.py install` should work too, no?
- I've also updated the documentation (central_scheduler.rst/example_top_artists.rst) about launching luigid a bit
- concerning the testing, I only added a test about running luigi as `luigi` (based on the test that runs it as `bin/luigi`)

erikbern:
Looks good!

Minor note: The word "cmdline" is a bit confusing in the Luigi context because there's a lot of code dealing with argument parsing etc. eg. test/cmdline_test.py. I should probably rename that test at some point

erikbern:
Also obviously you should fix the breaking unit test and squash your commits

Do you have issues running the unit tests locally? I've had some issues on and off (relying more and more on Travis, which is a bit frustrating)

.travis.yml
```diff
@@ -27,6 +27,7 @@ cache:
   - $HOME/.pip-cache
 
 install:
+  - pip install .
```
gpoulin: That shouldn't be install in the tox environment, or I'm missing something?

gpoulin:
Ok, ok, I give up :P. However, keep `bin/*` with deprecation. I suspect that there is some users that expects it, particularly that it was in the doc.

@erikbern what issue do you have with the unit test? The only issue that have is with python 2.6 since it was remove from Gentoo repo, however it's not an issue with luigi. I tried to use conda, but it doesn't seems to play nice with tox (I didn't try that hard thou). Other than that, the minicluster is sloooooow................ and when switching python version, the tox environment is overwritten, for `cdh` and `hdp` it means redownloading hadoop. I think the last point could be solved by handling python version the same way as snakebite ([tox](https://github.com/spotify/snakebite/blob/master/tox.ini), [travis](https://github.com/spotify/snakebite/blob/master/.travis.yml)).

gpoulin:
The issue with the unit tests is that the server start with the wrong option when we try to import luigid script (I don't there is any user that try to do that). Modify the `bin/luigid` by something like this and it should work.

```python
import warnings
import luigi.cmdline

def main(*args, **kwargs):
    warnings.warn("'bin/luigid' has moved to console script 'luigid'", DeprecationWarning)
    luigi.cmdline.luigid(*args, **kwargs)

if __name__ == '__main__':
    main()
```

That's pretty dummy but it's to be removed anyway.

.travis.yml
soxofaan: that is for installing the console scripts in the environment, so the tests can execute `luigid` (instead of `./bin/luigid`)

berkerpeksag: @gpoulin is right, `usedevelop = True` (it's equivalent of `pip install -e .`) in tox.ini should do the trick.

soxofaan:
- I rebased+squashed now on latest master
- tried to figure out the test failures from local test run. There are still some failures, but these seem not related to this issue at first sight, I'd first like to see what travis thinks

- > Minor note: The word "cmdline" is a bit confusing in the Luigi context

  would you prefer 'luigi.cli'? I

erikbern:
@gpoulin I've had some random issues running unit tests on my laptop, usually "too many files" and stuff like that. 

On a separate note, I don't know why Travis is ridiculously slow for the hadoop tests, but I wonder if it does something stupid like starting hadoop for every test.

Overall this looks great now – I re-triggered the Travis build because it was just failing on docs linkcheck

gpoulin:
@erikbern It's not failing because of the links but because luigi-grep is not python 3 compatible (I didn't tried this before). @soxofaan can you add the necessary parenthesis to the print statement? There will still be issues, I will solve them in another PR.

For the hadoop job, it's also slow on my local machine. I profiled the tests and `setUp` and `tearDown`take times but it's ~10%. Every call to hadoop seems to take ~2s, even for trivial call like `mkdir` and `stat`.

erikbern:
Oh ok, sorry @gpoulin – I'm so used to the "docs" test being flaky because of linkcheck that I didn't even bother looking – I just retriggered it.

The hadoop stuff is slow, but I think it only takes a minute or two on my machine, versus 20 min on Travis. Yeah every system call to hadoop takes several seconds because it has to spin up a JVM

gpoulin:
The entire hdfs test suite take you only 2 minutes? You are using the same hadoop as travis? On my local it takes about the same as travis. I looked `python -m cProfile -o luigi_test.prof test/runtests.py` and there is 259 calls to `luigi.hdfs.HdfsClient.exists` for a total of ~414s. It would be interesting to see what you get. I can try to find somewhere to upload my profile result. It's not so big ~2MB.

FYI, with hadoop 2.6.0 I'm getting:
```
HadoopJobError: ('Streaming job failed with exit code 1. Also, no tracking url found.', '', '-file option is deprecated, please use generic option -files instead.\nRequired argument: -input <name>\nTry -help for more information\nStreaming Command Failed!\n')
```

erikbern:
I ran it on a pretty beefy Linux machine a few weeks ago and it definitely ran in a few minutes, not 20.

I'll try to set up the minicluster on my laptop and run it.

soxofaan:
amended fix for py3 issue in luigi_grep

let's see what travis thinks now

soxofaan:
hmm again linkcheck failure from docs build:
> (line    1) broken    http://www.chiark.greenend.org.uk/ucgi/~cjwatson/blosxom/2009-07-02-python-sigpipe.html - timed out

this is apparently from `luigi/format.py`:
```python

    def create_subprocess(self, command):
        """
        http://www.chiark.greenend.org.uk/ucgi/~cjwatson/blosxom/2009-07-02-python-sigpipe.html
        """

        def subprocess_setup():
```

maybe that url should be moved out of the doc block (and just be a `#`-comment), so that builds don't fail if that blog is down/slow?

berkerpeksag:
> maybe that url should be moved out of the doc block (and just be a #-comment), so that builds don't fail if that blog is down/slow?

+1

erikbern:
I'm re-triggering the build.

I like the linkcheck but maybe we can change it so that it doesn't break the build if it fails

soxofaan:
or increase the linkcheck timeout  http://sphinx-doc.org/config.html#confval-linkcheck_timeout

erikbern:
Looks like the remaining issue is actually importing `urllib2` from `luigi_grep`.

We could probably just do what's happening in `luigi.rpc` and rely on `luigi.six`: https://github.com/spotify/luigi/blob/master/luigi/rpc.py#L27

gpoulin:
It's OK, I'm fixing `luigi_grep` now, will do a PR within the hour.

Issue #743, opened 2015-02-11
BoolParameters without config_path set to default value by set_global_parameters


daveFNbuck:
Since my latest upgrade, my snakebite has stopped working. Looking into it, I see that before set_global_parameters is called, snakebite_autoconfig is True and after set_global_parameters is called, snakebite_autoconfig becomes False, overriding the value from client.cfg and causing the wrong snakebite client to be used.

The problem seems to be that the command-line argument is overriding the value from client.cfg, but for boolean flags the command-line argument is present and False when not specified. This bug affects any BoolParameter that doesn't specify a config_path, as Parameter.value is called to find the default value and this doesn't pass a task_name or param_name to _get_value, causing it to ignore client.cfg.

It looks like snakebite_autoconfig, use_snakebite, and record_task_history are the only current parameters vulnerable to this bug. parallel_scheduling is ok because it has config_path set.

erikbern:
That's annoying. Let me add a test tonight and fix it

erikbern:
I haven't been able to reproduce this. My code:

```python
class SnakebiteConfigTest(unittest.TestCase):
    @with_config({"hdfs": {"snakebite_autoconfig": "true"}})
    def testBoolOverride(self):
        # See #743
        self.assertEqual(hdfs().snakebite_autoconfig, True)

        class MYTask(luigi.Task):
            pass

        luigi.run(['--local-scheduler', '--no-lock', 'MyTask'])
        self.assertEqual(hdfs().snakebite_autoconfig, True)
```

erikbern:
I haven't been able to repro. My code:

```python
class SnakebiteConfigTest(unittest.TestCase):
    @helpers.with_config({"hdfs": {"snakebite_autoconfig": "true"}})
    def testBoolOverride(self):
        # See #743
        self.assertEqual(hdfs.hdfs().snakebite_autoconfig, True)

        class MyTask(luigi.Task):
            pass

        luigi.run(['--local-scheduler', '--no-lock', 'MyTask'])
        self.assertEqual(hdfs.hdfs().snakebite_autoconfig, True)
```

erikbern:
Nvm realized I had already fixed the issue when I ran the test :)

Tarrasch:
Did you check that the test failed without https://github.com/spotify/luigi/commit/f6cb70fce4a6cbd64d26cc355a248c1791188d7c?

erikbern:
It did in Travis, but not on my machine. Really weird.

PR #840, opened 2015-03-10
Set cmdline argument defaults via task_name / param_name

dpkp:
This avoids side-effects caused by luigi.interface.set_global_parameters
where parsed cmdline arguments are set to parameter default values
without checking the config file.

To test, create a simple config file:
```
[hdfs]
client: snakebite
snakebite_autoconfig: True
```
and a simple task w/ luigi.hdfs.HdfsTarget:
```
class Foobar(luigi.Task):
  def output(self):
    return luigi.hdfs.HdfsTarget('foobar')
```

Without this patch, snakebite_autoconfig will be initialized to the task-level default=False set in luigi.hdfs.hdfs instead of the config file value True.

The reason for this, I believe, is that luigi.interface initializes global param values to the cmdline parsed values (set_global_parameters), but does not pass config file defaults to the cmdline parser.  Once the global value for a parameter is set, the config file is never again consulted (perhaps only affecting non-None values).

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -321,12 +321,18 @@ class Parameter(object):
             return
         flag = '--' + dest.replace('_', '-')
 
+        default = None
+        if self.has_task_value(task_name, param_name):
+            default = self.task_value(task_name, param_name)
+            if default is not None:
+                default = self.serialize(default)
+
         description = []
         description.append('%s.%s' % (task_name, param_name))
         if self.description:
             description.append(self.description)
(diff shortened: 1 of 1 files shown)


dpkp:
alternately #743

erikbern:
Yeah I thought I alraedy fixed this issue – is it still a problem?

dpkp:
appears fixed in 1.1.0.

erikbern:
thanks!

PR #842, opened 2015-03-11
Erikbern/more coverage

erikbern:
Just a bunch of trivial commits of things where it should be easy to get test coverage up

### DIFF
--- a/luigi/date_interval.py
+++ b/luigi/date_interval.py
@@ -129,7 +129,7 @@ class DateInterval(object):
             raise TypeError('Date interval type mismatch')
         return (self.date_a, self.date_b) < (other.date_a, other.date_b)
 
-    def __lte__(self, other):
+    def __le__(self, other):
         if not isinstance(self, type(other)):
             raise TypeError('Date interval type mismatch')
         return (self.date_a, self.date_b) <= (other.date_a, other.date_b)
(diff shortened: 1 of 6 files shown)


erikbern:
Somewhat relevant:

image

gpoulin:
@erikbern coveralls is at more than 86% :D. However, I need to say that this number is a little skew. There is some module that are never imported which mean there are not taken into account when computing the coverage. Here is a small list that I see:

  - mysqldb # should be easy to test in travis since mysql is already install
  - ftp # we could whether mock the server or use https://github.com/giampaolo/pyftpdlib
  - webhdfs # not sure if we can easily activate the http interface of hadoop. Otherwise, we can easily mock the http response
  - scadling # no idea how easy
  - sparkey # no idea there too

Maybe we could add some "dummy" import in the helpers just to make sure to capture those.

`@functools.total_ordering` interesting... I'm gonna remember.

erikbern:
That's a good point. If we add tests for those files it will probably drag down the average at first.

Those things you mentioned are great. I don't think we need to be super strict about tests for existing code in `luigi.contrib`, but going forward let's demand reasonable unit test coverage for all new code in `luigi.contrib`

For things in "core" Luigi I think we should strive for 95% coverage at least... shouldn't be too hard.

erikbern:
87% coverage now :)

Here's the updated breakdown:

image

I think we can do some simple stuff to `luigi.hdfs` and `luigi.hadoop` to bump the coverage...but I'll do that some other time

PR #844, opened 2015-03-12
Add a re-enable button on the dashboard

interskh:
So one can easily re-enable a task when it is disabled by scheduler (when too many failures).
re_enable


leafjungle:
I found that all the disabled tasks are displayed in the "Pending Tasks" list, instead of "Disabled Tasks" list.

PR #845, opened 2015-03-12
Doc fixes

jthi3rry:
Added class links to doc as suggested in previous PR and thought it'd be good to add Spark support as a selling point to the main readme.

Since PySparkTask and SparkSubmitTask are new features, should I update the version on the deprecation directives to 1.2.0 to follow semver?


erikbern:
Given that it tooks us two years to get from 1.0.x to 1.1.x I don't know if we should jump to 1.2.x after a few weeks only. Although I'm in favor of bumping version numbers faster.

For an 1.2.x release we should at least let the current master be out for a while and let people use it.

But I can release 1.1.1 in a few days if that makes sense.

I also think changes to "core" luigi should affect the version number more than `luigi.contrib`. There's some interesting work on re-thinking the whole execution flow (#479 #817 #818 #819) which if implemented should definitely imply going to 1.2.x

Anyway what do other people think? I don't have super strong opinions. In particular @Tarrasch and @freider usually has more opinions about what constitutes a release.

erikbern:
Btw i re-triggered the build

PR #846, opened 2015-03-12
Luigi grep py3k fix

gpoulin:
This PR also includes #839

### DIFF
--- a/bin/luigi
+++ b/bin/luigi
@@ -1,6 +1,14 @@
 #!/usr/bin/env python
 
-import luigi
+import sys
+import warnings
+import luigi.cmdline
+
+
+def main(argv):
+    warnings.warn("'bin/luigi' has moved to console script 'luigi'", DeprecationWarning)
+    luigi.cmdline.luigi_run(argv)
+
 
 if __name__ == '__main__':
-    luigi.run(use_dynamic_argparse=True)
+    main(sys.argv[1:])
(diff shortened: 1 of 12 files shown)


erikbern:
not sure why it's failing (something with bin_test) but will give it the benefit of the doubt and re-trigger

bin/luigi
```diff
@@ -1,6 +1,14 @@
 #!/usr/bin/env python
 
-import luigi
+import sys
+import warnings
+import luigi.cmdline
+
+
+def main(argv):
+    warnings.warn("'bin/luigi' has moved to console script 'luigi'", DeprecationWarning)
```
Tarrasch: @gpoulin, is there any documentation for me how to run this if `bin/luigi` is deprecated?

gpoulin: @Tarrasch, I was again deprecation :P , that was part of #839. I don't think is in the doc, but once you have installed Luigi, the command should be in your path.

``` bash
virtualenv luigi_env
. luigi_env/bin/activate
cd luigi
python setup.py install
luigi someTask
```

Tarrasch: Ah, I've always done tricks with `PYTHONPATH`, maybe I should install luigi the "proper" way. :)

Tarrasch: Ah, this was from when I was on vacation, that's why I don't remember it ^^

PR #847, opened 2015-03-12
Let tox handle the python version

gpoulin:
Also cache the hadoop distribution
Allow easier switching between python version
Benifit from the tox cache

Before
```bash
export EPYTHON=python2.7 # Gentoo way to set python version
tox -e cdh # install deps and download hadoop
export EPYTHON=python3.4
tox -e cdh # install deps and download hadoop
export EPYTHON=python2.7
tox -e cdh # install deps and download hadoop
```

After
```bash
tox -e py27-cdh # install deps and download hadoop
tox -e py34-cdh # install deps
tox -e py27-cdh # just run the tests
```

In addition, I think that coveralls is currently a little confused when producing the coverage diff, hopefully that will help. The main inconvenient is that it makes the travis conf a little verbos.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,5 @@
 language: python
 
-python:
-  - "2.7"
-  - "2.6"
-  - "3.3"
-  - "3.4"
-
 jdk: oraclejdk7
 
 services:
@@ -16,10 +10,16 @@ env:
     - PIP_DOWNLOAD_CACHE=$HOME/.pip-cache
     - POSTGRES_USER=postgres
   matrix:
-    - TOX_ENV=pep8
-    - TOX_ENV=cdh
-    - TOX_ENV=nonhdfs
-    - TOX_ENV=docs
+    - TOXENV=pep8
+    - TOXENV=docs
+    - TOXENV=py26-nonhdfs
+    - TOXENV=py27-nonhdfs
+    - TOXENV=py33-nonhdfs
+    - TOXENV=py34-nonhdfs
+    - TOXENV=py26-cdh
+    - TOXENV=py27-cdh
+    - TOXENV=py33-cdh
+    - TOXENV=py34-cdh
 
 sudo: false
 
@@ -27,7 +27,6 @@ cache:
   - $HOME/.pip-cache
 
 install:
-  - pip install .
   - pip install tox
 
 before_script:
@@ -39,7 +38,7 @@ before_script:
   - ssh -o StrictHostKeyChecking=no localhost true
 
 script:
-  - tox -e $TOX_ENV
+  - tox
 
 after_failure:
   - cat /home/travis/build/spotify/luigi/.tox/cdh/log/cdh-1.log
--- a/scripts/ci/setup_hadoop_env.sh
+++ b/scripts/ci/setup_hadoop_env.sh
@@ -42,8 +42,6 @@ while test $# -gt 0; do
     esac
 done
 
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -2,9 +2,6 @@ language: python
 
 python:
```
berkerpeksag: You can remove `python: "2.7"` too.

berkerpeksag:
You can now change

```
 script:
   - tox -e $TOX_ENV
```

to

```
script = tox
```

scripts/ci/setup_hadoop_env.sh
```diff
@@ -42,7 +42,7 @@ while test $# -gt 0; do
     esac
 done
 
-HADOOP_HOME=/tmp/hadoop-${HADOOP_DISTRO}
+HADOOP_HOME=${TOX_CACHE_FOLDER}/hadoop_cache/hadoop-${HADOOP_DISTRO}
```
berkerpeksag: You can also add this folder to Travis as a cache dir, but I'm not sure if there's too much to gain.

gpoulin:
We the fixup, we test pep8 only once and build the doc only once. Let me know if you think it's an issue I'll fix. On the other hand, we are 4 times less likely to have a broken link :)

scripts/ci/setup_hadoop_env.sh
gpoulin: Travis has better internet connection than me. So I don't think it makes so much difference. ~30s / 20min

gpoulin:
@berkerpeksag Wouldn't `script = tox` create a single travis build for all tox environments? If it's the case I'm against it. When I open travis, I want to see directly which environments have failed. So I simply click on the little travis :x: and say: "python 2.6... again"

erikbern:
We've actually caught one or two bugs by running docs separately for python 2 and 3... docs actually fires up a Python interpreter. But it's not super important to preserve.

We should also consider dropping some of the Python versions for cdh

berkerpeksag:
> Wouldn't script = tox create a single travis build for all tox environments?

No, see our .travis.yml we use in Gunicorn: https://github.com/benoitc/gunicorn/blob/master/.travis.yml Travis still creates separate builds for each environment: https://travis-ci.org/benoitc/gunicorn

gpoulin:
@berkerpeksag It was working for me because `.travis.yml` currently use `TOX_ENV` instead of `TOXENV`. I'll "fix" that.

@erikbern As far as I know the bugs that were catch by the doc was because a module was not importable. I would rather add a test that import all tests recursively.  That would reduce the coverage figure showed by coveralls, but will increase the actual coverage. If you are worry about the figure given by coveralls, I would rather omit the content of the contrib folder `.coveragerc`.

erikbern:
Yes it would be awesome to import all tests recursively. Maybe you're right that it was the issue

gpoulin:
@erikbern I added the test that import all modules. It "brings back" the coverage to [82%](https://coveralls.io/builds/2104770), but it's a more realistic figure. (I'm adding tests and the coverage go down...). I agree that we could be optimistic and run the hdfs test only for python 2.6 and python 3.4.

I also added `python --version` to tox scripts so we can see the [minor version used](https://travis-ci.org/spotify/luigi/jobs/54197225#L147).

If you are happy with the changes, it's ready for merging.

erikbern:
Will merge if Travis succeeds

erikbern:
Nice! I'm also considering dropping some of the python versions for CDH now.

Tarrasch:
Awesome job you guys!

PR #854, opened 2015-03-16
Fix for bin/luigi to make test/cmdline_test run on my local computer

erikbern:


gpoulin:
@erikbern How do you run the test? It shouldn't happen if you run the test with tox. Before #839 it was also possible to `PYTHONPATH=. nosetests test/cmdline_test`, but now Luigi needs to be install in virtual environment to pass `test_console_script_luigi`.

erikbern:
Haven't been using tox, just nosetests. Tried using tox to run tests but it's failing on my machine. I'll try to make it work. Would be great to avoid doing this stupid stuff with the paths.

erikbern:
ok, works with tox now (except on my machine some tests get "too many open files", but that's an os x issue i think). great! closing this

gpoulin:
With python3 nose [complain](https://travis-ci.org/spotify/luigi/jobs/54513463#L177) about unclosed files. Maybe trying to close all files during the test would solve your issue.

erikbern:
yeah i've seen that too. didn't connect it though. makes sense that's the issue

PR #861, opened 2015-03-18
Keeps assistants from becoming stakeholders

daveFNbuck:
If assistants live longer than other workers, we don't want them to keep tasks
alive unnecessarily. By preventing assistants from becoming stakeholders of
tasks they're working on, pruning of tasks done by the assistant is allowed when
the workers that actually scheduled them go away.


leafjungle:
I am wondering what is "stakeholders"? is it the task creator (the task initiator, the client)?

daveFNbuck:
Stakeholders are any workers that require the task to exist. So basically any process that schedules the task becomes a stakeholder.

PR #865, opened 2015-03-22
Various param cleanup

erikbern:
Allow the use of dashes in the config names

(but throw a warning saying you should use underscores)


erikbern:
There's some more commits coming soon, stay tuned

PR #867, opened 2015-03-24
Worker cleanup (WIP)

erikbern:
Have more to do (want to merge ExternalTaskProcess and TaskProcess etc).

Keeping an open PR to be able to keep an eye on Travis and see if people have strong opinions against anything I'm doing


erikbern:
Will squash once this is done

PR #871, opened 2015-03-30
Does not consider upstream disabled jobs as pending for worker keep-aliv...

daveFNbuck:
This is a simplified version of #686 which achieves the same goal without altering the status of any jobs.

If you schedule a worker with a pending task that depends on a disabled task and
have worker-keep-alive set to true, the worker will stay alive forever waiting
for a pending task that can never be scheduled. This updates the get_work
response not consider upstream disabled jobs as pending so that workers can die
once all potentially runnable jobs are complete.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -636,6 +636,7 @@ class CentralPlannerScheduler(Scheduler):
         best_task = None
         locally_pending_tasks = 0
         running_tasks = []
+        upstream_table = {}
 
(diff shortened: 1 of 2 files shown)


daveFNbuck:
@boosh I think this will fix your earlier issue from #686.

erikbern:
LGTM!

boosh:
Great, I think this will be a handy change and a lot simpler!

PR #872, opened 2015-03-30
Eliminate busy wait, fixing 100% CPU usage.

tomquisel:
Also remove unecessary communicate() call. It just prints "None" to the
log.

### DIFF
--- a/test/contrib/spark_test.py
+++ b/test/contrib/spark_test.py
@@ -193,7 +193,7 @@ class SparkSubmitTaskTest(unittest.TestCase):
         def interrupt():
             raise KeyboardInterrupt()
 
-        proc.return_value.poll = interrupt
+        proc.return_value.wait = interrupt
         try:
             job = TestSparkSubmitTask()
             job.run()
--- a/luigi/contrib/spark.py
+++ b/luigi/contrib/spark.py
@@ -235,9 +235,7 @@ class SparkSubmitTask(luigi.Task):
                                 universal_newlines=True)
         try:
             with SparkRunContext(proc):
-                while proc.poll() is None:
-                    pass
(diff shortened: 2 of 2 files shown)


Tarrasch:
Erm, I don't see why this PR was frogotten @tomquisel . But you'll have to squash the commits I think. Then I can merge once somebody else +1's. I don't know much about spark.

By the way, the commit header must name "spark" or something somewhere, as the commit message is unclear. :)

test/contrib/spark_test.py
```diff
@@ -193,7 +193,7 @@ def test_app_interruption(self, proc):
         def interrupt():
             raise KeyboardInterrupt()
 
-        proc.return_value.poll = interrupt
+        proc.return_value.wait = interrupt
```
erikbern: btw ideally we should use the mock library for this. but that's prob outside the scope of this PR

Tarrasch:
I wish you didn't merge this as I think the commit-quality was pretty poor :-/

erikbern:
I think it was fine. What was the issue? Not squashed? yeah maybe should have required squashing it. But I also think it would have been a bit rude to get back after almost a month of silence then asking for a rebase

tomquisel:
Totally not a big deal either way :). I'll squash commits and make the commit messages more clear in the future.

Tarrasch:
Sure, sorry for being not so responsive on the PR, don't hesitate to ping us @tomquisel :)

@erikbern My main issue is that the commit message doesn't mention "spark" *anywhere*. It sounds like it's fixing a critical bug in the scheduler or something. :)

erikbern:
yeah that's a good point. i was actually really confused too first :)

PR #874, opened 2015-04-03
Allow for workflows in which the assistant is the only long-term worker

daveFNbuck:
This encompasses two changes. First, it allows scheduling with --workers 0 so that after scheduling the worker will immediately die. Second, it keeps runnable jobs (and DONE jobs they depend on) from being pruned if there is an assistant. By combining these, we can create a workflows in which scheduling and running jobs are handled completely separately.

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -170,7 +170,8 @@ class Interface(object):
             success &= w.add(t, env_params.parallel_scheduling)
         logger = logging.getLogger('luigi-interface')
         logger.info('Done scheduling tasks')
-        success &= w.run()
+        if env_params.workers != 0:
+            success &= w.run()
         w.stop()
         return success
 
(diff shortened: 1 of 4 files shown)


erikbern:
interesting that this failed on 2.7 but no other version. maybe some flakiness in the tests for multiple workers?

erikbern:
btw very exciting to see this!

daveFNbuck:
Must be flaky, especially since 2.7 is what I'm using. Seems to have passed now.

I'm glad you like it. There are still a few issues that need to be ironed out before I could reasonably run everything this way, but I think it's getting close.

erikbern:
the reason i'm excited about the assistant is that it is a step towards a fully distributed architecture.

if we can get the assistant working, the only remaining thing is built-in scheduling

Tarrasch:
@erikbern Do you have a particular definition in mind when you say "fully distributed architecture"?  Is there a checklist anywhere on what is required to call an architecture "fully distributed"?

erikbern:
No. What I meant is the worker part will be horizontally scalable. Other
things are not. Eg scheduler is still spof

PR #875, opened 2015-04-06
Adding original template files & Correcting a lib path & Commenting a copyright

ystopia:
1. added copyright & author & description info to the db_task_history.py
2. readded three template files
3. modified the bootstrap lib path in layout.html

### DIFF
--- a/luigi/db_task_history.py
+++ b/luigi/db_task_history.py
@@ -19,6 +19,21 @@
 Provides a database backend to the central scheduler. This lets you see historical runs.
 See :ref:`TaskHistory` for information about how to turn out the task history feature.
 """
+#
+# Description: Added codes for visualization of how long each task takes
+# running-time until it reaches the next status (failed or done)
+# At "{base_url}/tasklist", all completed(failed or done) tasks are shown.
+# At "{base_url}/tasklist", a user can select one specific task to see
+# how its running-time has changed over time.
+# At "{base_url}/tasklist/{task_name}", it visualizes a multi-bar graph
+# that represents the changes of the running-time for a selected task
+# up to the next status (failed or done).
+# This visualization let us know how the running-time of the specific task
+# has changed over time.
+#
+# Copyright 2015 Naver Corp.
+# Author Yeseul Park (yeseul.park@navercorp.com)
+#
 
 import datetime
 import logging
--- a/luigi/templates/header.html
+++ b/luigi/templates/header.html
@@ -1,0 +1,4 @@
+<!doctype html>
+<title>Luigi Task History</title>
+<link href="/static/visualiser/lib/bootstrap/css/bootstrap.min.css" rel="stylesheet">
+<script src="/static/visualiser/lib/bootstrap/js/bootstrap.min.js"></script>
--- a/luigi/templates/layout.html
+++ b/luigi/templates/layout.html
@@ -82,10 +82,10 @@
       .lead { margin-top: -17px; margin-bottom: 13px; }
 
     </style>
-    <link href="{{ static_url("lib/bootstrap/css/bootstrap.css") }}" rel="stylesheet">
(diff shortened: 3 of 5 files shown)


luigi/db_task_history.py
```diff
@@ -19,6 +19,21 @@
 Provides a database backend to the central scheduler. This lets you see historical runs.
 See :ref:`TaskHistory` for information about how to turn out the task history feature.
 """
+#
```
erikbern: if you put the copyright within triple quotes it will appear in the documentation

luigi/db_task_history.py
```diff
@@ -19,6 +19,21 @@
...
+# At "{base_url}/tasklist", all completed(failed or done) tasks are shown.
+# At "{base_url}/tasklist", a user can select one specific task to see
+# how its running-time has changed over time.
+# At "{base_url}/tasklist/{task_name}", it visualizes a multi-bar graph
+# that represents the changes of the running-time for a selected task
+# up to the next status (failed or done).
+# This visualization let us know how the running-time of the specific task
+# has changed over time.
+#
+# Copyright 2015 Naver Corp.
```
erikbern: we already have a copyright at the top, should be enough

erikbern:
lgtm but i'm not sure about the extra copyright. note that the apache header is to give people the right to use the code. i'm not sure if copyright spotify has any legal bearing so we could probably remove it.

anyone who knows about licenses? @gpoulin @jcrobak @steenzout ?

luigi/db_task_history.py
ystopia: Oh, I see. Please let me know what to do then. :) 
It is required for me to put the copyright part "# Copyright 2015 Naver Corp. # Author Yeseul Park (yeseul.park@navercorp.com)" in the header for every files that I've modified. Moreover, I should write down the # Description part at least once.

steenzout:
I'm no expert but I think, in practice, whoever makes a significant contribution (or all contributions) should sign a CLA.
if you're working on behalf of a company (most of us are) then there could possibly be a need for a CCLA (help anyone?).
if one signs a CLA then the copyright notice I think would be Spotify based on

    "
    The purpose of a CLA is to ensure that the guardian of a project's outputs has the necessary
    ownership or grants of rights over all contributions to allow them to distribute under the chosen licence.
    In some cases this will mean that the contributor will assign the copyright in all
    contributions to the project owner;
    in other cases, they will grant an irrevocable licence to allow the
    project maintainer to use the contribution.
    "

http://en.wikipedia.org/wiki/Contributor_License_Agreement

I've played with this platform https://www.clahub.com/ to simplify CLA management on GitHub projects.
This is a short version of why CLAs : https://www.clahub.com/pages/why_cla .

this is where I drawn most of the information on this matter : http://oss-watch.ac.uk/resources/cla
more info and templates can be found here : http://harmonyagreements.org/index.html

gpoulin:
I'm not an expert either, but my understanding is that since Luigi is under Apache2 and no that CLA was agree upon, all contributors keep the copyright on their contribution, but every contribution is done under the [Apache License](https://github.com/spotify/luigi/blob/master/LICENSE#L130). It shouldn't be an issue as long as Apache2 license is compatible with all intended use of the code. The only minor issue I see it's when somebody contribute on behalf of a company, it's currently not clear who has the copyright.

For the license header, I'm not a big fan. When I open a file I want to know what the file is used, I already know that it is under Apache2. However, I'm not sure if it's not require by the Apache license. I would be against to add more copyright and changelog in the header. If all contributors start to add their little contribution in the header it would be a nightmare.

For code attribution and change log, git log is great. There's could be a general changelog file to easily track the change from version to version which is useful when upgrading, but that's another story.

steenzout:
@gpoulin the boilerplate on top of each source file seems to be a recommendation : http://choosealicense.com/licenses/apache-2.0/

PR #877, opened 2015-04-07
Fix external dynamic deps

f355:
Since running tasks communicate with worker via a queue, all dynamic dependencies that they yield must be serialized and then deserialized back. This doesn't work if a task has `run = NotImplemented`, since there was a specific check for that in Register for unclear reason.

This PR adds a test case to reproduce the issue and fixes it by removing the check.

### DIFF
--- a/luigi/task_register.py
+++ b/luigi/task_register.py
@@ -135,8 +135,6 @@ class Register(abc.ABCMeta):
         # We return this in a topologically sorted list of inheritance: this is useful in some cases (#822)
         reg = OrderedDict()
         for cls in cls._reg:
-            if cls.run == NotImplemented:
-                continue
             name = cls.task_family
 
             if name in reg and reg[name] != cls and \
--- a/test/task_test.py
+++ b/test/task_test.py
@@ -21,6 +21,7 @@ from datetime import datetime, timedelta
 
 import luigi
 import luigi.task
+from luigi.task_register import load_task
 
 
 class DummyTask(luigi.Task):
@@ -54,6 +55,10 @@ class TaskTest(unittest.TestCase):
         original = DummyTask(**params)
         other = DummyTask.from_str_params(original.to_str_params())
         self.assertEqual(original, other)
+
+    def test_external_tasks_loadable(self):
+        task = load_task("luigi", "ExternalTask", {})
+        assert(isinstance(task, luigi.ExternalTask))
 
     def test_id_to_name_and_params(self):
         task_id = "InputText(date=2014-12-29)"


f355:
Oh, it breaks tests, this behavior with `run = NotImplemented` seems to be intentional.
Could someone please explain the logic behind looking at `run` method existence to me? I'd also love to hear any ideas on how to fix the issue. Thanks a lot!

erikbern:
Yeah this sounds weird. I think the idea is you shouldn't be able to run external tasks from the command line. Not sure though. It seems like bad introspection which should be removed (I've seen same thing in a couple of places).

Do you have a breaking example with dynamic requirements btw?

f355:
Trivial example:
```(python)
import luigi

class Dependency(luigi.ExternalTask):
    def output(self):
        return luigi.LocalTarget("/tmp/i_am_not_here")

class MainTask(luigi.Task):
    def output(self):
        return luigi.LocalTarget("/tmp/i_am_not_here_as_well")

    def run(self):
        yield Dependency()

if __name__ == "__main__":
    luigi.run()
```

Running `MainTask` produces the following:
```
$ python test.py MainTask --local-scheduler
DEBUG: Checking if MainTask() is complete
INFO: Scheduled MainTask() (PENDING)
INFO: Done scheduling tasks
INFO: Running Worker with 1 processes
DEBUG: Asking scheduler for work...
DEBUG: Pending tasks: 1
INFO: [pid 45372] Worker Worker(salt=369848895, workers=1, host=f355-mbpr.local, username=konstantin, pid=45372) running   MainTask()
INFO: [pid 45372] Worker Worker(salt=369848895, workers=1, host=f355-mbpr.local, username=konstantin, pid=45372) new requirements      MainTask()
DEBUG: 1 running tasks, waiting for next task to finish
  File "test.py", line 15, in <module>
    luigi.run()
  File "/Users/konstantin/src/luigi/luigi/interface.py", line 425, in run
    return interface.run(tasks, worker_scheduler_factory, override_defaults=override_defaults)
  File "/Users/konstantin/src/luigi/luigi/interface.py", line 174, in run
    success &= w.run()
  File "/Users/konstantin/src/luigi/luigi/worker.py", line 743, in run
    self._handle_next_task()
  File "/Users/konstantin/src/luigi/luigi/worker.py", line 666, in _handle_next_task
    for module, name, params in new_requirements]
  File "/Users/konstantin/src/luigi/luigi/task_register.py", line 201, in load_task
    task_cls = Register.get_task_cls(task_name)
  File "/Users/konstantin/src/luigi/luigi/task_register.py", line 175, in get_task_cls
    raise TaskClassException('Task %r not found. Candidates are: %s' % (name, cls.tasks_str()))
luigi.task_register.TaskClassException: Task 'Dependency' not found. Candidates are: Config,MainTask,RangeBase,RangeDaily,RangeDailyBase,RangeHourly,RangeHourlyBase,Task,WrapperTask,core,scheduler,worker
```

f355:
Couple more thoughts on it:
1. If we remove the check, it will make this task ambiguous: https://github.com/spotify/luigi/blob/master/test/cmdline_test.py#L60 Even though picking the right class that way doesn't seem to be a good programming practice, some people might have existing code that relies on it, intentionally or not.
2. We still need to have a way of distinguishing external tasks here: https://github.com/spotify/luigi/blob/master/luigi/worker.py#L509 There are several ways of doing it, and I'm not sure the one in question is the worst of them. Ideas?

f355:
According to `git bisect`, the commit that broke external tasks is f8cdfedeee5894eb094d535972e95b75b77de255.

erikbern:
Yes that makes sense. I replaced a custom task importer with the standard one. But the standard one doesn't handle external tasks. Need to fix this

Tarrasch:
Nice catch @f355. Spotify haven't started to be power users of the dynamic require feature. Maybe @tommyengstrom want to comment? Anyway, kind of weird that we don't have tests for dynamically requiring a dynamic dependency.

tommyengstrom:
Well, I don't really have anything to add. I never considered this case but I agree that there should be tests for it.

erikbern:
I think we should break that case with non ambiguous class. This code was introduced in d1c5beae but I don't know what purpose it served – @freider ?

Tarrasch:
Also ping reviewer @anyman.

aldanor:
Hmm, wonder if it's still broken?

Tarrasch:
Sorry, nobody have worked on this. :-/

I most cases, lazy scheduling works, but there seem to be a problem with the external tasks. We can't merge the fix in #880 because it will break other things.

erikbern:
I can look at this at some point, had a bunch of ideas on how to clean up this code. Just need to find some time

Tarrasch:
Cool @erikbern, I know you can do it. I still think the configuration system you introuced is really good!

erikbern:
should be fixed now!

PR #878, opened 2015-04-10
Adding SequenceFileAsBinaryInputFormat iterator for hadoop mappers

jdangerx:
I'm currently running some Hadoop jobs on image data. They are `typedbytes` in a SequenceFile. 

I set the job's `-inputformat` to `org.apache.hadoop.mapred.SequenceFileAsBinaryInputFormat`, which worked great to pass in `typedbytes` to the mapper. 

However, I found that `JobTask._map_input` always splits on newlines. However, newlines occur in the middle of my image files, which causes failures.

Here's a janky fix that reads in the keys and values and their lengths, and splits the input stream based on that. I have a few questions before I would consider this worthy of merging, but since I'm already working on it I figured it should probably go in a PR.

1. The method by which I access (and check) the input_format is really janky. Is there a better way to set this up? Maybe when `job_runner().run_job(job)` is called, we can set `job.jobopts` to be a dict of all the options passed in to `job_runner`.

2. How would you suggest testing this? It looks like `test/hadoop_test.py` seems to be mostly tests of the "run jobs, then check the output" flavor. Would it be sufficient to just have a "`_binseqfile_iter(stdin)` takes binary input and outputs key-value pairs properly" test?

### DIFF
--- a/luigi/hadoop.py
+++ b/luigi/hadoop.py
@@ -39,6 +39,7 @@ try:
     from StringIO import StringIO
 except ImportError:
     from io import StringIO
+import struct
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


Tarrasch:
Closing to minimize noise. Open up again if you want to iterate on this.

PR #880, opened 2015-04-11
Fix #877 by breaking some backwards compatibility

erikbern:


Tarrasch:
What does it exactly mean when a task gets registered?

erikbern:
What do you mean?

This will expose external tasks over the CLI. But I can't think of why you would ever have two classes with the same name where one is external and the other not. That's the only case that breaks

erikbern:
@freider @anyman let me know what you think of this. I think it will simplify some stuff.

erikbern:
PING

erikbern:
oops caps lock

anyman:
We have a lot of cases with external tasks having the same name as the "real" class. That is common naming practice when adding classes to the external dataset repo.

How would this break that case?

erikbern:
It breaks it if you try to run any of those classes from cmd line.

How about we change the behavior to: return the class that was defined *last*, but throw a warning if there's ambiguity

Tarrasch:
Yea, say if we have 100 tasks at Spotify right now, we'll have one big file like this

external_tasks.py
```
Task1(ExternalTask):
    def output(self): return HdfsTarget('/some/path')

Task2(ExternalTask):
    def output(self): return HdfsTarget('/some/other/path')

Task3(ExternalTask):
    def output(self): return HdfsTarget('/some/third/path')

...
```

And then we have


task1.py
```
import external_tasks

Task1(luig.Task):
    def output(self): return HdfsTarget('/some/path')

    def run(self) hdfs.write('/some/path', 'content')

    def require(self): return external_tasks.Task2()
```

Tarrasch:
Why is it important for luigi's internal logic that External Tasks are exposed over the CLI? You would never want to run them I think.

erikbern:
It's not important. I just thought it was weird we treat external tasks differently. All the special cases also caused some issues.

I can fix this in other ways and refactor some shit. It's really confusing how there's all this code that does type checking and treats external tasks differently. Need to change it to methods and stuff

Tarrasch:
Maybe the namespace of an `ExternalTask` could be forced to be `external`. So if you create a task like

```
Task1(ExternalTask):
    def output(self): return HdfsTarget('/some/path')
```

You run it with `luigi --task external.Task1`. Would that work? The task can then be registered as it will not clash. Not sure if this is possible or if I've misunderstood the problem. Let me know what you think. :)

erikbern:
yeah that's another good idea. i'll look at it

anyman:
My thoughts on namespaces are that we should start using the python
namespace and remove the luigi internal namespace.

That should solve this as well I guess but is a much bigger change.
Den 18 apr 2015 06:49 skrev "Erik Bernhardsson" <notifications@github.com>:

Tarrasch:
@anyman, what other problems does it solve? Yes, using the python namespace would eliminate one concept, it would introduce these bad properties:

 * Code becomes file-path-sensitive. What a developer think is trivial refactoring will actually change behaviour.
 * It's more implicit over explicit, giving us less control and reduces grepability.
 * We can't make the assumptions like the namespace `external.*` is for external tasks.

@erikbern + everyone: By the way, if the CLI doesn't treat the external tasks differently, then doing `--help` would mean that all our (SpotifyInc's) 100 tasks would show up in the CLI. Let's keep that in mind when desinging a solution here.

anyman:
I agree with you that there is some issue with refactorings.

What I like with the idea of implicit namespaces is that we get it for free. Right now most Tasks don't have a namespace. We currently have some cases of duplicate naming as well.

Tarrasch:
I like that idea too. But I really think we have enough of python magic as it is. So I would prefer users specifying the `task_namespace` for their tasks.

erikbern:
Let me fix this one so that ExternalTask basically just overrides the namespace

I was looking at it but hte namespace code is pretty confusing so I'll give it another shot tomorrow

Tarrasch:
I know I proposed the namespace thingie. But I'm not sure that's the way to go. In the end we should have a function (or method or whatever) that is "the definition of external". Currently it is like `is_external(self): return self.run is NotImplementedError`. We're going to change that to `is_external(self): return 'external.' in self.task_namespace` or `is_external(self): return self.task_namespace.begins_with('external')`. I find that too be a bit fragile to look for a string in the task namespace. What if a user happens to override it? I mean like

```
class MyTask(ExternalTask):
     task_namespace = 'mynamespace'
```

One could use parameters, like changing the definition of `Task`

```
class Task(object):
     external = luigi.Parameter(default=False)
     ...


class ExternalTask(Task):
     external = luigi.Parameter(default=True)
     ...
```

Then we can have the more sensible `is_external(self): return self.external`. But we then have the problem of we've not solved the original problem of clashing of registered tasks. Hmm!

Maybe we should just add the `external.` to the task namespace and go with that. I suppose the "override" problem can be fixed in the meta class `TaskRegister`, hehe.

erikbern:
I think the problem is that a bunch of other places have to know. THey shouldn't have to know at all. worker.py and task_register.py and whatnot should all be agnostic of whether the task is external or not

Tarrasch:
Can you also articulate what places will need to know?

erikbern:
my goal is being external is just a property of the `Task` object that no one else has to care about. OOP :)

erikbern:
If you have two classes A and B with slightly different behavior then if you do OOP perfectly then there should never be any code outside A or B that tries to infer the type of them.

https://twitter.com/fulhack/status/587783554651856896 :)

erikbern:
And it's not just `isinstance`... trying to infer anything about the object by inspecting it is a form of cargo cult imho... anyway sorry for OOP ranting

erikbern:
But anyway going back to the original question: I'm increasingly convinced the right thing is to merge this PR as it is.

Having multiple tasks with same name isn't great – it screws up the idea of having a task register that can be used for multiple things (a) command line integration (b) dynamic requirements (c) assistant worker. I really think (a), (b), and (c) should be consistent.

It's not like this PR will break that much... the only thing it will break is if you *directly* invoke `TaskX` from the command line and there are more than one `TaskX` available but only one that implements the `run` method. You can still use it internally in Python using requires etc.

The easiest way to fix this in a one-liner for Spotify is to just add `task.namespace('external')` at the top of `external_tasks.py` and reset it at the end of the file. This isn't super beautiful, but does the trick. We can probably make the namespacing functionality better in a later PR

sisidra:
Not reading it all, I give +1 for @anyman and python namespace -> luigi namespace (at least by default). :)

Tarrasch:
Sure, I'm fine with merging this then.

By the way, it's almost a completely seperate issue, but we should only expose the `--help` flags for the `luigi.task.Config` and not `luigi.task.Task`.

erikbern:
we could have another task class property called `expose_cmdline` or something but it seems like it introduces more concepts than needed

Tarrasch:
@sisidra, well, then I guess I have to put my arguments down. Good point that it can be default but you can override it in a way so it's not file-path sensitive.

@erikbern, Currently, when you run `--help` you get:

```
❯ python external.py Foo --help
usage: external.py Foo [-h] [--scheduler-retry-delay SCHEDULER_RETRY_DELAY]
                       [--scheduler-remove-delay SCHEDULER_REMOVE_DELAY]
                       [--scheduler-worker-disconnect-delay SCHEDULER_WORKER_DISCONNECT_DELAY]
                       [--scheduler-state-path SCHEDULER_STATE_PATH]
                       [--scheduler-disable-window SCHEDULER_DISABLE_WINDOW]
                       [--scheduler-disable-failures SCHEDULER_DISABLE_FAILURES]
                       [--scheduler-disable-hard-timeout SCHEDULER_DISABLE_HARD_TIMEOUT]
                       [--scheduler-disable-persist SCHEDULER_DISABLE_PERSIST]
                       [--scheduler-max-shown-tasks SCHEDULER_MAX_SHOWN_TASKS]
                       [--scheduler-prune-done-tasks]
                       [--scheduler-record-task-history]
                       [--scheduler-visualization-graph SCHEDULER_VISUALIZATION_GRAPH]
                       [--worker-ping-interval WORKER_PING_INTERVAL]
                       [--worker-keep-alive] [--worker-count-uniques]
                       [--worker-wait-interval WORKER_WAIT_INTERVAL]
                       [--worker-max-reschedules WORKER_MAX_RESCHEDULES]
                       [--worker-timeout WORKER_TIMEOUT]
                       [--worker-task-limit WORKER_TASK_LIMIT]
                       [--worker-retry-external-tasks] [--local-scheduler]
                       [--scheduler-host SCHEDULER_HOST]
                       [--scheduler-port SCHEDULER_PORT]
                       [--lock-size LOCK_SIZE] [--no-lock]
                       [--lock-pid-dir LOCK_PID_DIR] [--workers WORKERS]
                       [--logging-conf-file LOGGING_CONF_FILE]
                       [--module MODULE] [--parallel-scheduling] [--assistant]
                       [--RangeBase-of RANGEBASE_OF]
                       [--RangeBase-start RANGEBASE_START]
                       [--RangeBase-stop RANGEBASE_STOP] [--RangeBase-reverse]
                       [--RangeBase-task-limit RANGEBASE_TASK_LIMIT]
                       [--RangeBase-now RANGEBASE_NOW]
                       [--RangeDailyBase-start RANGEDAILYBASE_START]
                       [--RangeDailyBase-stop RANGEDAILYBASE_STOP]
                       [--RangeDailyBase-days-back RANGEDAILYBASE_DAYS_BACK]
                       [--RangeDailyBase-days-forward RANGEDAILYBASE_DAYS_FORWARD]
                       [--RangeHourlyBase-start RANGEHOURLYBASE_START]
                       [--RangeHourlyBase-stop RANGEHOURLYBASE_STOP]
                       [--RangeHourlyBase-hours-back RANGEHOURLYBASE_HOURS_BACK]
                       [--RangeHourlyBase-hours-forward RANGEHOURLYBASE_HOURS_FORWARD]
                       [--Foo-par-sig FOO_PAR_SIG]
                       [--Foo-par-insig FOO_PAR_INSIG] [--par-sig PAR_SIG]
                       [--par-insig PAR_INSIG]

optional arguments:
  -h, --help            show this help message and exit
  --scheduler-retry-delay SCHEDULER_RETRY_DELAY
                        scheduler.retry_delay for all instances of class
                        scheduler [default: 900.0]
  --scheduler-remove-delay SCHEDULER_REMOVE_DELAY
                        scheduler.remove_delay for all instances of class
                        scheduler [default: 600.0]
  --scheduler-worker-disconnect-delay SCHEDULER_WORKER_DISCONNECT_DELAY
                        scheduler.worker_disconnect_delay for all instances of
                        class scheduler [default: 60.0]
  --scheduler-state-path SCHEDULER_STATE_PATH
                        scheduler.state_path for all instances of class
                        scheduler [default: /var/lib/luigi-
                        server/state.pickle]
  --scheduler-disable-window SCHEDULER_DISABLE_WINDOW
                        scheduler.disable_window for all instances of class
                        scheduler [default: 3600]
  --scheduler-disable-failures SCHEDULER_DISABLE_FAILURES
                        scheduler.disable_failures for all instances of class
                        scheduler [default: None]
  --scheduler-disable-hard-timeout SCHEDULER_DISABLE_HARD_TIMEOUT
                        scheduler.disable_hard_timeout for all instances of
                        class scheduler [default: None]
  --scheduler-disable-persist SCHEDULER_DISABLE_PERSIST
                        scheduler.disable_persist for all instances of class
                        scheduler [default: 86400]
  --scheduler-max-shown-tasks SCHEDULER_MAX_SHOWN_TASKS
                        scheduler.max_shown_tasks for all instances of class
                        scheduler [default: 100000]
  --scheduler-prune-done-tasks
                        scheduler.prune_done_tasks for all instances of class
                        scheduler [default: False]
  --scheduler-record-task-history
                        scheduler.record_task_history for all instances of
                        class scheduler [default: False]
  --scheduler-visualization-graph SCHEDULER_VISUALIZATION_GRAPH
                        scheduler.visualization_graph for all instances of
                        class scheduler [default: svg]
  --worker-ping-interval WORKER_PING_INTERVAL
                        worker.ping_interval for all instances of class worker
                        [default: 1.0]
  --worker-keep-alive   worker.keep_alive for all instances of class worker
                        [default: False]
  --worker-count-uniques
                        worker.count_uniques for all instances of class worker
                        [default: False]
  --worker-wait-interval WORKER_WAIT_INTERVAL
                        worker.wait_interval for all instances of class worker
                        [default: 1]
  --worker-max-reschedules WORKER_MAX_RESCHEDULES
                        worker.max_reschedules for all instances of class
                        worker [default: 1]
  --worker-timeout WORKER_TIMEOUT
                        worker.timeout for all instances of class worker
                        [default: 0]
  --worker-task-limit WORKER_TASK_LIMIT
                        worker.task_limit for all instances of class worker
                        [default: None]
  --worker-retry-external-tasks
                        worker.retry_external_tasks for all instances of class
                        worker [default: False]
  --local-scheduler     core.local_scheduler for all instances of class core
                        [default: False]
  --scheduler-host SCHEDULER_HOST
                        core.scheduler_host for all instances of class core
                        [default: localhost]
  --scheduler-port SCHEDULER_PORT
                        core.scheduler_port for all instances of class core
                        [default: 8082]
  --lock-size LOCK_SIZE
                        core.lock_size for all instances of class core
                        [default: 1]
  --no-lock             core.no_lock for all instances of class core [default:
                        False]
  --lock-pid-dir LOCK_PID_DIR
                        core.lock_pid_dir for all instances of class core
                        [default: /tmp/luigi]
  --workers WORKERS     core.workers for all instances of class core [default:
                        1]
  --logging-conf-file LOGGING_CONF_FILE
                        core.logging_conf_file for all instances of class core
                        [default: None]
  --module MODULE       core.module for all instances of class core [default:
                        None]
  --parallel-scheduling
                        core.parallel_scheduling for all instances of class
                        core [default: False]
  --assistant           core.assistant for all instances of class core
                        [default: False]
  --RangeBase-of RANGEBASE_OF
                        RangeBase.of for all instances of class RangeBase
  --RangeBase-start RANGEBASE_START
                        RangeBase.start for all instances of class RangeBase
  --RangeBase-stop RANGEBASE_STOP
                        RangeBase.stop for all instances of class RangeBase
  --RangeBase-reverse   RangeBase.reverse for all instances of class RangeBase
                        [default: False]
  --RangeBase-task-limit RANGEBASE_TASK_LIMIT
                        RangeBase.task_limit for all instances of class
                        RangeBase [default: 50]
  --RangeBase-now RANGEBASE_NOW
                        RangeBase.now for all instances of class RangeBase
                        [default: None]
  --RangeDailyBase-start RANGEDAILYBASE_START
                        RangeDailyBase.start for all instances of class
                        RangeDailyBase [default: None]
  --RangeDailyBase-stop RANGEDAILYBASE_STOP
                        RangeDailyBase.stop for all instances of class
                        RangeDailyBase [default: None]
  --RangeDailyBase-days-back RANGEDAILYBASE_DAYS_BACK
                        RangeDailyBase.days_back for all instances of class
                        RangeDailyBase [default: 100]
  --RangeDailyBase-days-forward RANGEDAILYBASE_DAYS_FORWARD
                        RangeDailyBase.days_forward for all instances of class
                        RangeDailyBase [default: 0]
  --RangeHourlyBase-start RANGEHOURLYBASE_START
                        RangeHourlyBase.start for all instances of class
                        RangeHourlyBase [default: None]
  --RangeHourlyBase-stop RANGEHOURLYBASE_STOP
                        RangeHourlyBase.stop for all instances of class
                        RangeHourlyBase [default: None]
  --RangeHourlyBase-hours-back RANGEHOURLYBASE_HOURS_BACK
                        RangeHourlyBase.hours_back for all instances of class
                        RangeHourlyBase [default: 2400]
  --RangeHourlyBase-hours-forward RANGEHOURLYBASE_HOURS_FORWARD
                        RangeHourlyBase.hours_forward for all instances of
                        class RangeHourlyBase [default: 0]
  --Foo-par-sig FOO_PAR_SIG
                        Foo.par_sig for all instances of class Foo [default:
                        hej]
  --Foo-par-insig FOO_PAR_INSIG
                        Foo.par_insig for all instances of class Foo [default:
                        yay]
  --par-sig PAR_SIG     Foo.par_sig [default: hej]
  --par-insig PAR_INSIG
                        Foo.par_insig [default: yay]
```

It would be nice to get rid of at least the `Range*` at least.

erikbern:
what's your suggestion for how to get rid of them though?

Tarrasch:
Have `class Task(): expose_cmdline = True`... `class Config(Task): expose_cmdline = False`. Does that make sense?

erikbern:
Yeah that would also work for `ExternalTask`

I just don't know if we should introduce even more mystical concepts :)

Tarrasch:
There's another thing I'm afraid will happen here @erikbern, while we could use task names to disambiguate tasks with the same task name. Luigi will still create conflicting parameters like this:

```
class TaskA(Task): 
     namespace = 'unique'
     date = luigi.Parameter()


class TaskA(ExternalTask):
     namespace = 'external'
     date = luigi.Parameter()
```

Now both the section `[TaskA] p: value` and the command line parameter for `p` is `--TaskA-p`. Not I wouldn't really mind if "both" were set when you pass the parameter. But I actually think luigi will crash and complain about duplicate parameters in this case.

erikbern:
I think that the namespace will be a part of the config section. If not, we should fix

Tarrasch:
Oh, let me double check. What about for parameters?

erikbern:
for class level cmdline params yes the same

i can fix if not

Tarrasch:
Cool. I'll test and check this on Monday then :)

Tarrasch:
The behavior works as you said @erikbern, I'm all for merging this then.

erikbern:
ok let me know if this leads to any issues. i have some other things i want to simplify with the `ExternalTask`/`Task` dichotomy

PR #881, opened 2015-04-11
Task has to be complete() to be considered DONE

### LINKED ISSUE #814: Yielding task with no output causes infinite loop
This is not an unlikely situation if task that is being yield is dynamic and constructed in runtime. I would expect to correctly yield and empty list of targets and keep moving forward. Do I miss something?

```
#!/usr/bin/env python

import luigi

class NoOutputTask(luigi.Task):
    def run(self):
        print 'Yo ho ho and the bottle of rum!'

    def output(self):
        return []

class EternityTask(luigi.Task):
    def run(self):
        yield NoOutputTask()

if __name__ == '__main__':
    luigi.run()
```

Run:
```
python eternity.py EternityTask --local-scheduler
```

erikbern:
This fixes #814 and also makes the behavior a lot more consistent between Task and ExternalTask

It does introduce a lot of backwards incompatible stuff though.

Let me know what you think about this change. It will break a lot of unit tests that I have to fix. I'm not going to do that until someone tells me it's a good idea :)

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -188,7 +188,13 @@ class TaskProcess(AbstractTaskProcess):
 
             new_deps = self._run_get_new_deps()
 
-            if new_deps is None:
+            if new_deps:
+                status = SUSPENDED
+                logger.info(
+                    '[pid %s] Worker %s new requirements      %s',
+                    os.getpid(), self.worker_id, self.task.task_id)
+
+            elif self.task.complete():
                 status = DONE
                 self.task.trigger_event(
                     Event.PROCESSING_TIME, self.task, time.time() - t0)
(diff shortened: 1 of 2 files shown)


Tarrasch:
So you abandoned this completely? :)

erikbern:
I realized I can merge `TaskProcess` and `ExternalTaskProcess` first. But I definitely want to do this later

aslafy-z:
It's a very good idea!

erikbern:
i might pick this up!

alexshires:
Hi Erik,
Is there any news on this? Otherwise, is there a way to manually re-run the complete method before the task is considered done?
Regards,
Alex

erikbern:
forgot about it. will look at it!

ThePletch:
@erikbern Has this been addressed elsewhere? I'm looking into ways to have tasks that still return false from `complete()` after running marked as failed.

PR #882, opened 2015-04-11
Merge TaskProcess and ExternalTaskProcess

erikbern:


Tarrasch:
Awesome so see so much red! Though  I would appreciate a brief summary why you do this change, and if any behaviour will change.

Tarrasch:
Other than that. LGTM.

erikbern:
I did it because there's a lot of code in worker.py and I want to clean it up.

There's no logic that should change. But I will double check. I noticed a lot of stuff isn't caught by unit tests because my first attempt definitely changed the logic and it didn't fail.

As a follow up I really want to change the logic so that for all tasks, they are not considered DONE unless they are `complete()`. This will break a lot of old stuff but I think it's worth doing it in the name of consistency

Tarrasch:
Sure, makes sense. :)

PR #885, opened 2015-04-13
Add ODBC target.

artiya4u:
Add ODBC target to support more RDBMS target like SqlServer, MySql, Oracle.
Using pyodbc library https://code.google.com/p/pyodbc/ .

### DIFF
--- a/luigi/contrib/odbc.py
+++ b/luigi/contrib/odbc.py
@@ -61,7 +61,7 @@ class ODBCTarget(luigi.Target):
         self.create_marker_table()
 
         if connection is None:
-            connection = self.connect(self.conn_str, autocommit=True)
+            connection = self.connect()
         connection.cursor().execute(
             """INSERT INTO {marker_table} (update_id, target_table, inserted)
                      VALUES (?, ?, ?);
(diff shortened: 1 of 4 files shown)


erikbern:
Looks great but please put it in `luigi.contrib.odbc` instead.

What's the difference from the existing database support? Can we have a unified interface?

gpoulin:
Correct me if I'm wrong, but sqlalchemy support pyodbc. Wouldn't it be better to improve the sqlalchemy target? I should have some time in the following weeks to work on this. Personally, I would deprecate `luigi.contrib.mysql`, `luigi.postgres`, `luigi.contrib.rdbms` and some part of `luigi.contrib.redshift` and try to handle all those cases with sqlalchemy. Just a suggestion.

Tarrasch:
If sqlalchemy covers all of them, then it makes sense to let that target rule them all, as long as it's proven to be stable. :)

artiya4u:
@gpoulin Thanks for suggestion. I would look at sqlalchemy.

Tarrasch:
Cool. Please contribute back your findings as well :)

PR #889, opened 2015-04-15
Hdfs py refactoring

Tarrasch:
I start with a small patch. If I get around to it, I'll move this to `contrib` and then split the file into multiple.


erikbern:
LGTM

PR #893, opened 2015-04-16
do not normalize remote path in RemoteFileSystem

### LINKED ISSUE #891: RemoteFileSystem uses wrong path separator
The RemoteFileSystem uses the client path  separators instead of the remote system path separators.

That means that when trying to write to a Linux RemoteTarget from a Windows Client, the path will be broken (backslashes instead of slashes as path separators, leading to wrong path names).

kutschkem:
fixes #891
This patch means people should/can not use a remote path like "path/.." because that would end up in a "mkdir .."

### DIFF
--- a/luigi/contrib/ssh.py
+++ b/luigi/contrib/ssh.py
@@ -40,6 +40,7 @@ import contextlib
 import os
 import random
 import subprocess
+import posixpath
 
 import luigi
 import luigi.format
@@ -173,7 +174,7 @@ class RemoteFileSystem(luigi.target.FileSystem):
 
     def put(self, local_path, path):
         # create parent folder if not exists
-        normpath = os.path.normpath(path)
+        normpath = posixpath.normpath(path)
         folder = os.path.dirname(normpath)
         if folder and not self.exists(folder):
             self.remote_context.check_output(['mkdir', '-p', folder])


erikbern:
lgtm

gpoulin:
Wouldn't it break backward compatibility for people that expect this function to normalize the path?

(Windows has ssh server?)

kutschkem:
The problem is Windows SSH clients connecting to Linux servers, not the other way around. (Probably Windows has SSH servers, too, but that's not the point). And yes, the proposed fix breaks backwards compatibility for people who expect normalization. Not sure how often normalization is needed, I guess people sometimes use relative paths. Relative remote paths are probably broken if this patch is merged.
Alternatives to not normalizing:
- normalize explicitly as linux file names, and hope the remote system is not windows
- somehow find out what os is on the remote side, and normalize accordingly
- make normalization opt-out

kutschkem:
I found that while the posix module is not available on a non-*nix OS, posixpath is - that means at least one can get back from not normalizing at all to normalizing explicitly as posix path. Windows SSH server exist, but then again Windows is less picky about path separators, I think.

Tarrasch:
The commits can be squashed I think. :)

Nice that you found `posixpath`, will that change backward compatibility? What different behaviour do they have?

gpoulin:
@Tarrasch `posixpath` === `path` on [posix environment](https://docs.python.org/2/library/undoc.html#platform-specific-modules), so no difference. It's just an implementation detail.

Tarrasch:
Cool. I'll be happy to merge this. Just an amend would be nice :)

kutschkem:
Squashing done

PR #897, opened 2015-04-17
remove deps in _serialize_task() when unnecessary

interskh:
I added this a while back in https://github.com/spotify/luigi/pull/484

Then I saw this got reverted in https://github.com/spotify/luigi/commit/4bcb2ca8623f807037db8e012950fef867b30357#diff-ec5c8f6c192cef5a08a5e22aa6273604L405

I put that change back in again. The goal is that when we call task_list() from dashboard, we don't need to load the entire deps for each task.

I also added a test in case we revert that again.


Tarrasch:
LGTM. Thanks for adding the test :)

PR #898, opened 2015-04-19
Add an option to ssh to skip host key check

interskh:


Tarrasch:
LGTM, but I think it's time to start using a NamedTuple here so we don't pass around so many arguments. Would you like to do that @interskh ?

interskh:
Yeah this makes sense. I can make the change.

interskh:
@Tarrasch I made the change. Please take a look. One thing i am not sure is how you guys want to handle backward compatibility. We can potentially keep all the existing args and only make new args into kwargs.

Also noticed `port` is prolly not working on master branch: https://github.com/spotify/luigi/blob/master/luigi/contrib/ssh.py#L134 This PR should fix it too.

Tarrasch:
LGTM. Thanks! :)

erikbern:
actually @Tarrasch what's the benefit of using **kwargs here?

not sure i got it

Tarrasch:
I reasoned that the line

`- def __init__(self, host, username=None, key_file=None, port=None):` 

becomes

`+ def __init__(self, host, **kwargs):`

which is shorter. Originally I intended for it to be

`+ def __init__(self, host, ssh_config):

where `type(ssh_config) === NamedTuple()`. Hmm... @interskh, any reason you went for `**kwargs` over a positional `NamedTuple`?

interskh:
Hmm I don't see benefits of going with NamedTuple over **kwargs. Won't that break more existing code?

Tarrasch:
Right, this is in code exposed to the user. Maybe it should stay as **kwargs then. Hmmm...

PR #899, opened 2015-04-19
add double quote on mapred.job.name

interskh:
so it won't break when there is space in self.task_id sometimes

### DIFF
--- a/luigi/hadoop.py
+++ b/luigi/hadoop.py
@@ -613,7 +613,7 @@ class BaseHadoopJobTask(luigi.Task):
 
     def jobconfs(self):
         jcs = []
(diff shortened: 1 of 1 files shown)


Tarrasch:
Can you add a test case? I don't even see why this would break, I mean we don't use `shell=True` anywhere right?

interskh:
I think this was breaking when I added https://github.com/spotify/luigi/pull/900 - maybe i should merge this as part of https://github.com/spotify/luigi/pull/900

Tarrasch:
I see. Yea, I guess it's easier for traceability of why you made the change if it's in the same PR. Well, they are linked now by GitHub anyway. So I think this is fine. :)

PR #900, opened 2015-04-19
Support running a remote hadoop jar command

interskh:
I am not sure if you guys would like this or whether this could be useful for other people. We have some requirements that needs to submit a hadoop job to a remote server so I added HadoopRemoteJarJobRunner and HadoopRemoteJarJobTask.

### DIFF
--- a/luigi/contrib/hadoop_jar.py
+++ b/luigi/contrib/hadoop_jar.py
@@ -83,7 +83,9 @@ class HadoopJarJobRunner(luigi.contrib.hadoop.JobRunner):
             arglist.append('{}@{}'.format(username, host))
         else:
             arglist = []
-            if not job.jar() or not os.path.exists(job.jar()):
+            if not job.jar():
+                raise HadoopJarJobError("Jar not defined")
+            if not os.path.exists(job.jar()):
                 logger.error("Can't find jar: %s, full path %s", job.jar(), os.path.abspath(job.jar()))
                 raise HadoopJarJobError("job jar does not exist")
 
--- a/luigi/contrib/hadoop_jar.py
+++ b/luigi/contrib/hadoop_jar.py
@@ -55,6 +55,10 @@ def fix_paths(job):
     return (tmp_files, args)
 
 
+class HadoopJarJobError(Exception):
+    pass
+
+
 class HadoopJarJobRunner(luigi.contrib.hadoop.JobRunner):
     """
     JobRunner for `hadoop jar` commands. Used to run a HadoopJarJobTask.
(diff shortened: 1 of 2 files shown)


erikbern:
I think this is great but would actually prefer if this is just a setting on the existing Hadoop job class so that everything else works like usual...

erikbern:
just wanted to ping and see what's the status.

is it possible for you to rebase? would also really appreciate it if you can make the changes in the existing `HadoopJobTask` class and just wrap the hadoop streaming command with an `ssh`

interskh:
Yeah I will make the change

interskh:
Hi @erikbern , I've made it in the existing HadoopJarJobTask. How does it look to you now?

Tarrasch:
Looks overall good to me :)

luigi/contrib/hadoop_jar.py
```diff
@@ -64,23 +64,40 @@ def __init__(self):
         pass
 
     def run_job(self, job):
+        ssh_config = job.ssh()
+        if ssh_config:
+            host = ssh_config.get("host", None)
+            key_file = ssh_config.get("key_file", None)
+            username = ssh_config.get("username", None)
```
Tarrasch: In the documentation you mentioned these to be required fields. So please don't use `get()` and silently ignore missing fields. Instead use `[]` to make sure they exist.

interskh: It is captured later in the if conditions.

Tarrasch: Sure, but `[]` would make the logic simpler right? Also you'll see form the stack trace exactly what attribute you're missing.

interskh:
Made the change :)

erikbern:
looks great! i wonder if there's some simple way to test this... probably not

Tarrasch:
Awesome! Anyway, can you find a way to test this @interskh? At least a little bit?

If definitely not, we can merge this I suppose.

interskh:
Looking at it again, I think I might be able to add some basic tests. However, I am not sure how you guys are running tests these days. Is there a doc for that?

Tarrasch:
Ah, finally made this page describing how to run the tests. https://github.com/spotify/luigi/blob/c2ef921ae355e04251dbc0a1673723514a58bcef/CONTRIBUTING.rst

(ping @erikbern, so you don't duplicate the effort)

interskh:
Hi @Tarrasch , I added some basic tests. But the test failed on something that doesn't appear to be related to my change ```ERROR:   pep8: commands failed```

erikbern:
It's the format checker

```
test/contrib/hadoop_jar_test.py:61:5: E303 too many blank lines (2)
```

Great if you can fix that

interskh:
oh lol! how come i didn't catch that.. Let me fix it.

interskh:
updated - let's see if it passes the test this time :sunny:

Tarrasch:
This looks very good to me now! Anyone else want to review?

erikbern:
LGTM let's merge this

PR #902, opened 2015-04-21
Fix #896: make table_exists case insensitive

### LINKED ISSUE #896: hive table_exists should be case insensitive?
Any thoughts on this one?

In https://github.com/spotify/luigi/blob/master/luigi/contrib/hive.py#L141
(possibly here too, but we're not on CDH: https://github.com/spotify/luigi/blob/master/luigi/contrib/hive.py#L192)

Sometimes we have tables that are defined as capitalized, rather than lower case underscored names.  These are easier to read in code if left capitalized, though hive is case insensitive, and will return them as lower case.

E.g. when checking for an existing table `table = 'FooBar'`, stdout will return with `foobar` and the test will fail

This wasn't an issue in older versions, which just checked for string "does not exist" or "Table not found" in stdout.

Would be easy to fix using `return stdout and table.lower() in stdout` or `return stdout and table.lower() in stdout.lower()`

Let me know your thoughts on this.  I can supply a pull request if necessary.

Thanks,
Lin.

nzlinus:
Check stdout against table.lower() to avoid case issues

### DIFF
--- a/luigi/contrib/hive.py
+++ b/luigi/contrib/hive.py
@@ -138,7 +138,7 @@ class HiveCommandClient(HiveClient):
         if partition is None:
             stdout = run_hive_cmd('use {0}; show tables like "{1}";'.format(database, table))
 
-            return stdout and table in stdout
+            return stdout and table.lower() in stdout
         else:
             stdout = run_hive_cmd("""use %s; show partitions %s partition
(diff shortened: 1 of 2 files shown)


erikbern:
lgtm

PR #903, opened 2015-04-21
Enables searching within groups too large to display in the visualiser

daveFNbuck:
This adds to the task_list api to allow search queries to be sent to the scheduler in order to retrieve a filtered list of tasks when the entire list is too large for the visualiser. This enables search results to be displayed for tasks in these groups, making the visualiser more useful when a lot of tasks are scheduled.

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -137,8 +137,12 @@ class RemoteScheduler(Scheduler):
     def inverse_dep_graph(self, task_id):
         return self._request('/api/inverse_dep_graph', {'task_id': task_id})
 
-    def task_list(self, status, upstream_status):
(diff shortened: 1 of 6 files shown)


Tarrasch:
Ah, so if you have configured a `max-shown-tasks` to be inifinite, you will not be affected by the patch I suppose?

Tarrasch:
Because you only do the api-call that does server-side filtering if you got a truncated result from the query-less api call. Right?

daveFNbuck:
Right. It only applies to groups that have been truncated. This makes using
max-shown-tasks much more tenable.

Tarrasch:
Cool. This looks good to me. I'll merge on another +1.

daveFNbuck:
ping for another +1

Tarrasch:
Thanks @daveFNbuck! :)

PR #904, opened 2015-04-21
Escapes search queries for the jquery selector in the visualiser.

daveFNbuck:
Search queries in the visualiser have been passed into jquery selectors
unmodified, leading to errors when they include special characters. This
corrects the errors by escaping all special characters.


Tarrasch:
LGTM. I guess there's no easy way to have a test case for this ...

daveFNbuck:
We should probably set up a javascript unit testing framework to make this easy to test.

PR #906, opened 2015-04-23
History subsecond timestamp.  Fixes #905

### LINKED ISSUE #905: Tasks that complete in <1s may display as RUNNING or PENDING in task history
Using MySQL backend, I am seeing tasks that have completed in either "RUNNING" or "PENDING" state in the task_history webpage.  Looking at the db there is a row marking these tasks as "DONE" but their timestamps clash with the displayed event.  Clearly the timestamps aren't sub-second precision.

E.g.

    mysql> select id, event_name, ts from task_events where task_id=61;
    +-----+------------+---------------------+
    | id  | event_name | ts                  |
    +-----+------------+---------------------+
    | 181 | PENDING    | 2015-04-22 15:10:02 |
    | 230 | RUNNING    | 2015-04-22 15:12:38 |
    | 231 | DONE       | 2015-04-22 15:12:38 |
    +-----+------------+---------------------+
    3 rows in set (0.00 sec)

Modern MySQL should support this.  I have found this stackoverflow question suggesting the issue is with SQLAlchemy: http://stackoverflow.com/questions/29711102/sqlalchemy-mysql-millisecond-or-microsecond-precision

Other than fixing this in SQLAlchemy it might work to simply sort task_event rows by timestamp *and* primary key, assuming the events are inserted in true order.

stephenpascoe:
Ensure MySQL TIMESTAMP columns are sub-second accurate.

SQLAlchemy has hooks to adapt column types to the dialect using the compiler extension.
Test included, which assumes a running MySQL.  Not sure how this fits with your test framework but it has demonstrated the fix works here.

### DIFF
--- a/test/db_task_history_test.py
+++ b/test/db_task_history_test.py
@@ -96,7 +96,7 @@ class MySQLDbTaskHistoryTest(unittest.TestCase):
         self.run_task(task)
 
         task_record = self.history.find_all_by_name('DummyTask').next()
-        print task_record.events
+        print (task_record.events)
         self.assertEqual(task_record.events[0].event_name, DONE)
 
     def run_task(self, task):
--- a/test/db_task_history_test.py
+++ b/test/db_task_history_test.py
@@ -82,3 +82,24 @@ class DbTaskHistoryTest(unittest.TestCase):
         self.history.task_scheduled(task.task_id)
         self.history.task_started(task.task_id, 'hostname')
         self.history.task_finished(task.task_id, successful=True)
+
+
+class MySQLDbTaskHistoryTest(unittest.TestCase):
+
+    @with_config(dict(task_history=dict(db_connection='mysql://@localhost/luigi_test')))
+    def setUp(self):
+        self.history = DbTaskHistory()
+
+    def test_subsecond_timestamp(self):
+        # Add 2 events in <1s
+        task = DummyTask()
+        self.run_task(task)
+
+        task_record = self.history.find_all_by_name('DummyTask').next()
+        print task_record.events
+        self.assertEqual(task_record.events[0].event_name, DONE)
+
+    def run_task(self, task):
+        self.history.task_scheduled(task.task_id)
+        self.history.task_started(task.task_id, 'hostname')
+        self.history.task_finished(task.task_id, successful=True)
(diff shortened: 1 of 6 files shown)


Tarrasch:
Cool @stephenpascoe , we use `tox` for running the tests. It seems like Travis CI has some comments on your tests.

Overall this patch looks good and I'll merge once you make Travis green. :)

test/db_task_history_test.py
```diff
@@ -82,3 +82,24 @@ def run_task(self, task):
         self.history.task_scheduled(task.task_id)
         self.history.task_started(task.task_id, 'hostname')
         self.history.task_finished(task.task_id, successful=True)
+
+
+class MySQLDbTaskHistoryTest(unittest.TestCase):
+
+    @with_config(dict(task_history=dict(db_connection='mysql+pymysql://travis@localhost/luigi_test')))
```
gpoulin: We are already using [mysql-connector](https://github.com/spotify/luigi/blob/master/luigi/contrib/mysqldb.py#L25), maybe we could stick with it? (You would need to modify tox to add `--allow-external mysql-connector-python` to [install_command](https://tox.readthedocs.org/en/latest/config.html?highlight=install_command#confval-install_command=ARGV))

Tarrasch:
Thanks for the comment @gpoulin! :)

Let us know when you've fixed that @stephenpascoe and when you've melded your last "Various changes" commits. You should be able to run `tox` locally btw! :)

stephenpascoe:
It turns out that MySQL 5.5 is the standard version in many ubuntu setups but I was testing with 5.6.  Subsecond timestamps are not supported in <5.6.  I am reverting to using the PK as a secondary search term as a work-around.

test/requirements.txt
```diff
@@ -8,3 +8,4 @@ boto
 sqlalchemy
 elasticsearch
 snakebite>=2.5.2
+mysql-connector-python
```
Tarrasch: fix newline :)

Tarrasch:
Cool, can you Meld your commits? The change looks pretty small but there are 9 commits ...

stephenpascoe:
I'm trying to meld these commits and ended up deleting my remote branch.  I'll re-submit the PR with the melded commits.

PR #907, opened 2015-04-23
Add more tests to test/parameter_test.py

Tarrasch:
Nothing special.  I just played around with the new parameter system and
thought that I had found a bug.  Luckily I found out that I was wrong
but at least I produced some tests I think are worth contributing.


erikbern:
excellent!

PR #911, opened 2015-04-27
Add a hard timeout for tasks.

boosh:
I've added a setting that allows a hard time limit for tasks to be set, ignoring the retry window logic. That's difficult to use when there are lots of tasks running in parallel which take different amounts of time.

Note: Please review this code, and if you're happy I'll write tests then. 

Thanks.


erikbern:
looks good by test please!

boosh:
OK will do :+1:

erikbern:
cool

luigi/scheduler.py
```diff
@@ -90,6 +90,8 @@ class scheduler(Config):
                                             config_path=dict(section='scheduler', name='disable-window-seconds'))
     disable_failures = parameter.IntParameter(default=None,
                                               config_path=dict(section='scheduler', name='disable-num-failures'))
+    disable_hard_timeout = parameter.TimeDeltaParameter(default=None,
+                                              config_path=dict(section='scheduler', name='disable-hard-timeout'))
```
Tarrasch: @erikbern, not related to this review. But why are these setting the `config_path`. Isn't it automatic to be `dict(section=$className, name=$fieldName)`?

erikbern: It should be automatic. Notice however it's spaces not underscore.

I'm not sure if I ever fixed it so that it tries both. Maybe I did (in which case the `config_path` argument does not matter)

boosh:
Tests added :+1:

boosh:
OK I've changed the setting to take the number of seconds as an int instead of a timedelta.

Tarrasch:
Cool, I think in this case all the 4 commits can be squashed into one. Would that make sense @boosh?

boosh:
How do I do that?

erikbern:
`git rebase -i HEAD~4`

boosh:
OK done, but for some reason it included the merge. I'm not really a rebaser as you might have gathered...

Tarrasch:
Erm now there are six commits. Something didn't work out. Note that you must eventually `--force` push.

boosh:
FFS

boosh:
See #916 instead.

PR #912, opened 2015-04-28
Move hadoop to contrib

Tarrasch:
These modules definetly are not essential for the scheduler. So I move them to contrib.

I thought 3 months is enough of a deadline.

Some more commits changing the `import luigi.hadoop_blah` will follow soon. (I want travis to trigger this build first though to make sure tests build)


erikbern:
This is great but I think we probably need to give it more than 3 months... maybe 6. No big deal though

what about Hdfs support?

Tarrasch:
>This is great but I think we probably need to give it more than 3 months... maybe 6. No big deal though 

Well, if somebody else complains we can change this.

> what about Hdfs support?

You mean `hdfs.py` and family? I'll have a seperate PR for that.

Tarrasch:
@erikbern, So how does the docs verification step for building the docs work?

See https://travis-ci.org/spotify/luigi/jobs/60362837#L316

Seems like I should ignore that issue for now, as merging this PR will create the link itself. Unbreaking the build. Hmm..

Tarrasch:
Or for next time, I wait with changing the references in the docs.

erikbern:
Yeah don't worry about this, will work if you merge & re-trigger

PR #914, opened 2015-04-28
Move luigi.hdfs to contrib

Tarrasch:
I'll stick with only this commit, lets merge this and #912 so I can fix the `import`s without getting conflicts.


Tarrasch:
Tests pass. Merging.

PR #916, opened 2015-04-28
Add a hard timeout for tasks

boosh:


erikbern:
looks great let's merge

Tarrasch:
I verified that the test failures are unrelated to this change. :)

gpoulin:
@Tarrasch The tests that are failing are due to this change. The tests that were added assume that `self.sch` is a `CentralScheduler` but those tests are also run with a `RemoteScheduler`.

Tarrasch:
Hmm, yea, I was confused by Travis here! When looking at the page for "https://travis-ci.org/spotify/luigi". I assumed it would show the latest build of luigi at that time. But it shows the latest PR. So I were looking at the same set of logs when compaing "before" and "after". D'oh!

Tarrasch:
Hmm, why are they not failing with `tox -e py27-nonhdfs test/central_planner_test.py` then?

I'll submit a patch to disble the tests.

gpoulin:
Because they are reused [here](https://github.com/spotify/luigi/blob/master/test/rpc_test.py#L27). Maybe we could simply disable them in `rpc_test`.

Tarrasch:
Awesome, @gpoulin  managed to disable the failing tests this patch introduced. See #923.

Still these tests are a bit unstable and shaky. When I ran this locally, I got this error :-/

```
FAIL: test_task_first_failure_time (central_planner_test.CentralPlannerTest)
  File "/home/arash/spotify/repos/luigis/github_luigi/test/central_planner_test.py", line 713, in test_task_first_failure_time
    self.assertGreater(time_after_failure, test_task.failures.first_failure_time)
AssertionError: 1430312828.010096 not greater than 1430312828.010096

```

Is it because my laptop happened to be ridiculously fast at that moment? @boosh ?

boosh:
Hmm we should probably change the assertion to assertGreaterEqual to fix this...

Tarrasch:
Cool, would you like to do that? :)

boosh:
I'm on it now :+1:

boosh:
See #924

Tarrasch:
@boosh, thanks for this patch. Just of out of curiosity. Why did you implement this? What's the actual use case? Do you still use luigi and how does this feature serve you today?

boosh:
@Tarrasch Hi yes, still using luigi. 

The use case for this was that certain third parties provide logs to us hourly. However, if their systems have issues (which seems to happen with some frequency) then instead of going back to populate missed logs (e.g. the log file for 3am) they'll just roll the data that would have been in that file into the next one they send (e.g. the one for 4am will contain the data for 3am and 4am). This meant that we were having luigi tasks waiting indefinitely for files that would never appear (e.g. for 3am in this scenario).

This patch means that we can set a parameter to say "keep trying for 2 days then just die". We monitor the output of our important workflows but for tasks like the above, since it's documented behaviour with our third party, there's not much else we can do but to just ignore these failures. Since we don't want processes piling up on our scheduler indefinitely, killing them after a hard timeout is the way to go.

Tarrasch:
Thanks! Your explanation serves as documentation for this change. And I'm glad that luigi still serves you well ^^

erikbern:
I don't see why this couldn't have been solved by some custom logic in the `requires()` method?

boosh:
@erikbern For a single task it could have. But it seems sensible to apply it to all tasks combined with some monitoring to prevent tasks with dependencies that are never likely to be satisfied from piling up.

PR #915, opened 2015-04-28
Disable tests that fail on Travis CI

Tarrasch:
I get an error like:

     ProgrammingError: (mysql.connector.errors.ProgrammingError) 1045 (28000): Access denied for user 'travis'@'localhost' (using password: NO)

We discussed this a bit in https://github.com/spotify/luigi/pull/913

PR #917, opened 2015-04-29
Skip DbTaskHistory test when is not MySQL

stephenpascoe:
See #915. If DbTaskHistory() fails to initialise assume MySQL is unavailable and skip.  This will allow the tests to pass in a vanilla tox environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,8 @@ before_script:
   - cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys
   - ssh -o StrictHostKeyChecking=no localhost true
 
-  - mysql -e 'create database IF NOT EXISTS luigi_test;' -uroot
+  # Create mysql database if possible but fail silently if not available.
+  - mysql -e 'create database IF NOT EXISTS luigi_test;' -uroot || true
 
 script:
   - tox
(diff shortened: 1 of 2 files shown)


Tarrasch:
Thanks @stephenpascoe ! :)

Tarrasch:
By the way, look at the result from the #915 patch: https://travis-ci.org/spotify/luigi/builds/60403388

There are still two failures happening due to it failing to setup mysql in the init setup, before any python code is run. Is that the case? I think you understand what happens better than I do. :)

stephenpascoe:
No idea why this is happening:

    ERROR: test_task_first_failure_time (rpc_test.RPCTest)
      File "/home/travis/build/spotify/luigi/.tox/py33-nonhdfs/lib/python3.3/site-packages/tornado/testing.py", line 120, in __call__
        result = self.orig_method(*args, **kwargs)
      File "/home/travis/build/spotify/luigi/test/central_planner_test.py", line 705, in test_task_first_failure_time
        test_task = self.sch._state.get_task('A')
    nose.proxy.AttributeError: 'RemoteScheduler' object has no attribute '_state'
    -------------------- >> begin captured logging << --------------------
    tornado.access: INFO: 200 POST /api/add_task (127.0.0.1) 0.75ms
    --------------------- >> end captured logging << ---------------------

Tarrasch:
I get the exact same error for my very orthogonal patch[1]. So it's not due to this. Thanks for fixing this @stephenpascoe! :)

[1]: https://github.com/spotify/luigi/pull/914

PR #927, opened 2015-04-29
adding ForwardAgent=yes for scp

iralls:
Very basic addition to allow for scp between two remote machines if ssh-agent key forwarding is setup.

You can pass the full host:file path as your "local_path" when using put(), rather than the path to the physically local file. A little bit of a hack, but works.
ex:

```
RemoteFileSystem(
    "somehost",
    username="myuser",
    ...,
    no_host_key_check=True
).put("myuser@host1:/tmp/file", "/tmp/file")
```

Might be good to eventually allow for many ssh_config values to be passed in a list

```
[
  "StrictHostKeyChecking=no",
  "UserKnownHostsFile=/dev/null", 
  "ForwardAgent=yes",
  ...
]
```


Tarrasch:
LGTM. Not sure how this affects backward compatibility though. Someone else who uses this contrib feature should comment.

PR #928, opened 2015-04-30
HdfsTarget: Use self.fs.<cmd> rather than <cmd>

Tarrasch:
I would consider this a bug fix that the HdfsTarget ignores the passed
`fs` parameter for half of it's invocations. But the actual reason I'm
writing this patch is to decouple this so I can refactor easier.


Tarrasch:
Merging, as tests are passing I want to avoid (merge) conflicts. :)

Tarrasch:
Please audit if you want to :)

ulzha:
Great +1

PR #929, opened 2015-04-30
hdfs.py refactoring: Move out configuration

Tarrasch:
By putting it in a seperate file, it's easier to refactor the rest of
contents of that file.

### DIFF
--- a/luigi/contrib/hdfs/config.py
+++ b/luigi/contrib/hdfs/config.py
@@ -1,0 +1,128 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2012-2015 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


erikbern:
nice. iirc there's some legacy stuff in the configuration that we should drop soon

Tarrasch:
Not sure about that. I did some in https://github.com/spotify/luigi/pull/889 I know. We should however use the new configuration system for the `[hadoop]` section.

By the way, I *really* like the new unified system for doing configuration. It's really nice! :)

luigi/contrib/hdfs/config.py
```diff
@@ -0,0 +1,128 @@
...
+import os
+import getpass
+
+try:
+    from urlparse import urlparse, urlunparse
+except ImportError:
+    from urllib.parse import urlparse, urlunparse
+
+
+class hdfs(luigi.Config):
```
ulzha: This leading to the repetitious "hdfs.config.hdfs()" is unfortunate. And how come it has to have a lovercase name?

Tarrasch: It's the new configuration system of luigi. This is why you get command line parameters for free, like `--hdfs-client` is `--<classname>-<paramname>`

PR #930, opened 2015-04-30
More hdfs.py refactoring

Tarrasch:
I think the `clients.py` can be split up even further, but this is a good start. At least the formats and the target and configuration are not put in seperate files and there are no cyclic dependencies.

Please review. :)


Tarrasch:
Yay, tests passes, want to review @erikbern ?

luigi/contrib/hdfs/__init__.py
```diff
@@ -19,856 +19,58 @@
...
-    from urlparse import urlsplit
-except ImportError:
-    from urllib.parse import urlsplit
-import warnings
 
-from luigi import six
+Currently (4th May) the :py:mod:`luigi.contrib.hdfs` module is under
+reorganization. We recommend importing the reexports from
+:py:mod:`luigi.contrib.hdfs` instead of the sub-modules, as we're not yet sure
+how the final structure of the sub-modules will be yet. Eventually this module
```
ulzha: "yet" twice

luigi/contrib/hdfs/__init__.py
```diff
@@ -19,856 +19,58 @@
...
+will be empty and you'll have to import directly from the sub modules like
+:py:mod:`luigi.contrib.hdfs.config`.
+"""
 
-import luigi.contrib.target
-import luigi.format
-from luigi.target import FileAlreadyExists, FileSystem, FileSystemTarget
+# Here follows a series of deprecated imports that have been moved out there own files.
```
ulzha: Confusing comment, unneeded.
* It looks more like saying that the contrib style of imports is deprecated, but that's the new shiny style of imports, right?
* Everything phrased as "here follows" is prone to become outdated because people who contribute to something far below in this code can't be expected to notice the comment at the top and add a "here the series of deprecated imports end. Here follow my lovely imports unrelated to that refactor".

Tarrasch: Good point. I just remove this.

Tarrasch:
I fix your comments in a seperate PR. It'll take at least an hour untangling the merge conflicts otherwise.

PR #931, opened 2015-05-01
Add ability to pass in connect_args to sql alchemy.

edkurowski:
I need to be able to pass some connection config into sqlalchemy.  I think this is the easiest way to do it.

### DIFF
--- a/test/contrib/sqla_test.py
+++ b/test/contrib/sqla_test.py
@@ -42,6 +42,7 @@ TEMPDIR = tempfile.mkdtemp()
 SQLITEPATH = os.path.join(TEMPDIR, "sqlatest.db")
 CONNECTION_STRING = "sqlite:///%s" % SQLITEPATH
 TASK_LIST = ["item%d\tproperty%d\n" % (i, i) for i in range(10)]
+CONNECT_ARGS = {'timeout': 5.0}
 
 
 class BaseTask(luigi.Task):
@@ -62,6 +63,7 @@ class SQLATask(sqla.CopyToTable):
         (["property", sqlalchemy.String(64)], {})
     ]
     connection_string = CONNECTION_STRING
+    connect_args = CONNECT_ARGS
     table = "item_property"
     chunk_size = 1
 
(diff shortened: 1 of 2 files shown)


Tarrasch:
Ah, here's the doc I see: http://docs.sqlalchemy.org/en/rel_1_0/core/engines.html#custom-dbapi-args

But shouldn't there be any way to set `connection_args` in the `__init__` of the class?

edkurowski:
Ooops, forgot to put that in.

Tarrasch:
That was fast! :D

Tarrasch:
If you could add a test for this as well. That would be appreciated!

edkurowski:
Sure.  I think I might update CopyToTable as well to pass this into the target it creates.

test/contrib/sqla_test.py
```diff
@@ -42,6 +42,7 @@
 SQLITEPATH = os.path.join(TEMPDIR, "sqlatest.db")
 CONNECTION_STRING = "sqlite:///%s" % SQLITEPATH
 TASK_LIST = ["item%d\tproperty%d\n" % (i, i) for i in range(10)]
+CONNECT_ARGS = {'separator': '\b'}
```
Tarrasch: Shouldn't this be used anywhere?

edkurowski: Yes, wow, today is not my day for properly committing things.

edkurowski:
So I'm not exactly sure what the best way to test this is.  It does not appear that sqlite has many options, and I could not find a good way to extract it from the engine object.  I added the connect_args to the SQLATask that is used in several other tests, and it does appear that the tests are passing.

Tarrasch:
How are you supposed to use the CopyToTable task here? Write a test or at least documentation of how the `connection_args` are set. It's also not clear to me why the default value sometimes is `{}` and sometimes `None`.

edkurowski:
Yeah, the inconsistency was not correct, I have fixed that.  I added some info in the documentation about it.  Its difficult to dive into too much detail because the way it works is very database specific.  The tests are setting the field, so it is technically being used.  Unfortunately, sqlite does not really have many connection settings, so I'm not sure how to verify this in a test.

Tarrasch:
Hmm, it seems like there's a test failure. Will you look at it?

edkurowski:
Are the tests in a good state right now?  I'm getting failures when running all the tests on master.  When I run only the sqla tests, they all pass.
```
PYTHONPATH=. nosetests
```
Is am running the tests correctly?

Tarrasch:
No, you should use the `tox` command. Something like `PYTHONPATH='' tox -e py27-nonhdfs test/my_test.py`

Tarrasch:
Looks better now! :)

It seems like there's an actual test using this stuff now. Good job! Should I merge?

erikbern:
lgtm

edkurowski:
If you were asking me, then yes I think you should merge. :)

PR #932, opened 2015-05-04
Fix deps py

Tarrasch:
Ok, this PR actually doesn't work yet. There's some bug when registering tasks it seems. So it will complain like this:

    deps.py: error: unrecognized arguments: --upstream-task-family hello

I'll look into this more later.


Tarrasch:
God I hate python sometimes...

Anyway. The parameters work now, so please review. :)

Tarrasch:
@rantav @zenlucent feel free to review, otherwise I'll just merge.

PR #934, opened 2015-05-05
Drop python 2.6 support

Tarrasch:
See this email thread as well:

https://groups.google.com/forum/#!topic/luigi-user/X8YA_JpQ8dQ

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,9 @@ env:
   matrix:
     - TOXENV=pep8
     - TOXENV=docs
-    - TOXENV=py26-nonhdfs
     - TOXENV=py27-nonhdfs
     - TOXENV=py33-nonhdfs
     - TOXENV=py34-nonhdfs
-    - TOXENV=py26-cdh
     - TOXENV=py27-cdh
     - TOXENV=py33-cdh
     - TOXENV=py34-cdh
--- a/debian/control
+++ b/debian/control
@@ -3,8 +3,7 @@ Section: libs
 Priority: extra
 Maintainer: Elias Freider <freider@spotify.com>
 Build-Depends:	debhelper (>= 7),
-		python-all (>= 2.6.6-3~),
-		python-argparse,
+		python-all (>= 2.7),
 		python-pip,
 		python-psycopg2,
 		python-pyparsing,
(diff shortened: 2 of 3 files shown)


Tarrasch:
Should I assume no complaints as OK to merge? :)

@jcrobak @daveFNbuck

Tarrasch:
Merging for now to speed up Travis build times. At least. Anyone feel free to audit this decision.

erikbern:
Is Python 2.6 support a big deal to maintain? I don't think it is

Tarrasch:
It was a big deal when Spotify was relying on it. I just wanted to give a fair chance for other big users of luigi to object.

By the way @erikbern, it's time for you to re-introduce your patch where you were about to remove OptParse support. Now there's no blockers to remove it :D

erikbern:
What I meant is what's the cost of maintaining 2.6 support? Is it very high? If so, we should drop it. If it's low, we should keep it.

It seems low to me but I could be wrong

erikbern:
The optparse issue was actually in 2.7.9 btw

erikbern:
afaik some things can be simplified if we drop 2.6, like we can use `functools.total_ordering` for `DateInterval`

Tarrasch:
Ah, I see what you mean now. I think we shouldn't claim to maintain it without anybody using it. It could lead to strange bugs that tests don't cover and affect only 2.6. As I mentioned in my mail to luigi-user@.

erikbern:
Well... we have 2.6 tests running in Travis so shouldn't that be almost enough? 

But if no one uses it then sure let's just drop it. I don't care too much

Tarrasch:
Nope, evidently it's not enough: https://github.com/spotify/luigi/pull/793/files

The other major benefit is the speedup in Travis build times. We're talking about a few minutes at least it seems per PR iteration. It's really annoying to wait for all the tests to pass when developing heavily on luigi.

erikbern:
But 2.6 tests are run in parallel right?

Anyway let's drop it and see if people complain... should be easy to add back alter

daveFNbuck:
2.6 is the main cause of test failures for my commits, because I like to
use test functions that aren't present. Glad to see it leaving.

Tarrasch:
> The optparse issue was actually in 2.7.9 btw

@erikbern: Does this mean that we need to maintain optparse still? I'm not sure what you meant there. I really want to remove it!

In #612 it sounded it was just because Spotify needed it for whatever reason. I'm pretty sure that's not the case for Spotify anymore. Any reason to not repoen #612?

erikbern:
We can prob just remove optparse, not sure why anyone would use it

PR #937, opened 2015-05-07
Split up hdfs clients

Tarrasch:


erikbern:
After you are done with all the HDFS refactoring it would be good to let it sit for a while then do a release to PyPI after a few weeks

Tarrasch:
Sure. But I'm not done yet, there will be a few more patches.

Just to be clear, you want it to sit for a while so some experimental users can try it out?

erikbern:
Yeah the code in Git is kind of bleeding edge and there's enough people running it to give it some exposure to real testers. I wouldn't want to upload it to PyPI without having it out on Github for a few weeks

Tarrasch:
Makes total sense. I'll let you know when I've settled with my refactoring :)

erikbern:
awesome

Tarrasch:
We should enable doctests by the way. Any ideas on how to do that @berkerpeksag? :)

Tarrasch:
Wohoo, tests finally passing (except intermittent doc failures), so this superseedes and closes #941. Will merge on Monday.

berkerpeksag:
@Tarrasch  I couldn't see any doctests, but it can be enabled by setting NOSE_DOCTEST_TESTS env variable at this line: https://github.com/spotify/luigi/blob/master/tox.ini#L21

By the way, we've converted doctests to proper test cases in CPython a while ago (except the tests for doctest itself), so if you have a chance to ignore them, I'd ignore them :) doctest was a good idea in 1999 :)

luigi/contrib/hdfs/error.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+
+
+class HDFSCliError(Exception):
+
+    def __init__(self, command, returncode, stdout, stderr):
+        self.returncode = returncode
+        self.stdout = stdout
+        self.stderr = stderr
+        msg = ("Command %r failed [exit code %d]\n" +
```
berkerpeksag: Nitpick: `+`s can be dropped.

```py
msg = ("foo %s"
       "bar %s") % (foo, bar)
```

is a valid Python code.

luigi/contrib/hdfs/hadoopcli_clients.py
```diff
@@ -0,0 +1,288 @@
...
+    """
+    version = hdfs_config.get_configured_hadoop_version()
+    if version == "cdh4":
+        return HdfsClient()
+    elif version == "cdh3":
+        return HdfsClientCdh3()
+    elif version == "apache1":
+        return HdfsClientApache1()
+    else:
+        raise Exception("Error: Unknown version specified in Hadoop version"
```
berkerpeksag: Nitpick: This could be a ValueError, too.

luigi/contrib/hdfs/hadoopcli_clients.py
```diff
@@ -0,0 +1,288 @@
...
+            for line in stderr.split('\n'):
+                if not_found_re.match(line):
+                    return False
+            raise hdfs_error.HDFSCliError(cmd, p.returncode, stdout, stderr)
+
+    def rename(self, path, dest):
+        parent_dir = os.path.dirname(dest)
+        if parent_dir != '' and not self.exists(parent_dir):
+            self.mkdir(parent_dir)
+        if type(path) not in (list, tuple):
```
berkerpeksag:
```py
if not isinstance(path, (list, tuple)):
```

luigi/contrib/hdfs/hadoopcli_clients.py
```diff
@@ -0,0 +1,288 @@
...
+            path = [path]
+        else:
+            warnings.warn("Renaming multiple files at once is not atomic.")
```
berkerpeksag: It's probably not important for this use case, adding `stacklevel=2` would be nice.

luigi/contrib/hdfs/hadoopcli_clients.py
```diff
@@ -0,0 +1,288 @@
...
+        Currently, the hadoop cli does not support this operation.
+
+        We keep the interface simple by just aliasing this to
+        normal rename and let individual implementations redefine the method.
+
+        rename2 -
+        https://github.com/apache/hadoop/blob/ae91b13/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/protocol/ClientProtocol.java
+        (lines 483-523)
+        """
+        warnings.warn("Configured HDFS client doesn't support rename_dont_move, using normal mv operation instead.")
```
berkerpeksag: Same as above. Can these be log.warning instead?

luigi/contrib/hdfs/hadoopcli_clients.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def getmerge(self, path, local_destination, new_line=False):
+        if new_line:
+            cmd = load_hadoop_cmd() + ['fs', '-getmerge', '-nl', path, local_destination]
+        else:
+            cmd = load_hadoop_cmd() + ['fs', '-getmerge', path, local_destination]
+        self.call_check(cmd)
+
+    def mkdir(self, path, parents=True, raise_if_exists=False):
+        if (parents and raise_if_exists):
```
berkerpeksag: Nitpick: `()` can be removed.

luigi/contrib/hdfs/hadoopcli_clients.py
```diff
@@ -0,0 +1,288 @@
...
+    """
+
+    def mkdir(self, path):
+        """
+        No -p switch, so this will fail creating ancestors.
+        """
+        try:
+            self.call_check(load_hadoop_cmd() + ['fs', '-mkdir', path])
+        except hdfs_error.HDFSCliError as ex:
+            if "File exists" in ex.stderr:
```
berkerpeksag: I'm not familiar with the codebase, but it seems like the exception handling logic could be moved to a private method to avoid code duplication. See line 188 for example.

luigi/contrib/hdfs/snakebite_client.py
```diff
@@ -0,0 +1,274 @@
...
+    """
+
+    def __init__(self):
+        super(SnakebiteHdfsClient, self).__init__()
+        self._bite = None
+        self.pid = -1
+
+    @staticmethod
+    def list_path(path):
+        if isinstance(path, list) or isinstance(path, tuple):
```
berkerpeksag: Nitpick:

```py
if isinstance(path, (list, tuple)):
```

luigi/contrib/hdfs/snakebite_client.py
```diff
@@ -0,0 +1,274 @@
...
+            return path
+        if isinstance(path, str) or isinstance(path, unicode):
```
berkerpeksag: If I'm not missing something, the unicode part of the if is not Python 3 friendly.

The following code will be check for `basestring` (equivalent to `(str, unicode)`) in Python 2, and `str` in Python 3:

```py
if isinstance(path, six.string_types):
```

test/minicluster.py
```diff
@@ -52,15 +52,15 @@ def setUp(self):
         self.fs = luigi.contrib.hdfs.client
         cfg_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "testconfig")
         hadoop_bin = os.path.join(os.environ['HADOOP_HOME'], 'bin/hadoop')
-        # TODO: Remove this hacky patching (@tarrasch 29-Apr-2015)
-        cmd = lambda: [hadoop_bin, '--config', cfg_path]
-        luigi.contrib.hdfs.load_hadoop_cmd = cmd
-        luigi.contrib.hdfs.clients.load_hadoop_cmd = cmd
-        luigi.contrib.hdfs.format.load_hadoop_cmd = cmd
+        cmd = "{0} --config {1}".format(hadoop_bin, cfg_path)
```
berkerpeksag: Since Python 2.6 support is removed, this can be written as:

```py
cmd = "{} --config {}".format(...)
```

Tarrasch: Good point! :)

luigi/contrib/hdfs/snakebite_client.py
Tarrasch: Aha, nice catch. Comments @gpoulin?

Tarrasch:
@berkerpeksag, about doctest. I really thought it still was a valid use case. You want to have documentation of how a method works, but you also want it to be statically checked. The idea makes sense to me at least.

In this particular case (#940), I think doctests really makes sense.

@berkerpeksag . I'll merge this anyway, and incorporate your changes into seperate commits. After all this PR was only to be about splitting up the modules not changing the classes.

berkerpeksag:
Yes, I agree that #940 is a good use case.

luigi/contrib/hdfs/hadoopcli_clients.py
Tarrasch: Not in this case right? The problem is actually with `rename_dont_move`, not it's caller.

Also, why `log.warning`? :)

berkerpeksag: Most of the time warnings produce too much information, and usually these information are misleading or unnecessary(e.g. line info, file name).

> Also, why log.warning? :)

Oh, I was wrong :) I should have read the code more carefully. In this case, logger.info or logger.debug would be more useful since the warning doesn't have any effect in program's workflow. It's just a message that can be useful for debugging(or for similar reasons). Otherwise, it would be better to raise an exception.

PR #938, opened 2015-05-07
CascadingClient: Avoid scary stack trace

Tarrasch:
The idea is that most users react to stack traces. The very common
scenario for the CascadingClient is that it's totally fine for the
non-last clients in the cascading chain to fail. So when there's an
error, they are likely to get blind by this actually harmless stack
trace.


Tarrasch:
No complaints ---> Merging

PR #939, opened 2015-05-08
Support DataInterchange with json

mvj3:
Our team use JSON format as data warehouse basic data format. We found that luigi use Python's builtin `eval` function to load data from `mapper` into `reducer`, it's slower than `json.loads` function.

Here provides a solution that adding a `data_interchange_format` attribute to Task class,  default value is "python", and another one is "json". The limitations are `set` data structure is not supported.

P.S. This patch treat `str` and `unicode` as the same type, see it at `test_unicode_job` test case.

### DIFF
--- a/luigi/hadoop.py
+++ b/luigi/hadoop.py
@@ -22,6 +22,5 @@ luigi.hadoop has moved to :py:mod:`luigi.contrib.hadoop`
 import warnings
 
 from luigi.contrib.hadoop import *
-
 warnings.warn("luigi.hadoop module has been moved to luigi.contrib.hadoop",
               DeprecationWarning)
--- a/.gitignore
+++ b/.gitignore
@@ -62,9 +62,11 @@ pip-delete-this-directory.txt
 htmlcov/
 .tox/
 .coverage
+.coverage.*
 .cache
 nosetests.xml
 coverage.xml
+my_dir
 
 # Translations
 *.mo
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,3 +48,4 @@ after_failure:
 branches:
   only:
     - master
+    - support_json_deserialization
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -46,6 +46,7 @@ import tempfile
 import warnings
 from hashlib import md5
 from itertools import groupby
+from cached_property import cached_property
 
 from luigi import six
 
@@ -677,10 +678,18 @@ class BaseHadoopJobTask(luigi.Task):
         else:
             return super(BaseHadoopJobTask, self).on_failure(exception)
 
+DataInterchange = {
+    "python": {"serialize": str, "deserialize": eval},
+    "json": {"serialize": json.dumps, "deserialize": json.loads}
+}
+
 
 class JobTask(BaseHadoopJobTask):
     n_reduce_tasks = 25
     reducer = NotImplemented
+
+    # available formats are "python" and "json".
+    data_interchange_format = "python"
 
     def jobconfs(self):
         jcs = super(JobTask, self).jobconfs()
@@ -689,6 +698,14 @@ class JobTask(BaseHadoopJobTask):
         else:
             jcs.append('mapred.reduce.tasks=%s' % self.n_reduce_tasks)
         return jcs
+
+    @cached_property
+    def serialize(self):
+        return DataInterchange[self.data_interchange_format]['serialize']
+
+    @cached_property
+    def deserialize(self):
+        return DataInterchange[self.data_interchange_format]['deserialize']
 
     def init_mapper(self):
         pass
@@ -737,7 +754,7 @@ class JobTask(BaseHadoopJobTask):
         """
         for output in outputs:
             try:
-                print("\t".join(map(str, flatten(output))), file=stdout)
+                print("\t".join(map(self.serialize, flatten(output))), file=stdout)
             except:
                 print(output, file=stderr)
                 raise
@@ -876,8 +893,8 @@ class JobTask(BaseHadoopJobTask):
         """
         Iterate over input, collect values with the same key, and call the reducer for each unique key.
         """
-        for key, values in groupby(inputs, key=lambda x: repr(x[0])):
-            for output in reducer(eval(key), (v[1] for v in values)):
+        for key, values in groupby(inputs, key=lambda x: x[0]):
+            for output in reducer(key, (v[1] for v in values)):
                 yield output
         if final != NotImplemented:
             for output in final():
@@ -917,7 +934,7 @@ class JobTask(BaseHadoopJobTask):
         Yields a tuple of python objects.
         """
         for input_line in input_stream:
-            yield list(map(eval, input_line.split("\t")))
+            yield list(map(self.deserialize, input_line.split("\t")))
 
     def internal_writer(self, outputs, stdout):
         """
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -679,8 +679,12 @@ class BaseHadoopJobTask(luigi.Task):
             return super(BaseHadoopJobTask, self).on_failure(exception)
 
 DataInterchange = {
-    "python": {"serialize": str, "deserialize": eval},
-    "json": {"serialize": json.dumps, "deserialize": json.loads}
+    "python": {"serialize": str,
+               "internal_serialize": repr,
+               "deserialize": eval},
+    "json": {"serialize": json.dumps,
+             "internal_serialize": json.dumps,
+             "deserialize": json.loads}
 }
 
 
@@ -702,6 +706,10 @@ class JobTask(BaseHadoopJobTask):
     @cached_property
     def serialize(self):
         return DataInterchange[self.data_interchange_format]['serialize']
+
+    @cached_property
+    def internal_serialize(self):
+        return DataInterchange[self.data_interchange_format]['internal_serialize']
 
     @cached_property
     def deserialize(self):
@@ -754,7 +762,14 @@ class JobTask(BaseHadoopJobTask):
         """
         for output in outputs:
             try:
-                print("\t".join(map(self.serialize, flatten(output))), file=stdout)
+                output = flatten(output)
+                if self.data_interchange_format == "json":
+                    # Only dump one json string, and skip another one, maybe key or value.
+                    output = filter(lambda x: x, output)
+                else:
+                    # JSON is already serialized, so we put `self.serialize` in a else statement.
+                    output = map(self.serialize, output)
+                print("\t".join(output), file=stdout)
             except:
                 print(output, file=stderr)
                 raise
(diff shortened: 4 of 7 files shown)


luigi/hadoop.py
```diff
@@ -689,6 +704,17 @@ def jobconfs(self):
             jcs.append('mapred.reduce.tasks=%s' % self.n_reduce_tasks)
         return jcs
 
+    @cached_property
+    def serialize(self):
+        return getattr(
+            DataInterchange, self.data_interchange_format)['serialize']
```
Tarrasch: I sense that you shouldn't use getattr in this case. Can't DataInterChange just be a dictionary? Or can't you change hte type of `data_interchange_format` to be the dict itself?

Tarrasch:
Cool! Can you look at comments and rebase on newest master? :)

mvj3:
@Tarrasch Your idea is awesome! Welcome to improve this PR code, such as variable naming, etc.

Tarrasch:
Thanks for the quick iteration @mvj3! :)

Can you please also add a test where you actually use this? I mean setting `data_interchange_format = 'json'`.

Also, I'm very excited to hear how much speed improvements you see by using the json format. And how many of your real world use cases uses `set`?

mvj3:
@Tarrasch Murphy's law rocks, my PR code had problems in logic, hope it's valid this time.

I use Python's builtin functions `str` and `eval` to mimic JSON module's `dumps` and `loads` functions.

`loads` has nearly 1 order of magnitude improvement, and `dumps` has 2 times improvement.

Here's the detailed benchmark result, gist url is https://gist.github.com/mvj3/02dca2bcc8b0ef1bbfb5

```txt
JSON Benchmark
2.7.9 (default, Jan 14 2015, 12:59:47) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.56)]
-----------------------------
ENCODING
simplejson: 0.200864s
cjson: 0.335475s
ujson: 0.169225s
jsonlib: 0.297341s
Python: 2.336938s
json: 0.548224s
yajl: 0.322109s
 
DECODING
simplejson: 0.302664s
cjson: 0.276507s
ujson: 0.115845s
jsonlib: 0.194297s
Python: 0.284689s
json: 0.250216s
yajl: 0.271826s
```

erikbern:
nice! should we change the default or is there some issue? eg does it work with all types

mvj3:
By the way, I really appreciate `luigi`, and our team build a project called `luiti`, luiti = luigi + time . It's a time task management framework, support multiple projects, the homepage is https://github.com/17zuoye/luiti. The bad thing is README is mostly written in Chinese right now, I'll translate it into English later.

The most used function is `@luigi.ref_tasks("TaskA", "TaskB", ...)` decorator, and the tasks are linked lazily, even refer to each other. Take a look at https://github.com/17zuoye/luiti/blob/master/luiti/luigi_decorators/ref_tasks.py and https://github.com/17zuoye/luiti/blob/master/tests/test_luiti_ext.py#L31

 Maybe luiti can merge into luigi in the future, or just acts as a extension of luigi :)

Tarrasch:
Wow! That's really cool @mvj3!! I had no idea you had such a big extension to luigi out there. I don't fully understand what you mean with time task management. Time for humans during a day or time measurement of task run times? I'm confused. Sorry.

You're welcome about luigi, and oh I wish I knew Chinese. :)

@erikbern, I think @mvj3 said it breaks `set()`. However, I'm tempted to having a default that tries both in a sense or something smart like that.

Tarrasch:
By the way, I was thinking more of benchmarks for running the whole hadoop jobs. I was wondering if doing the `eval` stuff was the major bottleneck for the python map reduce implementation in Python.

Tarrasch:
Hmm, what's the difference between `serialize` and `internal_serialize`?

mvj3:
@Tarrasch Thanks for your appreciation!  I have recently translated https://github.com/17zuoye/luiti README into an English version, which describes the ideas and structure of luiti better. I hope that you can understand it now :)

### Interchange data format
Yes, it breaks `set()`. Actually, I don't often use `set()` as a data interchange format, and I use it mostly just in a single function. Well, I don't know how about other people. I think JSON is a good enough for data exchange, and it's widely used.

### The major bottleneck in MapReduce
Yes, your idea is right and considerate. I'll do a word count MR job on a real YARN cluster.

### the difference between `serialize ` and `internal_serialize`
The first question to consider is that I want to support JSON feature quickly, and don't break related test cases meanwhile. In the default strategy, `serialize` means Python's `str` function, and `internal_serialize` means Python's `repr` function, at the same time, there exists `writer` and `internal_writer` functions. The tricky thing is that str("") return "", and repr("") return "''", so I had to compact with them. Yes, it's not a perfect solution, I think that the MR part of luigi maybe need to be refactored into more decoupled modules.

### Another YARN memory leak problem when there are fewer tasks number.
By the way, in our YARN production environment, luigi seems to have a memory leak problem, it reaches 6.6G threshold value when n_reduce_task is 5, and YARN had to kill it automatically. But if I set n_reduct_tasks to 25 or more, the problem was gone, no other code was changed.

I began to think that if `_map_input` and `_reduce_input` and other related functions don't release the object reference after they are already used, or keep them in a big list or dict until the total task is finished. After a period of investigation, I found `groupby` already guaranteed the data flow is a lazy generator, cause `groupby` requires the data flow is already sorted. So I still can't figure out what's the problem until now, but I know the working mechanism of luigi, and I thought `eval` is a bad thing Intuitively, so that's why I want to support json as a data interchange format in luigi.

hadesbox:
luiti seems very interesting, thanks for sharing.

Tarrasch:
Thanks for the translation! :)

Tarrasch:
Thsi makes sense to merge I think. Please let me know as soon as you've tested and seen if this improves running time on the cluster. If it does, I think this could be big news.

mvj3:
@Tarrasch @erikbern Here comes the detailed benchmark result. If it's a heavy CPU task, then `eval` is not the major bottleneck, otherwise `ujson` wins `eval` absolutely. See original benchmark code at https://gist.github.com/mvj3/72f34141a6e9c56695d5 


Local benchmark parameters | Speed
--------------- | -----------------
Pure count |
Char count | 3.8 MB/s
Word count | 21.8 MB/s
Pure decode |
Python | 7.5 MB/s
ujson | 70.1 MB/s
Word count + encode + encode |
Python | 9.6 MB/s
ujson | 19.1 MB/s
Char count + encode + encode |
Python | 3.5 MB/s
ujson | 3.7 MB/s


MapReduce parameters | value
--------------- | ----------------- 
 1. Files are already warmed by run reading. | 
 2. mapreduce.task.io.sort.mb | 512MB
 3. dfs.block.size | 128 MB
 4. n_reduce_tasks | 30

Conditions | Python | JSON
--------------- | ----------------- | --------------
CharCount + json 133 MB input file | 48s |   50s :confused: 
CharCount + json 1.1 GB input file | 2m:0s |   1m:30s
CharCount + json 16 GB input file | 3m:17s |   7m:13s :confused:
 | | 
CharCount + ujson 133 MB input file | 52s |   51s
CharCount + ujson 1.1 GB input file | 2m:2s |   1m:18s
CharCount + ujson 16 GB input file | 3m:13s |   2m:42s
 | | 
WordCount 133 MB input file | 30s |   30s
WordCount 1.1 GB input file | 1m:32s |   48s
WordCount 16 GB input file | 2m:22s |   1m:21s

Some thoughts:
1.  `List count` is really slowly than `Word count`. 
2. `CharCount + json`'s result is really wired, it may not be a problem

### Add more serialize  and deserialize methods
According to compare speed between builtin Python `eval` and json, serialize and deserialize is really not a major bottleneck in a total MapReduce task. Maybe we can support more complex object by adding [jsonpickle](http://jsonpickle.github.io/) to `luigi.contrib.hadoop.DataInterchange` .

```python
>>> import jsonpickle
>>> class Hello:
...   def __init__(self):
...     self.name = "world"
... 
>>> o1 = {"foobar": 42, "pi": 3.14, "object": Hello()}
>>> o2 = repr(o1)
>>> o2
"{'foobar': 42, 'pi': 3.14, 'object': <__main__.Hello instance at 0x105c845a8>}"
>>> o3 = eval(o2)
  File "<string>", line 1
    {'foobar': 42, 'pi': 3.14, 'object': <__main__.Hello instance at 0x105c845a8>}
                                         ^
SyntaxError: invalid syntax
>>> o4 = jsonpickle.dumps(o1)
>>> o4
'{"foobar": 42, "object": {"py/object": "__main__.Hello", "name": "world"}, "pi": 3.14}'
>>> o5 = jsonpickle.loads(o4)
>>> o5
{'foobar': 42, 'object': <__main__.Hello instance at 0x105c846c8>, 'pi': 3.14}
>>> 
```

I already add it to luiti, though it's not tested yet, https://github.com/17zuoye/luiti/commit/bae26321070750953bd84b87e41f7481a40edc77

mvj3:
@Tarrasch Is it ok to add `jsonpickle` as a dependency of `luigi`? I worry it may cause too many package dependencies.

Or use something like `cached_property` to load other Python module lazily?

Tarrasch:
Wait, what was `jsonpickle` related to? I thought we only needed `ujson`.

mvj3:
Yes, we only need `ujson` if we only want to support json.

Luigi can't support more complex data type between `mapper` and `reducer`, like the `Hello` example I posted in the "Add more serialize and deserialize methods" section. If we add it, user can change change `data_interchange_format` to "jsonpickle"。

Never mind, I already added it to `luiti`, and it's easy to add `jsonpickle` to `luigi` in the future.

mvj3:
It looks like your team are using [pyschema](https://github.com/spotify/pyschema) that built by yourself, to do serializing/deserializing data and more things. I also see there're some luigi related tests.

I wonder what's the difference between `pyschema` and `jsonpickle`, more controllable or strict?

Though I haven't dug too deep in `pyschema`, but I find some familiar feelings :smile: , e.g.

```python
# https://github.com/17zuoye/luiti/blob/master/luiti/utils/mr_utils.py#L32
    def is_mr_line(line1):
        is_true_2 = (line1.endswith("}") or line1.endswith("]"))
        is_true_3 = (not line1.startswith("{")) and (not line1.startswith("["))
        return is_true_1 and is_true_2 and is_true_3

# https://github.com/spotify/pyschema/blob/master/pyschema/core.py#L559
    if s.startswith(u"{"):
        json_dct = json.loads(s)
```

Tarrasch:
Hi again @mvj3.

I think @anyman and @freider know pyschema best.

@mvj3, how important is it to do the property caching? Does it make it noticeably faster even? We noticed that cached-property is not shipped on Ubuntu 14.04 and it would be nice to get rid of that extraneous dependency. We could just save the property ourselves in a private member.

Ping @rouzwawi

mvj3:
Hi @Tarrasch 

Nice suggestion! Though *cached_property* is one of my favourite tools, but it's really weird to use it only in luigi/contrib/hadoop.py, using private member is a good way to do it.

See another PR at https://github.com/spotify/luigi/pull/1308

Tarrasch:
Haha, yea I even started to rely on it but I found I could almost do as well without it too :)

PR #940, opened 2015-05-08
Make @with_config stackable

Tarrasch:


Tarrasch:
I merge for now to unblock #941, but please audit if you want. I'm a bit disappointed that it ended up with so much code. I also realized you can run the doc-tests as this as well:

    tox -e py27-nonhdfs -- --with-doctest test/helpers.py

PR #943, opened 2015-05-11
Enable doctests

Tarrasch:
We currently use doctests at some places like

 * luigi/deprecate_kwarg.py
 * test/helpers.py

And it works out very well for those cases. So I think it's worth having
in some situations.


Tarrasch:
Why the hell does enabling doc-tests fail the assistant tests. Sigh, you just gotta love python ...

Tarrasch:
Ok, any objections to merging this guys?

PR #944, opened 2015-05-11
Create an abstract hdfs client

Tarrasch:
Now clearer what client implements what. And we finally get rid of this weird
behaviour that the snakebite client inherits from HdfsClient.

This change kind of pushses for snakebite uses to start using the
snakebite_with_hadoopcli_fallback option, a many important file system
operations have not been ported to snakebite as of yet. But that's good, as that
configuration name better reflects what they actually use.

### DIFF
--- a/luigi/contrib/hdfs/abstract_client.py
+++ b/luigi/contrib/hdfs/abstract_client.py
@@ -1,0 +1,105 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2012-2015 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


erikbern:
Should we make an 1.1.2 release on Github and let it sit for a few days/weeks then push to PyPI soon?

Tarrasch:
Aha, you want to first put a release on GitHub and then push to PyPi? I never understood that. :)

I'm still doing some refactoring. I'm fine either way really, as long as I can keep making changes :)

erikbern:
I don't care, we don't have a well defined workflow. In the past I've made a release to both GH and PyPI simultaneously. When you release to PyPI you need to bump the version number in setup.py so I usually push that to GH right after.

erikbern:
Only thing I care about is we shouldn't push a bleeding edge version to GH. So either we slow down on pull requests or we push something that's a few weeks old

Tarrasch:
Push something a bit old then I think. :)

PR #950, opened 2015-05-13
Add Redis target support

pbamba:


erikbern:
@Tarrasch seems like tests have started failing a lot on Travis recently

erikbern:
otherwise looks good!

pbamba:
Thanks @erikbern, is there anything else I should do ?

Tarrasch:
@pbamba. Not really. I'll leave for long weekend now. So I won't be able to fix the travis build failures until then unfortunately. Sorry, it's usually always green :-/.

But if you want to track the build failure causes that would be awesome too ^^

Tarrasch:
Oh and yea, we can't merge this until the build failures (not caused by this PR) are fixed. :)

pbamba:
Thanks @Tarrasch, I'll take a look at the failures and see if I can help !

Tarrasch:
@pbamba nah I can fix it now I think. :)

Tarrasch:
Can you rebase on master and force push again?

pbamba:
@Tarrasch, it's done !

Tarrasch:
Cool! What about tests?

pbamba:
Should I add them through another pull request ?

Tarrasch:
I would much rather see them here if that's cool with you @pbamba :)

pbamba:
@Tarrasch, no problem, it's done ;)

Tarrasch:
Cool :). Will merge on Green

pbamba:
@Tarrasch, done ;)

Tarrasch:
Cool! (try to in the future to amend the commits so we don't get meaning less commits like "fix pep8 error")

Tarrasch:
Actually please do the amend and also move the test in `test/contrib/`. Is that ok?

pbamba:
yeah, sorry about that.

pbamba:
No problem, for the amend

Tarrasch:
Actually all 3 commits should be amended! I mean, adding tests should be part of the commit that adds the code itself.

pbamba:
Yeah, that's what I was planning to do

Tarrasch:
:+1:, will merge on Green

pbamba:
@Tarrasch , we should be good now. Thanks a lot for the feedback !

luigi/contrib/redis_store.py
```diff
@@ -0,0 +1,96 @@
...
+
+except ImportError:
+    logger.warning("Loading redis module without redis installed")
+
+
+class RedisTarget(luigi.Target):
+
+    """ Target for a resource in Redis."""
+
+    marker_prefix = luigi.configuration.get_config().get('redis',
```
Tarrasch: Whoups, didn't catch this until now, sorry. But you should use the new configuration system. See this [example](https://github.com/spotify/luigi/blob/b5ad3eba1501bdc25e91e98901bc781128f2d8a7/luigi/worker.py#L217)

luigi/contrib/redis_store.py
```diff
@@ -0,0 +1,96 @@
...
+
+import luigi
+
+logger = logging.getLogger('luigi-interface')
+
+try:
+    import redis
+
+except ImportError:
+    logger.warning("Loading redis module without redis installed")
```
Tarrasch: Is it ever sensible to do this? I would just remove the try-catch here. I think this causes more harm than good.

pbamba: I admit that it might not always be necessary, but I'm curious to know what harm could it cause ?

luigi/contrib/redis_store.py
pbamba: OK

luigi/contrib/redis_store.py
Tarrasch: Somebody who later studies this code and will think there is an actual reason for doing this. Does that make sense?

Tarrasch: It also silently defers the errors when importing `luigi.contrib.redis_store`

xcoderzach:
Just fyi, the test's class name is `ElasticsearchTargetTest` ;-)

Tarrasch:
Lol, nice catch @xcoderzach. This is what you get for breaking the DRY principle. Or that python allows more than one class per file. Hmm..

erikbern:
let's merge

Tarrasch:
@pbamba can you fix the configuration system thing so we can merge? :)

pbamba:
done @erikbern, @Tarrasch :)
(forgive me in advance if I forgot anything, it's pretty late here in France ha)

erikbern:
LGTM

Tarrasch:
It seems like you broke something in the build @pbamba :)

pbamba:
@Tarrasch , yeah I think removing the try / catch caused it. I've tried again but now the build sends back a " Failed connecting to remote scheduler 'localhost' " error :/

Tarrasch:
Oh ok, I guess it's an intermittent failure. Merging. :)

Issue #951, opened 2015-05-13
Test `test_get_info` is failing


Tarrasch:
See these logs: https://travis-ci.org/spotify/luigi/jobs/62354164#L1126

Any idea @gpoulin, you added it quite recently in #908. This error is one of many that suddenly just bursted out since yesterday. Travis probably made some internal change without announcing.

Tarrasch:
Hmm, maybe the locale changed on Travis? https://github.com/gabrielfalcao/HTTPretty/issues/223#issuecomment-101869731

Tarrasch:
Fixed in #952

PR #952, opened 2015-05-15
Test environment: Set encoding to utf8

Tarrasch:
This hopefully will make a few more tests pass.

See #951 and https://github.com/gabrielfalcao/HTTPretty/issues/223#issuecomment-102072192

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,6 +10,7 @@ deps=
   coverage>=3.6,<3.999
   coveralls
 setenv =
+  LC_ALL = en_US.utf-8
   cdh: HADOOP_DISTRO=cdh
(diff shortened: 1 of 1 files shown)


gpoulin:
Hadoop is still failing because of [this](https://travis-ci.org/spotify/luigi/jobs/62662866#L196). Travis seems to have few issues lately.

Tarrasch:
But this seems to fix at least so that py-{33,34}-nonhdfs works. So I'll merge this (unless objections). I'm not sure when I'll have time to take a stab at the HADOOP_HOME issue.

Yea, any idea why these just started to fail? I couldn't find other people/projects but us having issues with this.

PR #956, opened 2015-05-19
Snakebite cascading

daveFNbuck:
Raises NotImplementedError in unimplemented snakebite functions and adds getmerge to CascadingClient default names so that the snakebite client with hadoop cli fallback will work properly.


Tarrasch:
Ah, thanks!!

PR #957, opened 2015-05-19
Catch socket.timeout in RemoteScheduler to not crash in 2.7 and later

jooon:
* timeout is urllib2.URLError in python 2.6 and socket.timeout in 2.7
  and later

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -22,6 +22,7 @@ See :doc:`/central_scheduler` for more info.
 
 import json
 import logging
+import socket
 import time
 
 from luigi.six.moves.urllib.parse import urlencode
@@ -77,7 +78,7 @@ class RemoteScheduler(Scheduler):
             try:
                 response = urlopen(full_url, body, self._connect_timeout)
                 break
-            except URLError as e:
+            except (URLError, socket.timeout) as e:
                 last_exception = e
                 if log_exceptions:
                     logger.exception("Failed connecting to remote scheduler %r", self._host)


Tarrasch:
Awesome!

Can you just clarify what would happen in python 2.7? Also, since we're not supporting 2.6 anymore, should it only be `except socket.timeout as e:`?

jooon:
Nope. You need to catch both.

In 2.6 dns problems, connection problems, socket timeouts are all urllib2.URLError, so it was possible to catch only that one.

In 2.7 and later dns problems, connection problems are still URLError, while timeouts are socket.timeout

```
$ cat timeout.py 
#!/usr/bin/env python
from __future__ import print_function

from platform import python_version
from six.moves.urllib.error import URLError
from six.moves.urllib.request import urlopen
import socket

try:
    response = urlopen("http://localhost:8000/", timeout=1)
except (URLError, socket.timeout) as e:
    print("{0} URLError or socket.timeout: {1}".format(python_version(), type(e)))
```

Start a never answering tcp connection somewhere like:

```
$ while true; do nc -l 8000; done
```

and try it with different versions of python to see the exception for a timeout:

```
2.6.9 URLError or socket.timeout: <class 'urllib2.URLError'>
2.7.6 URLError or socket.timeout: <class 'socket.timeout'>
3.4.3 URLError or socket.timeout: <class 'socket.timeout'>
```

Kill the nc and see what happens with a connection problem.

```
2.6.9 URLError or socket.timeout: <class 'urllib2.URLError'>
2.7.6 URLError or socket.timeout: <class 'urllib2.URLError'>
3.4.3 URLError or socket.timeout: <class 'urllib.error.URLError'>
```

six already takes care about urllib2 vs urllib.error for us.

erikbern:
wow nice research. let's merge

Tarrasch:
Very, very nice research. Merging!

PR #958, opened 2015-05-19
Fix bug where command line Tasks skip config

Tarrasch:
Before this patch, when you run a task from the command line, it'll
ignore the defaults you've specified in the config. By config I mean the
`[MyClass] my_param: myvalue` syntax in `/etc/luigi/client.cfg`.

After this patch, the config is honored.

I also added another small and related fix, the default shown when doing
`--help` will now also honor the config and not only the `default=` value.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -279,7 +279,7 @@ class Parameter(object):
             return [str(v) for v in x]
         return str(x)
 
-    def parse_from_input(self, param_name, x):
+    def parse_from_input(self, param_name, x, task_name=None):
         """
         Parses the parameter value from input ``x``, handling defaults and is_list.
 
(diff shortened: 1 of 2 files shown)


Tarrasch:
@erikbern, The code for `parameter.py` is a bit too much, and I know that this patch makes it only worse. I'm also not sure if it makes sense to have `has_value` and `value` properties. As can been seen from this patch, you need to know the task name to calculate the value. What do you think?

Maybe we should start with (1) get rid of globals support as it will make code a level of magnitude easier to read and then (2) remove `value` and `has_value`, because their implementations will always be wrong unless you can somehow pass in the `task_name` to the `Parameter` instance themselves.

erikbern:
LGTM. Yeah `parameter.py` is complicated. We should simplify it. I already simplified a lot but I think by now we can remove some legacy crap like `is_global`

Tarrasch:
Cool. Remove the horrible the `is_global` stuff. We just need to unglobaify the remaining code in open source luigi (contrib.hadoop) that uses is.

PR #959, opened 2015-05-20
Use argparse for luigid

Tarrasch:
The motivation is simply that argparse is the new and better parser.
Also, argparse is shipped with 2.7 which is the lowest version of python
we support for now.

### DIFF
--- a/luigi/cmdline.py
+++ b/luigi/cmdline.py
@@ -1,5 +1,5 @@
 import os
-import optparse
+import argparse
 import logging
 import sys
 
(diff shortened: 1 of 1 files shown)


erikbern:
We should really use the configuration system though, not command line parsing

Tarrasch:
The same configuration system as we use for the client? Won't we unnessesarily couple the client with the server?

But this is OK to merge then I guess?

erikbern:
sure lgtm

PR #960, opened 2015-05-20
Promote retry-external-tasks to new config system

Tarrasch:
This basically allows you to also enable it as a command line parameter.
Run `luigi MyTask --worker-retry-external-tasks`.


erikbern:
i'm not sure about maintaining this logic since it creates a difference between `Task` and `ExternalTask`. There are already differences in how the worker treats them, and that leads to pretty messy code in `worker.py`. I started refactoring this with the goal of unifying them (`worker.py` should be agnostic imo)

anyway this particular PR doesn't really change anything about that situation, so it's fine to merge, just wants to point it up

Tarrasch:
I'm excited to see what you've planned to whip up for external tasks. :)

PR #964, opened 2015-05-21
Remove last is global usage

Tarrasch:
See individual commits.

After this is merged, there's no code using is_global within spotify/luigi. So then we can finally delete this annoying is_global logic that massively complicates things.

### DIFF
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -51,6 +51,7 @@ from luigi import six
 
 from luigi import configuration
 import luigi
+import luigi.task
 import luigi.contrib.hdfs
 import luigi.s3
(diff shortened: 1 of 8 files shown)


Tarrasch:
I also verified that this won't break in case you did this  kind of hack earlier:

```
class TaskNotInheritingFromBaseHadoopJobTask(luigi.Task):
    pool = luigi.hadoop.BaseHadoopJobTask.pool
```

Will not break. Even though it obviously will not honor the `--hadoop-pool` flag and corresponding config. @sisidra @plinton

erikbern:
LGTM

Tarrasch:
Also reviewed by @sisidra

Issue #953, opened 2015-05-18
Current build issues


erikbern:
Seems like Travis is acting up a lot lately

There's some issue with all Hadoop tests. This seems to be because JAVA_HOME is not set. I added an "env" command tot tox.ini that confirms this. We probably just have to do something under setenv that sets it to the right value

https://travis-ci.org/spotify/luigi/jobs/62960972

The relevant output is below. Snakebite tries to launch a namenode in the background, but the namenode fails starting because JAVA_HOME isn't set. However that doesn't trigger a failure. What happens later is Snakebite is unable to parse the namenode address from the namenode stdout and then it fails.

```
    self.hdfs_url = "hdfs://%s:%d" % (self.host, self.port)
TypeError: %d format: a number is required, not NoneType
-------------------- >> begin captured stdout << ---------------------
('Error: JAVA_HOME is not set and could not be found.\n',)
('',)
```

There's also currently something with installing HTTPretty but I think that's more intermittent (didn't see it yesterday)

gpoulin:
The issue with HTTPretty is due to ``LANG`` not being set correctly #952

Tarrasch:
Yea HTTPretty is solved. I'll merge that patch now!

erikbern:
Ok awesome. I'll play around a bit more with the java problems.

benjixx:
I ran into a similar issue some days ago with our company-internal Jenkins CI server. In this case the problem was related to ``tox`` which has changed the way environment variables are passed through.

You probably want to have a look at [tox 2.0 changelog](http://tox.readthedocs.org/en/latest/changelog.html). Especially have a look at the passenv setting.

Tarrasch:
@benjixx, aha!!!!!!!!!!!!!!!!!!!!!!!!!!! Thank you so much, I've been pulling my hair wondering why only we are affected. Thanks thanks thanks!

benjixx:
@Tarrasch Good to hear everything's working again.

PR #965, opened 2015-05-21
Don't run `env` from tox.

Tarrasch:
It was only temporarily added by @erikbern when he was debugging some
random failures (see #953). After this patch, we don't need to see this
anymore:

    WARNING:test command found but not installed in testenv
      cmd: /usr/bin/env
      env: /home/arash/spotify/repos/luigis/github_luigi/.tox/py27-nonhdfs
    Maybe you forgot to specify a dependency? See also the whitelist_externals envconfig setting.
    ...
    USER=arash
    LC_ALL=en_US.utf-8
    PATH=...
    COVERAGE_PROCESS_START=/home/arash/spotify/repos/luigis/github_luigi/.coveragerc

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -25,7 +25,6 @@ setenv =
 commands =
   cdh,hdp: {toxinidir}/scripts/ci/setup_hadoop_env.sh
   python --version
-  env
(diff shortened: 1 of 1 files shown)


erikbern:
sure but it might be useful :) but whatev :)

PR #972, opened 2015-05-25
Add tests for namespaces

Tarrasch:
See individual commits.

I basically just wanted so quickly check that namespaces work, but I ended up fixing other things that have bugged me for a while as well.


erikbern:
nice!

PR #973, opened 2015-05-25
Add support for appending to HDFS and a local file

hellais:
This PR superseeds: https://github.com/spotify/luigi/pull/910

### DIFF
--- a/luigi/contrib/hdfs/format.py
+++ b/luigi/contrib/hdfs/format.py
@@ -25,6 +25,9 @@ class HdfsAtomicWritePipe(luigi.format.OutputPipeProcessWrapper):
 
     TODO: if this is buggy, change it so it first writes to a
     local temporary file and then uploads it on completion
+
+    If the append keyword arugment is set to True it will appendToFile instead
+    of put.
     """
 
     def __init__(self, path, append=False):
--- a/luigi/contrib/hdfs/format.py
+++ b/luigi/contrib/hdfs/format.py
@@ -132,7 +132,7 @@ class CompatibleHdfsFormat(luigi.format.Format):
         return self.reader(input)
 
     def pipe_appender(self, output):
-        return HdfsAtomicWritePipe(output, True)
+        return HdfsAtomicWritePipe(path=output, append=True)
 
     def hdfs_writer(self, output):
         return self.writer(output)
--- a/luigi/contrib/hdfs/format.py
+++ b/luigi/contrib/hdfs/format.py
@@ -27,12 +27,15 @@ class HdfsAtomicWritePipe(luigi.format.OutputPipeProcessWrapper):
     local temporary file and then uploads it on completion
     """
 
-    def __init__(self, path):
+    def __init__(self, path, append=False):
         self.path = path
         self.tmppath = hdfs_config.tmppath(self.path)
         parent_dir = os.path.dirname(self.tmppath)
         mkdir(parent_dir, parents=True, raise_if_exists=False)
-        super(HdfsAtomicWritePipe, self).__init__(load_hadoop_cmd() + ['fs', '-put', '-', self.tmppath])
+        command = "-put"
+        if append:
+            command = "-appendToFile"
+        super(HdfsAtomicWritePipe, self).__init__(load_hadoop_cmd() + ['fs', command, '-', self.tmppath])
 
     def abort(self):
         logger.info("Aborting %s('%s'). Removing temporary file '%s'",
@@ -124,6 +127,9 @@ class CompatibleHdfsFormat(luigi.format.Format):
 
     def pipe_reader(self, input):
         return self.reader(input)
+
+    def pipe_appender(self, output):
+        return HdfsAtomicWritePipe(output, True)
 
     def hdfs_writer(self, output):
         return self.writer(output)
--- a/luigi/contrib/hdfs/format.py
+++ b/luigi/contrib/hdfs/format.py
@@ -35,10 +35,10 @@ class HdfsAtomicWritePipe(luigi.format.OutputPipeProcessWrapper):
         self.tmppath = hdfs_config.tmppath(self.path)
         parent_dir = os.path.dirname(self.tmppath)
         mkdir(parent_dir, parents=True, raise_if_exists=False)
-        command = "-put"
-        if append:
-            command = "-appendToFile"
-        super(HdfsAtomicWritePipe, self).__init__(load_hadoop_cmd() + ['fs', command, '-', self.tmppath])
+        command = [
+            'fs', "-appendToFile" if append else "-put", '-', self.tmppath
+        ]
+        super(HdfsAtomicWritePipe, self).__init__(load_hadoop_cmd() + command)
 
     def abort(self):
         logger.info("Aborting %s('%s'). Removing temporary file '%s'",
--- a/luigi/contrib/hdfs/format.py
+++ b/luigi/contrib/hdfs/format.py
@@ -34,7 +34,8 @@ class HdfsAtomicWritePipe(luigi.format.OutputPipeProcessWrapper):
         self.path = path
         self.tmppath = hdfs_config.tmppath(self.path)
         self.append = append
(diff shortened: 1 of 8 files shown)


luigi/contrib/hdfs/format.py
```diff
@@ -125,6 +131,9 @@ def pipe_writer(self, output):
     def pipe_reader(self, input):
         return self.reader(input)
 
+    def pipe_appender(self, output):
+        return HdfsAtomicWritePipe(output, True)
```
Tarrasch: please don't use a positional argument

luigi/file.py
```diff
@@ -95,17 +148,19 @@ def makedirs(self):
...
+        elif mode == 'a':
+            self.makedirs()
+            return self.format.pipe_writer(atomic_file_append(self.path, timeout=timeout))
         elif mode == 'r':
             fileobj = FileWrapper(io.BufferedReader(io.FileIO(self.path, 'r')))
             return self.format.pipe_reader(fileobj)
 
         else:
-            raise Exception('mode must be r/w')
+            raise Exception('mode must be r, w or a')
```
Tarrasch: Change to ValueError

Tarrasch: Or ArgumentError or whatever is most suitable. I don't know all exception types on top of my head :)

luigi/file.py
```diff
@@ -43,8 +49,55 @@ def move_to_final_destination(self):
...
+        self.lockfile = "%s.lock" % path
+        super(atomic_file_append, self).__init__(path)
+
+    def move_to_final_destination(self):
+        try:
+            self.acquire()
+            with open(self.tmp_path) as in_file:
+                with open(self.path, 'a') as out_file:
+                    shutil.copyfileobj(in_file, out_file)
+        except Exception as exc:
```
Tarrasch: Exception to broad

Tarrasch: Don't you want a try finally block here btw?

luigi/file.py
```diff
@@ -43,8 +49,55 @@ def move_to_final_destination(self):
     def generate_tmp_path(self, path):
         return path + '-luigi-tmp-%09d' % random.randrange(0, 1e10)
 
+class atomic_file_append(atomic_file):
+    """Use this to append to a temporary file and append to the final
```
Tarrasch: "append to the final" --> Isn't it "replace the final destination" something?

Tarrasch:
By the way, I see that the commits are very messy. In the second commit you almost remove all code from the first commit. Would it make sense to fix the history?

luigi/file.py
```diff
@@ -43,8 +49,55 @@ def move_to_final_destination(self):
...
+            os.unlink(self.tmp_path)
+            raise exc
+        self.release()
+        os.unlink(self.tmp_path)
+
+    def acquire(self):
+        start_time = time.time()
+        while True:
+            try:
+                self.fd = os.open(
```
Tarrasch: Are you assuming a local file system for everything here? Will this work for hadoop?

erikbern:
LGTM but please add tests. I hope this works with the minicluster!

erikbern:
Also this currently fails on pep checks

https://travis-ci.org/spotify/luigi/jobs/63956120

```
pep8 runtests: commands[0] | pep8 --exclude=luigi/six.py --ignore E501,E402,E731 luigi test examples bin
luigi/file.py:52:1: E302 expected 2 blank lines, found 1
luigi/file.py:99:1: E302 expected 2 blank lines, found 1
```

luigi/file.py
hellais: No, it is correct. It will append, that is the purpose of it to append to the final destination.

hellais:
Should now address the issues raised by both of you.

Tarrasch:
I think your 4 commits could be squashed into 1, right? When reading the log history you're usually not interested in "Fix pep8" like commits.

On the other hand, your first commit could be split into two commits, one for hdfs and one for local file system.

luigi/contrib/hdfs/hadoopcli_clients.py
```diff
@@ -146,6 +146,12 @@ def copy(self, path, destination):
     def put(self, local_path, destination):
         self.call_check(load_hadoop_cmd() + ['fs', '-put', local_path, destination])
 
+    def append(self, local_path, destination):
+        """
+        Requires Hadoop >= 2.3.0
```
Tarrasch: Good that you mention this constraint in the docstring :)

luigi/target.py
```diff
@@ -135,7 +139,9 @@ def isdir(self, path):
 
         *Note*: This method is optional, not all FileSystem subclasses implements it.
         """
-        raise NotImplementedError("isdir() not implemented on {0}".format(self.__class__.__name__))
+        raise NotImplementedError(
```
Tarrasch: Not sure why you changed this? :S

luigi/file.py
hellais: I am assuming a local file system because this is in fact a local file system append file (just like `atomic_file` that it inherits from).

hellais:
> I think your 4 commits could be squashed into 1, right? When reading the log history you're usually not interested in "Fix pep8" like commits.

I did that in a separate commit since it was a commit specifically aimed at resolving the issues raised by @erikbern in the comments to this PR (https://guides.github.com/introduction/flow/).

> On the other hand, your first commit could be split into two commits, one for hdfs and one for local file system.

I just rebased it all into 1 to avoid having too many commits (they were originally 2 commits, but then changes were made to upstream master that lead it to be broken and I had to patch them).

luigi/target.py
hellais: The line was too long.

hellais:
> LGTM but please add tests. I hope this works with the minicluster!

I'm not sure I will have time to add unittest soon. Will see what can be done.

luigi/target.py
Tarrasch: Ok :)

Tarrasch:
Given that this will be used in production, both by you and others. I think having some sort of tests is a given.

Thanks for the great work so far. Also I made some comments here: https://github.com/hellais/luigi/commit/dc1677d01dc451eeb1e1be9031461f472406db07 (sorry, I should have done it in the PR)

hellais:
/me crosses fingers hoping the HDFS unittests will pass :pray: :trollface:

luigi/target.py
```diff
@@ -30,6 +30,10 @@
 logger = logging.getLogger('luigi-interface')
 
 
+class FileLockException(Exception):
```
Tarrasch: There's still two of this?

hellais: Fixed and rebase.

hellais:
I can't at the moment debug the HDFS unittest issues since I don't have a dev environment setup with hadoop.

Tarrasch:
@hellais Have you read https://github.com/spotify/luigi/blob/master/CONTRIBUTING.rst? It'll download and setup the hadoop cluster for you. It shouldn't need any special system packages except for a java installation.

Tarrasch:
I also just added a section clarifying this: https://github.com/spotify/luigi/commit/17b255fc71f019fd1f1f6eb44655cd2848f9d560

hellais:
@Tarrasch I am having difficulties using the local minicluster. I think I will continue using this travis backed debugging approach in the meantime. When it's working I'll rebase the the last commits.

Tarrasch:
Fine by me. :)

hellais:
So after thinking more carefully of our data pipeline it seems like appending to files would be a bit fragile.

I also have come to believe that perhaps adding append support to luigi would lead people into doing a mistake similar to that which we were about to do and therefore it's perhaps ideal to NOT adding append support.
The risk is that adding append targets would encourage users of the library to not follow a best practice and could lead to ensurances expected from the luigi framework to no longer be satisfied.

I can perhaps write up more about why I believe this is the case if there is an interest for it. Anyways for the time being I will not continue development on this branch.

Tarrasch:
I totally see your point. And having it in the library would probably misguide users.

PR #976, opened 2015-05-26
Remove is_global support

Tarrasch:
Note that we sill support global parameters in the sense that they are
globally assigned (The `--ClassName-param-name param_value` syntax).
However, they don't clutter up the namespace anymore since they are
prepended with the task_family.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -114,8 +114,6 @@ class Parameter(object):
                              value of ``[]``.
         :param bool is_bool: specify ``True`` if the parameter is a bool value. Default:
                                 ``False``. Bool's have an implicit default value of ``False``.
-        :param bool is_global: specify ``True`` if the parameter is global (i.e. used by multiple
-                               Tasks). Default: ``False``. DEPRECATED.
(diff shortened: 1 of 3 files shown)


saline:
Missing some explanation on why we are doing this?

erikbern:
motivation is we're removing lots of complex logic that supported a pretty bad feature

Tarrasch:
Here's a real world example for Spotify's case: Across our hadoop-facing tasks, we want a uniform name of the `--pool` parameter (specifying the pool name in hadoop). For whatever reasons we didn't want to inherit from `HadoopBaseJobTask`, but we still wanted it's `pool` parameter. There's no clean solution when global parameters are involved. As `HadoopBaseJobTask.pool` is global and will conflict.

I'm hope other companies also would love to see this feature being removed :)

saline:
Good motivation.

Everything looks good to me. Only remark is maybe to add an assertRaises(DeprecationWarning) test as well.

Tarrasch:
Good idea about the `assertRaises(DeprecationWarning)`. Though can't be bothered with another iteration cycle. (I wonder if we can pay Travis to run our tests quicker..)

PR #978, opened 2015-05-26
Add http version of mortar help site back in to the docs

ddaniels888:


erikbern:
I keep getting this error when i go there – I think you might have to change the subdomain or something?

image

ddaniels888:
Hmm, I see that in Firefox but not in Chrome. I'll take a look.

Tarrasch:
Hmm, works for me on Firefox

ddaniels888:
I think this is an autofill feature in firefox, where if you've ever
visited the https version of a site it will force load https instead of
http, even if you type http. The url should be good; it's just the browser
that's being opinionated.

PR #980, opened 2015-05-26
Use keyword arguments for Schedulers

Tarrasch:
My main motivation for doing this was to rename 'worker' to 'worker_id',
increasing readability.

The other reason is that the test cases and the code in worker.py
becomes easier to follow when we're passing keyword arguments.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -535,7 +535,7 @@ class CentralPlannerScheduler(Scheduler):
             if t is not None and prio > t.priority:
                 self._update_priority(t, prio, worker)
 
-    def add_task(self, worker, task_id, status=PENDING, runnable=True,
+    def add_task(self, task_id=None, status=PENDING, runnable=True,
                  deps=None, new_deps=None, expl=None, resources=None,
                  priority=0, family='', module=None, params=None,
                  assistant=False, **kwargs):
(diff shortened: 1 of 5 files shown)


luigi/scheduler.py
```diff
@@ -546,7 +546,8 @@ def add_task(self, worker, task_id, status=PENDING, runnable=True,
         * add additional workers/stakeholders
         * update priority when needed
         """
-        self.update(worker)
+        worker_id = kwargs['worker']
```
saline: This assumes 'worker' is always in kwargs (otherwise will raise KeyError). Maybe it's enforced somewhere (I really have no idea)? If not, the function comment should mention that this kwarg is required. Maybe even add an assert.

Tarrasch: Well, I thought that it's "enough" that the first statement directly does access that attribute. Would it be OK to explicitly state that 'worker' must be passed as a kwarg?

Tarrasch: > Maybe even add an assert

Isn't doing `kwargs['worker']` some sort of an assertion?

saline: Well, it will raise an error I suppose. It might be good for readability of code to place an assert there. Or in a comment.

Otherwise everything else looks fine.

Tarrasch:
I merge since there was no strong objections. I'll fix the missing documentation later. (It's much easier to rebase and do git-stuff on my already done work if I don't amend here)

Tarrasch:
I meant that https://github.com/spotify/luigi/pull/986 depends on this.

PR #981, opened 2015-05-26
Task to upload a file to Salesforce

jeremykarn:
This PR adds a task that can be used to upload a file to Salesforce.  It uses a helper file target_factory that takes a local/s3 file path and returns the appropriate target object.

### DIFF
--- a/luigi/contrib/salesforce.py
+++ b/luigi/contrib/salesforce.py
@@ -1,0 +1,360 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2012-2015 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""
+Example configuration section in client.cfg::
+
+    [salesforce]
+    username: ${SF_USERNAME}
+    password: ${SF_PASSWORD}
+    security_token: ${SF_SECURITY_TOKEN}
+"""
+
+
+import requests
+import time
+try:
+    from urlparse import urlsplit
+except ImportError:
+    from urllib.parse import urlsplit
+import xml.etree.ElementTree as ET
+
+from luigi import configuration
+from luigi import BoolParameter, Parameter
+from luigi import target_factory
+from luigi import Task
+from luigi.s3 import S3Target
+
+
+import logging
+logger = logging.getLogger('luigi-interface')
+
+
+class UploadToSalesforceTask(Task):
+    """
+    This task will upload a file to Salesforce.  It requires the following fields
+    to be set in the [salesforce] section of the client.cfg file:
+
+    * username: The username of the SF user that will be doing the uploading.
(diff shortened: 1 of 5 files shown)


erikbern:
Looks great but let's remove the `target_factory` thing. The Luigi way is to subclass and implement a `requires` method and have an abstraction layer between tasks and paths

jeremykarn:
Do you mean to have users subclass this task and use the requires method to pass in the file that should be uploaded instead of passing in a file path as a parameter?  The problem with that is that the task may have multiple dependencies - only one of which is the file that should be uploaded.

Or am I misunderstanding what you meant?

The target_factory methods are pretty useful in other contexts as well.  In this case I want to write a file that indicates when the task has completed because there's no easy way to check for success/completeness with Salesforce directly.  By using the target_factory we can easily support both s3 and local paths.

jeremykarn:
What's the right way to fix: https://travis-ci.org/spotify/luigi/jobs/64256687  The Salesforce file depends on requests, but I'm not sure where to include it to install for the docs job.

erikbern:
No you understood what I meant.

How can the task have multiple dependencies? Looks like there's only one that's needed.

Parameters are not meant to provide the inputs to tasks. That goes against the whole idea with Luigi's dependencies. In this case there's no way to say that the salesforce import task depends on some report that has to be generated elsewhere.

To check whether the task has been run you also don't need the target factory – just rely on subclasses implementing an output() method.

Another thing is if you can use the new class level parameter instead of get_config()  – would be a bit cleaner.

Tarrasch:
Yes, please use the new config system :)

jeremykarn:
There's only one input file dependency, but you might want to add additional dependencies in terms of ordering the tasks that are running in your pipeline.  For example, if the pipeline is doing multiple uploads to Salesforce, maybe there is no logical dependency between those upload tasks but you want to ensure that they run serially to avoid api rate limits.  

The other issue is that in our pipeline we have this task depend on a Pig task that has multiple output files (it generates the same data but in multiple formats: Avro, CSV for Salesforce, split in s3 for Redshift).  Its nice to have the dependency on the Pig task as opposed to having the dependency on an external file task.  That allows the visualization/logging of the pipeline to more clearly show what's really happening and will allow a single invocation of Luigi to run all of the tasks.

I can remove the use of token files and rely on an output method in the subclass the user implements, but it seems like common functionality that is going to be needed as implementing an output() method that goes to Salesforce to figure out what was uploaded would be quite difficult.

I'll update to use the new config system.

jeremykarn:
@erikbern - Any thoughts on the above?  I can take a look at the failing tests if you're ok with the rest of the pull request.

erikbern:
Sorry I'm too busy these days so I forgot about this. Yes you should refactor this so that it looks like all the other tasks, i.e. no parameters defined on the base class, and then let subclasses implement requires() and output()

erikbern:
With all the other tasks I mean eg `HadoopJobTask`, `CopyToTable`, `PigJobTask` etc

erikbern:
It's arguable whether the subclass based API is the ideal, but that's just how everything else works, and it's been working pretty well so far

erikbern:
And please remove `TargetFactory` from this PR so that the PR is minimal and consistent

Tarrasch:
Are you still planning to iterate on this? Otherwise lets close for the moment.

Tarrasch:
Reopen ticket when you have time to iterate on this please. :)

jeremykarn:
Sorry, was offline for a few weeks.  I probably won't get to this too soon, so we can leave closed for now.

dlstadther:
@jeremykarn Once you finish the first complete revision of the SF bulk upload, I'd like to implement the bulk export. I will need the export functionality for a number of tasks. Thanks!

PR #986, opened 2015-05-28
Improve best case complexity for get_work

Tarrasch:
Yes, we only improve the best case complexity, not the worst case.
In summary, you'll get sublinear in aomunt of tasks if your worker
doesn't do any tasks that use global resources.

get_work has gotten way to complex once we added global resource
constraints, as the resource constraints interaction with priorities is
a hairy beast requiring a global perspective, not allowing for
worker-local optimization. See these related tickets which added the
expressiveness/complicated stuff:

  * spotify/luigi#445
  * spotify/luigi#462
  * spotify/luigi#496
  * spotify/luigi#513

Note that the optimization once done in spotify/luigi#584 which helped
@daveFNbuck's company a lot doesn't help Spotify's setup. Most of
Spotify's tasks lingering in the scheduler are pending tasks and not
done tasks.

The proposed hacky solution here, which I mostly see as a temporary
solution until we do a bigger rewrite and rethink of the central
scheduler is the following:

Categorize the get_work scenario in two different buckets, "trivial" and
"nontrivial". Trivial is when none of the tasks of the worker require
global resources, in that scenario, just loop over the tasks the worker
have registered for instead of all the tasks. For the non-trivial case,
we do what we've always done previously.

So in that sense this commit contains two parts:

 * Add an "index" so you can quickly retrieve the tasks for a worker
 * Do the special-casing of trivial and nontrivial in get_work

At Spotify, we don't use the resources feature that much, so we're
hoping for tremendous speed improvements, no more 30-second get-work
calls! :)

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -229,6 +229,7 @@ class Worker(object):
         self.reference = None  # reference to the worker in the real world. (Currently a dict containing just the host)
         self.last_active = last_active  # seconds since epoch
         self.started = time.time()  # seconds since epoch
+        self.tasks = set()  # task objects
         self.info = {}
 
     def add_info(self, info):
@@ -243,6 +244,19 @@ class Worker(object):
         # Delete workers that haven't said anything for a while (probably killed)
         if self.last_active + config.worker_disconnect_delay < time.time():
             return True
+
+    def get_pending_tasks(self):
+        return six.moves.filter(lambda task: task.status in [PENDING, RUNNING],
+                                self.tasks)
+
+    def is_trivial_worker(self):
+        """
+        If it's not an assistant having only tasks that are without
+        requirements
+        """
+        if self.assistant:
+            return False
+        return all(not task.resources for task in self.get_pending_tasks())
 
     @property
     def assistant(self):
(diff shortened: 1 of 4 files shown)


erikbern:
This is awesome, really always wanted to spend more time refactoring this code!

Do you have any performance numbers?

Tarrasch:
Not really, but here's a start at least: https://github.com/spotify/luigi/pull/986/files#diff-3425f46b36ec1572c55d6448f0ed6957R751

erikbern:
Should we merge this?

And again can you share some performance stats? I guess I can do it also

Tarrasch:
I think it can be merged if you've reviewed it as well (Dave has as well).

Well, I haven't tried this in a real environment yet. But clearly it's way faster according to the test I've added.

erikbern:
Happy to merge this. Would be great to have some numbers. But I can run later. I wonder if we should have some sort of benchmark tool

Tarrasch:
Less talking more merging. :p

I'll see if we can deploy this in the Spotify environment anytime soon. :)

erikbern:
nice!

PR #988, opened 2015-05-30
Add support for custom local-tmp-dir in LocalTarget

hellais:
This requires setting a client.cfg option:
[core]
local-tmp-dir: /media/ephemeral

### DIFF
--- a/luigi/file.py
+++ b/luigi/file.py
@@ -27,10 +27,16 @@ import io
 import sys
 import warnings
 
+import luigi
 import luigi.util
 import luigi.configuration
 from luigi.format import FileWrapper, get_default_format, MixedUnicodeBytes
 from luigi.target import FileSystem, FileSystemTarget, AtomicLocalFile
+
+
+class localfile(luigi.Config):
+    tmp_dir = luigi.Parameter(default=tempfile.gettempdir(),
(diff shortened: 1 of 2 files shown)


luigi/file.py
```diff
@@ -74,6 +75,9 @@ def __init__(self, path=None, format=None, is_tmp=False):
         if format is None:
             format = get_default_format()
 
+        tmp_dir = luigi.configuration.get_config().get("core", "local-tmp-dir", None)
```
Tarrasch: Don't use `core` for this and also use the new configuration system.

Tarrasch:
This looks good, but please fix the comment and add a test as well. :)

luigi/file.py
hellais: What section should it go under if not core?

Which new configuration system?

erikbern: you can define a `luigi.Config` class – see examples in the existing code

themalkolm:
Why don't you just provide it via TMPDIR env variable?

Tarrasch:
Can you please squash and (maybe) rebase this? I'll be very happy to merge this then!

Tarrasch:
Reopen ticket when you have time to iterate on this please. :)

PR #989, opened 2015-06-02
Add support for pseudo-tty allocation.

themalkolm:
This is required to be able run sudo command.

From man ssh

```
     -t      Force pseudo-tty allocation.  This can be used to execute arbitrary screen-based programs on a remote machine, which can be very useful, e.g. when
             implementing menu services.  Multiple -t options force tty allocation, even if ssh has no local tty.
```


Tarrasch:
LGTM

Issue #979, opened 2015-05-26
SQLAlchemyTarget only works with single SQLAlchemy engine


dylanjbarth:
I have a `ProcessDataTask(sqla.CopyToTable)` that processes some data and inserts it to `schemaA.TableA`. That task requires `BackupTask(sqla.CopyToTable)`, which copies Table A and inserts it into `schemaB.tableA_backup1` before the processing task dumps new data in. Both tasks return `SQLAlchemyTarget`s as output -- but because the targets are in different schemas, the one that get's defined first is where my backup is written. In other words, because `ProcessDataTask` creates a class level attribute `_engine` for the first schema, I can't have my backup task point at a different schema (unless I'm missing something, which is likely/possible). 

@gouthambs Happy to look at possible fixes and submit a PR if this is a priority or something you want to see fixed -- please correct me if I've overlooked something here or wasn't clear in my explanation.

gouthambs:
@dylanjbarth You have explained the issue very well. The `_engine` is a class variable to avoid creation of multiple engine instances. But the side effect of this is that one cannot use multiple schemas. 

If you have some ideas on fixing this issue, please take a stab at it. I can give you pointers if you need help.

One approach I can think of in order to address this issue is to replace `_engine` with `_engine_dict` which will store a `dict` of engine instances. The trick is then to chose an appropriate `key` for the `dict` and connect them intelligently with the task. Hope this helps!

dylanjbarth:
@gouthambs yeah the dict approach seems sensible. what do you think of this? https://github.com/dylanjbarth/luigi/commit/f026abc295defe0094381aac14011fc5e9d01558?diff=unified

Just using the connection string for the key, since that will be specified on a task by task basis and it's unique to the engine instance. Also, I subtracted storing the pid -- I wasn't too sure if there was a need for that anymore, let me know what you think.

sqla_tests are passing for me locally, happy to look at adding a test for multi-tables if you think this is in the right direction.

gouthambs:
@dylanjbarth The engine as a dict looks good. The `pid` logic is necessary in the multiprocessing scenario. I added that to address issues I was encountering in a production scenario with multiple processes. Unfortunately the tests don't always capture it I think. Issue #821 also mentions an issue which the `pid` logic fixes. Can you put the `pid` logic back?

dylanjbarth:
[...]

PR #991, opened 2015-06-03
Update SQLAlchemy target to work with multiple engines.

dylanjbarth:
Fix for https://github.com/spotify/luigi/issues/979.

Introduces an `engine_dict` to the SQLATarget class which allows the storage of multiple engine instances -- useful in scenarios when you need to write to tables in different schemas (and thus need more access to more than one engine).

### DIFF
--- a/luigi/contrib/sqla.py
+++ b/luigi/contrib/sqla.py
@@ -141,6 +141,7 @@ Date: 01/02/2015
 
 
 import abc
+import collections
 import datetime
 import itertools
 import logging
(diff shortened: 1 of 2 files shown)


Tarrasch:
I'll merge on green, since this is already reviewed in #979. :)

erikbern:
very nice!

Tarrasch:
Thanks for the patch!

dylanjbarth:
:+1:

PR #992, opened 2015-06-03
Fix for task history save problem on scheduler

hadesbox:
So today we were testing the task history (we wanted to enhance it) and found a problem related to the refactory of luigi's source code in many points. As we understand now with the new NameSpaces on the luigi clasess, this is part of the Luigis Task Names (hence id)... so there was a bug on the  luigi/tools/parse_task.py that was preventing from saving luigi's history in the enabled db. This mas mainly because of the introduction of a new character to the class namespace the period ".". 

To reproduce the bug just schedule any luigi task, with a period while having enabled the task_history feature. This will cause the scheduler to fail when it tries persist it in the db. To fix this we simply added the dot character to the valid chars parsed by pyparsing in parse_task utility class.

I can open a bug request also if needed.

### DIFF
--- a/luigi/tools/parse_task.py
+++ b/luigi/tools/parse_task.py
@@ -32,7 +32,7 @@ def id_to_name_and_params(task_id):
         DeprecationWarning,
         stacklevel=2)
 
-    name_chars = pp.alphanums + '_'
+    name_chars = pp.alphanums + '_.'
     # modified version of pp.printables. Removed '[]', '()', ','
     value_chars = pp.alphanums + '\'!"#$%&*+-./:;<=>?@\\^_`{|}~'
     parameter = (


Tarrasch:
Looks good, feel free to add a test case as well if you want.

erikbern:
we should really deprecate & remove parse_task....

hadesbox:
I second that... also I would like to have hashes (for example) as luigi task ids, instead of the classname+params. This of course will require a non-trivial amount of effort.. but I think that sooner or later its needed.

I really dont like Classnames and Luigiparams as concatenated string like ids... but thats just me I guess.

erikbern:
I don't think it would require lots of work.

I like the idea of task id's as human readable strings, but now the problem is they have been abused because you can parse the string and get back almost the same data.

Maybe we should make the task id's not contain all data to make it clear that they are just a way to identify a particular piece of work not as a serialization mechanism.

In general it's much better to store the task family and a k/v dict of the parameters

hadesbox:
Cool!
Anyway I'll try to write the test for when the task history is enabled.

@erikbern on a second thought of what you say, I agree that  human redable ids are better, my problem with the current state is that when you add the param values (such as luigiparams as strings, uris, /,&,$,{}) to the id, you get a super long and messy id. Maybe the id should be generate by the luigis classname and an integer number that represents the run date (i.e. 20150507) and the param names only, not the values, or maybe to have a paramname that is the "key" or identity of that param

```
class examples.Bar(luigi.Task):
    num = luigi.IntParameter(key_field=True)
    other = luigi.IntParameter()

    def run(self):
...

    def output(self):
....

```
This way we delegate the use to inform which luigi Params are the one that matter for identifying the class.

erikbern:
There's already such a thing – you can use the `significant` parameter!

hadesbox:
Then my feature request is done :). my ignorance, thanks for the info.

Issue #995, opened 2015-06-06
S3CopyToTable.s3_load_path TypeError


kianho:
I encountered this `TypeError` when subclassing `S3CopyToTable`:
```
...
Traceback (most recent call last):
  File "/home/kian/workspaces/contrib/luigi/luigi/worker.py", line 137, in run
    new_deps = self._run_get_new_deps()
  File "/home/kian/workspaces/contrib/luigi/luigi/worker.py", line 88, in _run_get_new_deps
    task_gen = self.task.run()
  File "/home/kian/workspaces/contrib/luigi/luigi/contrib/redshift.py", line 166, in run
    path = self.s3_load_path()
TypeError: 'str' object is not callable
INFO: Skipping error email. Set `error-email` in the `core` section of the luigi config file to receive error emails.
DEBUG: 1 running tasks, waiting for next task to finish
DEBUG: Asking scheduler for work...
INFO: Done
INFO: There are no more tasks to run at this time
INFO: Worker Worker(salt=532581476, workers=1, host=..., username=kian, pid=23435) was stopped. Shutting down Keep-Alive thread
```
which was fixed by changing https://github.com/spotify/luigi/blob/master/luigi/contrib/redshift.py#L166 from:
```python
path = self.s3_load_path()
``` 
to
```python
path = self.s3_load_path
```
(__I submitted this fix as PR #996__) 

as per the other class properties, which as far as i can tell, aren't explicitly referenced as methods either (otherwise they too would raise the above exception?). 

The snippet I used to generate the above error:
```python
import luigi
    
from luigi.s3 import S3Target, S3Client                                                                          
from luigi.contrib.redshift import S3CopyToTable                                                                 
    
    
class MyS3Task(luigi.Task): 
    local_tsv = luigi.Parameter()
    s3_load_path = luigi.Parameter()                                                                             
    client = luigi.Parameter()
    
    def output(self):
        return S3Target(self.s3_load_path, client=self.client)

    def run(self):
        self.client.put(self.local_tsv, self.output().path)
        return


class MyRedshiftTask(S3CopyToTable):
    host = luigi.Parameter()
    database = luigi.Parameter()
    user = luigi.Parameter()
    password = luigi.Parameter()
    table = luigi.Parameter()
    local_tsv = luigi.Parameter()

    aws_access_key_id = luigi.Parameter()
    aws_secret_access_key = luigi.Parameter()

    columns = [("x", "INT"),
               ("y", "INT")]

[...]

PR #996, opened 2015-06-06
Fix for #995.

kianho:
Simple correction to redshift.py.
See #995 for details.
@rantav @ddaniels888

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -163,7 +163,7 @@ class S3CopyToTable(rdbms.CopyToTable):
         if not (self.table):
             raise Exception("table need to be specified")
 
-        path = self.s3_load_path()
+        path = self.s3_load_path
         connection = self.output().connect()
         if not self.does_table_exist(connection):
             # try creating table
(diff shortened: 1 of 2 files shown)


Tarrasch:
A test case would be nice too of course. :)

kianho:
good idea, will address it over the next few days or so.

Tarrasch:
Yea, given how "trivial" mistake you've found, it seems like this module have been a bit under tested.

kianho:
Since this will be testing S3CopyToTable, which involves both s3 and redshift connectivity, should my test(s) be placed in `test/redshift_test.py` or `test/s3_test.py`?

Tarrasch:
I have no strong opinion. But I would say `test/contrib/retshift_test.py` (move it into `test/contrib/ as well, while you're at it), because the code seem to be defined there.

kianho:
The CI for this PR seems to be failing only on `py33-nonhdfs`, however it is due to a (seemingly) unrelated unit test: https://travis-ci.org/spotify/luigi/jobs/65873895#L1430
any thoughts on how to proceed?

Tarrasch:
Cool! Thanks for the patch! :)

kianho:
No worries, happy to help :)

PR #1000, opened 2015-06-09
Run gcs tests on Travis

Tarrasch:
The implementation uses Travis file encryption[1] and Google's default
way of setting the gcloud credentials described here[2].

[1]: http://docs.travis-ci.com/user/encrypting-files/
[2]: https://developers.google.com/identity/protocols/application-default-credentials


erikbern:
cool!

Tarrasch:
I can fix the nonhdfs errors. @mikekap, how easy would it be to make the tests be able to run in parallel? I think they fail because they assume there's only one integration test running at a time. Or maybe I should just setup different buckets for each test?

mikekap:
It shouldn't be too bad - we can just make each test work on some prefix of the bucket (they already sort of do that, but it's quite ad-hoc).

Tarrasch:
Relevant tests pass now. (I don't get why `bin_test.py` fails so much on only 3.3 recently, any idea @gpoulin?)

Does this look good @mikekap?

test/contrib/gcs_test.py
```diff
@@ -138,7 +150,18 @@ def test_put_file(self):
...
     def create_target(self, format=None):
         return gcs.GCSTarget(bucket_url(self.id()), format=format, client=self.client)
+
+    def test_text(self):
+        """
+        Skipping this test, not sure if UNICODE is working yet.
+
+        Here's a failing which caused me to mute this test for now:
+
+        https://travis-ci.org/spotify/luigi/jobs/66188889
```
mikekap: That looks pretty weird - I tested python3 locally and it passed for me. Could this be an eventual consistency failure? (it looks like GCS just failed to find the file that was just written)

test/contrib/gcs_test.py
```diff
@@ -58,10 +67,13 @@ def setUp(self):
 
             ATTEMPTED_BUCKET_CREATE = True
 
-        for item in self.client.listdir(bucket_url('')):
-            self.client.remove(item)
+        self.client.mkdir(bucket_url(''))
 
+    def tearDown(self):
+        self.client.remove(bucket_url(''), recursive=True)
```
mikekap: You might want to do this in setUp too (in case of failed tests).

mikekap:
LGTM

test/contrib/gcs_test.py
Tarrasch: Oh, I thought it ran regardless.

mikekap: I think in some cases (e.g. KeyboardInterrupt or kill -9) it might not. (Internally at Vine we treat tearDown as "best-effort" and addCleanup as "yes this will execute". The root cause of that was that any failure in setUp might cause tearDown to subsequently fail. We tend to favor addCleanup in setUp instead. This probably doesn't apply here, but it's proven a nice rule of thumb :))

PR #1001, opened 2015-06-10
Fix spark configuration property setup via client.cfg file

pbamba:
Removing the extra quotes when setting the spark configuration via client.cfg.

With the extra quotes the final flag for the spark-submit command might be:
`--conf '"spark.cores.max=2"'`
which is interpreted as 
`key: "spark.cores.max`
`value: 2"`
when it should be 
`--conf 'spark.cores.max=2'`

Since spark only accepts properties that start with the "spark." prefix, the property is ignored altogether.

### DIFF
--- a/luigi/contrib/spark.py
+++ b/luigi/contrib/spark.py
@@ -269,7 +269,7 @@ class SparkSubmitTask(luigi.Task):
         command = []
         if value and isinstance(value, dict):
             for prop, value in value.items():
-                command += [name, '"{0}={1}"'.format(prop, value)]
+                command += [name, '{0}={1}'.format(prop, value)]
         return command
 
     def _flag_arg(self, name, value):
--- a/test/contrib/spark_test.py
+++ b/test/contrib/spark_test.py
@@ -149,7 +149,7 @@ class SparkSubmitTaskTest(unittest.TestCase):
         self.assertEqual(proc.call_args[0][0],
                          ['ss-stub', '--master', 'yarn-client', '--deploy-mode', 'client', '--name', 'AppName',
                           '--class', 'org.test.MyClass', '--jars', 'jars/my.jar', '--py-files', 'file1.py,file2.py',
-                          '--files', 'file1,file2', '--archives', 'archive1,archive2', '--conf', '"Prop=Value"',
+                          '--files', 'file1,file2', '--archives', 'archive1,archive2', '--conf', 'Prop=Value',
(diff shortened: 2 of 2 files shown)


Tarrasch:
LGTM. Are there other spark users who want to comment?

(We really need to fix the intermittent 3.3 failures)

jthi3rry:
I had added the double quotes to cope with values that may have spaces. Never actually used the option as I tend edit the spark defaults directly.

The spark doc (https://spark.apache.org/docs/latest/submitting-applications.html) mentions:
> --conf: Arbitrary Spark configuration property in key=value format. For values that contain spaces wrap “key=value” in quotes

When removing the double quotes, it would end up running a command that looks like this:
`spark-submit ... --conf spark.cores.max=2 ...`
rather than:
`spark-submit ... --conf 'spark.cores.max=2' ...`

@pbamba, I assume it works for you when you remove the double quotes on this example, but what happens if you try with a value that contains spaces?

rapen:
@jthi3rry, shouldn't double quotes be on value and not on key=value? I mean instead of 
--conf 'spark.bla=value' it should be --conf spark.bla='value', thus if value contains spaces it will be ok, property name shouldn't contain spaces.

pbamba:
@jthi3rry, I agree with @rapen but to answer your question if I try
`--conf 'spark.test.test=with space' ` it still works for me.
You can try it yourself by passing this conf to the spark-shell and retrieve the value with
`sc.getConf.get("spark.test.test")`
which will return 
`res0: String = with space`

pbamba:
So basically, without the double quotes, Luigi still runs the command with simple quotes which is enough to handle spaces. I don't think simple or double quotes makes a difference. But when you got both, one kind simply gets ignored and is interpreted as being a part of the key and the value.

jthi3rry:
All good, I was just just wondering where these other quotes were coming from when python executes the command.

I gave it a try and it seems to be working both with spaces from client.cfg as well as when having something like:

```
class MyClass(SparkSubmitTask):
    conf = {'spark.test.test': 'with space'}
```

It looks like Popen handles it on its own when it's given the list of arguments. So you are right @pbamba, the double quotes are redundant and breaking.

Since Popen handles it, any quotes added to the value actually end up being passed to spark, so there's no need to wrap the value either @rapen.

This PR fixes the issue perfectly @Tarrasch.

Tarrasch:
Thanks for participating everyone. :)

pbamba:
Thanks everyone !
Thank you @jthi3rry for taking the time to find out where the simple quotes were coming from.

PR #1004, opened 2015-06-11
Remove optparse support

Tarrasch:
Since argparse is included in python >= 2.7, we don't add any
dependencies by adding this.

Trying to not break old code that much by allowing `use_optparse` to be
passed, but it will only be ignored and not affect anything.


erikbern:
We should break if use_optparse is passed – the interfaces are not compatible with each other. Let's throw an exception and say that it's removed.

erikbern:
other than that LGTM

Tarrasch:
Is the only difference that ArgParseInterface assumes `task == argv[0]` but OptParseInterface has the `--task TaskA` syntax? Maybe we could make ArgParse accept both so we don't break old code? Also was there a reason the explicit `--task` was removed?

erikbern:
the optparse interface doesn't have subcommands so we also register all parameters in a big unordered mess. so it's a bit messy

Tarrasch:
Ah I see

Tarrasch:
Closed in favor of #1036

PR #1005, opened 2015-06-11
fix missing package 'cached_property' in hadoop job pickle

teichmaj:
Simple Bugfix

Current master branch is not adding cached_property to the hadoop job pickle and therefore the task fails on Hadoop with an ImportError: No module named cached_property.
This PR adds the cached_property module to the job pickle.


erikbern:
LGTM

PR #1006, opened 2015-06-11
Fix reading ping_interval config

jkukul:
I just run into this while going through the code. Probably a result of 'copy-paste' mistake.


erikbern:
LGTM

Tarrasch:
Yep, this patch is correct I think. See: https://github.com/spotify/luigi/commit/0e6bec4d5c701a0180f23e67ecbcb96d32fa5654#diff-b438c8dc80f50172445c72006dc7c8f7L324

PR #1011, opened 2015-06-17
Adding support for schemas in checks for redshift table names.

dcheck:
The does_table_exist method doesn't work for table names that include a specific schema.  This patch adds support for redshift tables that specify a schema, without changing behavior for tables that do not specify a schema.

All other methods (create table, truncate) work fine with an explicit schema.

The big benefit of specifying a schema as a table name is that, when luigi creates a table, it can be created in an explicit, non-default schema.

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -216,12 +216,18 @@ class S3CopyToTable(rdbms.CopyToTable):
         """
         Determine whether the table already exists.
         """
-        query = ("select 1 as table_exists "
-                 "from pg_table_def "
-                 "where tablename = %s limit 1")
+
+        if '.' in self.table:
+            query = ("select 1 as table_exists "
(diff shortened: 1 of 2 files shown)


Tarrasch:
Thanks for the patch. Since I have no idea what redshift if, can you ping for other reviewers (potentially using git blame)

dcheck:
@ddaniels888 -- would you be willing to review this patch?  It's a small change to your does_table_exist method to support explicit schemas in table names for redshift.

ddaniels888:
:+1:

ddaniels888:
(we tested it out and it works for our code too)

Tarrasch:
So this is good to merge?

dcheck:
Thanks @ddaniels888!  Are you comfortable signing off on the merge?

ddaniels888:
@dcheck sure, :+1:

PR #1014, opened 2015-06-20
Adds pypy scheduler tests to tox

daveFNbuck:
I just started running my scheduler with PyPy, and it's reducing latency so I'd like to stick with it. I've added tests to tox.ini to ensure that this functionality is not lost with future updates. Psycopg2 isn't
compatible with PyPy, so I moved it out of requirements.txt and into tox.ini for compatible interpreters. If we wanted to make it work, psycopg2cffi is compatible with all versions but would require some code changes.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,33,34}-{cdh,hdp,nonhdfs}, docs, pep8
+envlist = py{27,33,34}-{cdh,hdp,nonhdfs}, pypy-scheduler, docs, pep8
 skipsdist = True
 
 [testenv]
@@ -7,6 +7,7 @@ usedevelop = True
 install_command = pip install {opts} --allow-external mysql-connector-python {packages}
 deps=
   -r{toxinidir}/test/requirements.txt
+  py{27,33,34}: psycopg2
   coverage>=3.6,<3.999
   coveralls
 passenv = USER JAVA_HOME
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -27,6 +29,9 @@ commands =
   python --version
   coverage run test/runtests.py -v --ignore-files='(^\.|^_|^setup\.py$)' \
                                    --ignore-files=not_imported.py \
+                                   pypy: --ignore-files=_mysqldb_test.py \
+                                   pypy: --ignore-files=db_task_history_test.py \
+                                   pypy: --ignore-files=sqla_test.py \
```
Tarrasch: Cool I didn't you know could do this!

Tarrasch:
So we then say that you may use the scheduler with pypy, but not the client code. Ok sure makes sense.

Have you by the way measure if ping and add_task gets faster?

Why do you care about latency by the way? Just so the server doesn't get overloaded or so that interactive users of luigi get a smoother experience?

Thanks for the patch!!

tox.ini
erikbern: instead of using `--ignore-files` is it possible to use `nose.plugins.attr` like we do for the minicluster?

daveFNbuck: @Tarrasch I didn't either, I just really hoped it would work and it did.

@erikbern we can't use nose.plugins.attr because it'll fail with an ImportError if it tries to load these files at all due to not installing them. Psycopg2 doesn't work with pypy but it turns out that sqlalchemy and mysql-connector-python do. Updating the PR to remove the ignores.

daveFNbuck:
I didn't look into ping and add_task. I don't think that ping was slow enough to make a difference on my setup. You can let me know if it seems to do anything for you.

I care because I noticed my latency spike into hundreds of milliseconds as the load increased and with enough workers and requests this was enough to cause the occasional timeout and kill some backfills. I also want to keep resource usage on the machine down as much as possible because I run all of my workers on it and it can get slow sometimes.

tox.ini
erikbern: I think there's some way for nose to ignore an entire module but I might
remember wrong. Wasn't able to find it from a few quick google searches
either. Let's merge this and I can look at it later. LGTM!

PR #1018, opened 2015-06-25
 Postgres port as keyword argument

Marco-Santoni:
Along the lines of https://github.com/spotify/luigi/pull/240

The idea is to simplify the `port` argument usage in Postgres. We can introduce keyword argument with default value set to `None`. By doing so, we achieve the goal of having an explicit argument and respecting backward compatibility as well. Furthermore, we avoid counter-intuitive usage as `localhost:port`.

### DIFF
--- a/luigi/postgres.py
+++ b/luigi/postgres.py
@@ -109,7 +109,9 @@ class PostgresTarget(luigi.Target):
     # Use DB side timestamps or client side timestamps in the marker_table
     use_db_timestamps = True
 
-    def __init__(self, host, database, user, password, table, update_id):
+    def __init__(
+        self, host, database, user, password, table, update_id, port=None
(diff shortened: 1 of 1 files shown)


Tarrasch:
Isn't this breaking backward compatibility? I mean, what happens now if `host="localhost:1234"`?

Marco-Santoni:
You're right, sorry for the confusion. I updated the PR according to you indications. Now, it will allow the usage of both `host:port` and `port=`. In case both are specified, the port in `host:port` is used.

Tarrasch:
Cool! Can you squash the commits please? :)

Marco-Santoni:
:+1:

Tarrasch:
Cool!. Will merge on +1

PR #1019, opened 2015-06-26
A Fix the TypeError whenever prune is called before update has been e…

weidongtoml:
Fix the TypeError whenever prune is called before update has been executed.

 Before worker.update has been called, self.last_active is None; and if
 worker.prune is called at this time, TypeError: unsupported operand type(s)
 for +: 'NoneType' and 'float' will be raised in worker.prune.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -224,7 +224,7 @@ class Worker(object):
     Structure for tracking worker activity and keeping their references.
     """
 
-    def __init__(self, worker_id, last_active=None):
+    def __init__(self, worker_id, last_active=time.time()):
         self.id = worker_id
         self.reference = None  # reference to the worker in the real world. (Currently a dict containing just the host)
         self.last_active = last_active  # seconds since epoch
--- a/test/scheduler_test.py
+++ b/test/scheduler_test.py
@@ -98,5 +98,14 @@ class SchedulerTest(unittest.TestCase):
             cps = reload_from_disk(cps=cps)  # This time without deleting
             self.assertEqual(cps.get_work(worker='D')['task_id'], '4')
 
+    def test_worker_prune_after_init(self):
+        worker = luigi.scheduler.Worker(123)
+
+        class TmpCfg:
+            def __init__(self):
+                self.worker_disconnect_delay = 10
+
+        worker.prune(TmpCfg())
+
 if __name__ == '__main__':
     unittest.main()


erikbern:
LGTM!

Tarrasch:
Can you fix the pep8 issuse?

Tarrasch:
Hmm, this haven't happened to us. Can you explain more as a user scenario when this happens?

weidongtoml:
After the client had request the schedule server to add the task, due to some reason, the client was killed, so the server did not receive any ping, therefore the worker.upate method would never be called. Whenever this worker's prune method is invoked, it will result in the above exception.

Tarrasch:
Ok. Great! But I would really like to see a test case added for this so we don't reintroduce the bug. Can you add that? :)

weidongtoml:
Ok, will try to do that over the weekend. Btw, this is my first time to contribute so I am not sure what you meant by the pep8 issue,  and the Travis CI build failed mentioned here is it due to unit test failure?

Tarrasch:
pep8 is just checking linting (like too many or few newlines). The actual tests did pass. :)

Here's a nice document mentioning how to test luigi locally: https://github.com/spotify/luigi/blob/master/CONTRIBUTING.rst :)

test/scheduler_test.py
```diff
@@ -98,5 +98,14 @@ def reload_from_disk(cps):
             cps = reload_from_disk(cps=cps)  # This time without deleting
             self.assertEqual(cps.get_work(worker='D')['task_id'], '4')
 
+    def test_worker(self):
```
Tarrasch: This name is not that descriptive. Can you rename it a bit :)?

Tarrasch:
Thanks! Can you just squash your commits? This small change don't warrant 5 commits. :)

weidongtoml:
Ok, have squashed them into one.

Tarrasch:
Thanks @weidongtoml ! :)

luigi/scheduler.py
```diff
@@ -224,7 +224,7 @@ class Worker(object):
     Structure for tracking worker activity and keeping their references.
     """
 
-    def __init__(self, worker_id, last_active=None):
+    def __init__(self, worker_id, last_active=time.time()):
```
erikbern: This doesn't look right to me – should be `last_active=None` and then
```python
if last_active is None:
    last_active = time.time()
```

Tarrasch: oh yea it is at module initialization time! Not worker creation time. But wait, if we just want to fix the type error, it might as well be `last_active=0.0`. But then `prune()` would kill the worker prematurely?

@weidongtoml, your thoughts?

weidongtoml: yes, erik is right. Sorry, my mistake, I was thinking __init__(self, worker, last_active=time.time()) would initialise last_active at object creation time. Shall I submit another pull request for a fix?

PR #1024, opened 2015-06-26
Fix #1023 add pruning on get_work of local scheduler

### LINKED ISSUE #1023: Local scheduler does not prune automatically
Take these script: a main task that depends on a task that only completes successfully after some time has passed since launch
```python
import time
import luigi

class DepThatNeedsSomeTime(luigi.Task):
    ready_time = luigi.FloatParameter()

    def run(self):
        assert self.complete()

    def complete(self):
        return  self.ready_time <= time.time()

class MainTask(luigi.WrapperTask):
    launch_time = luigi.FloatParameter(default=time.time())

    def requires(self):
        return [DepThatNeedsSomeTime(self.launch_time + 10)]

if __name__ == '__main__':
    luigi.run()
```

run it with these configs to simplify testing/debugging:
```
python script.py MainTask --scheduler-retry-delay 1 --worker-keep-alive
```

If you run this with central scheduler, the dependency will fail initially, but after some it will be retried because the server calls `scheduler.prune` every minute:
```python
    # prune work DAG every 60 seconds
    pruner = tornado.ioloop.PeriodicCallback(scheduler.prune, 60000)
```
and `scheduler.prune` causes rescheduling/retrying of the failed task (through triggering `SimpleTaskState.prune`)
 
So eventually all work will be done successfully.

However if you run this with the local scheduler (additional `--local-scheduler`), you end up in an endless loop because `scheduler.prune` is never called and the failed task never gets a chance on a retry.

(FYI: I lost quite some time figuring all this out when I was examining the retry behaviour of Luigi with the local scheduler, assuming it would work the same as the central scheduler)

soxofaan:
fix for #1023 

this idea of pruning on `scheduler.get_work` is inspired by `test_external_dependency_completes_later` from worker_external_task_test.py

### DIFF
--- a/test/worker_external_task_test.py
+++ b/test/worker_external_task_test.py
@@ -80,7 +80,8 @@ class WorkerExternalTaskTest(unittest.TestCase):
                                                  worker_disconnect_delay=3,
                                                  disable_persist=3,
                                                  disable_window=5,
-                                                 disable_failures=2)
+                                                 disable_failures=2,
+                                                 prune_on_get_work=True)
 
     def _assert_complete(self, tasks):
         for t in tasks:
@@ -126,15 +127,6 @@ class WorkerExternalTaskTest(unittest.TestCase):
         assert luigi.configuration.get_config().getboolean('core',
                                                            'retry-external-tasks',
                                                            False) is True
-
-        original_get_work = self.scheduler.get_work
-
-        def decorated_get_work(*args, **kwargs):
-            # need to call `prune()` to make the scheduler run the retry logic
-            self.scheduler.prune()
-            return original_get_work(*args, **kwargs)
-
-        self.scheduler.get_work = decorated_get_work
 
         tempdir = tempfile.mkdtemp(prefix='luigi-test-')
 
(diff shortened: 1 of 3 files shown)


soxofaan:
note sure why travis marked this one as failed, errors don't seem to be related with change

Tarrasch:
Hmm, does it work locally? This very innocent change have no build failures at least: https://travis-ci.org/spotify/luigi/builds/68708694 from https://github.com/spotify/luigi/pull/1025/.

soxofaan:
did minor addition to patch

test/worker_external_task_test.py
```diff
@@ -127,15 +128,6 @@ def test_external_dependency_completes_later(self):
                                                            'retry-external-tasks',
                                                            False) is True
 
-        original_get_work = self.scheduler.get_work
```
erikbern: consider using `mock.patch` instead

soxofaan: indeed,
but my patch removes that DIY monkey patching completely, so there is no need for either

erikbern: haha sorry i must have been color blind

Tarrasch:
Yea I got the same intermittent failure for "test_multiple_workers". So merging.

PR #1027, opened 2015-06-29
Fix get_work() slowdown introduced in #986

Tarrasch:
As daveFNbuck said in that PR. The patch in #986 slowed down his
get_work() calls. My theory is that I actually worsened the worst case
complexity in that patch because I assumed that tasks-per-worker is less
than total-pending-tasks. While that's definetly is the case at Spotify.
That is not the case at Houzz.

This patch tries to get the fastest speed in both settings. The setting
that will get a speed-up from this patch is when you have a very few
workers and most of your tasks are in the DONE state.

ping @daveFNbuck

### DIFF
--- a/test/central_planner_test.py
+++ b/test/central_planner_test.py
@@ -761,5 +761,26 @@ class CentralPlannerTest(unittest.TestCase):
             self.assertEqual(self.sch.get_work(worker=str(i))['task_id'], str(i))
             self.sch.add_task(worker=str(i), task_id=str(i), status=DONE)
 
+    def test_get_work_speed(self):
+        """ Test that get_work is fast for few workers and many DONEs.
+
+        In #986, @daveFNbuck reported that he got a slowdown.
+        """
+        # This took almost 4 minutes without optimization.
+        # Now it takes 10 seconds on my machine.
+        NUM_PENDING = 1000
+        NUM_DONE = 200000
+        assert NUM_DONE >= NUM_PENDING
+        for i in range(NUM_PENDING):
+            self.sch.add_task(worker=WORKER, task_id=str(i), resources={})
+
+        for i in range(NUM_PENDING, NUM_DONE):
+            self.sch.add_task(worker=WORKER, task_id=str(i), status=DONE)
+
+        for i in range(NUM_PENDING):
+            res = int(self.sch.get_work(worker=WORKER)['task_id'])
+            self.assertTrue(0 <= res < NUM_PENDING)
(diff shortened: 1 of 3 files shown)


erikbern:
This is why we should probably add some benchmark tests!

Tarrasch:
We could just do reasonable asserts in the tests I've added who are of a benchmarking nature. How do you imagine benchmark tests to look like?

erikbern:
Some script that runs a bunch of benchmarks and measures the total time :) nothing fancy

erikbern:
That way we can run the script before and after and compare

Tarrasch:
Well, I do run `tox -e py27-nonhdfs -- test/central_planner_test.py:CentralPlannerTest.test_quadratic_behavior` locally and compare before and after. But of course we don't find regressions for each test here.

test/central_planner_test.py
```diff
@@ -761,5 +761,26 @@ def test_quadratic_behavior(self):
             self.assertEqual(self.sch.get_work(worker=str(i))['task_id'], str(i))
             self.sch.add_task(worker=str(i), task_id=str(i), status=DONE)
 
+    def test_get_work_speed(self):
```
daveFNbuck: You don't seem to be calling get_work in this test. Did add task speed up too?

Tarrasch: I do see below vvv

test/central_planner_test.py
```diff
@@ -761,5 +761,26 @@ def test_quadratic_behavior(self):
...
+        NUM_DONE = 200000
+        assert NUM_DONE >= NUM_PENDING
+        for i in range(NUM_PENDING):
+            self.sch.add_task(worker=WORKER, task_id=str(i), resources={})
+
+        for i in range(NUM_PENDING, NUM_DONE):
+            self.sch.add_task(worker=WORKER, task_id=str(i), status=DONE)
+
+        for i in range(NUM_PENDING):
+            res = int(self.sch.get_work(worker=WORKER)['task_id'])
```
Tarrasch: Here @daveFNbuck :)

daveFNbuck: I guess I shouldn't be reviewing code first thing in the morning.

daveFNbuck:
Looks good to me. You could get slightly better performance in some cases by storing each worker's tasks hashed by state, but that's probably not worth it.

Tarrasch:
Great. Merging then. :)

PR #1028, opened 2015-06-29
DONTMERGE Dummy test to check if gcs work

Tarrasch:
I actually expect this to not work (so yes, build is broken). I need to do some hack similar to the one here: http://docs.travis-ci.com/user/pull-requests/


Tarrasch:
What on earth where Travis people thinking ... I assumed it would just "ignore" the decryption step automatically. I guess it has to be manually hacked. :-/

See for example: https://travis-ci.org/spotify/luigi/jobs/68813221

Issue #1029, opened 2015-06-29
Why is configuration file called 'client.cfg'


soxofaan:
Maybe a stupid question, but why is the configuration file called `client.cfg`, when it typically contains non-client (e.g. scheduler) stuff as well?

Wouldn't it be nicer to have at least some hint of "luigi" in the name, e.g. `luigi.cfg`?

Can be handy when you use the "use config from current working dir" from a directory with loads of other configs and files.

Tarrasch:
I guess this is for historical reasons.

But, I think, in the long run, that it should only contain client-side related config. So I'm not sure we should rename it as we might change our minds again. No strong opinions though.

Tarrasch:
As for a luigi hint in the name. It's by default stored in `/etc/luigi/client.cfg` no?

soxofaan:
indeed, for the global config `/etc/luigi/client.cfg`, it is,
but like I noted, you can also use `./client.cfg` (from working dir).
It's not that unusual to have your project's root as working dir, which is typically littered with all sorts of config files. Having a non descriptive name like `client.cfg` is not really helpful for overview :)

soxofaan:
Current version of pull request above (#1030) is the easy part: just adding support for `./luigi.cfg`

the harder part is deciding about:
- Also add support `/etc/luigi/luigi.cfg`?
- whether to choose one over the other and update documentation/code comments accordingly?
- also, about:

    > in the long run, that it should only contain client-side related config

   how to split client vs non-client, e.g. `luigi.cfg`, `luigid.cfg`? What with non-client stuff in `luigi.cfg` then?

Tarrasch:
Good points all over and I really appreciate the PR you sent.

I like the `luigid.cfg` idea. :)

soxofaan:
I went a step further and added a commit that updates to "luigi.cfg" in various parts of documentation. comments, ...

Feel free to comment

I was also considering to also add support for `/etc/luigi/luigi.cfg` to make the documentation less confusing. But I thought to first await some feedback from the maintainers about all this

leafjungle:
I find that both the client ( interface.py, cmdline.py, etc) and server (scheduler.py) will read the client.cfg.  
scheduler.py read client.cfg for "resources".

So it is a bit confused. what will luigi do if the client and server are deployed in different machines?  we have to put client.cfg in both machines?

Tarrasch:
Yea, I guess that's why they renamed it from `client.cfg`, as it's not only read by the client.

[...]

PR #1030, opened 2015-06-29
Issue 1029: add support for config file `luigi.cfg` in working dir

soxofaan:
See #1029


Tarrasch:
I think this makes total. Because I broke the build by mistake I'll have to fix that first though. :)

Tarrasch:
I'm all for this! Any comments from the community?

erikbern:
sounds good to me!

luigi/configuration.py
```diff
@@ -40,7 +40,7 @@
 class LuigiConfigParser(ConfigParser):
     NO_DEFAULT = object()
     _instance = None
-    _config_paths = ['/etc/luigi/client.cfg', 'client.cfg']
+    _config_paths = ['/etc/luigi/client.cfg', 'luigi.cfg', 'client.cfg']
```
erikbern: shouldn't this also include /etc/luigi/luigi.cfg?

soxofaan: indeed, that was one of my questions in #1029

erikbern: might be good if we can also show a warning on the cmd line that the location has moved

Tarrasch: Are we also planning to remove the old path?

doc/configuration.rst
```diff
@@ -4,7 +4,7 @@ Configuration
 All configuration can be done by adding a configuration files. They are looked for in:
 
  * ``/etc/luigi/client.cfg``
```
Tarrasch: @soxofaan Did you deliberately not change the documentation here? Because in the code you wrote that client.cfg is deprecated.

Tarrasch:
ping @soxofaan :)

doc/configuration.rst
soxofaan: I don't completely recall, but I guess that my main issue was that it is unobvious what `$CWD/client.cfg` is about, while `/etc/luigi/client.cfg` is more self-explanatory , so I focussed on the former one and left the latter as-is

soxofaan: but apart from that, I'd support changing `/etc/luigi/client.cfg` as well

PR #1032, opened 2015-06-29
travis: Dont run gcloud whithout credentials

Tarrasch:
This also unbreaks the build from the merge of spotify/luigi#1000.

Yes, this is a copy of #1031 kind of, but sent from the spotify/luigi and not from tarrasch/luigi.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,8 @@ install:
   - pip install tox
 
 before_install:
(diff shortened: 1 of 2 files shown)


Tarrasch:
Just merging to unbreak the build. Now it seem to never run the gcs tests. Annoyingly travis incorrectly "optimizes" so that it reuses the same build for the same commit, despite that they are run from different environments (with and without secrets).

PR #1034, opened 2015-06-29
Use a valid default value for connect_args of SQLAlchemyTarget.__init__

soxofaan:
```python
import luigi.contrib.sqla

target = luigi.contrib.sqla.SQLAlchemyTarget(
    connection_string="sqlite:///tmp.sql",
    target_table="foobar",
    update_id="foo1234",
)

print target.exists()
```

This fails with something like:
```
  File ".../poc.py", line 11, in <module>
    print target.exists()
  File ".../luigi.git/luigi/contrib/sqla.py", line 233, in exists
    self.create_marker_table()
  File ".../luigi.git/luigi/contrib/sqla.py", line 250, in create_marker_table
    engine = self.engine
  File ".../luigi.git/luigi/contrib/sqla.py", line 204, in engine
    echo=self.echo
  File ".../site-packages/sqlalchemy/engine/__init__.py", line 386, in create_engine
    return strategy.create(*args, **kwargs)
  File ".../site-packages/sqlalchemy/engine/strategies.py", line 83, in create
    cparams.update(pop_kwarg('connect_args', {}))
TypeError: 'NoneType' object is not iterable
```

commit fixes that

### DIFF
--- a/luigi/contrib/sqla.py
+++ b/luigi/contrib/sqla.py
@@ -160,11 +160,10 @@ class SQLAlchemyTarget(luigi.Target):
     to create a task to write to the database.
     """
     marker_table = None
-    connect_args = {}
     _engine_dict = {}  # dict of sqlalchemy engine instances
     Connection = collections.namedtuple("Connection", "engine pid")
 
-    def __init__(self, connection_string, target_table, update_id, echo=False, connect_args=None):
+    def __init__(self, connection_string, target_table, update_id, echo=False, connect_args={}):
         """
         Constructor for the SQLAlchemyTarget.
 
--- a/test/contrib/sqla_test.py
+++ b/test/contrib/sqla_test.py
@@ -35,40 +35,19 @@ from nose.plugins.attrib import attr
 if six.PY3:
     unicode = str
 
-#################################
-# Globals part of the test case #
-#################################
-TEMPDIR = tempfile.mkdtemp()
-SQLITEPATH = os.path.join(TEMPDIR, "sqlatest.db")
-CONNECTION_STRING = "sqlite:///%s" % SQLITEPATH
-TASK_LIST = ["item%d\tproperty%d\n" % (i, i) for i in range(10)]
-CONNECT_ARGS = {'timeout': 5.0}
-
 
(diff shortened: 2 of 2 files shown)


Tarrasch:
There are so many intermittent failures. This is quite bad at the moment.

soxofaan:
FYI: in a follow up commit I improved the unit tests of sqla which did hide the issue due to implicit/hidden dependencies/side effects between tests (global variable code smell)

Tarrasch:
Can you review @gouthambs ?

luigi/contrib/sqla.py
```diff
@@ -160,11 +160,10 @@ class SQLAlchemyTarget(luigi.Target):
     to create a task to write to the database.
     """
     marker_table = None
-    connect_args = {}
     _engine_dict = {}  # dict of sqlalchemy engine instances
     Connection = collections.namedtuple("Connection", "engine pid")
 
-    def __init__(self, connection_string, target_table, update_id, echo=False, connect_args=None):
+    def __init__(self, connection_string, target_table, update_id, echo=False, connect_args={}):
```
gouthambs: One suggestion I have here is to leave `connect_args` default value as `None` and assign inside the function as
`self.connect_args = {} if connect_args is None else connect_args`

This is to avoid having mutables as default values in function parameters.

gouthambs:
@soxofaan Thank you for cleaning up the test code. Looks tidy without all the globals. I am a little puzzled why the tests didn't catch this error earlier though. With your cleaned out tests, would the tests fail without the patch you have in the `sqla.py`? That would be a good proof that the tests are catching what they are meant to.

Tarrasch:
Thanks!!

soxofaan:
@gouthambs : the reworked test indeed failed on unpatched sqla.py. 

The problem was that the original tests had hidden interdependencies, through reusing the same temp dir for each test. Even though the temp dir was deleted between tests, the `_engine_dict` cache didn't reset, causing cache hits that shouldn't happen.
I refactored the test to make sure each tests starts from a unique temp dir to work in

If you would run each of the original tests in isolated process, I guess they would also have failed.

thanks for merging

PR #1035, opened 2015-06-30
Add packages config option to spark task.

blooshirt:
Add packages configuration option to spark task in order to support applications like:
https://github.com/databricks/spark-csv#using-with-spark-shell

### DIFF
--- a/luigi/contrib/spark.py
+++ b/luigi/contrib/spark.py
@@ -122,6 +122,10 @@ class SparkSubmitTask(luigi.Task):
         return self._list_config(configuration.get_config().get("spark", "jars", None))
 
     @property
+    def packages(self):
+        return self._list_config(configuration.get_config().get("spark", "packages", None))
+
+    @property
     def py_files(self):
         return self._list_config(configuration.get_config().get("spark", "py-files", None))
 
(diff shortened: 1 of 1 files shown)


luigi/contrib/spark.py
```diff
@@ -122,6 +122,10 @@ def jars(self):
         return self._list_config(configuration.get_config().get("spark", "jars", None))
 
     @property
+    def packages(self):
```
erikbern: can we make this cleaner using a `Config` class instead?

erikbern:
LGTM

luigi/contrib/spark.py
```diff
@@ -122,6 +122,10 @@ def jars(self):
...
+        return self._list_config(configuration.get_config().get("spark", "packages", None))
```
Tarrasch: Yea use `Config` class for this please.

blooshirt: Are you suggesting I make a new class just for the new parameter?

erikbern: Let's just merge this for now and deal with the config refactoring in a
separate PR

Tarrasch: Yes, see http://luigi.readthedocs.org/en/latest/api/luigi.configuration.html. As an advantage, you get command line parameters for free.

erikbern: yeah but this complexity has been there for a long time in the spark class, so we should fix it in a separate PR

blooshirt: Agreed Erik. I'll start working on a separate PR for refactoring the entire config of the spark class. In the mean-time this functionality is necessary so I think we should just merge this in.

Tarrasch: Ah I thought it was the first use of `get_config()`. Great then. :)

PR #1037, opened 2015-07-01
Alternative style for the visualiser 

stephenpascoe:
using DataTables and the AdminLTE dashboard style.

Displays all tasks in a single DataTable which can be filtered and ordered using the
search field.  Info boxes display totals for running/failed/pending/etc. tasks.

Exceptions are displayed in child rows that can be accessed using the RHS control.


stephenpascoe:
Sorry, accidentally submitted merge request.

PR #1038, opened 2015-07-01
Actually run BigQuery tests

Tarrasch:
Creating the PR  in name of @rtharper, I just amended his original commit with adding/removing blank lines to fix pep8 issues.


Tarrasch:
The failure is just a false alarm @rtharper. :) (we *really* need to fix it)

You can see that it actually runs the tests too [1]. Thanks!

[1]: https://travis-ci.org/spotify/luigi/jobs/69160201

PR #1044, opened 2015-07-06
Fix Worker initializtion bug.

weidongtoml:
In Worker.__init__, last_active parameter is default to None, and in actual initialisation, self.last_active will set to the time of Worker creation.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -226,7 +226,7 @@ class Worker(object):
     Structure for tracking worker activity and keeping their references.
     """
 
-    def __init__(self, worker_id, last_active=time.time()):
+    def __init__(self, worker_id, last_active=None):
         self.id = worker_id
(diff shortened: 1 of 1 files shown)


Tarrasch:
The way you revert this is a bit worrying. I thought the test in #1019 would fail if you reverted the change. No? Also, won't you reintroduce the `TypeError`?

Were you not supposed to do a `if last_active is None: last_active = time.time()` in the body of `__init__`?

weidongtoml:
Well, it isn't exactly a revert. because on line 232, there's this:
    self.last_active = last_active or time.time() 
already existed, therefore all I have to do is change the default value to None.  
Is it not?

Tarrasch:
Ah!! This is so confusing, this patch is also touching code close to #1024.

But you're right, the end result will be "correct" once this is merged.

PR #1045, opened 2015-07-06
Avoid intermediate variables in luigi/__init__.py for improved IDE-friendliness

soxofaan:
When I use luigi like this:
```python
import luigi.task
class MyTask(luigi.task.Task):
    def complete(self):
        ...
```
my IDE (PyCharm) provides "code intelligence" tricks like flagging the `complete` method as an override of the parent class' method, I can "go to declaration", etc

When I import luigi like this however:
```python
import luigi
class MyTask(luigi.Task):
    ...
```
all this fancy "code intelligence" does not work.

This pull request fixes that by avoiding intermediate variables in `luigi/__init__.py`

### DIFF
--- a/test/import_test.py
+++ b/test/import_test.py
@@ -42,3 +42,27 @@ class ImportTest(unittest.TestCase):
             for f in files:
                 if f.endswith('.py') and not f.startswith('_'):
                     __import__(package + '.' + f[:-3])
+
+    def import_luigi_test(self):
+        """
+        Test that the top luigi package can be imported and contains the usual suspects.
+        """
+        import luigi
+
+        # These should exist (if not, this will cause AttributeErrors)
+        expected = [
+            luigi.Event,
+            luigi.Config,
+            luigi.Task, luigi.ExternalTask, luigi.WrapperTask,
+            luigi.Target, luigi.LocalTarget, luigi.File,
+            luigi.namespace,
+            luigi.RemoteScheduler,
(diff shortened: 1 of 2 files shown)


gpoulin:
LGTM. However, it would be good if we could keep the comments.

Tarrasch:
I actually like that the comments disappeared. They were pretty useless and just drew attention from whoever was reading the source code.

test/import_test.py
```diff
@@ -42,3 +42,26 @@ def import_test(self):
...
+            luigi.namespace,
+            luigi.RemoteScheduler,
+            luigi.RPCError,
+            luigi.run, luigi.build,
+            luigi.Parameter,
+            luigi.DateHourParameter, luigi.DateMinuteParameter, luigi.DateParameter,
+            luigi.DateIntervalParameter, luigi.TimeDeltaParameter,
+            luigi.IntParameter, luigi.FloatParameter,
+            luigi.BooleanParameter, luigi.BoolParameter,
+        ]
```
erikbern: hm seems like something is missing here – where do you check that these are defined?

soxofaan: indeed, it might not look that obvious, but if one of these items does not exist, it will raise AttributeError, failing the test.
I added an extra assert and comment.

Tarrasch:
Again, this looks very good to me, thanks!. :)

PR #1048, opened 2015-07-07
Add SNS notifications

cpapazian:
Adding support for error notifications via Amazon SNS. Set up by adding a setting for error-sns-topic to [core] and optionally specify region and keys in [sns].

Note: SNS limits subject length to 100 characters, so I just truncated the subject passed in ... not the greatest solution.

### DIFF
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -157,6 +157,15 @@ def send_email(subject, message, sender, recipients, image_png=None):
         send_email_smtp(config, sender, subject, message, recipients, image_png)
 
 
+def send_notification(subject, message, topic):
+    import boto.sns
+    config = configuration.get_config()
+    con = boto.sns.connect_to_region(config.get('sns', 'region', 'us-east-1'),
+                                     aws_access_key_id=config.get('sns', 'AWS_ACCESS_KEY', None),
+                                     aws_secret_access_key=config.get('sns', 'AWS_SECRET_KEY', None))
+    con.publish(topic, message, subject[:100])
+
+
 def send_error_email(subject, message):
     """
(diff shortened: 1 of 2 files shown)


erikbern:
Looks good although at some point we should refactor to use the config object!

luigi/notifications.py
```diff
@@ -157,6 +157,15 @@ def send_email(subject, message, sender, recipients, image_png=None):
         send_email_smtp(config, sender, subject, message, recipients, image_png)
 
 
+def send_notification(subject, message, topic):
```
Tarrasch: Name is *way* to generic for this patch to be accepted I think. Can you rename it to maybe `send_sns_notification` or something else you find suitable?

Tarrasch:
Other than my comment. Looks good to me. :)

Tarrasch:
Also, please add some sort of a test case! This code might be re-factored one day and then suddenly this logic might stop working.

cpapazian:
@Tarrasch is https://github.com/spotify/luigi/blob/master/test/email_test.py the only test for this module? was looking around for something to start from, and that was all i found ... 

will change the name tomorrow morning.

Tarrasch:
I would rename `test/email_test.py` to `test/notifications_test.py` and then add your tests there. :)

cpapazian:
@Tarrasch not sure what these errors on Travis are about. unless the `@with_config` decorator is doing something weird, my tests should not call through to AWS. and the first test case confirms that `publish` is not called when the config is not set ... thoughts/

Tarrasch:
Can you rebase please? The build was failing at the time.

Tarrasch:

PR #1043, opened 2015-07-05
tox: Don't fail travis on linkchecks

Tarrasch:
This is yet another intermittent failure causing pain for a luigi maintainer. We can't make sure that external websites are up and responding.

From what I saw from the docs[1]. There's no feature to do retries or accept intermittent failures. Hence, let us just ignore failures. This commit was inspired by [2].

[1]: http://sphinx-doc.org/builders.html#sphinx.builders.linkcheck.CheckExternalLinksBuilder
[2]: https://travis-ci.org/spotify/luigi/jobs/69595616

PR #1051, opened 2015-07-08
Disable intermittent gcs failures

Tarrasch:
Inspired by the failures from https://github.com/spotify/luigi/pull/1043


Tarrasch:
Ping @mikekap, in case you want to look in to it. :)

PR #1056, opened 2015-07-10
SunGrid Engine support

jfeala:
- Submits luigi tasks to a SunGrid Engine cluster using qsub
- Executes "work" function on pickled task class
- Monitors task completion using qstat

Adapted from Alex Wiltschko's (@alexbw) implementation of LSF (https://github.com/dattalab/luigi/tree/lsf_submission).

### DIFF
--- a/luigi/contrib/sge.py
+++ b/luigi/contrib/sge.py
@@ -49,7 +49,7 @@ logger.propagate = 0
 
 POLL_TIME = 5  # decided to hard-code rather than configure here
 
-## Defaults for configuration parameters. Use StarCluster defaults. Number of CPUs is further configurable as task parameter
+# Defaults for configuration parameters. Use StarCluster defaults. Number of CPUs is further configurable as task parameter
 DEFAULT_HOME = configuration.get_config().get('sge', 'shared-tmp-dir', default=None) or '/home'
 DEFAULT_PE = configuration.get_config().get('sge', 'parallel-env', default=None) or 'orte'
 DEFAULT_NCPU = configuration.get_config().get('sge', 'n-cpu', default=None) or 2
@@ -57,12 +57,10 @@ DEFAULT_NCPU = configuration.get_config().get('sge', 'n-cpu', default=None) or 2
 
 def clean_task_id(task_id):
     '''Clean the task ID so qsub allows it as a "name" string. From the "name"
-    definition in sge_types(1):
+    definition in sge_types(1): The name may be any arbitrary alphanumeric ASCII
+    string, but may not contain "\n", "\t", "\r", "/", ":", "@", "\", "*", or "?".
 
-        The name may be any arbitrary alphanumeric ASCII string, but may not contain
-        "\n", "\t", "\r", "/", ":", "@", "\", "*", or "?".
-
-    Also remove some other punctuation and spaces just in case
+    Also remove some other punctuation and spaces just in case.
     '''
     for c in ['\n', '\t', '\r', '/', ':', '@', '\\', '*', '?', ',', '=', ' ', '(', ')']:
         task_id = task_id.replace(c, '-')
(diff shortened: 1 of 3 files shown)


Tarrasch:
Hmm, have you considered adding an entrypoint to the python packaging as well? So you can run it with something like `luigi-swf` instead of `python blah.py`? I much would prefer the former to be the standard option.

jfeala:
Just curious, why do you prefer that, since this is using the same model already followed by `mrrunner.py` and `pyspark_runner.py`? (Where the master node pickles a Task and its dependencies, and then the worker node unpacks it and runs with `python blah.py`).

Ok with me to add an entrypoint, but the tradeoff is that it adds a dependency that luigi is installed each worker environment.

Tarrasch:
Ah I see what you mean, but isn't there for a way for users to run `luigi-swf` binary and it packages stuff to run with `python whatever.py` from the workes nodes?

jfeala:
The user should sublass `SGEJobTask` in their luigi workflows, as in the
tests:
https://github.com/spotify/luigi/pull/1056/files#diff-927de6b27f9e382aeb183ce33ccfe360R61

I still don't understand why you need a separate binary instead of using
the standard luigi CLI (`luigi.run()`) in a workflow script that contains
`SGEJobTask`s. Is there a similar pattern you can point me to?

Also, why do you say `luigi-swf` when we're talking about SunGrid Engine
(SGE)? Maybe we're talking about different things?


On Fri, Jul 10, 2015 at 9:18 AM, Arash Rouhani <notifications@github.com>
wrote:

> Ah I see what you mean, but isn't there for a way for users to run
> luigi-swf binary and it packages stuff to run with python whatever.py
> from the workes nodes?
>
> —
> <https://github.com/spotify/luigi/pull/1056#issuecomment-120408621>.
>



-- 
Jacob Feala, PhD
Scientific Investigator II, Bioinformatics
H3 Biomedicine
300 Technology Square
Cambridge, MA 02139
(mobile) 858-366-2951
Jacob_Feala@H3Biomedicine.com

-- 
[This e-mail message may contain privileged, confidential and/or 
proprietary information of H3 Biomedicine. If you believe that it has been 
sent to you in error, please contact the sender immediately and delete the 
message including any attachments, without copying, using, or distributing 
any of the information contained therein. This e-mail message should not be 
interpreted to include a digital or electronic signature that can be used 
to authenticate an agreement, contract or other legal document, nor to 
reflect an intention to be bound to any legally-binding agreement or 
contract.]

luigi/contrib/sge.py
```diff
@@ -0,0 +1,257 @@
...
+                if not errors:
+                    logger.info('Job is done')
+                else:
+                    logger.error('Job has FAILED:\n' + '\n'.join(errors))
+                break
+            else:
+                logger.info('Job status is UNKNOWN!')
+                logger.info('Status is : %s' % sge_status)
+                break
+                raise Exception("job status isn't one of ['r', 'qw', 't']: %s" % sge_status)
```
erikbern: not reachable

Tarrasch:
Ah yea. I totally confused the acronyms. Nevermind me. I thought this was for whatever was discussed here:  https://groups.google.com/forum/#!topic/luigi-user/GOwAF3axF58 :)

Tarrasch:
Yea, ignore everything I've set up until this point ^^. :)

jfeala:
ha, no problem!

luigi/contrib/sge.py
jfeala: fixed, thanks!

holtgrewe:
:+1:

SGE support would be great!

jfeala:
resubmitted this PR as https://github.com/spotify/luigi/pull/1124

PR #1059, opened 2015-07-10
Fix luigi run use case

Tarrasch:
Please review @tym-oao :)

### DIFF
--- a/test/interface_test.py
+++ b/test/interface_test.py
@@ -20,14 +20,16 @@ from helpers import unittest
 import luigi
 import luigi.date_interval
 import luigi.notifications
+import sys
 from luigi.interface import core, Interface, WorkerSchedulerFactory
 from luigi.worker import Worker
-from mock import Mock
+from mock import Mock, patch
+from helpers import LuigiTestCase
 
 luigi.notifications.DEBUG = True
 
 
-class InterfaceTest(unittest.TestCase):
+class InterfaceTest(LuigiTestCase):
 
     def setUp(self):
         self.worker = Worker()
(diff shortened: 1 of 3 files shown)


test/interface_test.py
```diff
@@ -67,6 +70,19 @@ def test_interface_run_with_run_failure(self):
...
 
+    def test_just_run_main_task_cls(self):
+        class MyTestTask(luigi.Task):
+            pass
+
+        class MyOtherTestTask(luigi.Task):
+            my_param = luigi.Parameter()
+
+        with patch.object(sys, 'argv', ['my_module.py', '--no-lock', '--local-scheduler']):
+            luigi.run(main_task_cls=MyTestTask)
```
erikbern: you can just pass `cmdline_args` to `luigi.run` instaed of patching sys.argv

tym-oao:
Looks good! My original breaking case now works. Thanks for the fast attention!

tym-oao:
Personally, I will still move to preferring `luigi` script over `main_task_cls`, but I appreciate not having to scramble to do that immediately all at once :)

erikbern:
We could throw a deprecation warning to nudge people to move

test/interface_test.py
Tarrasch: Well yes, but the whole purpose is to test the behaviour when you don't pass that argument. As the commit message says. :)

Tarrasch:
Erm, I don't know why. But apparently there are these 3 test faillures that always happen. I don't dare to merge this PR just yet. :-/

Tarrasch:
Seem to be the same intermittent failures as I talked about here [1]. Merging to unbreak stuff for our beloved HEAD-users. :)

[1]: https://github.com/spotify/luigi/pull/1064#issuecomment-120597484

PR #1062, opened 2015-07-10
Handle UnicodeDecodeError on write to Postgres.

tym-oao:
Unfortunately, I don't have a good automated test to show this, but the following simple example returns a UnicodeDecodeError with the current `postgres.py` in master when it attempts to write the third row (or any text with extended Unicode characters).

```
#coding: utf-8

import luigi
import luigi.postgres
import logging

logger = logging.getLogger("luigi-interface")

testdata = ([99, 'My fake plants died because I did not pretend to water them.'],
       [100, 'I always arrive late at the office, but I make up for it by leaving early.'],
       [101, 'ஸ்றீனிவாஸ ராமானுஜன் ஐயங்கார்'])

class TrivialPgLoad(luigi.postgres.CopyToTable):
    host = 'localhost'
    database = 'spotify'
    user = 'spotify'
    password = 'guest'
    table = 'trivium'
    columns = [("trival_id", "INT"),
               ("description", "TEXT")]

    def rows(self):
        for row in testdata:
            yield row

if __name__ == '__main__':
    luigi.run(main_task_cls=TrivialPgLoad)
```

The traceback looks like:

```
  File "/Users/thomas/Dropbox/envs/luigi-pg-unicode/lib/python2.7/site-packages/luigi/worker.py", line 137, in run
  File "/Users/thomas/Dropbox/envs/luigi-pg-unicode/lib/python2.7/site-packages/luigi/worker.py", line 88, in _run_get_new_deps
    task_gen = self.task.run()
  File "/Users/thomas/Dropbox/envs/luigi-pg-unicode/lib/python2.7/site-packages/luigi/postgres.py", line 318, in run
    tmp_file.write(rowstr.encode('utf-8'))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe0 in position 4: ordinal not in range(128)
```

The load to Postgres succeeds with my fix.

### DIFF
--- a/luigi/postgres.py
+++ b/luigi/postgres.py
@@ -315,7 +315,10 @@ class CopyToTable(rdbms.CopyToTable):
                 logger.info("Wrote %d lines", n)
             rowstr = self.column_separator.join(self.map_column(val) for val in row)
             rowstr += "\n"
-            tmp_file.write(rowstr.encode('utf-8'))
+            try:
+                tmp_file.write(rowstr.encode('utf-8'))
+            except UnicodeDecodeError:
+                tmp_file.write(rowstr)
 
         logger.info("Done writing, importing at %s", datetime.datetime.now())
         tmp_file.seek(0)


erikbern:
Not thinking through this fully, but I would rather make the conversion explicit – it seems like either all the rows are utf-8, or none of them are.

tym-oao:
Yes, it actually works just as well for me to take the encode() call out of line 318 completely. I used the try/except block because I didn't feel 100% confident that doing it that wouldn't break other use cases.

tym-oao:
There was some discussion of this previously over here: https://groups.google.com/forum/m/#!searchin/luigi-user/Unicodedecodeerror/luigi-user/2UyGTJHFx_8 ... ISTM the issue has to do with the data getting converted to utf-8 twice in some cases. The issue appears if the source data is already Unicode. Pure Python str inserts work with the current code.

gpoulin:
The issue here (like at many place in Luigi) is that Binary and String data are missed. The postgres task is build to handle string and not binary data. However, in the example that you gave you use binary data  (I assume that you are using python2). The easy fix would be to use `unicode` instead of `str`.

```python
testdata = ([99, u'My fake plants died because I did not pretend to water them.'],
       [100, u'I always arrive late at the office, but I make up for it by leaving early.'],
       [101, u'ஸ்றீனிவாஸ ராமானுஜன் ஐயங்கார்'])
```

tym-oao:
Again, it is precisely the case where the data are already Unicode that fails the current code. Str data load fine as is.

gpoulin:
I had a closer look and there is indeed an issue, but it's not with the line in the traceback, but with this [one](https://github.com/spotify/luigi/blob/master/luigi/postgres.py#L265). I'll send a PR soon.

@erikbern I wouldn't require the user to encode explicitly the data since the text need to be encoded in utf-8 since it's what the connection [expect](https://github.com/spotify/luigi/blob/master/luigi/postgres.py#L197).  The current implementation doesn't seems to handle binary data, so the users should provide text data which is `unicode` for python2. Doing otherwise would create issue if we want to handle binary data.

tym-oao:
@gpoulin - thanks! I knew about line 265—that's what I was trying to say about unicode lines having encode() called twice. Since you seem to have a clearer idea how to fix this than I do, I will look forward to your PR, and we can close this one now.

PR #1063, opened 2015-07-10
Fix upload progress indicator in Google Cloud Storage target

tym-oao:
Line 144 of luigi/contrib/gcs.py in master tries to multiply 100 * a MediaUploadProgress object, resulting in TypeError. A little hard to detect, unless uploading a file that takes several chunks to upload. This sample code:

```
# coding: utf-8

import luigi
from luigi.contrib import gcs
import os

bucket_name = os.getenv('GSCBUCKET', 'warehouse-hadoop')

class GCSUpload(luigi.Task):

    def output(self):
        target = 'gs://%s/tmp/big' % bucket_name
        return luigi.contrib.gcs.GCSTarget(target)

    def run(self):
        f = self.output().open(mode='w')
        f.write("\0" * 10000000)
        f.close()
```
Results in the following error:
```
  File "/Users/thomas/Dropbox/envs/etl-luigi/lib/python2.7/site-packages/luigi/contrib/gcs.py", line 145, in _do_put
    logger.debug('Upload progress: %.2f%%', 100 * progress)
TypeError: unsupported operand type(s) for *: 'int' and 'MediaUploadProgress'
```
While patched gcs.py in my branch logs progess as expected.

### DIFF
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -139,9 +139,9 @@ class GCSClient(luigi.target.FileSystem):
         while response is None:
             error = None
             try:
-                progress, response = request.next_chunk()
-                if progress:
-                    logger.debug('Upload progress: %.2f%%', 100 * progress.progress())
+                status, response = request.next_chunk()
+                if status:
+                    logger.debug('Upload progress: %.2f%%', 100 * status.progress())
             except errors.HttpError as err:
                 error = err
                 if err.resp.status < 500:
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -140,8 +140,8 @@ class GCSClient(luigi.target.FileSystem):
             error = None
             try:
                 progress, response = request.next_chunk()
-                if progress is not None:
-                    logger.debug('Upload progress: %.2f%%', 100 * progress)
+                if progress:
+                    logger.debug('Upload progress: %.2f%%', 100 * progress.progress())
             except errors.HttpError as err:
                 error = err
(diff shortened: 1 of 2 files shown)


Tarrasch:
Thanks! But I would like to have a text case too, if possible. :)

tym-oao:
I'll see what I can do

tym-oao:
Ahem. The current `gcs_test.py` is like:

```
@attr('gcloud')
class GCSTargetTest(_GCSBaseTestCase, FileSystemTargetTestMixin):

    def test_text(self):
        """
        Skipping this test, not sure if UNICODE is working yet.

        Here's a failing which caused me to mute this test for now:

        https://travis-ci.org/spotify/luigi/jobs/66188889
        """
        pass

    def test_atomicity(self):
        """
        Intermittent failure in
        https://travis-ci.org/spotify/luigi/jobs/69597336
        """
        pass
```

Just sayin'. ;)
Still, I'll see if I can make a test that works. Despite the comment above, I don't think unicode is the issue here.

tym-oao:
OK, looking at this a little closer, there is a `test_put_file` in the `GCSClientTest` class that isn't just a `pass`. Right now the file just says "hi"—I think all I have to do is get it to generate a larger file, and it will show the TypeError.

tym-oao:
`gcs_test.py` updated with a > CHUNKSIZE write in `test_put_file`. Now fails with `gcs.py` in HEAD, and passes with my fix.

Tarrasch:
Great. Thanks @tym :)

PR #1064, opened 2015-07-11
Fix postgres encoding with python2

gpoulin:
Fix #1062

### DIFF
--- a/luigi/postgres.py
+++ b/luigi/postgres.py
@@ -256,12 +256,8 @@ class CopyToTable(rdbms.CopyToTable):
         """
         if value in self.null_values:
             return r'\N'
-        elif six.PY3:
-            return default_escape(str(value))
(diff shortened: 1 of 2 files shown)


gpoulin:
@Tarrasch the current master is breaking Travis?

Tarrasch:
Yea, I have no idea why that started to happening this Friday. I guess some library or something was released and the luigi package doesn't have proper upper bounds on its dependencies. I'll see if I can do something about this on Monday.

erikbern:
This is much better – thanks!

Tarrasch:
Oh, I forgot. Great thanks for you py2/3 and encoding ninja skills. I don't know what we would have done without you @gpoulin! :)

tym-oao:
Thanks @gpoulin! :+1:

PR #1066, opened 2015-07-12
feature at LocalTarget: support for directories in open('w') call

evgenyshulman:
based on https://groups.google.com/d/msg/luigi-user/oGy5Tyz7LAw/WNDzbJLIlf4J

Currently atomic_file/AtomicS3File objects works great if user want to write one file. However, if we could use .tmp_path of the return object and use it as it is, that would solve the problem. Currently that can not be done, as at the moment AtomicLocalFile is created,the file on disk is created as well. Thus we can not treat it as Directory. 
 
This PR fixes that by:  
  * is_dir flag at  LocalTarget object constractor, propagates it to atomic_file on write.
  * added is_dir parameter to AtomicFile constractor. If set, does not create IOFile, but fakes it with NonWritableFile object (raise exception on write). This way we can get object that has .tmp_path. That is the target directory.

Example
```
t = LocalTarget(path, is_dir=True):
with open(t,'w') as dp:
     write_something_to_the_dir(dp.tmp_path)
```

### DIFF
--- a/luigi/format.py
+++ b/luigi/format.py
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
+import glob
 import signal
 import subprocess
 import io
@@ -25,6 +25,7 @@ import tempfile
 import warnings
 
 from luigi import six
+from luigi.target import FileAlreadyExists, FileSystemException
 
 
 class FileWrapper(object):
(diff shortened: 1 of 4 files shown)


evgenyshulman:
Travis is failing, so I have rebased on master to be sure.

I see all other PR fail as well. Is there something wrong with my PR? 
Thanks!

Tarrasch:
Sorry no, something changed somewhere in a upstream package probably this Friday causing all builds to fail! Super sorry about this. I'll see if I can fix it on Monday. :)

luigi/target.py
```diff
@@ -23,8 +23,15 @@
...
 import os
 import random
+import shutil
 import tempfile
 import logging
+
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
```
Tarrasch: What is this uncommented import logic about???

evgenyshulman: Just copy paste from hadoop.py (the only place StringIO is used at luigi). Should I remove it?

Tarrasch: Oh I see. Hmm, I wouldn't dare touch it, kinda sad that it's not commented. Oh well.

erikbern: pretty sure this is a python 2/3 thing

evgenyshulman: if luigi is dropping python 2.6 support  we can use ` from io import StringIO` only as this module is supported both in 2.7 and 3+

evgenyshulman:
latest changes: 
* DirectoryFormat implementation:  can input multiple files in dir. can split output stream into multiple parts MAX_PART_SIZE each
* split is BSD by default.
* minor changes in OutputPipeProcessWrapper( error message, not requires next output_pipe to be stdout of the command, configurable write buffer size)

@Tarrasch  what do you think? 

The only test failing now is  `test_purge_hung_worker_override_timeout_time (worker_test.MultipleWorkersTest)`. I am not sure if my code is related to it at all. Please help.

luigi/target.py
Tarrasch: We have dropped 2.6 support. :)

luigi/format.py
```diff
@@ -499,6 +503,72 @@ def pipe_reader(self, input_pipe):
...
+
+    # on osx/other env we have different tools, we need the way to overload
+    # should be moved into general configuration or ToolsConfig class.
+    gnu_split = 'split'  # gnu split required (gsplit on OSX)
+    gnu_cat = 'cat'
+    support_bsd_split = True
+
+    def __init__(self, prefix='part-', max_part_size=None, suffix=None, writes_before_flash=None):
+        super(DirectoryFormat, self).__init__()
+        self.max_size = max_part_size
```
Tarrasch: Can you rename this to `max_part_size` too? It's very confusing when grepping otherwise.

Tarrasch: Also I don't know what it even is. Can you add docs for the `__init__` method at least?

luigi/format.py
```diff
@@ -499,6 +503,72 @@ def pipe_reader(self, input_pipe):
...
 
+
+class DirectoryFormat(Format):
+    input = 'bytes'
+    output = 'dir'
+
+    # on osx/other env we have different tools, we need the way to overload
+    # should be moved into general configuration or ToolsConfig class.
+    gnu_split = 'split'  # gnu split required (gsplit on OSX)
+    gnu_cat = 'cat'
```
Tarrasch: split and cat feels like very common tools that python has platform indepedent built in support for. No?

Tarrasch:
To be honest I've never fully understood what the `format=` thingie does. But can you somehow only use that so you don't need to reduntantly specify both `is_dir` and `format=DirectorySomethingFormat()`?

luigi/format.py
evgenyshulman: fixed, documentation is added

luigi/format.py
evgenyshulman: not exactly. there are 2 versions : gnu , extended ( with --additional-suffix) and BSD version ( really simple one)

evgenyshulman:
we find great us of format feature. We are replacing a lot of "Download/Unpack" tasks with simple .open functionality. 
Using it we are able to pipe data   cat->gzip->python. This brings us fastest implementation possible while reading the data. Same for output :  gzip->split is not simple implementation.

I have added default format value on is_dir=True && format = None

Tarrasch:
Hmm, I wanted the `is_dir` and `format=` to be either orthogonal or merged. It seems error prone that they vaguely correlate like they seem to do now.

evgenyshulman:
is_dir and format serves different purpose. we can create default format in case you have is_dir=True, and didn't specify the format, 
however, you really want  to be able to specify different formats. I don't know how to find out from the format user specified if it's going to be directory or not ( most of cases it's going to be ChainFormat). 

is_dir affects the way path is passed into pipeline
format affects the pipeline

Tarrasch:

I didn't have time to really understand this patch I feel. Have you guys used this in production?

Feel free to reopen at any time, I just close to clean up a bit.

PR #1067, opened 2015-07-13
Clarify the incomplete external task message

Tarrasch:
The idea is that most users of luigi aren't luigi developers and don't know all of the luigi terminology either. The new error message tries to describe the problem more in high level rather than explaining the issue in luigi terminology.


Tarrasch:
@petertoth @ulzha

medvind:
:+1: great!

petertoth:
I am not the @petertoth you are looking for ;)

PR #1068, opened 2015-07-13
Issue #1029/#1030 Follow up: add global luigi.cfg and add deprecation warning on old-style paths

soxofaan:
Follow up of some remarks on #1030:
- add support for /etc/luigi/luigi.cfg
- add deprecation warning about usage of old style paths


Tarrasch:
Nice :)

Tarrasch:
Sorry, not sure why we left this hanging.

PR #1069, opened 2015-07-13
Fix Mock dependency error

fabriziodemaria:
From version 1.1.1 of Mock, 'AttributeError' is reaised in case of attribute starting with 'assert' or 'assret'


Tarrasch:
Great! This is the "culprit" you pointed me to before: https://github.com/testing-cabal/mock/commit/7c530f0d9aa48d2538501761098df7a5a8979a7d#diff-332dfb235b0626ba759e2ff799faf592R706

PR #1071, opened 2015-07-13
Removes unnecessary quotes from around mapred.job.name

daveFNbuck:
In setting mapred.job.name="%s", we're actually including the quote marks in the
job name, as subprocess escapes everything passed in the arguments. Removing the
quotes results in job names that no longer have leading quote marks.


Tarrasch:
Thanks @daveFNbuck ! :)

PR #1072, opened 2015-07-14
GCS: Wait for eventual consistency before returning from various operations

mikekap:
This should help spurious test failures (and probably some prod scenarios). This obviously isn't a complete solution - just because GCS says something is
true once doesn't mean it's going to stay true, but at least we have a higher chance of succeeding various ops.

I'd love for the actual solution for the tests to include a faked client (a la moto), but it doesn't seem like that's anywhere on the roadmap :(.

### DIFF
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -17,11 +17,12 @@
 
 """luigi bindings for Google Cloud Storage"""
 
-import logging
 import io
+import logging
 import mimetypes
 import os
 import tempfile
+import time
 try:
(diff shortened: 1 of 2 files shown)


mikekap:
AFAICT the gcloud tests don't seem to be running under travis: https://travis-ci.org/spotify/luigi/jobs/70834039

Tarrasch:
Cool! Thanks! The tests will only run when run from the "spotify" GitHub organization. As otherwise Spotify would need to leak their credentials to the luigi-travis-testing project.

PR #1075, opened 2015-07-15
Add taskparameter type

Tarrasch:
Not done yet, I'll try to move over Range to use this before merging, just to see that it works in practice too. :)

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -27,6 +27,7 @@ try:
 except ImportError:
     from configparser import NoOptionError, NoSectionError
 
+from luigi import task_register
 from luigi import six
 
 from luigi import configuration
@@ -552,3 +553,26 @@ class TimeDeltaParameter(Parameter):
             return result
         else:
             raise ParameterException("Invalid time delta - could not parse %s" % input)
+
+
+class TaskParameter(Parameter):
+    """
+    A parameter that takes another luigi task class.
+
+    When used programatically, the parameter should be specified
+    directly with the :py:class:`luigi.task.Task` (sub) class. Like
+    ``MyMetaTask(my_task_param=my_tasks.MyTask)``. On the command line,
+    you specify the :py:attr:`luigi.task.Task.task_family`. Like
+
+    .. code:: console
+
+            $ luigi --module my_tasks MyMetaTask --my_task_param my_namespace.MyTask
+
+    When instantiated, it'll always instantiate to a task class.
+    """
+
+    def parse(self, input):
+        """
+        Parse a task_famly using the :class:`~luigi.task_register.Register`
+        """
+        return task_register.Register.get_task_cls(input)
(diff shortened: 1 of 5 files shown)


sisidra:
I suggest to remove Luigi custom namespace in favour of python package.module first and then task modules are not required to be imported first (can do it at runtime).
Otherwise this to work requires import everything. I don't see it as good practice.

Tarrasch:
Seems sensible, at least to make it a default. We discussed this earlier (forgot ticket number).

Anyway, this patch is not directly related to that. :)

Tarrasch:
Ok, please review folks! :)

@sisidra @ulzha @erikbern

luigi/parameter.py
```diff
@@ -552,3 +553,26 @@ def parse(self, input):
...
+    When used programatically, the parameter should be specified
+    directly with the :py:class:`luigi.task.Task` (sub) class. Like
+    ``MyMetaTask(my_task_param=my_tasks.MyTask)``. On the command line,
+    you specify the :py:attr:`luigi.task.Task.task_family`. Like
+
+    .. code:: console
+
+            $ luigi --module my_tasks MyMetaTask --my_task_param my_namespace.MyTask
+
+    When instantiated, it'll always instantiate to a task class.
```
sisidra: Add note that `MyTask`'s module *must* be imported in `my_tasks` module

sisidra:
I hold my previous comment about referencing task by python package.module.class name not Luigi's one.

luigi/parameter.py
Tarrasch: Good point! :)

Tarrasch:
> I hold my previous comment about referencing task by python package.module.class name not Luigi's one.

I agree that it makes sense in many settings. But I'm not sure we should remove the support for it, hence forcing that new convention. I'm afraid it have implications we haven't fully thought about. We could however:

 * Make it as a configurable default to set the `task_namespace` to be the same as package+module. Big organizations can then turn this on by default in config then.
 * Extend `TaskParameter` to silently "try" do the module loading. If we also change the current command line parsing (which I want to do anyway), making the `--task` param a `TaskParameter`, we would allow users to shorten cron lines from `--module my_module --task my_module.MyTask`  to only `--task my_module.MyTask`.

What do you think about that? I think we could keep the `task_namespace` concept in luigi though for those who want it.

Tarrasch:
@sisidra Does this look better now? :)

luigi/parameter.py
```diff
@@ -552,3 +553,29 @@ def parse(self, input):
...
+    When used programatically, the parameter should be specified
+    directly with the :py:class:`luigi.task.Task` (sub) class. Like
+    ``MyMetaTask(my_task_param=my_tasks.MyTask)``. On the command line,
+    you specify the :py:attr:`luigi.task.Task.task_family`. Like
+
+    .. code:: console
+
+            $ luigi --module my_tasks MyMetaTask --my_task_param my_namespace.MyTask
+
+    Where ``my_namespace.MyTask`` is defined in the ``my_tasks`` python module.
```
sisidra: It should be in registry, not only defining but also importing works :P

PR #1076, opened 2015-07-15
Linting: Nicer multiline exports in __init__.py

Tarrasch:
Just to rebase #1075 on later and make diffs look nicer :)


Tarrasch:
Does this look good @soxofaan ?

soxofaan:
yep, fine for me :)

PR #1079, opened 2015-07-17
Add MonthParameter and YearParameter

soxofaan:
Add MonthParameter and YearParameter


erikbern:
weird i thought we already had that :)

erikbern:
oh right we have the underlying `DateInterval` types but not the corresponding `Parameter` class

test/date_parameter_test.py
```diff
@@ -0,0 +1,140 @@
...
+    def test_parse_interface(self):
+        task = luigi.interface.ArgParseInterface().parse(["DateMinuteTask", "--dm", "2013-02-01T18H42"])[0]
+        self.assertEqual(task.dm, datetime.datetime(2013, 2, 1, 18, 42, 0))
+
+    def test_serialize_task(self):
+        t = DateMinuteTask(datetime.datetime(2013, 2, 1, 18, 42, 0))
+        self.assertEqual(str(t), 'DateMinuteTask(dm=2013-02-01T18H42)')
+
+
+class MonthTest(unittest.TestCase):
```
Tarrasch: Either stick with the "DateBlahTest" convention or change the other examples. :)

Tarrasch:
Thanks for the patch. Sorry for slow reply. Mind just renaming the test classes? I think this is good to merge otherwise. :)

soxofaan:
@Tarrasch : well, I chose `MonthParameter` and `YearParameter` (next to `DateParameter`, `DateHourParameter`, `DateMinuteParameter`), because something like `DateMonthParameter` seemed like needlessly redundant.
Likewise, I named the test classes `MonthTest` and `YearTest` and `DateTest` (next to existing `DateHourTest` and DateMinuteTest`)

Do you want me to change the former too (to `DateMonthParameter` and `DateYearParameter`)?

Tarrasch:
Ah, it was already mixed then. Nevermind then. LGTM. Thanks! :)

Tarrasch:
That said, it would of course be awesome if you renamed stuff to be consistent in the test file. :)

PR #1081, opened 2015-07-19
Make sure insignificant parameters are hidden in Scheduler's UI.

jkukul:
Documentation says:

```
 An insignificant Parameter might also be used to specify a password 
or other sensitive information that should not be made public via the scheduler.
```
However, currently, insignificant parameters are displayed in the tooltip when hovering over a task's node in the visualisation's graph. 

Another way to fix it could be to do the string magic and strip task parameters from the task id in the front-end part (it's already done like this elsewhere in the front-end code). However, it seems a bit wasteful, considering the `params` list is already passed via scheduler's api.

I'm not sure if there aren't any implications of the solutions I proposed, though.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -271,7 +271,7 @@ class Task(object):
         task_id_parts = []
         param_objs = dict(params)
         for param_name, param_value in param_values:
-            if dict(params)[param_name].significant:
+            if param_objs[param_name].significant:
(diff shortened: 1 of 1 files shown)


Tarrasch:
LGTM. Also nice that you fixed the surrounding code and even better that you kept the commits separate. :)

erikbern:
LGTM

PR #1082, opened 2015-07-20
connection_string should be a property

graingert:


Tarrasch:
ping @gouthambs

gouthambs:
This looks like an error that has long gone unnoticed. @graingert Thanks for catching it. How did you detect this issue? Any chance you can add some description for posterity!

graingert:
I'm thinking of integrating Luigi into a project, I'm currently reading through the code and documentation.

Tarrasch:
@graingert Did you just read through the code and noticed this or did you get some stack trace?

graingert:
Just reading the code

gouthambs:
Based on this question on SO, 
http://stackoverflow.com/questions/14441619/actual-difference-in-implementing-overriding-using-abstractproperty-and-abstra

both `abstractproperty` and `abstractmethod` adds a property to `__isabstractmethod__`. That probably explains why this did not hurt in any way. But for correctness sake, I like @graingert 's commit. If the tests pass, I think this is harmless to take.

Tarrasch:
Yea there's just an intermittent failure. Merging. Nice catch @graingert  and thanks for review @gouthambs!

PR #1083, opened 2015-07-20
Adding check that a module was defined on the command line before trying to import it

pedersen:
The following commands (from the command line) are broken:

````
luigi
luigi --help
````

Entering either of them results in an __import__ error. This patch checks to make sure that a module was specified on the command line before trying to import it. This allows expected behavior to work, and people can use the "--help" option to see the command line.

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -286,7 +286,8 @@ class DynamicArgParseInterface(ArgParseInterface):
         args, unknown = parser.parse_known_args(args=[a for a in cmdline_args if a != '--help'])
         module = args.module
 
-        __import__(module)
+        if module:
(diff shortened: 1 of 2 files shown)


erikbern:
LGTM

erikbern:
would be nice with a unit test btw

pedersen:
Local tests are passing, and the new unit tests have been added. Passing it upstream to be integrated when you're ready.

Tarrasch:
Can you just squash the 4 commits into 1?

pedersen:
commits have been squashed

Tarrasch:
Great! Awesome, thanks! :)

PR #1085, opened 2015-07-21
Allows non-atomic ftp

daveFNbuck:
Part of my pipeline requires uploading files to an external server via ftp. The
server requires specific file names and rejects and deletes the temporary files
uploaded by RemoteTarget. In order to get around this, I've added an optional
atomic parameter to RemoteTarget.put. By running it with atomic=false, the
the target file is written to directly.

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -147,7 +147,7 @@ class RemoteFileSystem(luigi.target.FileSystem):
 
         self.ftpcon.quit()
 
-    def put(self, local_path, path):
+    def put(self, local_path, path, atomic=True):
         # create parent folder if not exists
         self._connect()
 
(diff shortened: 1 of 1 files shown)


Tarrasch:
LGTM. but is there any way to test this? :)

daveFNbuck:
There are some tests in _test_ftp, but they won't actually run unless you set up an ftp server. We should probably add some more runnable tests that mock ftplib.

Tarrasch:
Well, I assume they worked for you. I'm fine with merging given that you've tested this somehow. Have you?

daveFNbuck:
Yeah, I'm currently using this code.

Tarrasch:
Cool! Thanks again @daveFNbuck

PR #1084, opened 2015-07-21
Add luigi.run() to task

harveyxia:
I needed to add these two lines to execute this task locally

PR #1087, opened 2015-07-22
Use luigi binary in docs

Tarrasch:
As discussed in #1084


Tarrasch:
@harveyxia. Does this look better? :)

harveyxia:
+1

PR #1089, opened 2015-07-22
Close pool to prevent stale processes

graingert:
Processes end up hanging around until the Pool is garbage collected. This could be a while, or never if the GC is disabled.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -172,6 +172,12 @@ class SingleProcessPool(object):
     def apply_async(self, function, args):
         return function(*args)
 
+    def close(self):
+        pass
+
+    def join(self):
+        pass
+
 
 class DequeQueue(collections.deque):
     """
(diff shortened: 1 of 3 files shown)


Tarrasch:
Very nice catch! Thanks for finding this serious resource leak! @daveFNbuck, you did stuff in code around here. Any thoughts?

But can you please just fix the pep8 issue? :)

https://travis-ci.org/spotify/luigi/jobs/72179182

By the way, how did you discover this issue? Did something crash/break for you?

graingert:
adding unittest

Tarrasch:
Just to be clear, you plan to add some test cases and then reopen the ticket?

graingert:
@Tarrasch yes :D. Well only one test case here.

daveFNbuck:
What an embarrassing oversight! The fix looks good to me.

graingert:
looks like my failing test isn't on on travis. I can assure you it is red/green on my local machine.

I'll try pausing the GC either side of the call to luigi

graingert:
@Tarrasch red and green builds complete!

Tarrasch:
For future reference (since GitHub ui tends to hide these links afterwards), here is what failed before: https://travis-ci.org/spotify/luigi/builds/72259810

Tarrasch:
Extremely awesome @graingert! Thanks! :)

erikbern:
nice!

PR #1090, opened 2015-07-23
'services two purposes' -> 'serves two purposes'

JackDanger:
'Services' is highly unusual as a verb in places where 'serves' would work. Also, this is my first PR to the project! :)


Tarrasch:
Thanks. Every great journey starts with a single step. :)

PR #1094, opened 2015-07-27
PR #1079 follow up: improve test class names

soxofaan:
follow up of #1079

@Tarrasch: you mean like this?


Tarrasch:
Yea I think it's better this way. :)

PR #1096, opened 2015-07-27
Corrected PR #996 according to comments in #995.

kianho:
- Changed S3CopyToTable.s3_load_path back to a method according to the
  suggestion: https://github.com/spotify/luigi/issues/995#issuecomment-123896142

- Modified the accompanying test case accordingly.

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -163,7 +163,7 @@ class S3CopyToTable(rdbms.CopyToTable):
         if not (self.table):
             raise Exception("table need to be specified")
 
-        path = self.s3_load_path
+        path = self.s3_load_path()
         connection = self.output().connect()
         if not self.does_table_exist(connection):
             # try creating table
--- a/test/contrib/redshift_test.py
+++ b/test/contrib/redshift_test.py
@@ -43,8 +43,10 @@ class DummyS3CopyToTable(luigi.contrib.redshift.S3CopyToTable):
 
     aws_access_key_id = AWS_ACCESS_KEY
     aws_secret_access_key = AWS_SECRET_KEY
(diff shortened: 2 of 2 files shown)


Tarrasch:
Please review @carlineng

carlineng:
Looks good, thank you @kianho

Tarrasch:
And there's no issues with breaking backward compatibility here guys? Or it just unbreaks what it previously broke?

carlineng:
Correct, this simply reverts the change made in spotify/luigi#996, which broke the redshift S3CopyToTable task.

kianho:
@Tarrasch, correct, it unbreaks #996

joeshaw:
`s3_load_path` is still decorated as `@abc.abstractproperty` -- should it be changed to `@abc.abstractmethod`?  I was using it as a property so this commit broke some of my code (was previously using 1.3.0 and upgraded to master.)

spiside:
@joeshaw I noticed that too re: `@abc.abstractproperty`. I'm thinking of making a couple more changes to the redshift module and will be addressing that point.

dlstadther:
It appears that @kianho 's PR intended to change `s3_load_path` back to a method (as noted by the used of `self.s3_load_path()`). However, [L72](https://github.com/spotify/luigi/commit/9136cd0424a62154799badd21a8b75a64d9689af#diff-778ea3db4cccaf4de6564889c5eb670fL72) shows that the decorator change never occurred.

No one (including myself) had picked up on the inconsistent use of an `@abc.abstractproperty` being called as a method (`self.s3_load_path()`).

I will update this in my existing pending PR #1393

kianho:
good spot @dlstadther, apologies for the oversight.

PR #1098, opened 2015-07-28
Make DateHourParameter subclass of DateParameter

Tarrasch:
In addition to saving a few lines of code. It also prepares for even
more code reuse.

Also removed ancient TODO from 2012:
https://github.com/spotify/luigi/commit/f65d4cac9d4d867a696df6d3f7f1a95c709de276

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -20,6 +20,7 @@ All Parameters sit on :class:`~luigi.task.Task` classes.
 See :ref:`Parameter` for more info on how to define parameters.
 '''
 
+import abc
 import datetime
 import warnings
 try:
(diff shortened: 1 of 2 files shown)


Tarrasch:
partially needed for #1091 like endeavours.

sisidra:
DateParameter create datetime.date instance. Date(Hour/Minute)Parameter creates datetime.datetime instance.
Semantically those should not be subclasses imo.
If you invent DatetimeParameterBase for all of those 3, it would look much better.

erikbern:
Yeah +1 for base class instead.

It's a classical OOP problem. Is a more-specified class a subclass of a less-specified class. Or the other way around? :)

Tarrasch:
Haha, damn smart people like you never let me take shortcuts.

I'll resubmit the proper PR. :)

erikbern:
Hehe ok just to explain what I mean: one classical example is: if you have two classes `Square` and `Rectangle` – which one inherits from which? `Square` is a special instance of `Rectangle` but you could also argue that `Rectangle` extends `Square`. So both ways kinda makes sense

erikbern:
https://en.wikipedia.org/wiki/Circle-ellipse_problem :)

Tarrasch:
Can you re review @erikbern @sisidra ?

erikbern:
LGTM

PR #1101, opened 2015-07-29
Fix DateMinuteParameter format

fphilipe:
Addresses points discussed in https://github.com/spotify/luigi/pull/625#issuecomment-125922059.

It was using H as the separator between hours and minutes, which is incorrect. The ISO 8601 standard requires a colon in the extended format or nothing in the basic format. Opt for basic format since colons don't play nice with file systems.

Added some specs for the parameter since there didn't seem to be any. Went for deprecation although I'm not entirely sure whether it is needed.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -446,11 +446,24 @@ class DateMinuteParameter(DateHourParameter):
     Parameter whose value is a :py:class:`~datetime.datetime` specified to the minute.
 
     A DateMinuteParameter is a `ISO 8601 <http://en.wikipedia.org/wiki/ISO_8601>`_ formatted
-    date and time specified to the minute. For example, ``2013-07-10T19H07`` specifies July 10, 2013 at
+    date and time specified to the minute. For example, ``2013-07-10T1907`` specifies July 10, 2013 at
     19:07.
     """
 
-    date_format = '%Y-%m-%dT%HH%M'  # ISO 8601 is to use 'T' and 'H'
+    date_format = '%Y-%m-%dT%H%M'
+    deprecated_date_format = '%Y-%m-%dT%HH%M'
+
+    def parse(self, s):
+        try:
+            value = datetime.datetime.strptime(s, self.deprecated_date_format)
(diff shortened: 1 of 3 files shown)


luigi/parameter.py
```diff
@@ -370,11 +370,12 @@ class DateHourParameter(Parameter):
     19:00.
     """
 
-    date_format = '%Y-%m-%dT%H'  # ISO 8601 is to use 'T'
+    date_format = '%Y-%m-%dT%H'
```
erikbern: why did you change this line?

luigi/parameter.py
```diff
@@ -394,11 +396,23 @@ class DateMinuteParameter(DateHourParameter):
...
+    def parse(self, s):
+        try:
+            return super(DateMinuteParameter, self).parse(s)
+        except ValueError:
+            warnings.warn(
+                'Using "H" between hours and minutes is deprecated, omit it instead.',
+                DeprecationWarning,
+                stacklevel=2
+            )
+            return datetime.datetime.strptime(s, self.deprecated_date_format)
```
erikbern: this will throw a deprecation warning for any misspecified string – maybe makes more sense to try deprecated first

erikbern:
LGTM

luigi/parameter.py
fphilipe: You're right, I'll fix it.

luigi/parameter.py
fphilipe: I didn't see the point of the comment and since I removed the one below, I thought I'd remove it here as well. ISO 8601 is being mentioned above anyways, so it doesn't add much IMO. Also, I think it's rather arbitrary to document `T` and not `YYYY-MM-DD`, for instance.

I can undo the change on this line if you prefer :blush:

Tarrasch:
Hmm, can you rebase this change. Unfortunately I made a patch that changes same places of code in #1098

fphilipe:
@Tarrasch Rebased.

PR #1104, opened 2015-07-30
Only require significant parameters to be supplied in the params_str …

ntim:
…when trying to create a task instance from the str->str hash.

Otherwise, the method results in a KeyError since insignificant parameters are not incorporated in the str->str hash created by the to_str_params method.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -294,8 +294,9 @@ class Task(object):
             params_str = {}
         kwargs = {}
         for param_name, param in cls.get_params():
-            value = param.parse_from_input(param_name, params_str[param_name])
-            kwargs[param_name] = value
+            if param.significant or param_name in params_str:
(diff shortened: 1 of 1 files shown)


erikbern:
Would be good to be able to set insignificant parameters though. Can we fix the underlying issue instead? I might be misunderstanding

ntim:
If the parameter is supplied in the params_str it will be set there. Maybe it would be nice to consider insignificant parameters in "to_str_params"

Tarrasch:
Sorry for our intermittent build falure. Since it's fixed, do you mind rebasing it? :)

jkukul:
@ntim I think you're right about considering insignificant parameters in `to_str_params` (or as @erikbern put it - fixing the underlying issue). I created #1109 that fixes the issue and it comes with a test case that would have avoided it in the first place.

Tarrasch:
Closed in favor of #1109

PR #1103, opened 2015-07-30
Fix dependencies mess

Tarrasch:
See individual commits.

PR #1106, opened 2015-07-31
Set upper bounds on packages, assuming semver

Tarrasch:
Every now and then some package somewhere in internet world changes,
causing our travis builds to fail. The only sensible thing to do is to
not pull the latest. We shouldn't bu unneccesarily strict either. So
assuming others use semver, I copied the current version of everything
that we currently use (based on a contemporary build [1]) and just
bumped the major number and set it as a strictly less-than constraint.

I got motivated to do this since #1103, but this is not the first
occurence.

[1]: https://travis-ci.org/spotify/luigi/jobs/73593492

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ cache:
   - $HOME/.pip-cache
 
 install:
-  - pip install tox
+  - pip install 'tox<3.0'
 
 before_install:
   - openssl aes-256-cbc -K $encrypted_e05f6ccc270e_key -iv $encrypted_e05f6ccc270e_iv -in test/gcloud-credentials.json.enc -out test/gcloud-credentials.json -d
--- a/setup.py
+++ b/setup.py
@@ -40,10 +40,10 @@ with open('README.rst') as fobj:
     long_description = readme_note + fobj.read()
 
 install_requires = [
-    'cached_property',
-    'pyparsing',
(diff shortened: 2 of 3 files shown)


Tarrasch:
The "build failures" are annoying intermittent failures. So this should be green.

Tarrasch:
Want to review @erikbern?

Tarrasch:
I merge for now, but please audit this. I just merge to avoid future merge conflicts.

erikbern:
Not convinced this is the best thing to do. I don't want Luigi to depend on a bunch of outdated packages that aren't maintained. Better to try to fix issues as they arise

Tarrasch:
I didn't find it acceptable that the luigi package "suddenly" starts to break. This will also mean that **our released package will also break**. I think it was great that we took luigi up a notch by starting to release versioned "known-to-work" packages of luigi. But those package will suddenly all be useless if somebody releases a package somewhere on the internet.

So yes, this is not only to make Travis builds more stable, it's basic versioning hygiene.

Are you convinced now? :)

sisidra:
I'm with @Tarrasch on this.
We can not afford anymore situation when something suddenly breaks on prod.
If you want to be on top of dependent packages - we need to do semantic versioning and change dependencies WITH fixes according to that.

PR #1108, opened 2015-07-31
contrib/hive: Escape names in Hive partition spec

bradenmacdonald:
In [newish versions of Hive](http://mail-archives.apache.org/mod_mbox/hive-commits/201503.mbox/%3C20150310220045.68294AC02E7@hades.apache.org%3E), if people use a word like `date`, `user`, or `interval` as a partition key, an error will occur:
```
Failed to recognize predicate 'DATE'. Failed rule: 'identifier' in ...
```

This change escapes partition key names when generating Hive DML partition specifications, allowing use of any word as a partition key.

This fix should be fully compatible with any version of Hive.

### DIFF
--- a/luigi/contrib/hive.py
+++ b/luigi/contrib/hive.py
@@ -158,7 +158,7 @@ class HiveCommandClient(HiveClient):
         """
         Turns a dict into the a Hive partition specification string.
         """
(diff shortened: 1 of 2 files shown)


bradenmacdonald:
Test failures appear unrelated, but let me know if I misinterpreted that.

Tarrasch:
Thanks!! :)

PR #1109, opened 2015-07-31
Fix task_to_str_to_task behaviour when a Task has insignificant parameter

jkukul:
Fixes issue introduced by a923a75 and adds a test case to avoid it in the future.

In a923a75, I removed insignificant parameters to be included in the output of `to_str_params`. It seems this was a bit careless, i.e. it causes an error when the serialised task is passed back to `from_str_params`.

The goal of a923a75 was to avoid passing insignificant params to scheduler and thus potentially avoid exposing it to users via UI. I guess I can come with another, less invasive way to achieve that. (possibly in another PR)

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -306,8 +306,7 @@ class Task(object):
         params_str = {}
         params = dict(self.get_params())
         for param_name, param_value in six.iteritems(self.param_kwargs):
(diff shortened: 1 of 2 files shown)


Tarrasch:
Thanks for reverting this! Hoping to see the underlying fixed too. :)

PR #1119, opened 2015-08-10
Allows assistant to run tasks with insignificant parameters

daveFNbuck:
If a task has an insignificant parameter, it should be ok for an assistant to
run it with the default value of that parameter. This change allows the
assistant to do that by ignoring when insignificant parameters are not present
in Task.from_str_params.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -294,8 +294,9 @@ class Task(object):
             params_str = {}
         kwargs = {}
         for param_name, param in cls.get_params():
-            value = param.parse_from_input(param_name, params_str[param_name])
-            kwargs[param_name] = value
+            if param.significant or param_name in params_str:
+                value = param.parse_from_input(param_name, params_str[param_name])
+                kwargs[param_name] = value
 
         return cls(**kwargs)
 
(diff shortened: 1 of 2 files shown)


erikbern:
Sounds good to me but what's the benefit? Just trying to understand what problem it solves

daveFNbuck:
Most of my jobs have an insignificant parameter with a default value. I started trying to run all my jobs on assistants last week and this was a major blocker. I'll put out some more PRs for this once I'm able to get things running more smoothly, but this one increased in importance after the same issue started causing some other bugs in production today.

erikbern:
Ok but what was the problem before? Or maybe it doesn't matter, I trust that this works

daveFNbuck:
Ah, sorry I didn't understand your question earlier. The problem before was that `value = param.parse_from_input(param_name, params_str[param_name])` called `params_str[param_name]` and params_str doesn't contain param_name for insignificant parameters because the scheduler doesn't know about them.

erikbern:
Ok, makes sense! Thanks

erikbern:
In general how is your experience with assistants now? I remember there was still a couple of bugs in the Worker class that I haven't had time to fix

daveFNbuck:
I have a branch in which assistants work mostly well but it somehow breaks how the scheduler keeps track of which workers are still running. I'm going to try to figure that out this week before making pull requests for the rest of my new changes.

erikbern:
Awesome

PR #1121, opened 2015-08-11
Fixed issue where string is appended to bytes.

pkbauer:


Tarrasch:
I would expect that you add:

 * An explanation when this happens
 * A test case or a motivation why this is hard to test.

Thanks. :)

pkbauer:
When this happens:
It happens when I try to write to hdfs.

    def output(self):
        import datetime
        today = datetime.datetime.now()
        output_file_name = '{}/{}{}'.format(self.output_path,
                                            today.strftime("%y-%m-%d-%H-%M"), ".out")
        if bool(self.local_mode):
            return S3Target(output_file_name)
        else:
            return luigi.contrib.hdfs.HdfsTarget(output_file_name)

Tarrasch:
I really don't get it from your example, there's no writing going on there. Can you just create a failing test case to show that this is not just an arbitrary change?

toidi:
I guess it will be better to integrate
```
from __future__ import unicode_literals
```
to whole luigi sources or at least to PR.

Tarrasch:

PR #1122, opened 2015-08-12
Fixing line wrapping in CONTRIBUTING.rst

JackDanger:
As a new contributor (who doesn't know about Tox) it was hard to figure out what I needed to `pip install` in order to get things going. Rather than making every developer look inside `.travis.yml` for the install command this PR uses the standard requirements.txt file for pip.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,9 +1,12 @@
 Running the tests
 -----------------
 
-We are always happy to recieve Pull Requests. When you open a PR, it will automatically build on Travis. So you're not strictly required to test the patch locally before submitting it.
+We are always happy to recieve Pull Requests. When you open a PR, it will
+automatically build on Travis. So you're not strictly required to test the
+patch locally before submitting it.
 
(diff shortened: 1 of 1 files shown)


Tarrasch:
I really like the line wrapping commit. And I like the splitting of the commits even more. :)

But I think a `requirements.txt` in the project root would be confusing. That's not the requirements for luigi, it's the requirements for running the tests only. Can you instead somehow make the "Running Unit Tests" refer to CONTRIBUTING.rst? Or somehow merge them? Does that make sense? :)

JackDanger:
@Tarrasch I've removed the second commit (the one that added the requirements.txt) and renamed this to be about fixing the line wrapping. I'll make a second PR that tries to address the root problem here. Thank you!

Tarrasch:
Awesomeness!

Tarrasch:
Also, when I say "merge", I would like to at least keep the file CONTRIBUTING.rst, as github integrates with it when contributors sends PRs.

JackDanger:
@Tarrasch I tried to address this in #1126. Let me know what you think!

PR #1125, opened 2015-08-12
Use the session.transaction context manager for nicer exceptions

graingert:


luigi/db_task_history.py
```diff
@@ -65,13 +65,8 @@ def _session(self, session=None):
             yield session
         else:
             session = self.session_factory()
-            try:
+            with session.transaction:
```
graingert: SQLAlchemy's session management does magic to make sure that if the rollback fails you still get the original exception that caused the rollback.

Also it looks nicer.

Tarrasch:
LGTM

Tarrasch:
I'm not sure of the test coverage of the task history as I don't use it. Have you either tested this in production or do we have unit tests covering your change @graingert? If so I can merge.

Tarrasch:
Closing as it's been stale for a while. Please reopen when you either have a test or at least verified it's running in production.

PR #1126, opened 2015-08-12
Centralizing test prep and execution in ./ci

JackDanger:
New developers now have a single command to run to both install the
necessary dependencies (i.e. tox) and execute the test suite.

This reuses code when possible so there's only ever one path through which the tests run.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -5,13 +5,17 @@ We are always happy to recieve Pull Requests. When you open a PR, it will
 automatically build on Travis. So you're not strictly required to test the
 patch locally before submitting it.
 
-Despite of this, it's good to test it locally first to speed up the iteration
-cycle. Here are four example commands:
+If you do want to run the tests locally you'll need to ``pip install tox`` and
+then run one of the tox commands below.
 
 .. code:: bash
 
+    # The first two are fast and will tell if you if you've broken
+    # something major:
     tox -e pep8
     tox -e py27-nonhdfs
+
+    # And some of the others involve downloading and running Hadoop:
     tox -e py33-cdh
     tox -e py34-hdp
 


gpoulin:
I'm not sure it's necessary. It's already one line `tox -e {ENV}`? I don't think it's a big entry barrier to ask contributor to be able to install tox. The doc it's pretty clear hoe to use tox. In addition, not everybody want necessarily want to install tox in pip user space. Maybe making the contributing section of the doc more visible.

JackDanger:
>  I don't think it's a big entry barrier to ask contributor to be able to install tox.

You may be right but I'm newish to Python so I'm not familiar with tox and I was confused there wasn't just one script I could run and check the exit status of. I was hunting for something like `./run_tests && git commit -m "works!"` but ended up having to read `.travis.yml` to figure it out.

I suspect anybody who knows about Tox and has a preference for where pip installs things won't need this script. Do you think it's bad to keep it for those of us who do?

JackDanger:
To rephrase: The reason I put together this PR is because it was quite difficult as a new contributor to figure out what one command I should run to ensure I haven't introduced a regression. If there's a better way than what I've proposed please teach me!

gpoulin:
If you think it's really useful, I would include a file named `test.sh` (`ci` is not really that obvious) and would just include something like this:

```bash
pip install 'tox<3.0'
tox -e py27-nonhdfs
#tox -e py34-nonhdfs
#tox -e py33-nonhdfs
...
```

I wouldn't modify `.travis.yml` just for this.

JackDanger:
Thanks, @gpoulin. I've done as you suggested and created an executable `./test.sh` that can be used by new developers like myself trying to get their feet wet. Take a look?

test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Run this file to quickly execute a subset of tests locally. This is not the
+# full test suite, look at .travis.yml for the complete matrix of dependencies
+# and combinations.
+pip install pep8
```
Tarrasch: I don't think you need this. Or at least you shouldn't right?

Tarrasch:
This looks better. :)

test.sh
JackDanger: Removed. Thanks!

Tarrasch:
@gpoulin, @erikbern, what do you guys think?

gpoulin:
I'm still not sure it's needed but I don't see harm in it. So I'll say merge it.

erikbern:
Would rather just put it in the documentation. But fine merging it too

Tarrasch:
Well, on second thought I must say I don't like this patch. It adds yet another file in the root folder. I would rather see documentation over this.

JackDanger:
Sounds fair. I've changed it to now only be a change to `CONTRIBUTING.rst`. Take a look?

erikbern:
LGTM let's merge

Tarrasch:
Thanks! I hope the many iterations didn't scare you away or anything @JackDanger :)

JackDanger:
> Thanks! I hope the many iterations didn't scare you away or anything @JackDanger :)

As long as there's a resolution eventually and the contributors are all giving constructive opinions then it feels worth it :)
Thanks for helping me make some initial contributions!

Tarrasch:
Actually @JackDanger, have you though of having something like `tox -e basic-tests` or maybe even make `tox` (passed without arguments) to do something like `tox -e pep8 && tox -e py27-nonhdfs`. Or maybe even  `tox -e pep8 && tox -e py27-nonhdfs test/paths/to/only/fast/tests`. Let us know if you come up with something interesting.

I just caught myself inefficiently doing pep8 followeb by py27-nonhdfs. :)

PR #1127, opened 2015-08-12
Atomic bigquery uploads

DeaconDesperado:
Stages uploads of data for a bigquery target in a temporary table which can be moved into place for the sake of atomicity.

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -19,6 +19,7 @@ import collections
 import logging
 import luigi.target
 import time
+import random
 
 logger = logging.getLogger('luigi-interface')
 
@@ -62,6 +63,11 @@ class BQTable(collections.namedtuple('BQTable', 'project_id dataset_id table_id'
     @property
     def dataset(self):
         return BQDataset(project_id=self.project_id, dataset_id=self.dataset_id)
+
+    @property
+    def uri(self):
+        return "bq://" + self.project_id + "/" + \
+               self.dataset.dataset_id + "/" + self.table_id
 
 
 class BigqueryClient(object):
@@ -276,8 +282,10 @@ class BigqueryClient(object):
 
 
 class BigqueryTarget(luigi.target.Target):
-    def __init__(self, project_id, dataset_id, table_id, client=None):
+    def __init__(self, project_id, dataset_id, table_id, client=None, tmp_dataset_id=None):
         self.table = BQTable(project_id=project_id, dataset_id=dataset_id, table_id=table_id)
+        tmp_table_id = "_" + table_id + "_%09" % random.randrange(0, 1e10)
+        self.tmp_table = BQTable(project=project_id, dataset=tmp_dataset_id or "_incoming", table_id=tmp_table_id)
         self.client = client or BigqueryClient()
 
     @classmethod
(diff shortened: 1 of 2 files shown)


luigi/contrib/bigquery.py
```diff
@@ -63,6 +64,11 @@ class BQTable(collections.namedtuple('BQTable', 'project_id dataset_id table_id'
     def dataset(self):
         return BQDataset(project_id=self.project_id, dataset_id=self.dataset_id)
 
+    @property
+    def uri(self):
```
Tarrasch: Is this used anywhere?

Tarrasch:
Nice, but can you squash the commits?

luigi/contrib/bigquery.py
DeaconDesperado: Not used in the load task, but could be a useful method for task implementors.

DeaconDesperado:
And so it is done!

luigi/contrib/bigquery.py
Tarrasch: I would strongly suggest that you add a test for this functionality then. Otherwise it might suddenly be carelessly "optimized away" in some re-factoring.

erikbern:
LGTM

luigi/contrib/bigquery.py
```diff
@@ -355,7 +363,14 @@ def run(self):
         if self.schema:
             job['configuration']['load']['schema'] = {'fields': self.schema}
 
-        bq_client.run_job(output.table.project_id, job, dataset=output.table.dataset)
+        logger.info('Loading data to temporary table %s.%s', output.tmp_table.dataset_id, output.tmp_table.table_id)
+        bq_client.run_job(output.tmp_table.project_id, job, dataset=output.tmp_table.dataset)
+        logger.info('Moving temporary table %s.%s to destination table %s.%s',
+                    output.tmp_table.dataset_id, output.tmp_table.table_id,
+                    output.table.dataset_id, output.table.table_id)
+        bq_client.copy(output.tmp_table, output.table)
```
Tarrasch: This is a reasonably fast operation, right?

DeaconDesperado: Yes, I believe it's just a reassignment of ids on googles side.  IIRC there is no actual copy of data involved, and [table operations are all atomic](http://stackoverflow.com/a/12514790/215608)

DeaconDesperado:
Added the requested test.

Also got gcloud tox environment working locally using the travis project, so it should work in build.

mikekap:
I think BQ is atomic by default:
> Each action is atomic and only occurs if BigQuery is able to complete the job successfully. Creation, truncation and append actions occur as one atomic update upon job completion.	
> https://cloud.google.com/bigquery/docs/reference/v2/jobs#configuration.load.createDisposition

This unfortunately breaks writeDisposition == WRITE_APPEND (granted that's a pretty weird use-case and requires overriding complete()) :(

DeaconDesperado:
@mikekap D'oh, it seems you are correct.  This is the result of some confusion between behaviors of the [BigQuery hadoop connector](https://github.com/GoogleCloudPlatform/bigdata-interop/tree/master/bigquery) and the http based bigquery client utilized in luigi.  The intention was to make Hadoop mapreduce jobs that load via the connector atomic using the swap-into-place pattern common throughout luigi, but it seems like when using the bigquery client from the `googleapis` package this is handled by default.  In the case of the connector, intermittent failures between mappers can cause partial uploads to the table, so swap in place via luigi after YARN marks the entire job as successful is the best way to ensure atomicity.

@Tarrasch I think the right way to fix this would be to drop the behavior from the canonical `BigqueryLoadTask` since it is redundant there, but leave the temporary table id generation for task implementors to do swap-in-place if needed.  This would support both use cases.  The target itself could also have a convenience method for doing the swap.

I'll work on another PR and we can optionally revert this.

Tarrasch:
Cool @DeaconDesperado. You seem to be on top of it. :)

Yea, either revert this or clean up the unneeded changes in a upcoming PR (preferred, I think)

Issue #1118, opened 2015-08-10
Support unix sockets for faster internal connections


graingert:
We plan to have one machine with luigid running, with the web-gui protected by Nginx auth.

It would be great if we could use a unix socket for improved internal networking performance both for the nginx proxy and the luigi workers.

erikbern:
Good idea – feel free to put together a pull request!

gpoulin:
@graingert Did you have a benchmark that show that the tcp connection is a bottleneck in your case? I have the impression it won't make that much a difference. Maybe for the worker using a connection pool (are we doing it at the moment?), would be easier the implement and would have benefice for more use cases.

graingert:
we also don't need to make a firewall rule

tashrifbillah:
Hi @graingert , can you give an example about how to write a unix-socket for luigid?

graingert:
@tashrifbillah sorry, I don't use this library anymore

tashrifbillah:
For future Googlers, here is my show and tell on this topic:

https://groups.google.com/g/luigi-user/c/PTLkefG8oNI

PR #1129, opened 2015-08-13
Add support for Unix sockets #1118

graingert:
This also adds support for TLS!

Simply use https://exmaple.com/ for TLS or [http+unix://%2Fvar%2Frun%2Fluigid%2Fluigid.sock](http:/%2Fvar%2Frun%2Fluigid%2Fluigid.sock) for Unix sockets

There doesn't seem to be any actual tests to see if the real fetch method works, as all the tests that cover RemoteScheduler patches "._fetch"

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -25,12 +25,27 @@ import logging
 import socket
 import time
 
-from luigi.six.moves.urllib.parse import urlencode
+from luigi.six.moves.urllib.parse import urlencode, ParseResult
 from luigi.six.moves.urllib.request import urlopen
 from luigi.six.moves.urllib.error import URLError
 
 from luigi import configuration
 from luigi.scheduler import PENDING, Scheduler
+
+
+HAS_UNIX_SOCKETS = True
+HAS_REQUESTS = True
+
+
+try:
+    import requests_unixsockets as requests
+except ImportError:
+    HAS_UNIX_SOCKETS = False
+    try:
+        import requests
+    except ImportError:
+        HAS_REQUESTS = False
+
 
 logger = logging.getLogger('luigi-interface')  # TODO: 'interface'?
 
@@ -42,32 +57,62 @@ class RPCError(Exception):
         self.sub_exception = sub_exception
 
 
+class FetcherException(Exception):
+    def __init__(self, original_exc):
+        self.original_exc = original_exc
+
+
+class URLLibFetcher(object):
+    def fetch(self, full_url, body, timeout):
+        try:
+            body = urlencode(body).encode('utf-8')
+            return urlopen(full_url, body, timeout).read().decode('utf-8')
+        except (URLError, socket.timeout) as e:
+            raise FetcherException(e)
+
+
+class RequestsFetcher(object):
+    def __init__(self, session):
+        self.session = session
+
+    def fetch(self, full_url, body, timeout):
+        from requests import exceptions as requests_exceptions
+        try:
+            resp = self.session.get(full_url, data=body, timeout=timeout)
+            resp.raise_for_status()
+            return resp.text
+        except (requests_exceptions.RequestException) as e:
+            raise FetcherException(e)
+
+
 class RemoteScheduler(Scheduler):
     """
     Scheduler proxy object. Talks to a RemoteSchedulerResponder.
     """
 
-    def __init__(self, host='localhost', port=8082, connect_timeout=None, url_prefix=''):
-        self._host = host
-        self._port = port
-        self._url_prefix = url_prefix
+    def __init__(self, url='http://localhost:8082/', connect_timeout=None):
+        assert (
+            not (url.startswith('http+unix://') and not HAS_UNIX_SOCKETS),
+            'You need to install requests-unixsocket for Unix socket support.',
+        )
 
+        self._url = url
         config = configuration.get_config()
 
         if connect_timeout is None:
             connect_timeout = config.getfloat('core', 'rpc-connect-timeout', 10.0)
         self._connect_timeout = connect_timeout
 
+        if HAS_REQUESTS:
+            self._fetcher = RequestsFetcher(requests.Session())
+        else:
+            self._fetcher = URLLibFetcher()
+
     def _wait(self):
         time.sleep(30)
 
-    def _fetch(self, url, body, log_exceptions=True, attempts=3):
-
-        full_url = 'http://{host}:{port:d}{prefix}{url}'.format(
-            host=self._host,
-            port=self._port,
-            prefix=self._url_prefix,
-            url=url)
+    def _fetch(self, url_suffix, body, log_exceptions=True, attempts=3):
+        full_url = self._url + url_suffix
         last_exception = None
         attempt = 0
         while attempt < attempts:
@@ -76,24 +121,23 @@ class RemoteScheduler(Scheduler):
                 logger.info("Retrying...")
                 self._wait()  # wait for a bit and retry
             try:
-                response = urlopen(full_url, body, self._connect_timeout)
+                response = self._fetcher.fetch(full_url, body, self._connect_timeout)
                 break
-            except (URLError, socket.timeout) as e:
-                last_exception = e
+            except FetcherException as e:
+                last_exception = e.original_exc
                 if log_exceptions:
-                    logger.exception("Failed connecting to remote scheduler %r", self._host)
+                    logger.exception("Failed connecting to remote scheduler %r", self._url)
                 continue
         else:
             raise RPCError(
                 "Errors (%d attempts) when connecting to remote scheduler %r" %
-                (attempts, self._host),
+                (attempts, self._url),
                 last_exception
             )
-        return response.read().decode('utf-8')
+        return response
 
     def _request(self, url, data, log_exceptions=True, attempts=3):
-        data = {'data': json.dumps(data)}
-        body = urlencode(data).encode('utf-8')
+        body = {'data': json.dumps(data)}
 
         page = self._fetch(url, body, log_exceptions, attempts)
         result = json.loads(page)
--- a/test/server_test.py
+++ b/test/server_test.py
@@ -17,6 +17,7 @@
 import os
 import multiprocessing
 import random
+import shutil
 import signal
 import time
 import tempfile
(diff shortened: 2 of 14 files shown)


Tarrasch:
Hmm, can you elaborate more on what this is and who would use it and why? Is this only related to the luigi server?

graingert:
@Tarrasch the discussion is in #1118

erikbern:
lgtm

test/worker_test.py
```diff
@@ -649,7 +649,7 @@ def tearDown(self):
 
     @email_patch
     def test_connection_error(self, emails):
-        sch = RemoteScheduler(host="this_host_doesnt_exist", port=1337, connect_timeout=1)
```
graingert: this_host_doesnt_exist is a perfectly valid option for a gTLD

luigi/rpc.py
```diff
@@ -42,32 +62,61 @@ def __init__(self, message, sub_exception=None):
...
+class FetcherException(Exception):
+    def __init__(self, original_exc):
+        self.original_exc = original_exc
+
+
+class URLLibFetcher(object):
+    def fetch(self, full_url, body, timeout):
+        try:
+            body = urlencode(body).encode('utf-8')
+            return urlopen(full_url, body, self._connect_timeout).read().decode('utf-8')
```
Tarrasch: Eh this is scary. There's no such thing as a `self._connect_timeout` in this context right? Is this code path really tested?

I know I've had to disable some tests on Travis. Can you make sure a normal `tox -e py27-nonhdfs` passes before we merge this?

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,33,34}-{cdh,hdp,nonhdfs,gcloud,postgres}, pypy-scheduler, docs, pep8
+envlist = py{27,33,34}-{cdh,hdp,nonhdfs,gcloud,postgres}{-requests,-unix,}, pypy-scheduler, docs, pep8
```
Tarrasch: I don't see any corresponding `attr('requests')`, nor did you add it to the `.travis.yml` right? Also i don't see why you need multiplicative expansion here. Why not just `py27-unix`? Also something more explicit over just `unix`. Like `py27-unixsockets` or something.

luigi/rpc.py
graingert: Ah woops, looks like something depends on requests in your test dependencies. Will fix

graingert:
@Tarrasch thanks for the review, I've made those changes.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,33,34}-{cdh,hdp,nonhdfs,gcloud,postgres}, pypy-scheduler, docs, pep8
+envlist = py{27,33,34}-{cdh,hdp,nonhdfs,gcloud,postgres}{-unixsockets,}, pypy-scheduler, docs, pep8
```
Tarrasch: Again, why multiplicative?

luigi/rpc.py
```diff
@@ -42,32 +62,61 @@ def __init__(self, message, sub_exception=None):
...
+class FetcherException(Exception):
+    def __init__(self, original_exc):
+        self.original_exc = original_exc
+
+
+class URLLibFetcher(object):
+    def fetch(self, full_url, body, timeout):
+        try:
+            body = urlencode(body).encode('utf-8')
+            return urlopen(full_url, body, timeout).read().decode('utf-8')
```
Tarrasch: Awesome that you fixed this! But you checked somehow that this code path is tested before fixing it right? :)

luigi/rpc.py
```diff
@@ -25,13 +25,33 @@
...
+    import requests_unixsockets as requests
+except ImportError:
+    HAS_UNIX_SOCKETS = False
+    try:
+        import requests
+    except ImportError:
+        HAS_REQUESTS = False
+
+try:
+    from requests import exceptions as requests_exceptions
```
Tarrasch: Hmm, if you don't have it installed, won't you just get an exception like "no variable defined `requests_exceptions
Markdown supported `"? Can't you just do `from requests import exceptions as requests_exceptions` without try-catch inside the `fetch()` method?

test/server_test.py
```diff
@@ -25,8 +26,15 @@
 import luigi.rpc
 import luigi.server
 from luigi.scheduler import CentralPlannerScheduler
+from luigi.six.moves.urllib.parse import urlencode, ParseResult
 
 from tornado.testing import AsyncHTTPTestCase
+from nose.plugins.attrib import attr
+
+try:
+    from unittest import mock
```
Tarrasch: You should be able to assume that `unittest` is installed I believe.

test/server_test.py
```diff
@@ -72,19 +81,50 @@ def test_api_404(self):
         self._test_404('/api/foo')
 
 
+class INETServerClient(object):
```
Tarrasch: What is "INET"?

Tarrasch:
Starting to look better, but I want to make sure that this part of the code hold high quality as it's so central. :)

test/server_test.py
graingert: Socket family

https://docs.python.org/2/library/socket.html#socket.AF_UNIX
https://docs.python.org/2/library/socket.html#socket.AF_INET

test/server_test.py
graingert: Python 3.4 puts mock on unittest.mock

luigi/rpc.py
graingert: https://github.com/spotify/luigi/pull/1129/files#diff-9b595b54a300214f2172cde2e9c1eec4R169

luigi/rpc.py
graingert: yep that makes more sense

graingert:
@Tarrasch thanks!

test/server_test.py
Tarrasch: Ah ok :)

Tarrasch:
Looks good to me now :)

test/server_test.py
erikbern: internet :)

it's pretty cool, you should try it!

luigi/rpc.py
```diff
@@ -42,32 +57,62 @@ def __init__(self, message, sub_exception=None):
...
         time.sleep(30)
 
     def _fetch(self, url, body, log_exceptions=True, attempts=3):
-
-        full_url = 'http://{host}:{port:d}{prefix}{url}'.format(
-            host=self._host,
-            port=self._port,
-            prefix=self._url_prefix,
-            url=url)
+        full_url = self._url + url
```
Tarrasch: Erm, there's no way this line can be "right". Are the variable names extremely confusing? Or you actually string-concatenating two urls?

graingert: `full_url = self._url + url_suffix`?

Tarrasch: Yea, that's *much* clearer.

Tarrasch:
Ok, can somebody else give a somewhat careful review and then we can merge? Maybe you can add some documentation @graingert?

graingert:
@Tarrasch docs added

erikbern:
LGTM

Tarrasch:
Cool, let's see how this does in the wild now :)

graingert:
@Tarrasch do you know when the next release might be?  Is there an official release schedule?

Tarrasch:
No, it'll take a while (month?) from now I think. Unless somebody really asks for it. I'll have to release a 2.0.0 since things have changed.

luigi/rpc.py
```diff
@@ -42,32 +57,62 @@ def __init__(self, message, sub_exception=None):
...
-        self._host = host
-        self._port = port
-        self._url_prefix = url_prefix
+    def __init__(self, url='http://localhost:8082/', connect_timeout=None):
+        assert (
+            not (url.startswith('http+unix://') and not HAS_UNIX_SOCKETS),
+            'You need to install requests-unixsocket for Unix socket support.',
+        )
 
+        self._url = url
```
paxan: If trailing `/` is present on `url`, fetcher gets lots of 404s for path `/app/...`
So it's better to do:

    self._url = url.rstrip('/')

Tarrasch: Would be nice if you could whip up a PR with testcase+fix. :)

graingert: This was possible before with the url_prefix kwargument, so I did this intentionally. But yes the other behaviour makes way more sense

geowurster:
@graingert @Tarrasch @erikbern I think this PR may have broken something involving the worker-master communication, although it could be an issue with my cluster configuration.  For reference, I'm using Google's [`bdutil 1.3.1 (latest)`](https://cloud.google.com/hadoop/bdutil?hl=en) to configure and deploy a hadoop cluster on Google Compute Engine.  After installing the latest commit on master my workers are not able to connect to the scheduler:

```python
ERROR:luigi-interface:Failed connecting to remote scheduler 'http://luigi-master:8082/'
  File "/usr/local/lib/python2.7/dist-packages/luigi/rpc.py", line 124, in _fetch
  File "/usr/local/lib/python2.7/dist-packages/luigi/rpc.py", line 71, in fetch
    raise FetcherException(e)
FetcherException
WARNING: Failed pinging scheduler
WARNING:luigi-interface:Failed pinging scheduler
```

I tried several different commits and I believe fb9a0a2370f47d238e8df9234de8a3f6dbb332e9 is the latest functional commit and 0a3797911425f6f16812ac441f14c4e5e0a2e4c8 introduced the change.

The [blame](https://github.com/spotify/luigi/blame/master/luigi/rpc.py#L71) shows that the object raising this exception was added in a commit in this PR. The annotated screenshot might help to resolve the timeline.

Is anyone else on GCE experiencing this behavior?

commits_ _spotify_luigi

graingert:
@geowurster what's the "original_exc" from the worker?

geowurster:
@graingert `HTTP Error 404: Not Found`.

graingert:
that's odd

luigi/rpc.py
geowurster: @graingert I think I'm seeing this problem.  Looking at the `URLLibFetcher()`:

```python
class URLLibFetcher(object):
    def fetch(self, full_url, body, timeout):
        try:
            body = urlencode(body).encode('utf-8')
            return urlopen(full_url, body, timeout).read().decode('utf-8')
        except (URLError, socket.timeout) as e:
            raise FetcherException(e)
```

The `full_url` is `http://luigi-master:8082//api/add_task`, which has a double slash after the port.  Changing that to a single slash seems to fix the problem.

paxan: @Tarrasch here you go:
https://github.com/spotify/luigi/pull/1147

Tarrasch:
@graingert Can you commit to fixing this? I think it's pretty urgent if people can't rely on HEAD anymore! Is #1147 by @paxan related?

PR #1130, opened 2015-08-13
Revert "Atomic bigquery uploads"

DeaconDesperado:
Reverts spotify/luigi#1127

Investigating broken master build.

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -19,7 +19,6 @@ import collections
 import logging
 import luigi.target
 import time
-import random
 
 logger = logging.getLogger('luigi-interface')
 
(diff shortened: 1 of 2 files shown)


Tarrasch:
There's usually no extreme hurry to fix build errors if they only affect a small part of luigi like gcloud.

Was it just an intermittent failure? Unfortunately there sometimes are intermittent failures in the luigi build causing extreme confusion. :-/

DeaconDesperado:
@Tarrasch Wasn't sure and sort of panicked, but it looks like everything is ok now! :smile:

Tarrasch:
Great, keep on contributing awesome patches @DeaconDesperado  :)

PR #1133, opened 2015-08-14
Remove is_list support for Parameters

Tarrasch:
The main motivation is that they are harder to reason about. For example
it's unclear how command-line parameters and config interplay (replacing
semantics or additive semantics? If additive, in what order?)

It was proposed in this Google Group thread[1]. Indeed, this does break
backward compatibility, but it's quite a burden to maintain in addition
to having unclear semantics.

As a workaround, you can for example have comma-seperated parameters.
One can potentially implement a listify-factory for Parmeters objects
too.

[1]: https://groups.google.com/forum/#!topic/luigi-user/kywHGzDD02c

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -101,16 +101,13 @@ class Parameter(object):
     """non-atomically increasing counter used for ordering parameters."""
 
     @deprecate_kwarg('is_boolean', 'is_bool', False)
-    def __init__(self, default=_no_value, is_list=False, is_boolean=False, is_global=False, significant=True, description=None,
+    def __init__(self, default=_no_value, is_boolean=False, is_global=False, significant=True, description=None,
                  config_path=None, positional=True):
         """
         :param default: the default value for this parameter. This should match the type of the
                         Parameter, i.e. ``datetime.date`` for ``DateParameter`` or ``int`` for
                         ``IntParameter``. By default, no default is stored and
                         the value must be specified at runtime.
-        :param bool is_list: specify ``True`` if the parameter should allow a list of values rather
-                             than a single value. Default: ``False``. A list has an implicit default
-                             value of ``[]``.
         :param bool is_bool: specify ``True`` if the parameter is a bool value. Default:
                                 ``False``. Bool's have an implicit default value of ``False``.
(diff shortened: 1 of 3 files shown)


erikbern:
I like it! But would prefer it if we can deprecate then for a few months first

Tarrasch:
@erikbern I thought of that too, but I couldn't be bothered given how little response there was in that google group. I thought of the next release of Luigi could be 2.0.0, removing quite some things, like opt_parse and is_list and revamping the way luigi tasks are instantiated. Maybe we can also have #1086 merged then too?

erikbern:
that would be nice! but i still think we should deprecate it for a month or something

Tarrasch:
Hmm, let's first see if anybody complains in this thread first? If somebody is using HEAD they should expect some things changing right?

Tarrasch:
On second thought, if I'm going to rewrite most of the parameter parsing logic anyway, I might just add a deprecation warning and remove the feature in the big re-factoring anyway. Hmm...

erikbern:
yeah it's fine for people using HEAD but then that means we can't push to pypi without breaking compatibility still

Tarrasch:
Well, we already have the opt_parse removal which breaks backward compatibility. We kind of removed that without headsup deprecation warnings iirc.

Anyway, next release should be 2.0 probably

erikbern:
ok you decide, i'm just a grumpy lazy person!

Tarrasch:
I'm merging, because the interface for luigi Parameters will change in many other ways too. So it's not reasonable to keep backward compatibility.

erikbern:
ok cool and then let's bump to 2.0.0 for next pypi release

jfeala:
Hi guys, I just was testing HEAD to try out the new visualizer and this issue broke my code. Searching brought me here. I agree with the thoughts above and am fine to switch to comma-separated args. 

But I'm wondering if there is an active changelog somewhere so we can anticipate breaking changes without waiting for the release notes?

Tarrasch:
That's actually a good point @jfeala. I should create, if not only for myself as it'll be a long list of PRs to go through to summarize all the changes for upcoming 2.0.0.

erikbern:
yeah good points. we've been bad at making formal releases.

PR #1134, opened 2015-08-14
More informative error messages for RemoteContext commands

daveFNbuck:
When I get an error e-mail with text like Command '['test', '-e', 'filename']'
returned non-zero exit status 255 from a remote filesystem operation, it's a bit
difficult to debug without knowing which machine the command was run on. This
improves the situation by adding the hostname to the error message.


Tarrasch:
LGTM

PR #1135, opened 2015-08-14
Enables task failure on worker death while assistant is running

daveFNbuck:
We avoid pruning PENDING and RUNNING tasks while an assistant is live, as the
assistant is considered a stakeholder for all runnable tasks. Unfortunately,
the prune function is where we fail RUNNING tasks for which the running worker
has died. So when an assistant is alive, it becomes impossible for a RUNNING
task whose worker has died to ever change status.

This fixes the issue by separating the function to mark RUNNING tasks as FAILED
from the pruning function and ensuring that it gets called regardless of the
presence of assistants.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -415,16 +415,7 @@ class SimpleTaskState(object):
         self._status_tasks[new_status][task.id] = task
         task.status = new_status
 
-    def prune(self, task, config, assistants):
-        remove = False
-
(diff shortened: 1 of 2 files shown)


Tarrasch:
Ah, that must have been causing lots of stale tasks!

daveFNbuck:
It was pretty weird, as long-dead workers would keep popping up in the worker list due to the still RUNNING tasks.

PR #1136, opened 2015-08-14
Allows assistant to work on orphaned tasks

daveFNbuck:
Assistants keep PENDING tasks alive in order to work on them, but aren't
actually able to work on them because we check whether tasks have live workers
before we give them to assistants. This is to prevent assistants from working on
external tasks. In order to allow assistants to work on tasks with no live
workers but not work on external tasks, we store whether a task is runnable in
the scheduler.

To keep compatibility with old pickle files, we treat tasks without runnable set
as runnable. This should minimize the problems caused be the changed schema.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -194,6 +194,7 @@ class Task(object):
         self.disable_hard_timeout = disable_hard_timeout
         self.failures = Failures(disable_window)
         self.scheduler_disable_time = None
+        self.runnable = False
 
     def __repr__(self):
         return "Task(%r)" % vars(self)
@@ -640,6 +641,7 @@ class CentralPlannerScheduler(Scheduler):
         if runnable:
             task.workers.add(worker_id)
             self._state.get_worker(worker_id).tasks.add(task)
+            task.runnable = runnable
 
         if expl is not None:
             task.expl = expl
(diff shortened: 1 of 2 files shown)


luigi/scheduler.py
```diff
@@ -170,7 +170,8 @@ def _get_default(x, default):
 class Task(object):
 
     def __init__(self, task_id, status, deps, resources=None, priority=0, family='', module=None,
-                 params=None, disable_failures=None, disable_window=None, disable_hard_timeout=None):
+                 params=None, disable_failures=None, disable_window=None, disable_hard_timeout=None,
+                 runnable=False):
```
Tarrasch: Where do you set this kwarg?

Tarrasch:
Looks sensible, but I don't see how it works. (See my inline comment)

luigi/scheduler.py
daveFNbuck: It gets set in add_task post-initialization.

Tarrasch: If it's post-initialization (not in the __init__), why do you need it as a kwarg in the __init__? Can you give a line-hyperurl?

daveFNbuck: I guess it doesn't need that at all.

daveFNbuck: The line is at https://github.com/spotify/luigi/pull/1136/files#diff-ec5c8f6c192cef5a08a5e22aa6273604R643

Tarrasch:
Sweet! :)

PR #1137, opened 2015-08-14
Enables disabling of old workers while allowing them to finish currently running tasks

daveFNbuck:
This PR is split into two commits. The first one adds a signal handler for SIGUSR1 to workers, which prevents them from calling get_work. This allows old workers to be replaced without killing currently running tasks. By doing this, we prevent old code from being run going forward without losing progress on long-running tasks.

The second commit adds a --take-lock argument that causes a job to send SIGUSR1 to jobs that would otherwise occupy its lock. I've been using this successfully with an assistant that is automatically run with code deploys to ensure that the latest code is always being run without continuous deployment constantly killing all of my jobs.

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -24,6 +24,7 @@ import logging.config
 import os
 import sys
 import tempfile
+import signal
 
 from luigi import configuration
 from luigi import lock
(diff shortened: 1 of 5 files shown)


Tarrasch:
LGTM, can you just fix the small comment on variable name?

daveFNbuck:
Fixed the variable name.

PR #1138, opened 2015-08-15
Cleanup files created during tests

graingert:


test/bin_test.py
```diff
@@ -32,4 +38,11 @@ class LuigidDaemonTest(server_test.ServerTestRun):
 
     @mock.patch('daemon.DaemonContext')
     def run_server(self, daemon_context):
-        luigi.cmdline.luigid(['--port', str(self._api_port), '--background', '--logdir', '.', '--pidfile', self.id() + '.pid'])
+        tempdir = tempfile.mkdtemp(prefix=self.id())
```
erikbern: hey you should probably use a context manager so you don't have to use `addCleanup`

https://docs.python.org/3/library/tempfile.html#examples

i've had some issues with tempfile before so not 100% that it works well

test/server_test.py
```diff
@@ -95,6 +96,7 @@ def stop_server(self):
 
     def setUp(self):
         state_path = tempfile.mktemp(suffix=self.id())
+        self.addCleanup(functools.partial(os.unlink, state_path))
```
erikbern: same here

test/bin_test.py
graingert: It's difficult to use a context manager with unittest's method call interface, especially in this case where this is an overridden method

graingert: Also python 2 doesn't have a tempdir context manager

erikbern: ok got it – if it's not in python 2 then nvm!

Tarrasch:
Sweeeeet

PR #1139, opened 2015-08-17
Properly handle space in ssh command for remote hadoop jar job

interskh:
Esp when task_id has space inside, the remote command will fail. By using pipes.quote(), it can properly quote the args that passed in ssh command.


Tarrasch:
Can you add a test case as well?

interskh:
Test case added

erikbern:
LGTM

PR #1142, opened 2015-08-18
Clean up docs for the execution_summary module

Tarrasch:


Tarrasch:
It looked good when I `tox -e docs`ed it. So I'll just merge. :)

PR #1145, opened 2015-08-19
Moved atomic swaps out of BigQueryLoadTask

DeaconDesperado:
Per conversation in #1127, this restores the original behavior of BigQueryLoadTask but also leaves a stub implementation of swap-in-place on tables for task implementors using the hadoop to bigquery connector.

@Tarrasch @mikekap

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -282,9 +282,9 @@ class BigqueryClient(object):
 
 
 class BigqueryTarget(luigi.target.Target):
-    def __init__(self, project_id, dataset_id, table_id, client=None, tmp_dataset_id=None):
+    def __init__(self, project_id, dataset_id, table_id, client=None, tmp_dataset_id=None, tmp_salt=None):
         self.table = BQTable(project_id=project_id, dataset_id=dataset_id, table_id=table_id)
-        tmp_table_id = "_" + table_id + "_%09d" % random.randrange(0, 1e10)
+        tmp_table_id = "_" + table_id + (tmp_salt or "_%09d" % random.randrange(0, 1e10))
         self.tmp_table = BQTable(project_id=project_id, dataset_id=tmp_dataset_id or "_incoming", table_id=tmp_table_id)
         self.client = client or BigqueryClient()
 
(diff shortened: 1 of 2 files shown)


Tarrasch:
Looks good, but would be nice if you also added a test for `move_tmp_table` I suppose. I'll let @mikekap chime in before we merge this time. :)

DeaconDesperado:
@Tarrasch Added a test using a subclass of BigqueryLoadTask.

Also made the random salt for the temporary table configurable.

luigi/contrib/bigquery.py
```diff
@@ -282,9 +282,9 @@ def copy(self,
 
 
 class BigqueryTarget(luigi.target.Target):
-    def __init__(self, project_id, dataset_id, table_id, client=None, tmp_dataset_id=None):
+    def __init__(self, project_id, dataset_id, table_id, client=None, tmp_dataset_id=None, tmp_salt=None):
         self.table = BQTable(project_id=project_id, dataset_id=dataset_id, table_id=table_id)
-        tmp_table_id = "_" + table_id + "_%09d" % random.randrange(0, 1e10)
+        tmp_table_id = "_" + table_id + (tmp_salt or "_%09d" % random.randrange(0, 1e10))
```
mikekap: Hmm since most tasks that use BQ directly don't need this, what do you think about having an interface like:

```python
target = BigqueryTarget(...)
temp_table = target.temp_table(salt=...)
# temp_table is a BQTable
DO_HADOOP_CONNECTOR_JOB()
```

It might be nice to do it as a context manager as well, but I think with the way luigi classes are usually implemented, you might need the temp table name outside of the run() method - it'll get weird.

Separately, I'd suggest using a sha hash of the table id or something - using a non-consistent table name might lead to subtle bugs (i.e. two BigqueryTarget() instances pointing to the same project/dataset/table don't have the same temp table id)

DeaconDesperado: @mikekap, I like it, thanks for the feedback.  I had encountered the the access outside run in writing the test so this refactor makes a lot of sense.

DeaconDesperado:
Updated per @mikekap's suggestions.

luigi/contrib/bigquery.py
```diff
@@ -33,6 +33,8 @@
     logger.warning('Bigquery module imported, but google-api-python-client is '
                    'not installed. Any bigquery task will fail')
 
+IMPORT_TIME_SALT = random.randrange(0, 1e10)
```
Tarrasch: Any reason you don't do this once per target?

mikekap: I think I gave that idea - I thought it might be better given the way luigi targets tend to get re-constructed (which would cause two BigqueryTargets pointing to the same table to have two different temp table names). It feels as though maybe the nicest way would be to either have a completely consistent table name (but one that may exist so maybe blow it away at use time), or return a new table name every time the function is called.

But really, I think any of these approaches are fine :)

Tarrasch: Ah, I see! Can you add that explenation @DeaconDesperado? Then I think this can be merged :)

Tarrasch:

DeaconDesperado:
Will do @Tarrasch.  Sorry to have missed the pings.  I'll be looking at the issue with GCS later tonight.

DeaconDesperado:
@Tarrasch As it turns out, this logic is no longer necessary and was the result of some misunderstandings we were seeing using it with the java bigquery connector.

I have a local branch I'll check in removing it that just depends of #1420 for test fixes.  Thanks for being patient!

Tarrasch:
Wow, you really seem to be on top of things now. Great! :)

PR #1146, opened 2015-08-19
Fix error when connecting to the central luigi scheduler

graingert:


Tarrasch:
Can you please also add a test case? I merge now as I assume you've tested this somehow and the build is still green. But I expect you to add a follow up test case. Thanks! :)

PR #1147, opened 2015-08-19
RemoteScheduler: use urljoin on base URL & suffix to form a full URL

paxan:


luigi/rpc.py
```diff
@@ -24,6 +24,7 @@
 import logging
 import socket
 import time
+from urlparse import urljoin
```
graingert: Python 3?

Tarrasch: Yea, you can iterate locally to catch these errors early @paxan, see [CONTRIBUTING](https://github.com/spotify/luigi/blob/master/CONTRIBUTING.rst) file. :)

paxan: My bad. I iterated locally only with py27 :(

Tarrasch:
I usually wants commits to be squashed, but I might make an exception here as I suppose this is pretty urgent.

Thanks. :)

paxan:
i will squash!

Tarrasch:
Awesome, thanks for the quick fix @paxan, I hope this didn't have drastic consequences for your or anybody else production environments!

paxan:
:+1:  (i've worked around it by ensuring scheduler_url arg has no trailing `/`)

Tarrasch:
But after this patch it works regardless, right?

geowurster:
I'll check master against my setup later today.

geowurster:
@Tarrasch @paxan @graingert Looks like this is functioning correctly for me.  Thanks :+1:

paxan:
@tarrasch right.

PR #1150, opened 2015-08-20
Clarified the purpose of the configuration files

Yoone:


Tarrasch:
Actually it's read by *both* the client and the server (yea, this is a bit unfortunate). Can you send a new PR with that in mind? :)

Yoone:
Oh, I didn't understand that. Wouldn't it be clearer to specify which categories are for which (luigid or the client)?

Reading the docs I initially thought that `default-scheduler-host` and `default-scheduler-port` were meant for luigid, until I saw `luigid -h` and its `--port` and `--address` parameters.

Yoone:
New PR here: https://github.com/spotify/luigi/pull/1159

PR #1153, opened 2015-08-21
Prefer list2command over ' '.join()

Tarrasch:
Just a nice convenienve. The motivation is that it becomes easier to
copy these logs and paste directly into the terminal without manually
escaping spaces and whatnot.


ulzha:
list2cmdline is undocumented, https://bugs.python.org/issue11827, but also unlikely to disappear without a replacement in the future (so I conclude from the discussion on the same issue).

PR #1154, opened 2015-08-21
Unexport private parts of interface.py

Tarrasch:
The purpose is to make the documentation page pleaner[1] and to not make
it easy for users of luigi to misuse luigi and start relying on things
so we don't dare changing it later.

This will of course break for people who've considered those exports to
be sane to use, but I think that's fine since we're making proper
releases of the luigi package.

[1]: http://luigi.readthedocs.org/en/latest/api/luigi.interface.html

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -106,7 +106,7 @@ class core(task.Config):
         description='Configuration file for logging')
     module = parameter.Parameter(
         default=None,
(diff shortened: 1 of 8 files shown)


Tarrasch:
Want to review @erikbern ?

erikbern:
LGTM

PR #1155, opened 2015-08-21
Use ExternalTask for a external Target objects

bmaggard:
Using 'Task' results in 'Unfulfilled dependency at run time'


Tarrasch:
Thanks!! :)

PR #1157, opened 2015-08-21
Unexport parameters stuff

Tarrasch:
The most of the content is in the second commit. Please see the commit message there.

@erikbern can you review as you wrote much of this?

For later when I actually refactor the code, I know have a rough idea of how many test cases that needs to be changed with that refactoring.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -299,9 +299,6 @@ class Parameter(object):
         else:
             return self.parse(x)
 
-    def serialize_to_input(self, x):
-        return self.serialize(x)
-
     def parser_dest(self, param_name, task_name, glob=False, is_without_section=False):
         if is_without_section:
             if glob:
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -190,7 +190,7 @@ class Parameter(object):
         return _no_value
 
     @property
-    def has_value(self):
+    def _has_value(self):
         """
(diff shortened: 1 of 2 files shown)


Tarrasch:
Also, by hiding the stuff now, it's unblocking a potential 2.0.0 release and the actual refactoring can happen under the hood not requiring yet another major version bump.

erikbern:
LGTM

jkukul:
@Tarrasch - this PR removed some public apis from `Parameter` class, e.g. `value`, `has_value`, `reset_global`, etc. For people who access them in workflows, upgrading from `1.3` to `2.0.0` will break things.  
 
I guess it's ok, since it's a major upgrade, but it would have been nicer if these methods were deprecated in one of `1.*` releases (just something to maybe keep in mind for the future).

Tarrasch:
@jkukul - Then the cost for the maintainers is higher. I agree it would be nicer with deprecations in-between, but then we wouldn't have time to do other nice things. Also, *all of those methods were inherently buggy*. Their behaviour was undefined really.

I'm sorry if things broke for you. :-/

Tarrasch:
Also this PR is related to #1182

jkukul:
@Tarrasch no problem! I completely understand that the development speed is on the other side of the trade-off and that finding the right balance is hard.

PR #1159, opened 2015-08-24
Clarified the purpose of the configuration files

Yoone:
Added info about the fact that both the client and `luigid` load configuration files.


Tarrasch:
LGTM

PR #1164, opened 2015-08-25
luigid: Add format conversion code for #916

Tarrasch:
In spotify/luigi#916 all (scheduler) tasks started to have the
disable_hard_timeout attribute which the code assumes. But the scheduler
will crash if you load an old pickle file.

Yea I know it was ages ago we merged that PR but Spotify still runs some
luigid with old state files.

I did test that this works. But I did not add any unit test, as this is
not any compatibility we want to maintain in the long run. To make the
code base a bit cleaner, I dated all the format conversion code.


sisidra:
:+1:

PR #1167, opened 2015-08-26
Refactor Parameter._get_value()

Tarrasch:
The purpose is to clarify where in the code the actual ordering happens.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -163,31 +163,31 @@ class Parameter(object):
         return self.parse(value)
 
     def _get_value(self, task_name=None, param_name=None):
-        if self.__global != _no_value:
-            return self.__global
-        if task_name and param_name:
-            v = self._get_value_from_config(task_name, param_name)
-            if v != _no_value:
-                return v
-            v = self._get_value_from_config(task_name, param_name.replace('_', '-'))
-            if v != _no_value:
-                warnings.warn(
-                    'The use of the configuration [%s] %s (with dashes) should be avoided. Please use underscores.' %
-                    (task_name, param_name), DeprecationWarning, stacklevel=2)
-                return v
-        if self.__config:
-            v = self._get_value_from_config(self.__config['section'], self.__config['name'])
-            if v != _no_value and task_name and param_name:
-                warnings.warn(
-                    'The use of the configuration [%s] %s is deprecated. Please use [%s] %s' %
-                    (self.__config['section'], self.__config['name'], task_name, param_name),
-                    DeprecationWarning, stacklevel=2)
-            if v != _no_value:
-                return v
-        if self.__default != _no_value:
-            return self.__default
-
+        for value, warn in self._value_iterator(task_name, param_name):
+            if value != _no_value:
+                if warn:
+                    warnings.warn(warn, DeprecationWarning, stacklevel=2)
+                return value
         return _no_value
+
+    def _value_iterator(self, task_name, param_name):
+        """
+        Yield the parameter values, with optional deprecation warning as second tuple value.
+
+        The parameter value will be whatever non-_no_value that is yielded first.
+        """
+        yield (self.__global, None)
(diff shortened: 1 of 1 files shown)


Tarrasch:
Can you do a quick review @erikbern ?

luigi/parameter.py
```diff
@@ -163,31 +163,31 @@ def _get_value_from_config(self, section, name):
         return self.parse(value)
 
     def _get_value(self, task_name=None, param_name=None):
-        if self.__global != _no_value:
-            return self.__global
+        for value, warn in self._value_iterator(task_name, param_name):
+            if value != _no_value:
```
sisidra: Cries for Optional.{hasValue,get,...} container in python... :(

luigi/parameter.py
```diff
@@ -163,31 +163,31 @@ def _get_value_from_config(self, section, name):
...
-            if v != _no_value:
-                return v
-            v = self._get_value_from_config(task_name, param_name.replace('_', '-'))
-            if v != _no_value:
-                warnings.warn(
-                    'The use of the configuration [%s] %s (with dashes) should be avoided. Please use underscores.' %
-                    (task_name, param_name), DeprecationWarning, stacklevel=2)
-                return v
+            yield (self._get_value_from_config(task_name, param_name), None)
+            warn = 'Configuration [{}] {} (with dashes) should be avoided. Please use underscores.'.format(
```
sisidra: variable warn is not used

luigi/parameter.py
```diff
@@ -163,31 +163,31 @@ def _get_value_from_config(self, section, name):
...
-                    DeprecationWarning, stacklevel=2)
-            if v != _no_value:
-                return v
-        if self.__default != _no_value:
-            return self.__default
-
-        return _no_value
+            warn = (task_name and param_name and
+                    'The use of the configuration [{}] {} is deprecated. Please use [{}] {}'.format(
+                    self.__config['section'], self.__config['name'], task_name, param_name))
```
sisidra: Shouldn't this warning be in constructor?
In other languages I would see overloaded constructor being the one with deprecation warning.

luigi/parameter.py
Tarrasch: Copy paste ftw! I'll fix this :)

luigi/parameter.py
Tarrasch: Yea, in most cases you can use `None` so I guess they didn't bother.

luigi/parameter.py
Tarrasch: The constructor of `Parameter()` or of `Task()`? Currently this code path will be executed when you instantiate a task (the constructor of `Task()`). I suppose it could be in the parameter's constructor too I guess. But I'm actually striving for making a `Parameter()`-object only data without side effects. I wouldn't like creating a `Parameter()` to have side effects.

Tarrasch:
I realized that having those `warn` local variables are really error prone, and the compiler nor pep8 complained about the unused variable. So I wrote it in a way where it's bound to be used.

luigi/parameter.py
sisidra: In Java you have null, that's not the point.. ;) Offtopic anyway.

Tarrasch:
Does the patch look good now @sisidra ? :)

luigi/parameter.py
sisidra: I see this as no side effect but as Python way to do @Deprecated.
And it is constructor with config_path parameter that is deprecated, not some method invocation after that.

sisidra:
I would love to see unittests as you made mistake before :tongue: 
But i'm not strongly insisting.
lgtm

Tarrasch:
Hmm, I haven't checked how to tests with deprecation warnings. Oh well, no non-warning-related functionality have changed so old tests pass. I also tried removing random `yield` statements and then the tests failed. So I have some confidence in the code. :)

PR #1169, opened 2015-08-27
Remove Parameter.{_value,_has_value}

Tarrasch:
Pretty much a continuation of #1157. See also commit messages.

PR #1170, opened 2015-08-27
Simplify Parameter internals

Tarrasch:
It's now possible (since spotify/luigi#1169) to assume that task_name
and param_name are given when doing `_get_value()`.


sisidra:
:+1:

PR #1173, opened 2015-08-28
Removed hardcoded delimiter option from base copy method

spiside:
The base S3CopyToTable class included a hardcoded delimiter option as part of the COPY command used with Redshift. This was causing issues when trying to copy a CSV file after adding the CSV copy option. Since it is a base class, I felt that the delimiter command should be passed in as an option rather than required in the copy method. 

Status: **Requesting Merge**
Reference: [Redshift COPY Command Documentation](http://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html)

## Changes
- Removed hardcoded delimiter command
- Added example of delimiter command into copy_option comment

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -99,15 +99,18 @@ class S3CopyToTable(rdbms.CopyToTable):
         * IGNOREHEADER 1
         * TRUNCATECOLUMNS
         * IGNOREBLANKLINES
+        * DELIMITER '\t'
         """
         return ''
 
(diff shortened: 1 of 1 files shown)


Tarrasch:
Cool, try to find somebody who knows s3+redshift to review this. :)

spiside:
Awesome, thanks!

spiside:
@steenzout @rantav @erikbern

erikbern:
sounds good but better if someone else can review

PR #1174, opened 2015-08-29
Make luigi package flake8 compliant

toidi:
`flake8` perform more source code checks then `pep8`.

Despite of there is some upcoming work, I made the step to improve code quality.

BTW, could you tell what is wrong with `doc/settings.py`? This file contain invalid python code.

PR #1175, opened 2015-08-30
Make luigi package more flake8 compliant

toidi:
Motivation of this PR was to make flake8 checks as strict as possible.

Before: 
```
commands = flake8 --max-line-length=384 --exclude=doc,test,luigi/six.py
  flake8 --max-line-length=100 --ignore=E265 doc
  flake8 --max-line-length=252 --ignore=F401,F841 test
```
After:

```
commands = flake8 --max-line-length=160 --exclude=doc,luigi/six.py
   flake8 --max-line-length=100 --ignore=E265 doc
```

This will protect luigi package from govnokod.

In future `max-line-length` should be less then 120 chars for better readability.

### DIFF
--- a/luigi/execution_summary.py
+++ b/luigi/execution_summary.py
@@ -33,10 +33,13 @@ def _partition_tasks(worker):
     pending_tasks = {task for(task, status, ext) in task_history if status == 'PENDING'}
     set_tasks = {}
     set_tasks["completed"] = {task for (task, status, ext) in task_history if status == 'DONE' and task in pending_tasks}
-    set_tasks["already_done"] = {task for (task, status, ext) in task_history if status == 'DONE' and task not in pending_tasks and task not in set_tasks["completed"]}
+    set_tasks["already_done"] = {task for (task, status, ext) in task_history
+                                 if status == 'DONE' and task not in pending_tasks and task not in set_tasks["completed"]}
     set_tasks["failed"] = {task for (task, status, ext) in task_history if status == 'FAILED'}
-    set_tasks["still_pending_ext"] = {task for (task, status, ext) in task_history if status == 'PENDING' and task not in set_tasks["failed"] and task not in set_tasks["completed"] and not ext}
-    set_tasks["still_pending_not_ext"] = {task for (task, status, ext) in task_history if status == 'PENDING' and task not in set_tasks["failed"] and task not in set_tasks["completed"] and ext}
+    set_tasks["still_pending_ext"] = {task for (task, status, ext) in task_history
+                                      if status == 'PENDING' and task not in set_tasks["failed"] and task not in set_tasks["completed"] and not ext}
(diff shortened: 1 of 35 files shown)


luigi/execution_summary.py
```diff
@@ -97,7 +101,8 @@ def _get_str(task_dict, extra_indent):
             break
         if len(tasks[0].get_params()) == 0:
             row += '- {0} {1}()'.format(len(tasks), str(task_family))
-        elif _get_len_of_params(tasks[0]) > 60 or (len(tasks) == 2 and len(tasks[0].get_params()) > 1 and (_get_len_of_params(tasks[0]) > 40 or len(str(tasks[0])) > 100)) or len(str(tasks[0])) > 200:
+        elif _get_len_of_params(tasks[0]) > 60 or len(str(tasks[0])) > 200 or \
```
Tarrasch: I thought `\` were unpythonic and one preferred parenthesis? No?

Tarrasch:
Awesome progress, for future reference, this a continuation of #1174.

Tarrasch:
Thanks so much for caring about the code quality of luigi! :)

Tarrasch:
What's "govnokod" by the way? I only found results in Russian. :)

luigi/execution_summary.py
toidi: I guess there is no more or less pythonic. In my opinion - pythonic mean better readable.

In some cases parenthesis better readable but in this case - this condition `\` looks good. Condition must no be over-parenthesised.


https://www.python.org/dev/peps/pep-0008/#maximum-line-length
> Backslashes may still be appropriate at times. For example, long, multiple with -statements cannot use implicit continuation, so backslashes are acceptable:

toidi:
> Thanks so much for caring about the code quality of luigi! :)
We use `luigi` and loved it. I wish bright future for project. You are welcome.

> What's "govnokod" by the way?
Well.. It is very specific term. Usually govnokod generated by anything but not developers head. It is like monkey code but a little bit different.

Talk is cheap. I show you the code: http://govnokod.ru :-))

test/execution_summary_test.py
```diff
@@ -166,8 +173,17 @@ def run(self):
...
+                      '* 2 were left pending, among these:\n'
+                      '    * 1 were being run by another worker:\n'
+                      '        - 1 LockTask()\n'
+                      '    * 1 had dependencies that were being run by other worker:\n'
+                      '        - 1 ParentTask()\n', s)
+        self.assertIn('\n\nThe other workers were:\n'
+                      '    - other_worker ran 1 tasks\n\n'
+                      'Did not run any tasks\n'
+                      'This progress looks :) because there were no failed '
+                      'tasks or missing external dependencies\n', s)
```
Tarrasch: Ah this looks so much nicer! :)

Tarrasch:
This is *so* awesome! Thanks :)

PR #1176, opened 2015-08-30
Support unix sockets for faster internal connections (MK II)

graingert:


Tarrasch:
Thanks @graingert !

Tarrasch:
When running the tests, don't you have the habit of doing "(a) try to ruin something in implementation (b) confirm that test fails now"? I'm still a bit worried that this could have happened.

Tarrasch:
What's going on @graingert? Is this the first time you actually run the tests??? :s

graingert:
I got them working with just requests, then backported the old code with a load of "refactors" on both the tests and code

PR #1177, opened 2015-08-31
Deprecate 'use_dynamic_argparse' keyword

Tarrasch:
Dynamic argparse is the feature enabling `--module` command line option.

The motivation is that there probably is no real use case where you want
it to be False. From now on assume it to be True.

This commit also changes the behavior of luigi.run (which is discouraged
as direct invocation anyway) in that now we always allow dynamic
argparse.


lenaspotify:
LGTM

PR #1180, opened 2015-08-31
Simplify interface.py

Tarrasch:
Previous code had unnecessary OOP abstractions. Also there were
countless of unrelated occurrences of the method "run".

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -129,68 +129,62 @@ class _WorkerSchedulerFactory(object):
             scheduler=scheduler, worker_processes=worker_processes, assistant=assistant)
 
 
-class _Interface(object):
+def _schedule_and_run(tasks, worker_scheduler_factory=None, override_defaults=None):
+    """
+    :param tasks:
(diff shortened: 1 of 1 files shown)


Tarrasch:
Oh, and the reason there previously was `class Interface(object)` was probably that we once supported both optparse and argparse. See #1004

erikbern:
LGTM but might be worth slowing down with the refactoring soon so that people have time to run it locally!

Tarrasch:
Well, my intention is a to do a quite big re-factoring. I try to split up the commits to speed up my own development and make it much easier to review.

Most of the PRs I've sent don't change any of the public behavior (that's why I started off with a PR turning everything into private, so people don't get affected by my PRs on a daily basis :))

erikbern:
ok haven't you been removing lots of functionality though?

Tarrasch:
I meant that I removed (hide) the functionalities in #1154 and #1157. So commits like this one doesn't touch any of luigi's public API.

PR #1181, opened 2015-08-31
Move retry_time logic to its own method.

dhruvg:
This allows subclassing schedulers to modify the retry times on a task-by-task basis, which can be useful for exponential back offs.


Tarrasch:
Looks good! Just out of curiosity, do you have your own scheduler or something? :)

dhruvg:
Nah we use `CentralPlannerScheduler` for the most part, but we want the flexibility to retry different tasks differently. Many of our tasks talk to external APIs and we want to be able to exponentially backoff on task failures if we get rate limited.

Tarrasch:
Cool, thanks for you contribution! :)

PR #1186, opened 2015-09-01
Fix error message naming / docs. Set task.expl earlier.

dhruvg:
Very minor cleanup and ordering change.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -116,7 +116,7 @@ class TaskProcess(multiprocessing.Process):
             random.seed((os.getpid(), time.time()))
 
         status = FAILED
-        error_message = ''
+        expl = ''
         missing = []
         new_deps = []
         try:
@@ -146,7 +146,7 @@ class TaskProcess(multiprocessing.Process):
             elif status == DONE:
                 self.task.trigger_event(
                     Event.PROCESSING_TIME, self.task, time.time() - t0)
-                error_message = json.dumps(self.task.on_success())
+                expl = json.dumps(self.task.on_success())
                 logger.info('[pid %s] Worker %s done      %s', os.getpid(),
                             self.worker_id, self.task.task_id)
                 self.task.trigger_event(Event.SUCCESS, self.task)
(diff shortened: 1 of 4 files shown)


luigi/worker.py
```diff
@@ -159,13 +159,15 @@ def run(self):
             self.task.trigger_event(Event.FAILURE, self.task, ex)
             subject = "Luigi: %s FAILED" % self.task
 
-            error_message = notifications.wrap_traceback(self.task.on_failure(ex))
+            raw_error_message = self.task.on_failure(ex)
+            notification_error_message = notifications.wrap_traceback(raw_error_message)
+            expl = json.dumps(raw_error_message)
```
Tarrasch: I don't  know this code really. But didn't you add ` json.dumps()` call here? So it's not just refactoring here right?

Tarrasch:
Cool! :)

luigi/worker.py
dhruvg: You're right, I forgot to mention this. I think this is the right behavior because we do it for `on_success` (i.e. `json.dumps(self.task.on_success())`).

Tarrasch:
I want to merge. Anyone else wants to review?

Tarrasch:
@dhruvg, since you're actually changing behaviour here. Can you somehow test, preferably by addinga test case, that your logical change (additional `json.dumps()`) fixes an existing bug?

dhruvg:
@Tarrasch Added tests, please merge at your convenience.

Tarrasch:
"Add tests." will won't be that useful when scrolling through a commit log. Just merge the two commits I think. :)

dhruvg:
Done. Thanks!

Tarrasch:
By the way, if you (hypothetically) would revert `worker.py`, would the new test cases fail?

dhruvg:
Yes, the FailTask case would fail. The different between using `json.dumps` and not using it is very subtle in the test, but important if we return `dicts` in the `on_failure` calls.

F...
FAIL: test_update_result_queue_on_failure (__main__.TaskProcessTest)
  File "test/worker_task_test.py", line 86, in test_update_result_queue_on_failure
    mock_put.assert_called_once_with((task.task_id, FAILED, json.dumps("test failure expl"), [], []))
  File "/Library/Python/2.7/site-packages/mock.py", line 835, in assert_called_once_with
    return self.assert_called_with(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/mock.py", line 824, in assert_called_with
    raise AssertionError(msg)
AssertionError: Expected call: put(('FailTask()', 'FAILED', '"test failure expl"', [], []))
Actual call: put(('FailTask()', 'FAILED', 'test failure expl', [], []))

Ran 4 tests in 0.054s

FAILED (failures=1)

Tarrasch:
Cool! Thanks! :)

PR #1187, opened 2015-09-01
Clean up worker subprocesses on timeout

ChrisBeaumont:
Luigi uses the multiprocessing module to manage worker tasks, and uses `multiprocessing.Process.terminate()` to implement worker timeout. However, as [mentioned in the docs](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.Process.terminate), terminate orphans subprocesses spawned by the task. Maybe there are good reasons for orphaning these processes instead of killing them. However, it was leading to large resource leaks in my application -- I'm assuming most luigi workflows want subprocesses to be cleaned up during a worker timeout.

The psutil module seems to make recursively terminating a process and its subprocesses straightforward (though I've only tested this on one machine). I'm not sure if you'd be keen on adding `psutil` as a required dependency, so I've proposed a solution that falls back to the default terminate behavior if it isn't available.

Here's a script to demonstrate the issue

```python
from subprocess import check_call
import luigi


class Worker(luigi.Task):

    foo = luigi.Parameter()

    def output(self):
        return luigi.LocalTarget(self.foo)

    def run(self):
        check_call(['python', '-c', 'import time; time.sleep(1000)'])


class Test(luigi.Task):

    def requires(self):
        return [Worker(foo="_%i" % i) for i in range(5)]

    def output(self):
        return luigi.LocalTarget('done')


if __name__ == "__main__":
    luigi.run()
```
Invoked via
```
python test.py Test --worker-timeout=3 --workers 5 --local-scheduler
```

With this patch, `ps` shows 5 processes of `python -c import time; time.sleep(1000)` while the pipeline is running, and 0 when completed. Without the patch, those 5 processes stay alive + orphaned.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -167,6 +167,23 @@ class TaskProcess(multiprocessing.Process):
             self.result_queue.put(
                 (self.task.task_id, status, error_message, missing, new_deps))
 
+    def _recursive_terminate(self):
+        import psutil
+
+        parent = psutil.Process(self.pid)
+        for child in parent.children(recursive=True):
+            child.kill()
+
+        parent.kill()
+
+    def terminate(self):
+        """Terminate this process and it's subprocesses."""
+        # default terminate() doesn't cleanup child processes, it orphans them.
+        try:
+            return self._recursive_terminate()
+        except ImportError:
+            return super(TaskProcess, self).terminate()
+
 
 class SingleProcessPool(object):
     """


Tarrasch:
Awesome finding! LGTM

PS. Damn intermittent Travis failure. #1188

luigi/worker.py
```diff
@@ -167,6 +167,23 @@ def run(self):
...
+        import psutil
+
+        parent = psutil.Process(self.pid)
+        for child in parent.children(recursive=True):
+            child.kill()
+
+        parent.kill()
+
+    def terminate(self):
+        """Terminate this process and it's subprocesses."""
```
erikbern: minor typo: *its* subprocesses

but no blocker. otherwise LGTM!

Tarrasch:
@ChrisBeaumont, by the way, could you add a test case to this too? Just to ensure that we don't happen to remove this functionality in a future refactoring.

ChrisBeaumont:
@Tarrasch sure thing

Issue #1189, opened 2015-09-02
Log also missing targets, not only missing tasks? (In RuntimeError: Unfulfilled dependency at run time)


samuell:
It seems that probably the most commonly occuring problem for us, using luigi for commandline and file-system based workflows, is [1]:

```log
RuntimeError: Unfulfilled dependency at run time: SomeTask(...etc etc ...
```

It also turns out that it can often be quite hard to figure out which of the outputs of that task (since we typically have multiple outputs), that is missing.

So, I would suggest if we could get the missing targets also logged somehow? ... either in the Exception error message, or as a warning log before?

----
[1] Probably this is extra common in our case, since some of our commandline apps unfortunately produce outputs of which we can not control the file name, so we have to just try to create a luigi output that matches the file name that is created, and sometimes this goes wrong.

Lucas-C:
What's the status of this @samuell ? It looked really promising

PR #1190, opened 2015-09-02
Log missing targets on unfulfilled dependencies

samuell:
This is a proposed solution for issue #1189

It seems that probably the most commonly occuring problem for us, using luigi for commandline and file-system based workflows, is `RuntimeError: Unfulfilled dependency at run time: ...`. [1]

It also turns out that it can often be quite hard to figure out which of the outputs of that task (since we typically have multiple outputs), that is missing.

To that end, I suggest something like this, where the missing targets are at least logged (with 'WARNING' level in this case), before firing the "Unfulfilled dependency" error.

Maybe it is even better to include them in the error message, but it seems that it will get very long?

For a nice string representation of the missing targets, I also added a __str__() representation for the LocalTarget class. One should implement similar for the other target classes, to get a nice output, but I don't know them well enough to do that off-arm right now.

An example output of the logs that this gets us (with our custom log formatting though, but you should get the idea ... and with some parameters cleared out for brevity):

```log
2015-09-02 22:35:13 |  WARNING | Missing Target: /exp/20150627-crossval/data/mm_test_small.smi.h1_3.sign.r1.10_50_rand_trn.csr.logg
2015-09-02 22:35:13 |    ERROR | [pid 20380] Worker Worker(...blabla...) failed    UnGzipFile(...blabla...)
  File "/home/samuel/.pyenv/versions/2.7.8/lib/python2.7/site-packages/luigi/worker.py", line 132, in run
    raise RuntimeError('Unfulfilled %s at run time: %s' % (deps, ', '.join(missing_taskids)))
RuntimeError: Unfulfilled dependency at run time: CreateSparseTrainDataset(...blabla...)
```

... so, you see that we get a `WARNING` about the missing file target, just before the `ERROR`, which would make life much easier for us.

What do you think, except that my code probably could use some improvements?

----
[1] Probably this is extra common in our case, since some of our commandline apps unfortunately produce outputs of which we can not control the file name, so we have to just try to create a luigi output that matches the file name that is created, and sometimes this goes wrong.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -117,14 +117,19 @@ class TaskProcess(multiprocessing.Process):
 
         status = FAILED
         error_message = ''
-        missing = []
+        missing_tasks = []
+        missing_targets = []
         new_deps = []
         try:
             # Verify that all the tasks are fulfilled!
-            missing = [dep.task_id for dep in self.task.deps() if not dep.complete()]
-            if missing:
-                deps = 'dependency' if len(missing) == 1 else 'dependencies'
-                raise RuntimeError('Unfulfilled %s at run time: %s' % (deps, ', '.join(missing)))
+            missing_deps = [dep for dep in self.task.deps() if not dep.complete()]
+            missing_targets = [tgt for tgt in flatten(dep.output()) for dep in missing_deps if not tgt.exists()]
+            for missing_target in missing_targets:
+                logger.warn('Missing Target: %s' % missing_target)
+            missing_taskids = [dep.task_id for dep in missing_deps]
+            if missing_taskids:
+                deps = 'dependency' if len(missing_taskids) == 1 else 'dependencies'
+                raise RuntimeError('Unfulfilled %s at run time: %s' % (deps, ', '.join(missing_taskids)))
             self.task.trigger_event(Event.START, self.task)
             t0 = time.time()
             status = None
@@ -165,7 +170,7 @@ class TaskProcess(multiprocessing.Process):
             notifications.send_error_email(subject, formatted_error_message, self.task.owner_email)
         finally:
             self.result_queue.put(
-                (self.task.task_id, status, error_message, missing, new_deps))
+                (self.task.task_id, status, error_message, missing_taskids, new_deps))
 
     def _recursive_terminate(self):
         import psutil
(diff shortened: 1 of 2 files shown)


Tarrasch:
Eventually I would expect a test case and the commits to be merged. But I like the idea.

luigi/worker.py
```diff
@@ -117,14 +117,19 @@ def run(self):
...
+        missing_targets = []
         new_deps = []
         try:
             # Verify that all the tasks are fulfilled!
-            missing = [dep.task_id for dep in self.task.deps() if not dep.complete()]
-            if missing:
-                deps = 'dependency' if len(missing) == 1 else 'dependencies'
-                raise RuntimeError('Unfulfilled %s at run time: %s' % (deps, ', '.join(missing)))
+            missing_deps = [dep for dep in self.task.deps() if not dep.complete()]
+            missing_targets = [tgt for tgt in flatten(dep.output()) for dep in missing_deps if not tgt.exists()]
```
Tarrasch: Can you assume that there is a `output()` defined on all tasks? I don't think so, it's just a common practice but in general the interface is only the run-requires-complete triple iirc.

samuell:
Ah, yes! I'll have a look at that.

luigi/worker.py
samuell: Aha ... good point!

samuell:
Ah, yes (about test case, and merged commits)! I'm happy to fix this.

But we could of course discuss whether the change is at all feasible first :)

luigi/worker.py
samuell: Though, it seems tasks will always at least return an empty sequence / list:
https://github.com/spotify/luigi/blob/master/luigi/task.py#L385-L401

Tarrasch: Ah, that's right. I guess we can assume it to be an iterable of Targets.

Tarrasch:
By the way, why is this happening for you in the first case? Are you `complete()` methods inconsistent?

Another thought. Maybe it's only worthwhile printing out the individual missing targets is `output()` contains at least 2 elements, otherwise you can deduce yourself which the missing target is.

Tarrasch:

PR #1191, opened 2015-09-02
Test for orphaned subprocesses

ChrisBeaumont:
This adds the test requested in https://github.com/spotify/luigi/pull/1187#issuecomment-137130124 .

I've tweaked the implementation of `_recursive_terminate` to send SIGINT instead of SIGKILL (which allows processes to cleanup after themselves). While using this I also ran into a race condition where a subprocess finishes between the time terminate() is called and the signal is sent, so I added some try/catch blocks to handle this gracefully.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -172,14 +172,25 @@ class TaskProcess(multiprocessing.Process):
     def _recursive_terminate(self):
         import psutil
 
-        parent = psutil.Process(self.pid)
-        for child in parent.children(recursive=True):
-            child.kill()
+        try:
+            parent = psutil.Process(self.pid)
+            children = parent.children(recursive=True)
 
-        parent.kill()
+            # terminate parent. Give it a chance to clean up
+            super(TaskProcess, self).terminate()
+            parent.wait()
+
+            # terminate children
+            for child in children:
+                try:
+                    child.terminate()
+                except psutil.NoSuchProcess:
+                    continue
+        except psutil.NoSuchProcess:
+            return
 
     def terminate(self):
-        """Terminate this process and it's subprocesses."""
+        """Terminate this process and its subprocesses."""
         # default terminate() doesn't cleanup child processes, it orphans them.
         try:
             return self._recursive_terminate()
--- a/test/worker_task_test.py
+++ b/test/worker_task_test.py
@@ -15,10 +15,14 @@
 # limitations under the License.
 #
 import multiprocessing
+from subprocess import check_call
+import sys
 
 from helpers import unittest
-import mock
 import json
+import mock
+from psutil import Process
+from time import sleep
 
 import luigi
 import luigi.date_interval
@@ -50,6 +54,13 @@ class FailTask(luigi.Task):
 
     def on_failure(self, exception):
         return "test failure expl"
+
+
+class HangingSubprocessTask(luigi.Task):
+
+    def run(self):
+        python = sys.executable
+        check_call([python, '-c', 'while True: pass'])
 
 
 class WorkerTaskTest(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


luigi/worker.py
```diff
@@ -170,14 +170,25 @@ def run(self):
...
+            parent = psutil.Process(self.pid)
+            children = parent.children(recursive=True)
+
+            # terminate parent. Give it a chance to clean up
+            super(TaskProcess, self).terminate()
+            parent.wait()
 
-        parent.kill()
+            # terminate children
+            for child in children:
```
Tarrasch: Shouldn't this for loop be after (and outside) the try except?

luigi/worker.py
```diff
@@ -170,14 +170,25 @@ def run(self):
...
+                try:
+                    child.terminate()
+                except psutil.NoSuchProcess:
+                    continue
+        except psutil.NoSuchProcess:
+            return
```
Tarrasch: ... but then this should be `pass` and not `return` I suppose.

Tarrasch:
I'm not really that comfortable with signals and unix processes. But shouldn't you first try to terminate and then kill after a while if the process still doesn't terminate?

luigi/worker.py
ChrisBeaumont: The reasoning here is that if one child.terminate() fails, we should still try to terminate the others. So the try/except should be inside the loop

luigi/worker.py
ChrisBeaumont: This outer try/except catches the case where assigning parent or children fails (and, potentially, parent.wait() but i'm not sure). This most often happens if the parent finishes before we enter try block. In that case, there's nothing to do, so we exit the method

Tarrasch:
Ok, can you rebase this please? Sorry for the conflict. :)

ChrisBeaumont:
> shouldn't you first try to terminate and then kill after a while if the process still doesn't terminate

Yeah you could do that -- I switched from `kill()` to `terminate()` (with no eventual kill) because that was more consistent with the name of the method, and the behavior of the superclass. To my knowledge multiprocessing.Process sends a single SIGTERM signal and doesn't ever send SIGKILL. 

I could imagine situations where the cleanup after a terminate request legitimately takes a while (For example in python you can intercept SIGTERM and do as much work as you want), so choosing a proper timeout interval is tricky, and potentially something that should be specifiable as a task attribute and/or command line argument.  Maybe that's best left to a dedicated PR?

Tarrasch:
Yea, I'm fine with keeping it simple and only doing SIGTERM. You seem to know what you're doing. :)

ChrisBeaumont:
@Tarrasch ok, this is rebased and green!

Tarrasch:
Can you merge (amend) the commits? It's not nice useful to see "Rebase cleanup" in the commit log. :)

luigi/worker.py
```diff
@@ -172,14 +172,25 @@ def run(self):
...
-        parent = psutil.Process(self.pid)
-        for child in parent.children(recursive=True):
-            child.kill()
+        try:
+            parent = psutil.Process(self.pid)
+            children = parent.children(recursive=True)
+
+            # terminate parent. Give it a chance to clean up
+            super(TaskProcess, self).terminate()
+            parent.wait()
```
Tarrasch: Wait, is it really a good idea to block here? What if it's a process that never terminates even though we've "asked for it"? Won't the whole worker-loop block on this then? @ChrisBeaumont

ChrisBeaumont: I think my idea here is that we're giving the parent the opportunity to
clean up after itself, before taking the more blunt approach of terminating
children. I can't remember if I hit an actual issue without this step, or
if it was just a precaution. I can rerun my problematic pipeline without
the wait, to check for any bad behavior.

Tarrasch: Cool, thanks @ChrisBeaumont ! :)

PR #1182, opened 2015-08-31
Major command line parsing overhaul (internal)

Tarrasch:
See commit message.

PR #1193, opened 2015-09-03
Remove luigi.contrib.client, lazify fs operations

Tarrasch:
This patch makes two related changes. It removes luigi.contrib.client
and turns the functions luigi.contrib.{exists,remove,...} into
functions that creates a new autoconfigured hdfs client for each
invocation.

The problem with the luigi.contrib.client convenience object is that it
requires effects when you load the module. That is bad not only because
of the side effects, but it also means that it'll ignore command line
parameters. That is considered a bug. It'll mean that for example
`--hdfs-client hadoopcli` will not be honored. This is also why I
lazified the file system operations, so that the configs are read after
the command line parameters have been parsed (as opposed to at module
load time)


erikbern:
nice!

Tarrasch:
Sorry for being lazy and not adding tests, but this class of bugs can't be introduced once #1182 is merged.

PR #1197, opened 2015-09-03
tests: Start using LuigiTestCase

Tarrasch:
It's required for spotify/luigi#1182 since from that PR an on all tasks
need to be instantiated in a context where command line parameters have
already been parsed, and LuigiTestCase will provide an "empty" parsing
context.

### DIFF
--- a/test/_test_ftp.py
+++ b/test/_test_ftp.py
@@ -18,7 +18,7 @@
 import datetime
 import ftplib
 import os
-from helpers import unittest
+from helpers import unittest, LuigiTestCase
 try:
     from cStringIO import StringIO
 except ImportError:
@@ -36,7 +36,7 @@ USER = "luigi"
 PWD = "some_password"
 
 
-class TestFTPFilesystem(unittest.TestCase):
+class TestFTPFilesystem(LuigiTestCase):
 
     def setUp(self):
         """ Creates structure
@@ -89,7 +89,7 @@ class TestFTPFilesystem(unittest.TestCase):
         self.assertFalse("test" in list_dir)
 
 
-class TestFTPFilesystemUpload(unittest.TestCase):
+class TestFTPFilesystemUpload(LuigiTestCase):
 
     def test_single(self):
         """ Test upload file with creation of intermediate folders """
@@ -119,7 +119,7 @@ class TestFTPFilesystemUpload(unittest.TestCase):
         ftp.close()
 
 
-class TestRemoteTarget(unittest.TestCase):
+class TestRemoteTarget(LuigiTestCase):
 
     def test_put(self):
         """ Test RemoteTarget put method with uploading to an FTP """
--- a/test/contrib/hadoop_jar_test.py
+++ b/test/contrib/hadoop_jar_test.py
@@ -18,9 +18,9 @@
 import luigi
 import tempfile
 import shlex
-from helpers import unittest
 from luigi.contrib.hadoop_jar import HadoopJarJobError, HadoopJarJobTask
 from mock import patch
+from helpers import LuigiTestCase
 
 
 class TestHadoopJarJob(HadoopJarJobTask):
@@ -48,7 +48,7 @@ class TestRemoteHadoopJarTwoParamJob(TestRemoteHadoopJarJob):
     param2 = luigi.Parameter()
 
 
-class HadoopJarJobTaskTest(unittest.TestCase):
+class HadoopJarJobTaskTest(LuigiTestCase):
     @patch('luigi.contrib.hadoop.run_and_track_hadoop_job')
     def test_good(self, mock_job):
         mock_job.return_value = None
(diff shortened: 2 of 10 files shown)


Tarrasch:
Does this sound sane to you @erikbern?

Tarrasch:
By the way this is far from all the uses of `unittest.TestCase`, so I'll update this PR again soon. Is anyone up for the challenge of finding a clever way to automate this refactoring? I'm using vim :)

erikbern:
Strongly against this. You shouldn't have to require any global context to define a class.

erikbern:
I might be misunderstanding what you're doing though.

Tarrasch:
I want to enforce you to have a global parsing context (see #1182) to *instantiate* luigi classes (i.e. `MyTask()`). The reason? You already do in some sense now (since we set the `__global` field of Parameters during parsing). Only now that it will work regardless if you've have done the parsing or not.

Imagine this code:

```

class MyTask(luigi.Task):
    my_param = Parameter(default='hi')

print(MyTask().my_param))

class OtherTask(luigi.Task):
     run(self):
          print(MyTask().my_param)
```

Invoked something like this `luigi OtherTask --MyTask-my-param hello`. What output do you expect? Currently you'll see something like:

```
hi
hello
```

To me that's a major issue with luigi, it's causing bugs like the one just fixed in #1193.

erikbern:
I would expect the output you provided.

What if you don't use the command line interface to run Luigi? I think it's important to be able to run Luigi without using the cmdline interface

Tarrasch:
You'll still be able to use `luigi.build` if you want to use luigi programatically over the `luigi` binary. Note that `luigi.build` already does "magic" like `--no-lock`-ing it for you. So in that sense you can still use luigi without the cli.

But isn't it bothering you that `MyTask().my_param` has different values? In fact it's worse, if `MyTask` has only hashable params, it'll \*surprise\*  remain as "hi", but if it contains (or inherits) a non-hashable parameter, it'll be "hello".

erikbern:
But are you solving this by disallowing you to instantiate objects before the parameters are set? If so, then you're not solving the problem really

Sorry for grumpy

Tarrasch:
> But are you solving this by disallowing you to instantiate objects before the parameters are set? If so, then you're not solving the problem really

Why isn't that solving the problem?

Python has no type-checke. But you'll get something quite similar to static checking by not allowing instantiations of objects in an "indeterministic" manner.

Also, to be clear, you can still instantiate tasks if you pass *all* parameters (as args+kwargs). It's only in the case where you fall back on the "cli-->config-->default" chain that a surprising behavior can happen.

Tarrasch:
And, again, in 90% cases users won't be affected. In the other 9% I think they'll be happy that the fraework found a bug for them.

The other motivation to do this is that the parameter.py and interface.py code is just horrible. These changes will greatly reduce complexity and code size.

erikbern:
Should be possible to clean up parameter.py and interface.py without disallowing this kind of stuff though?

but maybe using `luigi.build` is enough for this case

Tarrasch:
1. I'm more than happy to hear a design proposal of how to implement the Parameter - commandlline relation

 * We currently put state in all Parameter objects. That is just asking for subtle bugs.
 * My current implementation tries to narrow state to only one singleton. You also get the (imo) benefit of that you can verify that the singleton is alive before instantiating tasks.
 * Is there any implementation options I have overlooked?

2. Can you think of any use case where you want to instantiate Tasks before `luigi.run` or `luigi.build`? If not, then I think we shouldn't allow it as it is probably a user bug then.

3. The plan is to extend command line parsing after the overhaul. In particular we at Spotify have seen the need for luigi to semantically understand task parameters. For example, we would like `Range --of MyTask --my-param 123` to work. Currently we've tragically been forced to do whatever ugly metamagic and annotate "global" on parameters.

Tarrasch:
@erikbern if you veto against this. I could still go on and overhaul the underlying implementation and still permit that you instantiate task without the global ArgumentParsing instance existing. I'll do that to unblock myself. Closing for now, but please think about the bullet points I wrote.

erikbern:
For (2): `luigi.build` takes a task as an argument, so yes you need to instantiate first.

I would prefer it if you just ignore global command line arguments if they haven't been set yet. That's how the current implementation works. Does that add a lot of complexity?

Tarrasch:
> For (2): luigi.build takes a task as an argument, so yes you need to instantiate first.

Yea, you're right actually. There's no way to lazify that computation huh.

> I would prefer it if you just ignore global command line arguments if they haven't been set yet. That's how the current implementation works. Does that add a lot of complexity?

Not really, but it was one of the big benefits of the refactoring. But that's fine, let me push for the refactoring and we can then discuss if we want the sanity-checking feature or not.

erikbern:
Cool great if you can still do the refactoring. But yeah great if we don't rely on the command line being instantiated

PR #1204, opened 2015-09-10
tests: Various fixes for external task tests

Tarrasch:
I found a lot of errors and irrelevant code when I started to
investigate why these tests were failing for me in spotify/luigi#1182.
This patch basically splits an existing test into two tests and also
clarifies how keep_alive and retry_external_tasks differ.

### DIFF
--- a/test/worker_external_task_test.py
+++ b/test/worker_external_task_test.py
@@ -74,20 +74,12 @@ class TestTask(luigi.Task):
 
 class WorkerExternalTaskTest(unittest.TestCase):
 
-    def setUp(self):
-        self.scheduler = CentralPlannerScheduler(retry_delay=0.01,
-                                                 remove_delay=3,
-                                                 worker_disconnect_delay=3,
-                                                 disable_persist=3,
-                                                 disable_window=5,
-                                                 disable_failures=2,
-                                                 prune_on_get_work=True)
-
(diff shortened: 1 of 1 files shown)


Tarrasch:
Since you wrote this stuff, can you look at this @boosh? (Also, you should associate you email with your github account, so there's no question mark next to your name. see https://github.com/Tarrasch/luigi/commit/825b91bb8eff5c139f0aa43e959be2d6a46eace1)

Tarrasch:
Also, this tests started to hang for me for python-34, so I have no hopes of the tests to pass on Travis. But I just want to confirm it's not a local issue.

erikbern:
Separately it kind of annoys me how ExternalTasks are "special". I tried to clean up that code in the worker but never finished. Would be good to do that at some point

Tarrasch:
@erikbern not sure how you imagine them to not be special. Doesn't seem trivial/possible to me at first thought. I'm happy to be convinced otherwise. :)

Tarrasch:
Excellent, the builds are failing. By actually fixing whatever is wrong here I hope I can unblock #1182. Damn this bug has killed my productivity :-/

erikbern:
Should just be a subclass of Task and we shouldn't have any code doing `isinstance(ExternalTask, task)`. But maybe a separate discussion

Issue #1201, opened 2015-09-09
Add an option to prevent Worker from catching exceptions from Task.run()


belevtsoff:
In our project this problem arises when unit testing tasks with dynamic dependencies (i.e. their `run()` method yields a generator). Because `luigi` has it's own mechanism of dealing with such cases, we use `worker.run()` to test a task. However, if one wants to test that a task raises under certain conditions, `worker.run()` catches that exception and logs, making it problematic to make sure a right kind of error was raised. It would be great to have smth like `Worker(..., raise_on_error=False)` kw argument of the `Worker`'s constructor.

If there are no objections, I can put together a PR

Tarrasch:
We would happily accept such a PR. :)

I'm also interested to hear how you use dynamic dependencies.

erikbern:
You mean the default would be to catch the exception and log it, but it's possible to override so that the Worker re-raises the exception? Yeah makes sense for unit tests

erikbern:
it might be a bit messy get it working with the multiprocessing RPC but not too hard

belevtsoff:
@Tarrasch cool. We have a kind of luigi pipeline generator that uses information from a kind of `nx.DiGraph` and then puts together a corresponding luigi pipeline. That's a static part. During the process of putting together luigi tasks, the generator might find that two nodes in the graph are pointing to the same target (a) -> (c) <- (b). In this case, the data from those deps has to be joined in a certain way, and that requires pulling additional data from a source:
(a) -> (c) <- (b)
(src)   _/
 Now, we don't want users to create the (src)->(c) link manually, but be do want the process of pulling the additional data to appear as a proper luigi.Task dependency on the final pipeline. That's where dynamic deps are used: when such situation arises, (c).run() pulls (src) dep dynamically

belevtsoff:
@erikbern hm, I might have underestimated the amount of work required... But I guess someone will help me on the PR-review board

erikbern:
Yeah the Worker code is pretty messy right now, for a reason – turns out there's a million ways things can go wrong (exceptions, signals, etc) and you need to deal with them in different ways. But I think you can essentially just pickle the exception and put it onto the queue back and then unpickle and raise.

belevtsoff:
Closed by #1205

daveFNbuck:
#1205 just got reverted. It turns out that custom exceptions often can't be unpickled due to a bug in python 2, so it was breaking my pipelines in a big way

PR #1205, opened 2015-09-10
Add an option to prevent Worker from catching exceptions from Task.run()

belevtsoff:
Addresses #1201 

Here I used the fact that `multiprocessing.Queue` does basic serialization itself, which should work well for exceptions. I also wanted to create a container class TaskProcessResult to formalize a bit the growing set of attributes that is passed over the queue and read in several places, but I have no experience with mock and didn't know how to adjust the tests. Do you think that would be an overkill?

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -119,6 +119,8 @@ class TaskProcess(multiprocessing.Process):
         expl = ''
         missing = []
         new_deps = []
+        exception = None
+
         try:
             # Verify that all the tasks are fulfilled!
             missing = [dep.task_id for dep in self.task.deps() if not dep.complete()]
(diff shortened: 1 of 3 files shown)


erikbern:
LGTM

luigi/worker.py
```diff
@@ -700,6 +709,10 @@ def _handle_next_task(self):
                 continue
             self._running_tasks.pop(task_id)
 
+            # raise exception if needed
```
Tarrasch: I would just remove this comment. The code is definitely readable enough. :)

Tarrasch:
LGTM. May I just ask that you merge the commits?

belevtsoff:
@Tarrasch do you mean you'd like to have a single commit for the whole pull request?

luigi/worker.py
belevtsoff: fair enough :)

Tarrasch:
@belevtsoff Yea, I think that keeps the commit log clener. And when you git blame the test lines, you'll also see the implementation that came with it, and vice verse.

belevtsoff:
@Tarrasch done

Tarrasch:
Thanks @belevtsoff!

PR #1211, opened 2015-09-12
Fixing escape for null sequence in postgres COPY FROM

moritzschaefer:
copy_from (and as such the whole CopyTable class in postgres) doesn't work because \N isn't recognized as null sequence. Adding a second backslash fixes this. 
It should be possible to choose an own null-sequence anyway when using CopyTable..

### DIFF
--- a/luigi/postgres.py
+++ b/luigi/postgres.py
@@ -258,7 +258,7 @@ class CopyToTable(rdbms.CopyToTable):
         Default behaviour is to escape special characters and identify any self.null_values.
         """
         if value in self.null_values:
-            return r'\N'
+            return r'\\N'
         else:
             return default_escape(six.text_type(value))
 
@@ -286,7 +286,7 @@ class CopyToTable(rdbms.CopyToTable):
             column_names = [c[0] for c in self.columns]
         else:
             raise Exception('columns must consist of column strings or (column string, type string) tuples (was %r ...)' % (self.columns[0],))
-        cursor.copy_from(file, self.table, null=r'\N', sep=self.column_separator, columns=column_names)
+        cursor.copy_from(file, self.table, null=r'\\N', sep=self.column_separator, columns=column_names)
 
     def run(self):
         """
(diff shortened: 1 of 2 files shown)


Tarrasch:
Thanks for you PR. We are happy to merge this once you've added a test case. :)

Also, it would be nice to associate your email with your GitHub account.

Tarrasch:
By the way, since we're kind of copying code with the null character like that. Would it be better to extract it out to a member like `default_null_value`?

moritzschaefer:
I would just test, that the default argument is \\N. How would you test this fix? The default_null_value is better I think as well because we are using \\N twice..

Tarrasch:
I don't understand this change (I don't know postgres). But I would say that your test should be alike the program that made you discover this bug.

moritzschaefer:
The problem is, that Postgres detects the error: luigi and psycopg2 generate a Query for postgres which is sent to the server and the server complains about it, so it's hard to test it like this. 
In fact one could mock the psycopg2 function that is called and check if it is called with the correct escape sequence. Is this appropriate for this bug?

Tarrasch:
I don't understand why you can't check it against postgres without mocking. There already are tests running against a postgres instance. Like `tox -e py27-postgres`. See these resources:

https://github.com/spotify/luigi/blob/06f177b83d92653a76fa822b64bee96c6b54f25a/tox.ini#L2

https://github.com/spotify/luigi/blob/06f177b83d92653a76fa822b64bee96c6b54f25a/.travis.yml#L46

moritzschaefer:
Hey,

thanks for hint!

Arash Rouhani <notifications@github.com> schrieb am Di., 15. Sep. 2015 um
08:51 Uhr:

moritzschaefer:
ok, now the new(/modified) test passes with the fix and fails without the fix.

Tarrasch:
> now the new(/modified) test passes with the fix and fails without the fix.

Perfect, that's exactly what I wanted! :)

Thanks for this PR. But can you please just squash these 3 commits into one to keep the history clean?

moritzschaefer:
Done :)

Tarrasch:
Nice!

PR #1216, opened 2015-09-14
None (null) becomes 'None' (string) when yielding a task.

jeremykarn:
When you yield a task with a parameter value of None (null) the new task has the parameter value of 'None' (string).  

If you return a task in the requires method with a parameter of None (null) that gets passed as None (null), which seems like the correct behaviour.

### DIFF
--- a/test/parameter_test.py
+++ b/test/parameter_test.py
@@ -25,6 +25,7 @@ import luigi.interface
 import luigi.notifications
 from luigi.mock import MockTarget
 from luigi.parameter import ParameterException
+from luigi.worker import Worker
 from worker_test import email_patch
 
 luigi.notifications.DEBUG = True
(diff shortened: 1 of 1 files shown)


jeremykarn:
This is just a test case showing the error.  I'm not sure exactly what/where the fix should be.

erikbern:
https://github.com/spotify/luigi/blob/master/luigi/worker.py#L104

It's a bit tricky. The idea with `to_str_params` and `from_str_params` is that the type is defined by the parameter class (i.e. `IntParameter` etc). But `None` is not the same type as the parameter. You would see the same error if you passed a string to an `IntParameter` etc.

Not sure how to solve. Probably the only way to solve it is to avoid using `to_str_params` and `from_str_params` and use some other serialization mechanism. Alternatively I wonder if we should just enforce type adherence and actually throw an error if types are not matching (eg `to_str_params` would only serialize strings and nothing else)

test/parameter_test.py
```diff
@@ -688,3 +689,45 @@ class OtherTask(luigi.Task):
         self.assertEqual(MetaTask.saved_value, MetaTask)
         self.assertTrue(luigi.run(['--local-scheduler', '--no-lock'] + 'mynamespace.MetaTask --a other_namespace.OtherTask'.split()))
         self.assertEqual(MetaTask.saved_value, OtherTask)
+
+
+class ForNone(luigi.Task):
+    p = luigi.Parameter(default=None)
+    has_run = False
```
Tarrasch: If you want to you can use this test-helper: https://github.com/spotify/luigi/blob/master/test/helpers.py#L123

Tarrasch:
I'm not sure if it makes sense for `Parameter` to "be"  `None`. All parameters must be serialize-able and parse-able back and forth.

@erikbern, I don't think the task-level `Task.to_str_params` and `Task.from_str_params` are the culprit here. The whole logic fails already at parameter level with the `Parameter.serialize` and `Parameter.parse` methods.

erikbern:
ok yeah i think you're right. those are the guilty ones. we should probably catch type mismatch there

jeremykarn:
Being able to set a parameter to None has been pretty useful.  It looks like its used a few times in the included Luigi tasks as well.

Should there be a special type of nullable (Optional?) parameter?

test/parameter_test.py
jeremykarn: Thanks, I should have guessed that existed somewhere.

erikbern:
I'm torn about to what extent we should re-invent a type system. Maybe nullability should be allowed, maybe not. I can't decide :)

Tarrasch:

Also closing because this is a wontfix. You're (we should probably document this) not allowed to pass `None` to string-parameters now.

PR #1219, opened 2015-09-15
Parameterize GCSClient chunksize for better control when uploading or downloading data

### LINKED ISSUE #1218: GCSClient.download()/.put() is unusably slow for large files
**TL;DR**: It takes 20 _MINUTES_ to download a `1.7 GB` files with `GCSClient()` but `gsutil` does it in about 20 seconds.  Working on a PR to parameterize `chunk_size` and suggest the default be changed to `10 MB`.  Not sure why the Google docs recommend such a small chunk size.

`GCSClient.download()` downloads files in [`1 MB` chunks](https://github.com/spotify/luigi/blob/672856d3b01428681c390773f5e15346c49f3ec5/luigi/contrib/gcs.py#L336), which is what the [Google Cloud Storage API docs use](https://cloud.google.com/storage/docs/json_api/v1/objects/get).  Select `python` from the menu under `Examples`.  `GCSClient.put()` uses a [`2 MB`](https://github.com/spotify/luigi/blob/672856d3b01428681c390773f5e15346c49f3ec5/luigi/contrib/gcs.py#L245) [chunk size](https://github.com/spotify/luigi/blob/672856d3b01428681c390773f5e15346c49f3ec5/luigi/contrib/gcs.py#L55).

Obviously this works but is incredibly slow for large files as each API call costs about `0.5` seconds.  I ran some download tests against `gsutil` on a [`n1-standard-2`](https://cloud.google.com/compute/docs/machine-types) machine with a `50GB` boot disk in zone `us-central-a` with a file that is `1.7 GB`.  Did not test upload but it appears to be about as slow.

| Chunk Size | Time (m:s) |
|:---------------:|:-------:|
| gsutil | 00:21 |
| 1 mb | 20:04 |
| 10 mb | 01:01 |
| 100 mb | 00:44 |

Test code:

```python
import json
import httplib2
import oauth2client.client

from googleapiclient import errors
from googleapiclient import discovery
from googleapiclient import http
import io
from apiclient.http import MediaIoBaseDownload

oauth_credentials = oauth2client.client.GoogleCredentials.get_application_default()

client = discovery.build('storage', 'v1', credentials=oauth_credentials, http=httplib2.Http())

bucket_name='{bucket}'
object_name='2015-01-01.nmea'

# Get Metadata
req = client.objects().get(
        bucket=bucket_name,
        object=object_name)
resp = req.execute()
#print json.dumps(resp, indent=2)

# Get Payload Data
req = client.objects().get_media(
        bucket=bucket_name,
        object=object_name)
# The BytesIO object may be replaced with any io.Base instance.
fh = io.FileIO(object_name, 'w')
downloader = MediaIoBaseDownload(fh, req, chunksize=10*1024*1024)
done = False
while not done:
    status, done = downloader.next_chunk()
```

geowurster:
Closes #1218 

I upped the default `chunksize` to `10 mb` and parameterized as `GCSClient(chunksize=10 * 1024 * 1024)`.

Creating `GCSClient()` instances can be expensive if you hit the API at the wrong time, so the user can also override the chunksize when calling the `download()` or `put()` methods directly rather than instantiating another instance of `GCSClient()` with a different chunk size.

@erikbern ready for review.  I didn't see a docstring convention mentioned anywhere so I wasn't sure how to format.

### DIFF
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -97,6 +97,9 @@ class GCSClient(luigi.target.FileSystem):
              scope='https://www.googleapis.com/auth/devstorage.read_write')
          client = GCSClient(oauth_credentials=credentails)
 
+        The chunksize parameter specifies how much data to transfer when downloading
+        or uploading files.
+
     .. warning::
       By default this class will use "automated service discovery" which will require
       a connection to the web. The google api client downloads a JSON file to "create" the
@@ -104,7 +107,9 @@ class GCSClient(luigi.target.FileSystem):
       contents of this file (currently found at https://www.googleapis.com/discovery/v1/apis/storage/v1/rest )
       as the ``descriptor`` argument.
     """
-    def __init__(self, oauth_credentials=None, descriptor='', http_=None):
+    def __init__(self, oauth_credentials=None, descriptor='', http_=None,
+                 chunksize=10 * 1024 * 1024):
+        self.chunksize = chunksize
         http_ = http_ or httplib2.Http()
 
         if not oauth_credentials:
@@ -238,11 +243,12 @@ class GCSClient(luigi.target.FileSystem):
 
         return False
 
-    def put(self, filename, dest_path, mimetype=None):
+    def put(self, filename, dest_path, mimetype=None, chunksize=None):
+        chunksize = chunksize or self.chunksize
         resumable = os.path.getsize(filename) > 0
 
         mimetype = mimetype or mimetypes.guess_type(dest_path)[0] or DEFAULT_MIMETYPE
-        media = http.MediaFileUpload(filename, mimetype, chunksize=CHUNKSIZE, resumable=resumable)
+        media = http.MediaFileUpload(filename, mimetype, chunksize=chunksize, resumable=resumable)
 
         self._do_put(media, dest_path)
 
@@ -320,7 +326,8 @@ class GCSClient(luigi.target.FileSystem):
         for it in self._list_iter(bucket, obj_prefix):
             yield self._add_path_delimiter(path) + it['name'][obj_prefix_len:]
 
-    def download(self, path):
+    def download(self, path, chunksize=None):
+        chunksize = chunksize or self.chunksize
         bucket, obj = self._path_to_bucket_and_key(path)
 
         with tempfile.NamedTemporaryFile(delete=False) as fp:
@@ -333,7 +340,7 @@ class GCSClient(luigi.target.FileSystem):
                 return return_fp
 
             request = self.client.objects().get_media(bucket=bucket, object=obj)
-            downloader = http.MediaIoBaseDownload(fp, request, chunksize=1024 * 1024)
+            downloader = http.MediaIoBaseDownload(fp, request, chunksize=chunksize)
 
             attempts = 0
             done = False
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -52,7 +52,7 @@ else:
 NUM_RETRIES = 5
 
 # Number of bytes to send/receive in each request.
-CHUNKSIZE = 2 * 1024 * 1024
+CHUNKSIZE = 10 * 1024 * 1024
 
 # Mimetype to use if one can't be guessed from the file extension.
 DEFAULT_MIMETYPE = 'application/octet-stream'
@@ -108,7 +108,7 @@ class GCSClient(luigi.target.FileSystem):
       as the ``descriptor`` argument.
     """
     def __init__(self, oauth_credentials=None, descriptor='', http_=None,
-                 chunksize=10 * 1024 * 1024):
+                 chunksize=CHUNKSIZE):
         self.chunksize = chunksize
         http_ = http_ or httplib2.Http()
 


luigi/contrib/gcs.py
```diff
@@ -104,7 +107,9 @@ class GCSClient(luigi.target.FileSystem):
       contents of this file (currently found at https://www.googleapis.com/discovery/v1/apis/storage/v1/rest )
       as the ``descriptor`` argument.
     """
-    def __init__(self, oauth_credentials=None, descriptor='', http_=None):
+    def __init__(self, oauth_credentials=None, descriptor='', http_=None,
+                 chunksize=10 * 1024 * 1024):
```
erikbern: you should default this to CHUNKSIZE and set CHUNKSIZE to 10*1024*1024

Tarrasch:
Do @mikekap want to quickly review this too? This looks good to me. Just see Erik's comment. :)

geowurster:
@erikbern @Tarrasch Moved default chunksize to global variable.

erikbern:
LGTM

I realize maybe we should have made chunksize configurable but it's no biggie. let's merge

mikekap:
LGTM as well.

A small note - I didn't expect making a GCSClient to be a heavy op. I don't use the default client myself (to pass in a local descriptor, I have a globally memorized one). Perhaps if the default construction is that expensive, we should memorize it in the BigqueryTarget constructor?

geowurster:
@mikekap It's only heavy because of the [API discovery](https://github.com/spotify/luigi/blob/master/luigi/contrib/gcs.py#L113).  If you hit the Google API at a time with high latency then it can be expensive to build a `GCSClient()`.  Instantiating a client with every target creates more opportunities for encountering latency.

A generic luigi solution might be to just instantiate a `GCSClient()` in a `gcs.GCS_CLIENT` global variable and have the `GCSTarget()` grab that one rather than creating a new instance if a client is not supplied.

Our solution was to attach a singleton `GCSClient()` to a subclassed `luigi.task.Config` object paired with a `DynamicTarget()` that always gives us GCS targets with the instantiated client or a more appropriate target based on the `schema://`.  It also gave us the opportunity to create a `GCSClientStub()` that translates GCS paths to local paths for running locally and testing.  We have not started integrating `BigqueryTargets()` but will before too long and will report back if we find some latency.  The `BenthospConfig()` object always references a single client in a class variable no matter how many times it is instantiated.

```python
def DynamicTarget(path, **kwargs):

    """
    A single interface for dynamically retrieving luigi target objects based on
    the input path's schema.

        file://path = LocalTarget
        gs://path = GCSTarget

    The ``BenthospConfig()`` object is used to determine if a stubbed client
    should be used.

    Parameter
    ---------
    path : str
        Behaves just like ``path`` for any other target, except it requires a
        schema.
    kwargs : **kwargs, optional
        Additional keyword arguments for the desired target.
    """

    cfg = settings.BenthospConfig()

    if path.startswith('gs://'):
        kwargs.update(client=cfg.gcs_client)
        return gcs.GCSTarget(path, **kwargs)

    elif path.startswith('file://'):
        return luigi.LocalTarget(path[7:], **kwargs)

    else:
        raise ValueError("Unrecognized schema: {}".format(path))
```

```python
class GCSClientStub(object):

    """
    Drop-in replacement for luigi.contrib.gcs.GCSClient
    Maps a given GCS path to a given local file path so that any request using a gcs path
    will be executed on the local file system
    """

    def __init__(self, local_root=os.getcwd(), **kwargs):

        """
        Create a special GCS client that operates locally.
        Parameters
        ----------
        local_root : str, optional
            Consider this location to be the root of GCS.  Any sub-directories
            will be considered buckets.  Creating a bucket creates a sub-dir in
            this location.
        kwargs : **kwargs, (does nothing)
            Quietly eat arguments that would be passed to ``GCSClient()`.
        """

        local_root = os.path.abspath(local_root)
        if not os.path.exists(local_root):
            os.makedirs(local_root)

        self._local_root = local_root

        self.fs = LocalFileSystem()

    def _gcs2local(self, path):

        """
        Convert a GCS path to a local path
        """

        split = urlparse.urlsplit(path)
        bucket = posixpath.join(self._local_root, split.netloc)
        path = posixpath.join(bucket, split.path[1:])  # Slice off leading slash
        if not os.path.exists(bucket):
            os.makedirs(bucket)
        return path

    def _local2gcs(self, path):
        path = path.replace(self._local_root, '', 1)[1:]  # Slice off leading slash
        return 'gs://' + path

    def _path_to_bucket_and_key(self, path):
        (scheme, netloc, path, _, _) = urlsplit(path)
        assert scheme == 'gs'
        path_without_initial_slash = path[1:]
        return netloc, path_without_initial_slash


    def copy(self, source_path, destination_path):
        source_path = self._gcs2local(source_path)
        destination_path = self._gcs2local(destination_path)
        LocalTarget(source_path).copy(destination_path)

    def download(self, path):
        # GCSClient() downloads the file to a temp location and returns an open
        # file pointer, so we do the same thing here

        path = self._gcs2local(path)
        assert os.path.exists(path), "File does not exist: {}".format(path)
        with tempfile.NamedTemporaryFile(delete=False) as fp:
            LocalTarget(path).copy(fp.name)
            return gcs._DeleteOnCloseFile(fp.name, 'r')

    def exists(self, path):
        return self.fs.exists(self._gcs2local(path))

    def isdir(self, path):
        return self.fs.isdir((self._gcs2local(path)))

    def listdir(self, path):
        lpaths = self.fs.listdir(self._gcs2local(path))
        return (self._local2gcs(p)for p in lpaths)

    def mkdir(self, path, **kwargs):
        return self.fs.mkdir(self._gcs2local(path), **kwargs)

    def put(self, filename, dest_path, **kwargs):
        LocalTarget(filename).copy(self._gcs2local(dest_path))

    def put_string(self, contents, dest_path, **kwargs):
        dest_path = self._gcs2local(dest_path)
        bname = os.path.basename(dest_path)
        if not os.path.exists(bname):
            os.makedirs(bname)
        with LocalTarget(dest_path).open('w') as f:
            f.write(contents)

    def remove(self, path, **kwargs):
        # GCSClient does not fail on remove() if the path does not exist, but LocalFileSystem does,
        # so check to make sure it exists before attempting to remove
        if self.exists(path):
            return self.fs.remove(self._gcs2local(path), **kwargs)
        else:
            return False

    def rename(self, source_path, destination_path):
        source_path = self._gcs2local(source_path)
        destination_path = self._gcs2local(destination_path)
        LocalTarget(source_path).move(destination_path)
```

Tarrasch:
Thanks for the patch. I guess when users find that the value should be parametrized they'll add it :)

PR #1220, opened 2015-09-16
tests: Fix intermittent failures

Tarrasch:
The actually tests that are fixed are those (that are not even disabled)
in remote_scheduler_test.py. But the errors are easily reproduceable by
increasing the number of DummyTasks and decreasing the default value of
ping_interval.

### DIFF
--- a/test/server_test.py
+++ b/test/server_test.py
@@ -21,7 +21,6 @@ import shutil
 import signal
 import time
 import tempfile
-
 from helpers import unittest, skipOnTravis
 import luigi.rpc
 import luigi.server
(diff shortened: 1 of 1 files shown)


Tarrasch:
Want to review guys? @erikbern @gpoulin

erikbern:
looks good to me. we could always replace the KeepAliveThread with some tornado-specific thing.

Tarrasch:
> we could always replace the KeepAliveThread with some tornado-specific thing.

I'm not sure I understand what you're suggesting. Do you mean that we change in the testing/mocking code or that we change the KeepAliveThread class in `worker.py`?

Tarrasch:
Merging, in hope of that this is what caused the extremely surprising errors in #1182

erikbern:
was suggesting we change the keepalivethread to use tornado threads during tests

PR #1222, opened 2015-09-16
Allow gcs._DeleteOnCloseFile().close() to be called multiple times

### LINKED ISSUE #1221: gcs._DeleteOnCloseFile().close() cannot be called more than once
The issue is that [`close()`](https://github.com/spotify/luigi/blob/master/luigi/contrib/gcs.py#L371-L373) does not to check to see if the file exists before trying to delete, so an exception is raised if the file doesn't exist.

geowurster:
Closes #1221 

Could someone point me towards a unittest that I can use as an example for how to structure a test?  Where should I put a test?

### DIFF
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -370,7 +370,12 @@ class GCSClient(luigi.target.FileSystem):
 class _DeleteOnCloseFile(io.FileIO):
     def close(self):
         super(_DeleteOnCloseFile, self).close()
-        os.remove(self.name)
+        try:
+            os.remove(self.name)
+        except OSError:
+            # Catch a potential threading race condition and also allow this
+            # method to be called multiple times.
+            pass
 
     def readable(self):
         return True
--- a/test/contrib/gcs_test.py
+++ b/test/contrib/gcs_test.py
@@ -159,3 +159,19 @@ class GCSTargetTest(_GCSBaseTestCase, FileSystemTargetTestMixin):
 
     def create_target(self, format=None):
         return gcs.GCSTarget(bucket_url(self.id()), format=format, client=self.client)
+
+    def test_close_twice(self):
+        # Ensure gcs._DeleteOnCloseFile().close() can be called multiple times
+        tgt = self.create_target()
+
+        with tgt.open('w') as dst:
+            dst.write('data')
+        assert dst.closed
+        dst.close()
+        assert dst.closed
+
+        with tgt.open() as src:
+            assert src.read().strip() == 'data'
+        assert src.closed
+        src.close()
+        assert src.closed


Tarrasch:
Sorry about the intermittent fail from Travis. The build should be green.

Anyhow, this implementation doesn't look to follow the LBYL principle that python usually follows. Can't you try to catch the Exception type raised by a file not existing instead?

geowurster:
@Tarrasch We have a domain specific I/O library that behaves like Python's `open` and `file`, so I can't catch the exception on close, because I might be suppressing other valid exceptions.

Your comment reminded me that I ended up in this situation because I thought a luigi file-like object's `.close()` was not being called in `__exit__()`, so I have been doing:

```python
with target.open('w') as f, io_library.open(f, 'w', driver='name') as src:
    for line in src:
        # do something
```

So `.close()` was being called explicitly once when my I/O library hits `__exit__()` and again in the target's `__exit__()`.  I verified that I can just pass `target.open()` to my I/O library so this is no longer an issue, but Python's `io.FileIO()`, `gzip.GzipFile()`, and `bz2.BZ2File()` libraries all allow `.close()` to be closed multiple times, so this seems like an acceptable pattern.

What about the code below?  More sane logic based on whether the file is already closed rather than checking to see if the file exists.

```python
    def close(self):
        if not self.closed:
            super(_DeleteOnCloseFile, self).close()
            os.remove(self.name)
```

sisidra:
@Tarrasch it is common in most languages to specify `close()` as callable multiple times. 
For example in Java: http://docs.oracle.com/javase/7/docs/api/java/io/Closeable.html#close()
Python: https://docs.python.org/2/library/stdtypes.html?highlight=close#file.close
C#: https://msdn.microsoft.com/en-us/library/system.idisposable.dispose(v=vs.110).aspx
etc :)

Tarrasch:
Sorry I'm a bit unclear on what I wanted. I'm totally agreeing that `close()` should be callable twice. However, the implementation in this PR is not following the "don't LBYL" principle.

So change

```
        if os.path.exists(self.name):
            os.remove(self.name)
```

to something like

```
        try:
            os.remove(self.name)
        except GCPFileDoesNotExist:
            pass
```

LBYL is described here: https://docs.python.org/2/glossary.html. I just made `GCPFileDoesNotExist` up. I'm fine with the current implementation if there's no defined exception subclass for "FileDoesNotExist", but if there is, let's do it the python way.

geowurster:
@Tarrasch Agreed - here's my suggested revision from above.  The logic only gets executed if the file is not closed.

```python
    def close(self):
        if not self.closed:
            super(_DeleteOnCloseFile, self).close()
            os.remove(self.name)
```

Tarrasch:
Does that not have the race-like problem? But sure. That sounds good if there's not `GCPFileDoesNotExistException`, have you verified that there is not?

geowurster:
@Tarrasch I'm not following your comment about a "race-like condition" or where `GCPFileDoesNotExistException` fits into this.

Checking to see if the object is already closed and only attempting to remove the file if the object is still open seems to be the most Pythonic and still follows LBYL.

Tarrasch:
By race I mean that two thread first check `if not self.closed:` condition. And then they both run `os.remove(self.name)` and then the second thread to do it fails.

geowurster:
@Tarrasch Ah ok your point about threading makes sense - is that something that we need to handle here?  You are more familiar with Luigi's internals so I defer to your judgement.

When you say `GCPFileDoesNotExist`, this is a local file so `os.remove()` just raises an `OSError` if it doesn't exist, unless I'm missing something.

Tarrasch:
Well, it's better be safe than sorry I think. Oh, I didn't see it was implemented with a local file. Yes, then please use `OSError`. Thanks! :)

geowurster:
@Tarrasch Updated.  Any guidance for a unittest, or are you happy without one?

Tarrasch:
I usually suggest you should add a test case that's similar to how you encountered the bug. Would that be possible?

geowurster:
Yeah I meant more information about where luigi's test data lives in GCS.
To replicate all I need to be able to do is open a test file in GCS with
GCSTarget(), but I don't which bucket to pull from. Is there a similar test
I can look at?

Tarrasch:
Didn't you see https://github.com/spotify/luigi/blob/master/test/contrib/gcs_test.py?

geowurster:
@Tarrasch Thanks.  Updated with a test and squashed my two commits into one.

luigi/contrib/gcs.py
```diff
@@ -370,7 +370,12 @@ def download(self, path, chunksize=None):
 class _DeleteOnCloseFile(io.FileIO):
     def close(self):
         super(_DeleteOnCloseFile, self).close()
-        os.remove(self.name)
+        if not self.closed:
```
Tarrasch: Do you still need this? Who sets this property by the way? If it's inherited, won't it be always set to `True` when you call the parent's `close()` like you do on the line above?

geowurster: @Tarrasch Good point.  It's a standard property on `io.FileIO` that is always set to `True` or `False` depending on the state of the object.  I switched to just a `try / except`.

Tarrasch:
@geowurster, since they are in the same PR, it makes sense to merge the commits. :)

geowurster:
@Tarrasch Yep forgot to do it a second time.  Too much multitasking.

Tarrasch:
There's a flake8 error. Sorry you have to rebase again :p

geowurster:
@Tarrasch Fixed it.

test/contrib/gcs_test.py
```diff
@@ -159,3 +159,19 @@ class GCSTargetTest(_GCSBaseTestCase, FileSystemTargetTestMixin):
 
     def create_target(self, format=None):
         return gcs.GCSTarget(bucket_url(self.id()), format=format, client=self.client)
+
+    def test_close_twice(self):
+        # Ensure gcs._DeleteOnCloseFile().close() can be called multiple times
```
Tarrasch: You don't need to fix this (you've done so many iterations). But for next time, you can make this into a docstring so it'll have a nicer descriptive name when the tests are running.

geowurster: Will do.  I noticed none of the other tests had docstrings so I didn't use one.

Tarrasch:
Thanks!!

PR #1229, opened 2015-09-18
Redshift support for safe data cleanse (copy to temp table)

dlstadther:
Added support for S3 copy to temp table, allowing an arbitrary number of
addition sql statements to be executed.

To note, these changes lack one feature which may prove to be useful... The redshift marker table will note the copy into the temp table, but not additional inserts/changes to additional tables within this transaction.


dlstadther:
I am unsure why Travis failed on the last commit...

luigi/contrib/redshift.py
```diff
@@ -150,12 +162,22 @@ def create_table(self, connection):
...
+                    table=self.table,
+                    coldefs=coldefs,
+                    table_attributes=self.table_attributes())
+            else:
+                query = ("CREATE TABLE "
+                         "{table} ({coldefs}) "
+                         "{table_attributes}").format(
+                    table=self.table,
+                    coldefs=coldefs,
+                    table_attributes=self.table_attributes())
```
Tarrasch: Surely you don't need to copy this much code.

Tarrasch:
Thanks for your PR. For this to be merged you'll need to:

 * Find reviewers who know redshift
 * Fixup commits that add no historic value like "fixed formatting"
 * And add tests of course. :)

moandcompany:
Dillon, this is great. I was recently implementing something very similar after reviewing Luigi's Redshift support and had the same thoughts about finding a way to use the marker table for recording arbitrary queries.

I believe the marker table can be used to record and prevent duplicate query executions by encapsulating the query in a Luigi Task with a Redshift/Postgres Target output; the query and its related parameters can be used to compose the `update_id` value used for creating the Redshift Target. I believe `update_id` is the sole record value checked in the marker table when the Luigi scheduler checks to see if an Redshift/Postgres task has completed/has already been completed.

moandcompany:
After some additional review, the rdbms CopyToTable abstract task we inherit from provides a method for performing operations (i.e. queries) prior to the table copy `init_copy`. Perhaps we should add `post_copy` to the ancestor.

Also for a query we want to perform after the copy, if the query isn't required to act upon or reference a temporary table, we can record the query's execution as a note in the marker table by creating a Query abstract task.

dlstadther:
@moandcompany, I will soon address your general comments.

As for your specific code comments, please reference the most up-to-date code here. https://github.com/republic-analytics/luigi/commit/9787270c8457c969cd673ecdd81b8da892eca846

dlstadther:
@moandcompany I agree with implementing a post_copy method in rdbms to be overridden in redshift.

Regarding updating the marker table for non-temp table queries, it may be good to separate them from redshift copy completely.  Correct me if I'm wrong, but there doesn't seem to be any existing classes/methods to perform any independent queries on a postgres/redshift db in luigi. (This supports your suggestion to create a Query abstract task).

moandcompany:
@dlstadther Cool. I created a Query abstract task last night as well as child classes for postgres and Redshift. I can put them together for a PR later this evening.

moandcompany:
@dlstadther `Query`, `PostgresQuery` and `RedshiftQuery` classes have been added in PR: https://github.com/spotify/luigi/pull/1289

dlstadther:
Note: i'm working on updated tests

PR #1233, opened 2015-09-18
Revert exception pickling

### LINKED ISSUE #1227: Workers dying on errors that can't be unpickled
If I have a task that raises an error on run that can't be unpickled, it kills the whole worker immediately. Example code:

```
import luigi

class MyError(BaseException):
    def __init__(self, a, b):
        pass


class MyTask(luigi.Task):
    def run(self):
        raise MyError(1, 2)


if __name__ == '__main__':
    luigi.run()
```

If I run this with the latest master, the whole worker dies on the error with the following traceback:

```
  File "test.py", line 14, in <module>
    luigi.run()
  File "/Users/buck/workspace/luigi/luigi/interface.py", line 331, in run
    return _schedule_and_run(tasks, worker_scheduler_factory)
  File "/Users/buck/workspace/luigi/luigi/interface.py", line 184, in _schedule_and_run
    success &= w.run()
  File "/Users/buck/workspace/luigi/luigi/worker.py", line 790, in run
    self._handle_next_task()
  File "/Users/buck/workspace/luigi/luigi/worker.py", line 690, in _handle_next_task
    timeout=self._config.wait_interval))
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/multiprocessing/queues.py", line 135, in get
    res = self._recv()
TypeError: ('__init__() takes exactly 3 arguments (1 given)', <class '__main__.MyError'>, ())
```

whereas if I run with older code, it recovers from the failure and carries on. The problem seems to be that the exception is now put into the result queue and cannot be successfully unpickled when it's retrieved from the queue in the parent process. This is the same error you get if you pickle and try to unpickle the error.

daveFNbuck:
This reverts #1205 in order to resolve #1227. WorkerTest.test_fail is also modified such that it fails without the reversion.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -128,8 +128,6 @@ class TaskProcess(multiprocessing.Process):
         expl = ''
         missing = []
         new_deps = []
-        exception = None
-
         try:
             # Verify that all the tasks are fulfilled!
             missing = [dep.task_id for dep in self.task.deps() if not dep.complete()]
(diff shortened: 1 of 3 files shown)


erikbern:
thanks this is better than #1232

daveFNbuck:
Oh, I missed that you had done that! Let's get one of these merged then.

Tarrasch:
Thanks @daveFNbuck! Sorry I was busy today and didn't have time to do this revert.

PR #1234, opened 2015-09-18
Uses popovers instead of tooltips for svg graph

daveFNbuck:
The graph tooltips don't display very well for long names. This splits up the
names and puts them in a popover for a nicer display.

Before:
screen shot 2015-09-18 at 3 28 06 pm

After:
screen shot 2015-09-18 at 3 28 32 pm

The only downside is that the params appear in arbitrary order, but this could be fixed by sending the right order from the scheduler if it bugs anyone enough.

### DIFF
--- a/luigi/static/visualiser/js/graph.js
+++ b/luigi/static/visualiser/js/graph.js
@@ -180,14 +180,15 @@ Graph = (function() {
                 .attr("data-task-id", node.taskId)
                 .appendTo(g);
 
-            var titleText = node.taskId;
+            var titleText = node.name;
+            var content = $.map(node.params, function (value, name) { return name + ": " + value; }).join("<br>");
             g.attr("title", titleText)
-                .tooltip({
+                .popover({
+                    trigger: 'hover',
                     container: 'body',
                     html: true,
-                    title: function() {
-                        return $(this).prop('title');
-                    }
+                    placement: 'top',
+                    content: content
                 });
         });
 


Tarrasch:
Man you guys are just so awesome! @stephenpascoe @daveFNbuck @toidi

Tarrasch:
Order doesn't matter I think. Can anyone who knows frontend review? :)

toidi:
`node.params` expose sensitive information(`significant=False` job parameters).

toidi:
This is not that simple. Web UI via `RPCHandler` access `CentralPlannerScheduler` methods.

It could be tricky not to send sensitive params and do not affect to other code.

daveFNbuck:
We had tried not sending sensitive params to the scheduler before and I think it only broke assistants but there's a fix in for that already. We should be able to do it now. Alternatively, we could just tell the scheduler which parameters are insignificant and not send those in graph requests. That would completely avoid any possible side-effects.

Tarrasch:
I merge this because it looks nicer.

To be honest. I don't think we should send insignificant parameters to the scheduler too. It exposes useful information sometimes (like how many mappers you use). Also, it's not clear from that name that it should be used for sensitive data.

@toidi, if you have the use case to not send sensitive data. May I suggest that we instead the `Parameter` class to take an optional kwarg like `sensitive=True`, it can only used in conjunction with `significant=False` since it's not being sent to the scheduler. Does that make sense?

daveFNbuck:
I can put out a PR later today to not send significant params to the scheduler. I think it can be done safely since #1119

toidi:
@Tarrasch , `sensitive` parameter looks good to me.

daveFNbuck:
sensitive seems to mean the same thing as insignificant, as they're both about not displaying the parameter in the visualiser

Tarrasch:
@daveFNbuck. Well I would argue they are not. Significant is about if a parameter is significant to determine what will `run()`, when it's considered `complete()` etc, so for example `--num-mappers 100` should be considered `signifcant=False, sensitive=False`. I want to see how many mappers the job is using from the scheduler, but I don't want the scheduler for another worker to run the exact same task with just different amount of mappers.

daveFNbuck:
Sounds reasonable now. I hadn't thought of that use case.

PR #1240, opened 2015-09-22
add parameter name parameter to RangeBase

joeshaw:
When set, use this name as a keyword argument for the parameterized
value (date, hour, whatever) instead of using the first positional
argument.

--

This is particularly useful when your tasks derive from other tasks and
you don't have control over the ordering of parameters.

### DIFF
--- a/luigi/tools/range.py
+++ b/luigi/tools/range.py
@@ -101,6 +101,10 @@ class RangeBase(luigi.WrapperTask):
     now = luigi.IntParameter(
         default=None,
         description="set to override current time. In seconds since epoch")
+    param_name = luigi.Parameter(
+        default=None,
+        description="parameter name used to pass in parameterized value. Defaults to None, meaning use first positional parameter",
+        positional=False)
 
     @property
     def of_cls(self):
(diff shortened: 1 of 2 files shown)


luigi/tools/range.py
```diff
@@ -101,6 +101,9 @@ class RangeBase(luigi.WrapperTask):
     now = luigi.IntParameter(
         default=None,
         description="set to override current time. In seconds since epoch")
+    param_name = luigi.Parameter(
+        default=None,
```
Tarrasch: Maybe add `positional=False`. It does not need it more than the other parameters, but one must start somewhere.

Tarrasch:
Looks really good! I really have wanted this for a while! Thanks!!!

Though, you will have to add a test case :)

Tarrasch:
Also it would be nice to associate your email with your github account

luigi/tools/range.py
joeshaw: How would `positional=False` interact with `param_name`?  I don't understand its benefit.

`positional=False` + `param_name is not None` -> `task_cls(**{param_name: param})`
`positional=True` + `param_name is None` -> `task_cls(param)`
`positional=False` + `param_name is None` -> ?
`positional=True` + `param_name is not None` -> ?

joeshaw:
oops, forgot the format of `user.email` when i was overriding it.  I'll fix that and add a test.

joeshaw:
force pushed an update that adds a test and fixes my git email address.

luigi/tools/range.py
joeshaw: Oh, `positional` is a flag to `luigi.Parameter`.  I didn't know about that.  I'll fix that now.

joeshaw: Ok, pushed a commit which adds that.

Tarrasch:
Looks good to me. But can you fix the flake8 errors so the build becomes green?

joeshaw:
@Tarrasch should be all set now, thanks!

Tarrasch:
LGTM. Want to review @ulzha?

ulzha:
:+1:

Tarrasch:
Thanks @joeshaw !!

PR #1241, opened 2015-09-23
add AdRoll to the list of Luigi users

tuulos:
we published a blog article about our stack that is based on Luigi and Docker today at http://tech.adroll.com/blog/data/2015/09/22/data-pipelines-docker.html


Tarrasch:
Cool!

PR #1243, opened 2015-09-23
Use default number of attempts (3) for get_work rpc method.

jkukul:
Our workers die from time to time because of:
```
luigi.rpc.RPCError: Errors (1 attempts) when connecting to remote scheduler 'localhost'
```

This happens a lot more often when calling the `get_work` method than other RPC methods. Unlike all other RPC methods, `get_work` uses only one retry attempt and I don't really understand if there's a reason for it. I suspect additional two attempts were unintentionally cut off in 725172e3d8d970a505931a2556cda9a235dd2373

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -147,7 +147,7 @@ class RemoteScheduler(Scheduler):
         return result["response"]
 
     def ping(self, worker):
-        # just one attemtps, keep-alive thread will keep trying anyway
+        # just one attempt, keep-alive thread will keep trying anyway
         self._request('/api/ping', {'worker': worker}, attempts=1)
 
     def add_task(self, worker, task_id, status=PENDING, runnable=True,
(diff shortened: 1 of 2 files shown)


Tarrasch:
Wait, what?

This looks great and all and I would like to merge. But can you add a test case as well @jkukul?

jkukul:
@Tarrasch I pushed again, with a test case this time.

Tarrasch:
Thanks. I'll wais for @erikbern to review this though. He might remember why he did 725172e back in the days. :)

Tarrasch:
There's a flake8 error @jkukul

jkukul:
@Tarrasch fixed, sorry about that

erikbern:
did you try this in production yet?

jkukul:
Not yet. I'll deploy it to production and get back to you after it will have been working smoothly for a while.

Tarrasch:
Thanks @jkukul. This change makes sense to me. But it doesn't hurt to be on the safe side.

The bad thing I think can happen is that you keep asking for `get_work` but the client doesn't get the replies, so it keeps asking for it. Doing two `add_task` isn't harmful, but two `get_work` will alter state more than just one `get_work` would.

Tarrasch:
Haha, there's so many ex-spotifiers in #174 :)

toidi:
@jkukul , I have a worry about `log_exceptions=False`. Why did you drop it?

Tarrasch:
@toidi, The reason `log_exceptions=False` was true initially was that it was `=False` for the initial call, and then retried 2 times with `=True`. See the commits and surrounding history that @jkukul linked to.

jkukul:
So this has been running for us in production for more than a week now, without any side effects.

PR #1244, opened 2015-09-23
fix pickling for HdfsTarget

davemt:
A simple HdfsTarget instance no longer is pickleable due to instance methods being stored as attributes on CompatibleHdfsFormat (since aab71dac71091e8794e235358caeafb7240a648a).

We have some scenarios where an HdfsTarget is stored as an attribute on a hadoop task instance, which then must be pickled to submit the job to the cluster.

### DIFF
--- a/test/contrib/hdfs_test.py
+++ b/test/contrib/hdfs_test.py
@@ -19,6 +19,7 @@ import functools
 import re
 from helpers import unittest
 import random
+import pickle
 
 import helpers
 import luigi
@@ -449,6 +450,10 @@ class HdfsTargetTests(MiniClusterTestCase, FileSystemTargetTestMixin):
     def test_tmppath_username(self):
         self.assertRegexpMatches(hdfs.tmppath('/path/to/stuff', include_unix_username=True),
                                  "^/tmp/[a-z0-9_]+/path/to/stuff-luigitemp-\d+")
+
+    def test_pickle(self):
+        t = hdfs.HdfsTarget("/tmp/dir")
+        pickle.dumps(t)
 
 
 @attr('minicluster')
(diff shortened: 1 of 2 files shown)


erikbern:
thanks! maybe it would have been simpler to remove the reader and writer attributes?

what's the breaking test?

davemt:
I don't see an obvious way to remove need to store reader/writer methods on CompatibleHdfsFormat until support for hdfs_writer and hdfs_reader methods goes away (after that can just store format).

test/contrib/hdfs_test.py
```diff
@@ -450,6 +451,13 @@ def test_tmppath_username(self):
         self.assertRegexpMatches(hdfs.tmppath('/path/to/stuff', include_unix_username=True),
                                  "^/tmp/[a-z0-9_]+/path/to/stuff-luigitemp-\d+")
 
+    def test_pickle(self):
+        t = hdfs.HdfsTarget("/tmp/dir")
+        try:
+            pickle.dumps(t)
```
Tarrasch: Why not just `pickle.dumps(t)` without the `try-except`?

davemt: I'll simplify to that. I didn't have a good reason

Tarrasch:
@davemt. By the way. As I think @erikbern meant, can you show the real code you're running that fails. That could server as another test case. It also serves as documentation accompanying this PR.

davemt:
This is an example of the task type that was hitting the error:

davemt:
I should add that these pickling errors are extremely non-specific about what particular object/attribute is failing, so on chance it does come up for people it can be a real headache to track down.

PR #1246, opened 2015-09-24
prevent infinite recursion for _input_pipe and _output_pipe

demoray:
Basically, the same problem corrected for _process in 9f6c33f5c9247e931e9993f1146632d2f651d63a

This can be exposed by trying to use luigi.contrib.ssh.RemoteTarget to upload to a nested directory without permissions to create the dirname of the path, such as the following, as a user without write access to "/etc" on the remote host.

luigi.contrib.ssh.RemoteTarget("/etc/this/fails", hostname)

### DIFF
--- a/luigi/format.py
+++ b/luigi/format.py
@@ -144,7 +144,7 @@ class InputPipeProcessWrapper(object):
             self._finish()
 
     def __getattr__(self, name):
-        if name == '_process':
+        if name in ['_process', '_input_pipe']:
             raise AttributeError(name)
         try:
             return getattr(self._process.stdout, name)
@@ -225,7 +225,7 @@ class OutputPipeProcessWrapper(object):
         self._finish()
 
     def __getattr__(self, name):
-        if name == '_process':
+        if name in ['_process', '_output_pipe']:
             raise AttributeError(name)
         try:
             return getattr(self._process.stdin, name)
--- a/test/contrib/test_ssh.py
+++ b/test/contrib/test_ssh.py
@@ -27,7 +27,7 @@ import subprocess
 from helpers import unittest
 import target_test
 
-from luigi.contrib.ssh import RemoteContext, RemoteFileSystem, RemoteTarget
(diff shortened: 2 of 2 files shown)


erikbern:
LGTM. test would be nice

Tarrasch:
Thanks. Looks nice. But please add a test. :)

demoray:
I've added a unit test per request to test/contrib/test_ssh.py, using the ssh host defined elsewhere in the script, following the existing tests.

Note though, I am unable to run the script from that location as is, as it imports helpers & target_test, which are in a directory above test/contrib, which isn't in the import path.  Perhaps whatever test runner used by travis will pick it up correctly, but py.test locally does not.

When I rename test/contrib/test_ssh.py to test/test_ssh2.py (as test/test_ssh.py exists for testing mock interfaces), then the module works as expected.

Tarrasch:
@demoray have you used `tox` as described in CONTRIBUTIN.rst file?

demoray:
Works within tox.  (well, coverage as used by tox)

Tarrasch:
So the test failed prior to your change? I'm bit confused since the exception type you raise in the logic is not the same one as in the test. No?

demoray:
Prior to the change, the above code would cause an infinite recursion within __del__ of luigi.contrib.ssh.AtomicRemoteFileWriter due to __getattr__ within luigi.format.OutputPipeProcessWrapper.

The test was a simple mechanism showed the underlying issue that didn't touch private variables (those prefixed with underscore).

Running the test without the change results in a never-ending loop of:
Exception RuntimeError: RuntimeError('maximum recursion depth exceeded in cmp',) in <bound method AtomicRemoteFileWriter.__del__ of <luigi.contrib.ssh.AtomicRemoteFileWriter object at 0xa2fa3ec>> ignored

Run with the change, and that does not occur.

Tarrasch:
Perfect! Thanks @demoray. Thanks for clarifying.

PR #1249, opened 2015-09-26
worker.py: Remove unnecessary (== RUNNING) check

Tarrasch:
It must be from old times or something. RUNNING is only a concept of
luigid, you never add_task(status=RUNNING) so to say.  The status is
taken from the task queue, and the task queue is of course only
populated from the client.


Tarrasch:
You start writing a small patch and end up touching 15 different modules :p

Tarrasch:
ping @ulzha @erikbern

PR #1250, opened 2015-09-26
Deprecate SUSPENDED status

Tarrasch:
The reason is just to simplify the code, it's confusing when there are
so many statuses. And it's not obvious at all how it's utilized and that
it's only associated with the new_deps parameter.

This makes so new clients will not send the status SUSPENDED. Server
will still just convert it to PENDING as previously.

Hence, it will work for "old client new server", but not vice verse.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -389,10 +389,6 @@ class SimpleTaskState(object):
         if new_status == FAILED:
             assert config is not None
 
-        # not sure why we have SUSPENDED, as it can never be set
-        if new_status == SUSPENDED:
-            new_status = PENDING
-
         if new_status == DISABLED and task.status == RUNNING:
             return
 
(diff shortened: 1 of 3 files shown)


Tarrasch:
Can you review @tommyengstrom?

erikbern:
isn't this needed for dynamic dependencies? but i guess not

Tarrasch:
@erikbern. Nope, since you can just check truthness of `new_deps` if you should behave as the old SUSPENDED behavior or not.

ulzha:
My code goggles are somewhat convinced that this preserves behavior ok...

Unimpressed with the concept of a task becoming PENDING instead of DONE after returning from run(). Is that simple to interpret for everyone? or is perhaps the SUSPENDED state necessary to keep the dynamic dependencies world as simple as possible, but not simpler?

Tarrasch:
> Is that simple to interpret for everyone

Not sure who is "everyone". The visualiser will show it as pending either way. Do you mean for people who dig into the `worker.py` code?

It would just be confusing if it returned something like DONE. It's not done, it has just done a `yield` from its `run()`. it has not finished it's `run()`. For the status to be PENDING is more intuitive than SUSPENDED for me. PENDING I think of something that will "run again, just isn't running yet". SUSPENDED sounds like it have been behaving bad and got suspended (which one might confused with the DISABLED status). Less states less confusion was my motivation for this PR.

ulzha:
Wait, you're saying it hasn't finished its run()? If that's true, then that's the point I'm missing and LGTM. No need of convincing me on the points that I inferred from that false premise.

Tarrasch:
Stupid human communication. :p 

Thanks for the review! :)

PR #1194, opened 2015-09-03
Add main module for luigi package

toidi:
I have an issue launching luigi jobs in development environment.

The purpose of this PR better to describe on an example:
```
$ pwd
/home/e.iskandarov/workspace/mario-jobs
$ ls mario_jobs
common.py  false_positive  __init__.py  return_path  sort_out_inbox_senders  tests
$ luigi --module mario_jobs.false_positive FalsePositiveLauncher --site=mail.ru --local-scheduler
Traceback (most recent call last):
  File "/home/e.iskandarov/.virtualenvs/luigi/bin/luigi", line 9, in <module>
    load_entry_point('luigi==1.3.0', 'console_scripts', 'luigi')()
  File "/home/e.iskandarov/.virtualenvs/luigi/lib/python2.7/site-packages/luigi/cmdline.py", line 13, in luigi_run
    luigi.interface.run(argv, use_dynamic_argparse=True)
  File "/home/e.iskandarov/.virtualenvs/luigi/lib/python2.7/site-packages/luigi/interface.py", line 421, in run
    tasks = interface.parse(cmdline_args, main_task_cls=main_task_cls)
  File "/home/e.iskandarov/.virtualenvs/luigi/lib/python2.7/site-packages/luigi/interface.py", line 329, in parse
    __import__(module)
ImportError: No module named mario_jobs.false_positive
```

This happened because in that case current working directory was not in `PYTHONPATH`.

To overcome this problem I add `__main__` module for `luigi` package and it works perfectly fine for development purposes.

```
$ python -m luigi --module mario_jobs.false_positive FalsePositiveLauncher --site=mail.ru --local-scheduler
DEBUG Checking if FalsePositiveLauncher(site=mail.ru) is complete
...
```

PR #1252, opened 2015-09-28
Various doc fixes

Tarrasch:
See individual commits


Tarrasch:
@erikbern

doc/command_line.rst
```diff
@@ -16,13 +18,10 @@ Any task can be instantiated and run from the command line:
         def run(self):
             print self.x + self.y
```
ulzha: Annoying that running the examples doesn't in fact print this on console.

doc/command_line.rst
```diff
@@ -16,13 +18,10 @@ Any task can be instantiated and run from the command line:
...
+    $ luigi --module my_module MyTask --x 123 --y 456 --local-scheduler
 
-The other way to run a Luigi task is to use the builtin *luigi* script.
-This will be default on your path and
-can be run by providing a module name.
-The module will imported dynamically::
+Or alternatively like this:
 
-    $ luigi --module my_module MyTask --x 123 --y 456 --local-scheduler
+    $ python -m luigi --module my_module MyTask --x 100 --local-scheduler
```
ulzha: I got this (after `sudo pip install luigi`):
```
uldis@uldis-ubuntu:~/tmp$ PYTHONPATH=. python -m luigi --module my_module MyTask --x 100
/usr/bin/python: No module named luigi.__main__; 'luigi' is a package and cannot be directly executed
```
Did you mean `-m my_module`?

Tarrasch: Nope. This works in fact, see https://github.com/spotify/luigi/pull/1194. I ran something like `PYTHONPATH='' pip install --upgrade $HOME/spotify/repos/luigis/github_luigi/` when I tested this. I think you can also specify a git url pointing direclty to github's HEAD.

But it's expected to not work with latest stable version (we should release 2.0.0 soon)

doc/programmatic_execution.rst
```diff
@@ -1,25 +0,0 @@
...
-The code will look something like
-
-.. code:: python
-
-    task = MyTask(123, 'xyz')
-    interface.setup_interface_logging()
-    sch = scheduler.CentralPlannerScheduler()
-    w = worker.Worker(scheduler=sch)
-    w.add(task)
-    w.run()
```
ulzha: I think we still support this as the programmatic worker and scheduler API, so deleting the section doesn't seem warranted? Just the "As seen above" sentence is bad, there's nothing above in http://luigi.readthedocs.org/en/latest/programmatic_execution.html

ulzha:
LGTM

doc/programmatic_execution.rst
Tarrasch: We can re-add this section in a next iteration. But the content in the current section is all outdated and discouraged. Having `sch = scheduler.CentralPlannerScheduler()` is like encouraging `--local-scheduler`.

erikbern: I sort of disagree. I want Luigi to be a library as much as a framework. If people want to invoke Luigi pipelines from other places then that should be supported. We shouldn't make command line the only entry point

erikbern: but let's remove it if the text is outdated

Tarrasch: I'm not against programmatic use as a library. We've recently improved the docs for `interface.py`. Hopefully it will suffice with that, but if need arises, let's recreate this page with accurate information.

Issue #1253, opened 2015-09-29
Composition of parameters


riga:
Consider this small example:

```python
class A(Task):
    log = Parameter(default="/var/log.txt")

class B(Task):
    logPath = Parameter(default="~/")

class C(A, B):
    logFile = Parameter(default="log.txt")
```

Now, I want to tell task ``C`` to overwrite its inherited ``log`` parameter, which should be ``os.path.join(logPath, logFile)``. To be consistent with the instance caching, I cannot use the
``__init__`` method. However, I could extend the default ``get_param_values`` method, but
that does not seem clean enough to me. In cases like this I would propose something like a
hook to ensure parameter flexibility when e.g. working in such inheritance scenarios. If ``A.log``
had not default (which might be the case if you want to define ``A`` as some kind of interface),
the extension of ``get_param_values`` wouldn't work anyway.

Edit: solved in PR #1254.

Tarrasch:
Do you really need to inherit from `A`? I'm not that fond of parameters being dependent on each other within one Task. Can't you just override `log` with a property in `C`?

riga:
I see your point w.r.t. that minimal example.

However, our setup is a way more complex (we use luigi to structure a physics analysis of data from the LHC at CERN), which makes it difficult and unmanageable for us to simply override parameters of super tasks.

E.g. one of our tasks (``CustomTask``) defines 9 parameters. Its super tasks (we have several inheritance layers and multiple inheritance at some points) contains a task (``BaseTask``) that defines only one parameter (``myParam``) that has no default value. For logical reasons, ``CustomTask`` uses 6 of its 9 parameters to compose the value of its (inherited) ``myParam`` parameter. Of course, there might be some redundance, but that's wanted. Other tasks inheriting from ``BaseTask`` define ``myParam`` in a different way or even leave it blank to require explicit definition.

Tarrasch:
> However, our setup is a way more complex (we use luigi to structure a physics analysis of data from the LHC at CERN), which makes it difficult and unmanageable for us to simply override parameters of super tasks.

LHC?! That's cool! Didn't know it used luigi. :)

Have you thought of `CustomTask` not inheriting from `BaseTask`? Or maybe have a `BaseTaskWithMyParam` that inherits from `BaseTask`, so you could then have something like:

```
class BaseTask(luigi.Task):
    @abstractproperty
    def my_param(self):
         pass

    def calculate_energy(self):   # Or whatever physicist use luigi for ;)
[...]

PR #1254, opened 2015-09-29
Parameter hook

riga:
This PR addresses issue #1253. It adds a classmethod ``param_hook`` to ``Task`` to allow for composition of parameters in accordance with instance caching.

The example use case in #1253 becomes:

```python
class A(Task):
    log = Parameter(default="/var/log.txt")

class B(Task):
    logPath = Parameter(default="~/")

class C(A, B):
    logFile = Parameter(default="log.txt")

    @classmethod
    def param_hook(cls, result, args, kwargs):
        log = os.path.join(result["logPath"], result["logFile"])
        # now, we can either set a default log value
        result.setdefault("log", log)
        # or force it
        result["log"] = log
```

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -27,6 +27,7 @@ except ImportError:
 import logging
 import traceback
 import warnings
+from collections import OrderedDict
 
 from luigi import six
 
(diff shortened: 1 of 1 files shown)


Tarrasch:
Closing for now, as a workaround have been suggested in #1253 already.

PR #1255, opened 2015-09-29
Removes status selectors from graph and worker viz tabs

daveFNbuck:
The status selectors aren't used on these tabs, so they serve only to confuse.


Tarrasch:
LGTM. Can you review @stephenpascoe?

Tarrasch:
Makes sense. Thanks

PR #1256, opened 2015-09-29
Removes showWorker argument from taskToDisplayTask

daveFNbuck:
This argument is no longer used, but was causing a bug because maps were passing
indices to it, causing it to evaluate to true when this is never intended.


Tarrasch:
Cool. Was the bug making the browser crash or what?

daveFNbuck:
No, it was just adding worker information to the running time in places it shouldn't.

PR #1258, opened 2015-09-29
Sorts running tasks on worker page by running time

daveFNbuck:
This makes it easier to keep track with what workers are doing over time.


Tarrasch:
Makes sense. Thanks!

PR #1262, opened 2015-10-01
Change S3 references to GCS

dimfeld:
A few docstrings and comments in the GCS target still referenced S3. This just changes them to say GCS instead.


Tarrasch:
I have no idea how this was overlooked. Thanks @dimfeld! :)

Issue #1261, opened 2015-09-30
execution summary limits are hard coded


demoray:
The number of lines that an execution summary will print is hard coded to 5, with a comment that specifies to alter the hard coded value to change the limits [0].

It would be useful if this limit was configurable.

0 - https://github.com/spotify/luigi/blame/master/luigi/execution_summary.py#L97

Tarrasch:
Hmm, can you send a PR? What number have you found to be more suitable?

demoray:
Happy to, just didn't want to spend the time if nobody agreed with me.

For my work, I'd rather not have it limited.  A configurable limit would solve that for me.

Tarrasch:
@demoray I suppose that could be marked by just having a very high integer?

Looking forward to your patches @demoray :)

demoray:
See #1263 for PR.

demoray:
this was merged

Issue #1260, opened 2015-09-30
execution summary output is very different depending on number of results


demoray:
The execution summary is very different based on how many times a task was executed.  Given TaskA running once, TaskB running twice, and TaskC running thrice, the output looks like this:

\- 1 TaskA(arg=1)
\- TaskB(arg1=1, arg2=1) and TaskB(arg1=1, arg2=2)
\- TaskC(arg=/long/arg/here) and 30 other TaskC

It would be nice if the output was visually similar regardless of the number of times it was executed, such as:
\- 1 TaskA(arg1)
\- 2 TaskB(arg1=1, arg2=1), TaskB(arg1=1, arg2=2)
\- 30 TaskC(arg=/log/arg/here) ...

Tarrasch:
I agree on it should always be a leading count-number for consistency. Would you like to send a PR? :)

demoray:
Happy to, just didn't want to spend the time if nobody agreed with me.  PR will be coming shortly.

demoray:
See #1263 for PR.

erikbern:
awesome

demoray:
this was merged

PR #1263, opened 2015-10-01
normalize execution summary (addresses #1260 and #1261)

demoray:
* allow configuration of the number of tasks to show in the execution summary
* always preface each task in the execution summary with the number of times the task was called

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -603,3 +603,13 @@ Parameters controlling storage of task history in a database
 db_connection
   Connection string for connecting to the task history db using
   sqlalchemy.
+
+
+[execution_summary]
+-------------------
+
+Parameters controlling execution summary of a worker
+
+summary-length
+  Maximum number of tasks to show in an execution summary.  If the value is 0,
+  then all tasks will be displayed.  Default value is 5.
--- a/luigi/execution_summary.py
+++ b/luigi/execution_summary.py
@@ -25,6 +25,12 @@ import textwrap
 import collections
 import functools
 
+import luigi
+
+
+class execution_summary(luigi.Config):
+    summary_length = luigi.IntParameter(default=5)
+
 
 def _partition_tasks(worker):
     """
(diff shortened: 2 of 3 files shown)


luigi/execution_summary.py
```diff
@@ -89,34 +91,39 @@ def _get_str(task_dict, extra_indent):
     """
     This returns a string for each status
     """
+    summary_length = int(configuration.get_config().get('core', 'summary-length', 5))
```
Tarrasch: This is not how we do configuration in luigi. :)

http://luigi.readthedocs.org/en/latest/configuration.html#configuration-classes

luigi/execution_summary.py
```diff
@@ -89,34 +91,39 @@ def _get_str(task_dict, extra_indent):
...
+        prefix_size = 4
         if extra_indent:
-            row += '    '
-        if len(lines) >= 5:
-            """
-            This is how many rows will be printed for each status. If you want fewer rows you can lower the limit.
-            """
-            row += '...'
-            lines.append(row)
+            prefix_size = 8
```
Tarrasch: I would just write `prefix_size = 8 if extra_indent else 4` :)

luigi/execution_summary.py
```diff
@@ -126,15 +133,15 @@ def _get_str(task_dict, extra_indent):
...
                 if len(tasks) == 1:
-                    row += '- {0} {1}'.format(len(tasks), tasks[0])
+                    line = prefix + '- {0} {1}'.format(len(tasks), tasks[0])
                 if len(tasks) == 2:
-                    row += '- {0} and {1}'.format(tasks[0], tasks[1])
+                    line = prefix + '- {0} {1} and {2}'.format(len(tasks), tasks[0], tasks[1])
                 if len(tasks) > 2:
-                    row += '- {0} and {1} other {2}'.format(tasks[0], len(tasks) - 1, task_family)
-        lines.append(row)
+                    line = prefix + '- {0} {1} ...'.format(len(tasks), tasks[0])
```
Tarrasch: Hmm, Wasn't it nicer and shorter to just do `row += blah` instead of `line =  prefix + blah`. I don't mind the new variable name, but I think it looked ok with `+=`, why did you remove that?

test/execution_summary_test.py
```diff
@@ -69,16 +69,90 @@ def requires(self):
...
+                    'This progress looks :( because there were failed tasks',
+                    '',
+                    '===== Luigi Execution Summary =====',
+                    '']
+        result = summary.split('\n')
+        self.assertEqual(len(result), len(expected))
+        for i, line in enumerate(result):
+            self.assertEqual(line, expected[i])
+
+    @with_config({'core': {'summary-length': '1'}})
```
Tarrasch: Yea, don't use section `[core]` for this. Maybe use `execution_summary`, and then the cli parameter will be something like`--execution-summary-length 3`. Right?

test/execution_summary_test.py
```diff
@@ -616,21 +718,33 @@ def requires(self):
...
+                    '    * 1 were missing external dependencies:',
+                    '        - 1 MyExternal()',
+                    '    * 21 had missing external dependencies:',
+                    '        - 10 DateTask(date=1998-03-23...1998-04-01, num=5)',
+                    '        - 1 EntryPoint()',
+                    '        - 10 Foo(num=100, num2=12) ...',
+                    '',
+                    'This progress looks :| because there were missing external dependencies',
+                    '',
+                    '===== Luigi Execution Summary =====',
```
Tarrasch: Nice! :)

Tarrasch: Actually, an this be even dicer with:

```
dedent("""
Scheduled 218 tasks ...
* blah blah
...
""")
```

test/execution_summary_test.py
```diff
@@ -700,3 +814,6 @@ class Foo(RunOnceTask):
         self.assertNotIn('The other workers were', s)
         self.assertIn('This progress looks :) because there were no failed ', s)
         self.assertNotIn('\n\n\n', s)
+
+if __name__ == '__main__':
+    unittest.main()
```
Tarrasch: Why? Just run `tox -e py27-nonhdfs test/execution_summary_test.py`?

doc/configuration.rst
```diff
@@ -243,6 +243,9 @@ worker-wait-interval
   for another job after the scheduler has said that it does not have any
   available jobs.
 
+summary-length
+  Maxinum number of tasks to show in an execution summary.  If the value is 0,
+  then all tasks will be displayed.  Default value is 5.
```
Tarrasch: Thanks for also writing the docs :)

luigi/execution_summary.py
demoray: Is it?  I can only find this in use by HDFS.  Every other instance I can find of configuration does it in a similar fashion.

From [notifications.py](https://github.com/spotify/luigi/blob/master/luigi/notifications.py#L45):
        return configuration.get_config().get('core', 'email-type', 'plain')

From [postgres.py](https://github.com/spotify/luigi/blob/master/luigi/postgres.py#L107)
    marker_table = luigi.configuration.get_config().get('postgres', 'marker-table', 'table_updates')

[interface.py](https://github.com/spotify/luigi/blob/master/luigi/interface.py#L156)
    if not configuration.get_config().getboolean(
            'core', 'no_configure_logging', False):

That said, I should have used 'getint' rather than int(... get_config.get(...))  An update will be coming shortly.

luigi/execution_summary.py
demoray: Shorter, yes.

The same variable was being appended across 37 lines of code, which was more than would fit in one terminal.  It wasn't immediately clear as to what was appending to row and when.  My personal taste is that _get_msg could use additional refactoring, but this change made it easier to trace the content of 'line' through the flow.

test/execution_summary_test.py
demoray: Alright.  Next update will adjust this.

test/execution_summary_test.py
demoray: Alright.

luigi/execution_summary.py
Tarrasch: They are wrong. I'll try to update them as I see it can be very confusing.

luigi/execution_summary.py
Tarrasch: Ok. :)

Tarrasch:
Can you also squash the commits please? I think it's enough with only 1 here. Unless you are willing to split them into logical commits (right now they are more of the "whoups, forgot XYZ" kind :)).

demoray:
Squashed per request.

luigi/execution_summary.py
```diff
@@ -89,34 +91,37 @@ def _get_str(task_dict, extra_indent):
     """
     This returns a string for each status
     """
+    summary_length = configuration.get_config().getint('execution_summary', 'summary-length', 5)
```
Tarrasch: This is still blocking merge. Please do it the luigi way as in the docs I linked to

doc/configuration.rst
```diff
@@ -603,3 +603,13 @@ Parameters controlling storage of task history in a database
...
   sqlalchemy.
+
+
+[execution_summary]
+--------------
+
+Parameters controlling execution summary of a worker
+
+summary-length
+  Maxinum number of tasks to show in an execution summary.  If the value is 0,
```
Tarrasch: Maximum*

Tarrasch:
This looks great, and thanks for making the summary more deterministic as well. But there just seem to be some small build issues blocking this from getting merged. Will you fix those so we can merge? :)

Tarrasch:
Sorry for asking for another iteration. But can you please squash the commits? :)

This looks really great! Can't wait to get this merged!

demoray:
Squashed again, per request.

Tarrasch:
Thanks!!!

Issue #687, opened 2015-01-30
Task scripts should return non-zero if any tasks were DISABLED or FAILED


boosh:
This would make error handling much simpler because the return code of a workflow would indicate whether it completed successfully. Workflows could then be run in a shell script which could check the return code and take appropriate action if anything failed (e.g. try again, trigger alerts, etc.).

Tarrasch:
Hmm, maybe that should be configurable from a command line flag to the `luigi` command. But I don't think we dare change the current behavior.

erikbern:
I actually agree with this,  think it would make more sense. What's the current behavior?

Tarrasch:
The current behavior is that we return 0, because the scheduler didn't crash.

Tarrasch:
I've changed my mind on this. I'm all for this being implemented. :)

themalkolm:
I would say it should be optional. There is a difference between luigi & tasks failing.

Maybe some option like in salt i.e. --retcode-passthrough to pass return code from task to the runner.

erikbern:
I doubt we're going to break anything if we just exit 1. Should we try?

Tarrasch:
Follow progress in #1264

PR #1264, opened 2015-10-02
Add semantic exit codes

Tarrasch:
This commits adds a long requested feature, to be able to tell what the luigi
client did (or didn't) based on its exit code. This closes spotify/luigi#687.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -411,6 +411,47 @@ hive resources and 1 mysql resource:
 Note that it was not necessary to specify the 1 for mysql here, but it
 is good practice to do so when you have a fixed set of resources.
 
+.. _retcode-config:
+
+[retcode]
+----------
+
+Configure return codes for the luigi binary. In the case of multiple return
+codes that could apply, for example a failing task and missing data, the
+*numerically greatest* return code is returned.
+
+We recommend that you copy this set of exit codes to your ``luigi.cfg`` file:
+
+::
+
+  [retcode]
+  # The following return codes are the recommended exit codes for luigi
+  # They are in increasing level of severity (for most applications)
+  already_running: 10
+  missing_data: 20
+  task_failed: 30
+  unhandled_exception: 40
+
+unhandled_exception
+  For exceptions during scheduling (if you raise from the ``complete()`` or
+  ``requires()`` methods for instance) or for internal luigi errors.  Defaults
+  to 4, since this type of error probably will not recover over time.
+missing_data
+  For when an :py:class:`~luigi.task.ExternalTask` is not complete, and this
+  caused the worker to give up.  As an alternative to fiddling with this, see
+  the [worker] keep_alive option.
+task_failed
+  For signaling that there were last known to have failed. Typically because
+  some exception have been raised.
+already_running
+  This can happen in two different cases. Either the local lock file was taken
+  at the time the invocation starts up. Or, the central scheduler have reported
+  that some tasks could not have been run, because other workers are already
+  running the tasks.
+
+If you customize return codes, prefer to set them in range 128 to 255 to avoid
+conflicts. Return codes in range 0 to 127 are reserved for possible future use
+by Luigi contributors.
 
 [scalding]
 ----------
--- a/luigi/cmdline.py
+++ b/luigi/cmdline.py
@@ -3,14 +3,15 @@ import argparse
 import logging
 import sys
 
+import luigi.interface
 import luigi.server
 import luigi.process
 import luigi.configuration
-import luigi.interface
+from luigi.retcodes import run_with_retcodes
 
 
 def luigi_run(argv=sys.argv[1:]):
-    luigi.interface.run(argv)
+    run_with_retcodes(argv)
 
 
 def luigid(argv=sys.argv[1:]):
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -167,7 +167,7 @@ def _schedule_and_run(tasks, worker_scheduler_factory=None, override_defaults=No
     kill_signal = signal.SIGUSR1 if env_params.take_lock else None
     if (not env_params.no_lock and
             not(lock.acquire_for(env_params.lock_pid_dir, env_params.lock_size, kill_signal))):
-        sys.exit(1)
+        raise PidLockAlreadyTakenExit()
 
     if env_params.local_scheduler:
         sch = worker_scheduler_factory.create_local_scheduler()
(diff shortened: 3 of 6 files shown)


erikbern:
LGTM although I think this further solidifies the special property of `ExternalTask` which I don't think is great (we already have a bunch of code in the worker class that makes distinction based on those etc)

Tarrasch:
Not sure what to do about the concept of ExternalTasks. Isn't it natural that it exists in real-world settings? I mean, eventually there will be a data source you rely on that is not meant to be created within the luigi framework.

erikbern:
I like `ExternalTask`. But I think it should just be a subclass of `Task` and there shouldn't be a bunch of magic that detects whether a certain task is a `Task` or an `ExternalTask`

doc/configuration.rst
```diff
@@ -411,6 +411,35 @@ hive resources and 1 mysql resource:
 Note that it was not necessary to specify the 1 for mysql here, but it
 is good practice to do so when you have a fixed set of resources.
 
+.. _retcode-config:
+
+[retcode]
+----------
+
+Configure return codes for the luigi binary. In the case of multiple return
```
ulzha: What use is there for making them configurable? An unusual enabler of accidental complexity.

Perhaps declare in documentation a "reserved" range of return codes that Luigi might define later, the rest being available for application-specific usage.

Tarrasch: It adds complexity (for users), but I think it's justified since real-world scheduling is complex, and whoever uses luigi in a big setting should actively think about the corner cases. Keep in mind that:

 * People might already rely on the current behaviour (exit code 1 on unhandled exceptions).
 * You want different return codes in different contexts.
   * Sometimes you simply want the "run until you succeed" behaviour that is the default (exit code 0 for failures/missingdata/alreadyrunning). 
   * Sometimes you want to run a command line, and feel confident that if I get exit code 0, my root task is in fact complete (then you would configure all exit codes).
   * If you use Jenkins as a cron-invoker, you would like task_failed to have a dedicated error code, causing "red" in Jenkins. Meanwhile already running or missing data exit codes should indicate "pending", or indicating your Jenkins dependencies mismatch what's configured in luigi.

That said. I would be thrilled to see this being simplified (convention over configuration), but I don't want to render it useful for only 50% of the users who want exit codes.

> Perhaps declare in documentation a "reserved" range of return codes that Luigi might define later, the rest being available for application-specific usage.

I'm not sure what you mean. You basically want to enforce that one can only choose exit codes within the range [0, 4] or something and that [5..255] are reserved for luigi itself?

ulzha:
We do have some beginnings of an event system. Event handlers are a preferred way of programmatically obtaining rich information from innards of a complex system in a uniform way. This implementation (run_with...) adds a ad-hoc layer for that, while the "running with events" layer/convention, which we already have, can serve the same purpose, and supports more than 8 bits of bandwidth, so why not use that. (Almost. Currently the event context only can be a Task. Adjusting it to support something like a Worker-wide event is needed. Also a way to configure loading of user modules dynamically, so any Luigi user can develop proprietary handlers easily.) The users who want to customize return codes then can load handlers that do sys.exit(x)...

Tarrasch:
Sounds sensible @ulzha. But It seems like `run_with...` can just be made private `_run_with...`, and we could change the underlying implementation to use the event system (once we have it working for worker-wide events) but still expose the same cli/config API for the exit codes -- if we are happy with the configurable complex API this PR is suggesting.

stowell:
I'll leave the actual implementation details and style to the _Guardians of Luigi (tm)_. This change would get us what we need: a way to determine from the shell if a task failed to complete for any reason.

doc/configuration.rst
ulzha: "Enforce" sounds as if it were a burden. It is a best practice that avoids problems in the future (custom thingies clashing with vendor/Luigi thingies). Various kinds of specifications, when dealing with limited amount of things, do make such RFU provisions.

I meant neither of those ranges to be so small as [0, 4]... Say, [0..127] reserved for Luigi, and [128..255] for anyone who wraps or forks or plugs into it (so they are promised to never conflict with something from vanilla Luigi).

ulzha: Yeah, I do understand both of the black bullet points. My question was different - I'm questioning why we seem to encourage the concrete values of "different return codes" to vary from one environment to another. Better standardize on one meaning of each return code across all Luigi community?

doc/configuration.rst
```diff
@@ -411,6 +411,46 @@ hive resources and 1 mysql resource:
...
+  For signaling that there were last known to have failed. Typically because
+  some exception have been raised.
+already_running
+  This can happen in two different cases. Either the local lock file was taken
+  at the time the invocation starts up. Or, the central scheduler have reported
+  that some tasks could not have been run, because other workers are already
+  running the tasks.
+
+Please set these exit codes within the range of 0 to 127, inclusive.  Other exit
+codes are reserved for yet to be defined purposes.
```
ulzha: Reads like the opposite thing of what we want... It's the 0 to 127 codes that we are reserving (hoping to standardize usage of any newly introduced codes in that range in the future), and the rest we are allowing to use freely.

"If you introduce custom exit codes, prefer to set them in range 128 to 255 to avoid conflicts. Exit codes in range 0 to 127 are reserved for possible future use by Luigi."

doc/configuration.rst
```diff
@@ -411,6 +411,46 @@ hive resources and 1 mysql resource:
...
+
+::
+
+  [retcode]
+  # The following return codes are the recommended exit codes for luigi
+  # They are in increasing level of severity (for most applications)
+  already_running: 1
+  missing_data: 2
+  task_failed: 3
+  unhandled_exception: 4
```
ulzha: Prefer to spread them out, 10 20 30 40 like BASIC line numbers or some other scheme? to give us more leeway in introducing codes for nuances of any of the conditions in the future, without breaking the increasing level of severity (approximate) invariant.

doc/configuration.rst
Tarrasch: I see what you mean now. Thanks for clarifying by example.

ulzha:
LGTM. Would be cool if requesters of https://github.com/spotify/luigi/issues/687 (@boosh) gave a look too.

I caught myself in too many experts exchanges today about "standard" return code allocation practices existing out there. E.g. [this](http://tldp.org/LDP/abs/html/exitcodes.html) is insightful...

Tarrasch:
Boom! @ulzha I already pinged them in #687. I suppose they are busy :)

PR #1271, opened 2015-10-05
Allow default BoolParameter to be True.

dhruvg:


test/parameter_test.py
```diff
@@ -173,6 +177,9 @@ def test_bool_true(self):
         self.run_locally(['Baz', '--bool'])
         self.assertEqual(Baz._val, True)
 
+    def test_bool_default_true(self):
```
Tarrasch: Also add a `with_config` decorator, setting the default to false, and check that it ranks stronger than `default=`

dhruvg: Done. Lmk if there is anything else.

test/parameter_test.py
```diff
@@ -45,7 +49,7 @@ class Foo(luigi.Task):
 
 
 class Baz(luigi.Task):
-    bool = luigi.BoolParameter()
+    bool = luigi.BoolParameter(default=False)
```
Tarrasch: You are not allowed to change this. :)

luigi/parameter.py
```diff
@@ -137,11 +137,7 @@ def __init__(self, default=_no_value, is_boolean=False, is_global=False, signifi
                                 as positional arguments become very tricky when
                                 you have inheritance and whatnot.
         """
-        if is_boolean:
-            self.__default = False
```
Tarrasch: You still need this code path in some way. If somebody doesn't explicitly set `default=True/False`, it should default to false.

I think it should be something like `self.__default = default or False`. As long as the tests pass (without you changing the current tests :)), it should be OK.

Tarrasch:
Thanks. Sorry for the messy confusing code for the Parameter bool logic. I should refactor it once we've merged this. :)

dhruvg:
Good to merge?

Tarrasch:
This is good to merge to me, except that you need to amend all 5 commits into one, so the git history remains clean.

dhruvg:
done

On Wednesday, October 7, 2015, Arash Rouhani <notifications@github.com>
wrote:

> This is good to merge to me, except that you need to amend all 5 commits
> into one, so the git history remains clean.
>
> —
> <https://github.com/spotify/luigi/pull/1271#issuecomment-146119770>.
>


-- 
Dhruv Garg
@drgarg <https://twitter.com/drgarg> | www.dhruvgarg.com

PR #1274, opened 2015-10-06
cli: Merge local and global cli parser flags

Tarrasch:
This patch is all about using argparse in a more native way.  This
doesn't change luigi much. But users of --help will now notice that
flags like `--of` and `--RangeHourly-of` are not merged. Like this:

    ./bin/luigi --local-scheduler RangeHourly --help
    ...
      --of RANGEHOURLY_OF, --RangeHourly-of RANGEHOURLY_OF
                             RangeHourly.of task name to be completed. The task
                             must take a single datetime parameter
    ...

Further, it has these advantages:

 * Less logic, internals are simpler
 * Slightly less output in --help
 * In the future. It could help us leverage the built in conflict
   handling in argparse

### DIFF
--- a/luigi/cmdline_parser.py
+++ b/luigi/cmdline_parser.py
@@ -20,7 +20,6 @@ be considered internal to luigi.
 """
 
 import argparse
-import functools
 from contextlib import contextmanager
 from luigi.task_register import Register
 import cached_property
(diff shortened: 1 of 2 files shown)


Tarrasch:
Also, this get rids of the very unintuitive variable name "glob" in luigi's parameter.py. Something I have wanted to do for a long time.

erikbern:
Nice! lgtm

PR #1276, opened 2015-10-06
cli: Make --help useful and add --help-all

Tarrasch:
Before this commit, running `luigi --help` is going to print 300 lines
of probably not useful flags, mostly being about all the combinations of
the registered various versions of the Range flags. Instead, I changed
--help to just print.

    $ ./bin/luigi --help
    usage: luigi [--local-scheduler] [--module CORE_MODULE] [--help] [--help-all]

    optional arguments:
      --local-scheduler     Use a in-memory central scheduler. Useful for testing.
      --module CORE_MODULE  Used for dynamic loading of modules
      --help                Show most common flags and all task-specific flags
      --help-all            Show all command line flags
    Exiting due to --help was passed

And when you're interested in an actual tasks parameter, they will be
displayed when specifying that task.

    $ ./bin/luigi --module examples.foo examples.Bar --help
    ...
    --num EXAMPLES.BAR_NUM, --examples.Bar-num EXAMPLES.BAR_NUM
    ...

### DIFF
--- a/luigi/cmdline_parser.py
+++ b/luigi/cmdline_parser.py
@@ -23,6 +23,7 @@ import argparse
 from contextlib import contextmanager
 from luigi.task_register import Register
 import cached_property
+import sys
 
 
 class CmdlineParser(object):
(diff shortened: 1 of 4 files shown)


luigi/cmdline_parser.py
```diff
@@ -130,6 +132,6 @@ def _possibly_exit_with_help(parser, known_args):
         """
         Check if the user passed --help, if so, print a message and exit.
         """
-        if known_args.core_help:
+        if known_args.core_help or known_args.core_help_all:
             parser.print_help()
             raise SystemExit('Exiting due to --help was passed')
```
ulzha: This message is incorrect now ("--help-all" != "--help") and I think no message should be printed at all. Spammy unlike any other program I've seen. I find it hardly likely for a person to need an explanation as to why the program printed something that looked like help and exited.

luigi/interface.py
```diff
@@ -72,7 +72,8 @@ class core(task.Config):
 
     local_scheduler = parameter.BoolParameter(
         default=False,
-        description='Use local scheduling')
+        description='Use a in-memory central scheduler. Useful for testing.',
```
ulzha: "an"

ulzha:
Apart from that, LGTM.

luigi/cmdline_parser.py
Tarrasch: Good point!

PR #1275, opened 2015-10-06
Cleaned up temp_dir function in test/simulate

Yoone:
Cleaned up test code from https://github.com/spotify/luigi/pull/1245

PR #1277, opened 2015-10-06
Cleaned up temp_dir function in test/simulate (branch renamed)

Yoone:
Cleaned up test code from #1245.
Duplicate for #1275. Renamed the branch, just in case it's the reason Travis doesn't run the tests.

### DIFF
--- a/test/simulate_test.py
+++ b/test/simulate_test.py
@@ -25,8 +25,6 @@ import tempfile
 
 
 def temp_dir():
-    if os.getenv('TRAVIS') == 'true':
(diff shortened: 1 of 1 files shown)


Tarrasch:
It seem to run it now. Nice :)

Tarrasch:
Thanks for the audit. But I hoped for you to be able to get rid of the `@is_writeable` decorator to. :)

Yoone:
Oh but sadly the tests don't pass on `py27-nonhdfs` without it, and I really haven't found a better way.

Tarrasch:
Hmm. OK

Issue #1278, opened 2015-10-06
debian/control should include python-setuptools in Build-Depends


demoray:
Unless python-setuptools is installed, then /usr/bin/luigi and /usr/bin/luigid do not get installed appropriately.

Given that setuptools is a build prerequisite rather than a runtime prerequisite, there is almost no impact to the end user systems.

erikbern:
Great can you send a pull request?

demoray:
sure.  Will do shortly.

demoray:
the PR was merged

PR #1279, opened 2015-10-06
add setuptools as a build prereq

demoray:
Add python-setuptools as a build prereq per spotify/luigi#1278


nailor:
Since we now explicitly define python-setuptools as a build requirement (which makes sense), could you extend your pull request to remove the conditional import in https://github.com/spotify/luigi/blob/master/setup.py#L20 too?

This stuff won't even fly with distutils, since `entry_points` is not supported there.

Good fixing! :)

Tarrasch:
Thanks! Feel free to ping me (write a message in the PR) when you've rebased, I don't get notifs on just pushes.

Issue #1280, opened 2015-10-06
python-pip is an "install requirement"


demoray:
The debian package has python-pip as a install time requirement.  

Typically, any required python library should be a package prerequisite, and the package should not install/update any other libraries either during installation or runtime.

Is this an oversight?

erikbern:
No strong feelings, sounds like you know more about Debian so you are probably right :)

demoray:
the PR I submitted was merged

Tarrasch:
Thanks for cleaning up the issues @demoray :)

PR #1281, opened 2015-10-06
remove python-pip as a build or install requirement

demoray:
'pip' should not be used to install prereqs, these should be done via debian packages, both during build time as well as runtime.

spotify/luigi#1280


nailor:
:+1: 

Looks good to me and indeed it makes sense not to depend on pip

nailor:
Btw, merging this is blocked on #1279, since the build will most likely fail if this is merged in before we explicitly depend on the setuptools.

Tarrasch:
Thanks @demoray and @nailor!

Issue #1283, opened 2015-10-07
Visualization gets messy in some cases


samuell:
We have a problem that with certain dependency graph topologies, they get very very very long (on the page) ... seemingly due to how text labels are shifted to be stacked diagonally upon each other.

In our case, this seems to create a "domino" effect, with longer and longer connections between up/downstream stages, to the extent that it gets unreadable. See example here (this is only the "top of the iceberg" :P ): 
![image](https://cloud.githubusercontent.com/assets/125003/10343710/7145c920-6d1f-11e5-8f84-44641a126c75.png)

erikbern:
did you try the new visualizer?

samuell:
@erikbern Ah, no ... that's news to me ...gotta check ... any pointer?

Tarrasch:
@samuell it's in Luigi HEAD. Not on pypi yet.

Though you'll still have this issue in the svg visualiser

samuell:
@Tarrasch Ok, thanks ... will have a look though!

Tarrasch:
I don't know how to solve this though. @stephenpascoe and @daveFNbuck were talking about introducing the new filtering on the graphs too. It could solve this problem.

daveFNbuck:
I have this problem as well, and the D3 graph doesn't help because it's way too slow on larger graphs. Filtering by status wouldn't really help in the pictured example, as the graph is all pending nodes anyway. My graphs tend to be dominated by done nodes and have a fairly pending subgraph. One thing that I think could help here is to add a depth limit to the graph. You won't be able to see everything at once, but it'll load much faster and you can always click on the nodes you're interested in to see more detail. It would be much more useful than a lot of my graphs in which 99% of the scrolling is just pages and pages of lines with no nodes.

daveFNbuck:
Also, we should be able to swap in a better graph layout algorithm. There's no reason that the portion of the graph shown in this example needs to use so much vertical space.

samuell:
Good description of the problem, @daveFNbuck !

But then, comparing to some other workflow tools, like [galaxy](http://galaxyproject.org), I realize luigi does visualization on the task level, while I think many other tools just do it on the process level (one box for each task *class*, not each *instance*) ...

samuell:
... that is, one box per task class, with indicators telling how many instances of that task class are pending, running finished etc.

samuell:
Or maybe, only on clicking one such, it is expanded to task instances. THat would be a sort of "defult grouping". Would be cool!

[...]

PR #1286, opened 2015-10-07
Shortens task graphs by counting each node only once

daveFNbuck:
In computeDepth in graph.js, the entire dependency tree is traversed and each
node is given a depth and position within its depth. This computation is a bit
off, however, as a node may be encountered multiple times during traversal. This
could be easily fixed by stopping the recursion when we hit a seen node, but
then we could end up with a node depending on a node at a higher level than
itself.

Instead, I turn this into a two-stage affair. In the first stage, we do the full
traversal and only remember the maximum depth for each node. In the second stage
we traverse the whole tree again and only place nodes when we hit them for the
first time at the right depth. The second step requires a tree traversal rather
than just iterating over the nodes array because it groups child nodes together
and results in a much more readable and nicer-looking graph.

This shrinks each level's height by a factor of about the average in-degree, which partially addresses #1283.

Before:
screen shot 2015-10-07 at 4 47 04 pm

After:
screen shot 2015-10-07 at 4 47 38 pm

### DIFF
--- a/luigi/static/visualiser/js/graph.js
+++ b/luigi/static/visualiser/js/graph.js
@@ -58,21 +58,35 @@ Graph = (function() {
     /* Compute the maximum depth of each node for layout purposes, returns the number
        of nodes at each depth level (for layout purposes) */
     function computeDepth(nodes, nodeIndex) {
-        var rowSizes = [];
         function descend(n, depth) {
-            n.depth = depth;
+            if (n.depth === undefined || depth > n.depth) {
+                n.depth = depth;
+                $.each(n.deps, function(i, dep) {
+                    if (nodeIndex[dep]) {
+                        descend(nodes[nodeIndex[dep]], depth + 1);
+                    }
+                });
+            }
+        }
+        descend(nodes[0], 0);
+
+        var rowSizes = [];
+        function placeNodes(n, depth) {
             if (rowSizes[depth] === undefined) {
                 rowSizes[depth] = 0;
             }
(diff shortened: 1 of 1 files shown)


luigi/static/visualiser/js/graph.js
```diff
@@ -58,21 +58,35 @@ Graph = (function() {
     /* Compute the maximum depth of each node for layout purposes, returns the number
        of nodes at each depth level (for layout purposes) */
     function computeDepth(nodes, nodeIndex) {
-        var rowSizes = [];
         function descend(n, depth) {
-            n.depth = depth;
+            if (n.depth === undefined || depth > n.depth) {
```
Tarrasch: This looks quadratic in worst malicious cases. But whatever.

Tarrasch:
The code looks sane to me, and this seem to have been validated in by both you and Samuel. So I'll merge. :)

luigi/static/visualiser/js/graph.js
daveFNbuck: It is, but so was the previous code. The SVG rendering is prohibitively slow on larger graphs, so I'll look into lowering the complexity of this.

Tarrasch: So is the slow user experience because of this? I assumed it was slow because of rendering...

daveFNbuck: No, after some testing I don't think that this would account for much of the overall lag.

PR #1287, opened 2015-10-07
Displays SVG graphs in visualiser more densely

### LINKED ISSUE #1283: Visualization gets messy in some cases

daveFNbuck:
The SVG graphs only show one item on each horizontal line, which can result in
very tall graphs. This shows multiple items per line in order to compress the
amount of vertical space needed to display a graph. Each node is given 200
pixels width to display names in order to minimize overlaps.

Together with #1286 I believe that this resolves #1283.

Here are some before and after pictures (both include the #1286 so that they're not too crazy):

Before:
screen shot 2015-10-07 at 4 13 37 pm

After:
image

### DIFF
--- a/luigi/static/visualiser/js/graph.js
+++ b/luigi/static/visualiser/js/graph.js
@@ -10,6 +10,12 @@ Graph = (function() {
     /* Line height for items in task status legend */
     var legendLineHeight = 20;
 
+    /* Height of vertical space between nodes */
+    var nodeHeight = 10;
+
+    /* Amount of horizontal space given for each node */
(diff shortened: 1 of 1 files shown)


Tarrasch:
LGTM! Thanks!

PR #1288, opened 2015-10-08
Removes leftover popovers when updating graph data

daveFNbuck:
This is necessary because the popover that's visible when you click on a node
to view its subgraph does not go away on its own. When it remains, it cannot be
removed through normal user actions. It can either obscure the graph or linger
outside of the bounds of the new graph, causing the page to scroll below where
it normally would to display the popover. By destroying the popovers, this
problem is solved.

Before:
image


Tarrasch:
Thanks! Nice catch

PR #1289, opened 2015-10-08
Add RDBMS Query abstract class, PostgresQuery and RedshiftQuery subclasses

moandcompany:
Implements RDBMS `Query` abstract class and related subclasses `PostgresQuery`, `RedshiftQuery` as discussed in PR https://github.com/spotify/luigi/pull/1229 comments

`PostgresQuery` and `RedshiftQuery` provide task templates that will execute a query and result in marker table update(s) if the query is successful.

* If the query has already been executed successfully, and a marker table entry already exists, the Luigi scheduler will use this entry to prevent a future instance of the task from being executed, maintaining idempotence.

* When using these task templates, Luigi task parameters or an `update_id` method override should be used to ensure the query task will execute (i.e. `run`) during periodic, scheduled task workflow executions, but not more than once for a particular workflow execution.
* Example: use 'date' as a task parameter for a query task that is meant to execute as part of a daily batch process; the parameter is not required to be used in the actual query. One use case might include updating a report table with a summary of the prior day's data.

* `table` is a required instance property, as the `table` value is used when making marker table updates to the `target_table` column; `table` is not required to be used in the actual query

* It's worth noting that the marker table's `target_table` value is not used when the Luigi scheduler evaluates if a PostgresTarget or RedshiftTarget already exists; this strictly considers `update_id` values. `None` or an empty string appear to be valid `table` values

* If a `PostgresQuery` and `RedshiftQuery` based task is used in multiple workflows, or has a dynamic `target_table`, this needs to be provided to the task as a parameter. This will result in a distinct `update_id` task signature and enables execution on different tables. Otherwise, the Luigi scheduler will conclude that the task has already finished successfully.

The class docstrings provide usage notes, but it would be great if we could update project-level documentation to better explain how the marker tables work and can be used.


Tarrasch:
How is this working in relation to `CopyToTable`?

moandcompany:
I believe `CopyToTable` from Postgres and Redshift could be viewed and refactored as subclasses of `Query`

dlstadther:
@moandcompany Has there been any advances on this?

erikbern:
Can you fix the breaking tests?

alexdebrie:
@moandcompany I think the changes merged from #1390 will fix the broken tests. Can you pull those in?

erikbern:
yes would be great if you could rebase on latest master just to verify.

otherwise looks great!

erikbern:
can you fix the pep8 issues? https://travis-ci.org/spotify/luigi/jobs/84743046

Tarrasch:
Closing because of being inactive for a month. Please feel free to reopen whenever you have time to work on it. :)

PR #1290, opened 2015-10-08
Make `tox -e docs` so doesn't create temp-files

Tarrasch:
Otherwise your directory becomes unclean, and the tox logic is much
simpler and safer now.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -14,6 +14,22 @@
 
 import sys
 import os
+import sphinx.environment
+from docutils.utils import get_source_line
+
+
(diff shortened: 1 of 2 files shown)


Tarrasch:
I just merge since it's a pain to develop on the docs without this ...

erikbern:
Did you verify that this works with readthedocs?

Tarrasch:
I more went for "push and see". But it seems to not have broken anything: http://luigi.readthedocs.org/en/latest/

erikbern:
great!

PR #1293, opened 2015-10-08
cli: Fail on unknown param and show TASK in --help

Tarrasch:
This commit solves two different but implementation wise related issues:

 * When you do `luigi --help`, it does not even mention that you should
   supply a positional task argument. I also wonder if it's better to
   just mark it as a named argument (like OptParser had with --task)
 * Make luigi fail when one passes an unrecognized argument. It's a
   serious problem when the argument has a default value, since then the
   task will get instantiated anyway. For example, today somebody
   mentioned that they misspelled --hours-back to --hours_back in the
   luigi user group. This bug was probably introduced during the
   pretty recent command line parsing overhaul.

### DIFF
--- a/luigi/cmdline_parser.py
+++ b/luigi/cmdline_parser.py
@@ -22,7 +22,6 @@ be considered internal to luigi.
 import argparse
 from contextlib import contextmanager
 from luigi.task_register import Register
-import cached_property
 import sys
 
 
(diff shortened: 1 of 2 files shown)


Tarrasch:
@ulzha, the cmdline_parser code is even less stateful/declarative and more procedural with big __init__ now. Enjoy red bugfix :)

luigi/cmdline_parser.py
```diff
@@ -61,15 +60,14 @@ def __init__(self, cmdline_args):
         """
         Initialize cmd line args
         """
-        self.cmdline_args = cmdline_args
         known_args, _ = self._build_parser().parse_known_args(args=cmdline_args)
         self._attempt_load_module(known_args)
+        known_args, _ = self._build_parser().parse_known_args(args=cmdline_args)
```
rouzwawi: why parse twice?

test/cmdline_test.py
```diff
@@ -235,6 +236,15 @@ def test_python_module_help_class(self):
...
 
+    def test_bin_luigi_options_before_task(self):
+        args = ['./bin/luigi', '--module', 'cmdline_test', '--no-lock', '--local-scheduler', '--FooBaseClass-x', 'hello', 'FooBaseClass']
+        returncode, stdout, stderr = self._run_cmdline(args)
+        self.assertEqual(0, returncode)
+
+    def test_bin_fail_on_unrecognized_args(self):
+        returncode, stdout, stderr = self._run_cmdline(['./bin/luigi', '--no-lock', '--local-scheduler', 'Task', '--unknown-param', 'hiiii'])
+        self.assertNotEqual(0, returncode)
+
```
rouzwawi: Is there a test that asserts the task name was correctly picked up?

Tarrasch: There are hundreds of those tests already. :)

luigi/cmdline_parser.py
Tarrasch: Add a comment in the code. The `test_bin_luigi_options_before_task` would break if you delete this line.

Tarrasch: But yea. Allowing for the task to be specified positionally was a costly mistake, would be better if we had kept the `--task` way.

rouzwawi:
:+1:

PR #1294, opened 2015-10-08
make the wait_interval jitter configurable

demoray:
This change allows for the configuration of how much a randomness to add to the
sleep value for a worker.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -278,6 +278,8 @@ class worker(Config):
                                   'well as having keep-alive true')
     wait_interval = FloatParameter(default=1.0,
                                    config_path=dict(section='core', name='worker-wait-interval'))
+    wait_jitter = FloatParameter(default=5.0)
+
     max_reschedules = IntParameter(default=1,
                                    config_path=dict(section='core', name='worker-max-reschedules'))
     timeout = IntParameter(default=0,
@@ -341,6 +343,7 @@ class Worker(object):
         self._config = worker(**kwargs)
 
         assert self._config.wait_interval >= _WAIT_INTERVAL_EPS, "[worker] wait_interval must be positive"
+        assert self._config.wait_jitter >= 0.0, "[worker] wait_jitter must be equal or greater than zero"
 
         self._id = worker_id
         self._scheduler = scheduler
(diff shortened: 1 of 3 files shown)


luigi/worker.py
```diff
@@ -278,6 +278,9 @@ class worker(Config):
                                   'well as having keep-alive true')
     wait_interval = FloatParameter(default=1.0,
                                    config_path=dict(section='core', name='worker-wait-interval'))
+    wait_jitter = FloatParameter(default=5.0,
+                                 config_path=dict(section='core', name='worker-wait-jitter'))
```
Tarrasch: Don't add a `config_path`. It's only for old deprecated values.

Tarrasch:
Looks nice! But can you add a test case too?

Tarrasch:
Just out of curiosity, how is your setup so that you felt a need to parametrize this? What value do you plan to configure it to?

demoray:
I have two setups I'm working in.  

In one, I'd rather remove the jitter all together.  As such, I'd set it to 0.

In another, I'd prefer to shrink it.  Having jitter can be useful, but I've seen multiple workers end up with nearly an additional 5 seconds of wait time.

demoray:
Added unit tests per request, though it may false positive on over-loaded machines.  I'd welcome suggestions as to making the tests more reliable.

test/worker_test.py
```diff
@@ -968,3 +968,39 @@ def test_asserts_for_worker(self):
         """
         Worker(wait_interval=1)  # This shouldn't raise
         self.assertRaises(AssertionError, Worker, wait_interval=0)
+
+
+class WorkerWaitJitterTest(unittest.TestCase):
+    @with_config({'worker': {'wait_jitter': '0.0', 'wait_interval': "%f" % luigi.worker._WAIT_INTERVAL_EPS}})
```
Tarrasch: Isn't it better to just not set `wait_interval`?

Tarrasch:
Maybe it's better to write the tests with the mocking library? You can mock the sleep function, and dome something like `.assert_called_with(0, 10)`. The last tests in `test/worker_external_task_test.py` have something like that. You can also `git grep assert_called_with` for some inspiration.

Tarrasch:
Also. While I know a lot of the luigi tests are running slow, lets not add even more tests that do sleeping. It's nice if running the whole test suite goes fast. :)

demoray:
Updated the tests to use mock, for both uniform.random and time.sleep.  

I squashed the commits, as that will likely be your next comment.  ;)

Tarrasch:
I'm so predictable ain’t I? :p

Tarrasch:
Thanks for your patch @demoray !

PR #1298, opened 2015-10-09
Wire up tracking

daveFNbuck:
This adds an optional tracking_url_callback parameter to the run function of
tasks. The worker passes a callback that updates the scheduler with the url. The
scheduler can then display the url in the visualiser.

I combined this with a second commit to pass the tracking_url_callback to run_and_track_hadoop_job so that hadoop tracking urls will show up in the scheduler.

image
image

### DIFF
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -402,7 +402,7 @@ class HadoopJobRunner(JobRunner):
         self.end_job_with_atomic_move_dir = end_job_with_atomic_move_dir
         self.tmp_dir = False
 
-    def run_job(self, job):
+    def run_job(self, job, tracking_url_callback=None):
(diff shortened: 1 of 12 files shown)


Tarrasch:
Looks really cool!

But I won't merge this until we've released 2.0. If that's OK.

daveFNbuck:
Is that happening today or something? This would be nice for a new release, but I guess it's a bit new if the release is soon. It seems to be broken in Python 3 anyway.

Tarrasch:
Enschuldigung for the late review. But better late than never I suppose. :)

Tarrasch:
Ok. I looked through the implementation now. It looks good to me. Have you used this in production already? Does it work? :)

daveFNbuck:
Yeah, I've been using it in production and it works. It makes debugging of broken MR jobs _much_ easier.

PR #1299, opened 2015-10-09
Captures tracking urls more broadly

daveFNbuck:
Tracking url lines in my current hadoop job runner don't look like

   tracking url: http://whatever

so in order to be able to match how they're output in different systems, I've
broadened the matching to use a regex that includes the format I'm seeing.


Tarrasch:
Awesome! Thanks!

erikbern:
would be nice with unit tests for this but not urgent

daveFNbuck:
oops! I'll try to add some today

PR #1300, opened 2015-10-09
Make luigi always log task status changes

Tarrasch:
I find debugging luigi logs a bit hard. Most task status changes are
emitted when luigi client says "Scheduler MyTask() (DONE)". But not all
status changes are logged (but I still expect them to be). I've made it
output what it does on every call to /api/add_task.


Tarrasch:
Merging. I think this should make debugging easier.

PR #1302, opened 2015-10-09
clarify #569 multiple workers limitation

thatneat:
I stumbled over this for a while - updated docs to help the next person:
- added mention of the name of the config item (dashed not underscored was another stumbling point)
- multiple workers caveat is important
- mention of config file name wasn't the whole story so removed it


Tarrasch:
Thanks!

Issue #1301, opened 2015-10-09
New svg layout algorithm broken


Tarrasch:
The old graph looks like this:

![selection_079](https://cloud.githubusercontent.com/assets/294349/10395674/844577b6-6e9e-11e5-91ab-1a80cd0347ee.png)

and the new one looks like this:

![selection_081](https://cloud.githubusercontent.com/assets/294349/10395679/8cec2d9c-6e9e-11e5-8192-a2af0b5709d4.png)

@daveFNbuck, do you see directly what might have caused this? If not I can try to make it reproducible and  whatnot. I suppose this blocks 2.0 release.

daveFNbuck:
That's a big bug! It looks like every row ended up with 1 node. Not sure why they're all right-oriented. It looks like the graph is wider. Did you make the window wider before generating the second graph? I also see at least one node didn't get laid out, but still somehow has an edge. I don't know why this would happen. Does the bug remain if you revert my complexity fix?

Could you share a pickle file or some code to produce this dependency graph?

Tarrasch:
Hi @daveFNbuck 

Ok, here a starting point:

```python
import luigi

class FinalTask(luigi.Task):
    def requires(self):
        return []

class A(luigi.Task):
    def requires(self):
        yield FinalTask()

class B(luigi.Task):
    def requires(self):
        yield FinalTask()

class Recursive(luigi.Task):
    param = luigi.IntParameter()

    def requires(self):
        if self.param:
            yield Recursive(param=self.param - 1)
        else:
            yield FinalTask()

class Range(luigi.Task):
    def requires(self):
        for i in range(2):
            yield Recursive(param=i)
```

This is how you do it:

    python -m luigi --module hello Range && python -m luigi --module hello A

Then check out the *reverse* dependency graph of `FinalTask`

![selection_089](https://cloud.githubusercontent.com/assets/294349/10399064/ccedee50-6eb0-11e5-8053-3ce5bd19b6aa.png)

Actually you can just start debugging without also launching the `A` task.

![selection_090](https://cloud.githubusercontent.com/assets/294349/10399103/fae4c7c0-6eb0-11e5-86a3-c446ca4d7bbe.png)

The problem here is that they are all on the same line, though they shouldn't be.

daveFNbuck:
[...]

PR #1303, opened 2015-10-09
Reverts #1296 "Makes computeDepth take linear time"

daveFNbuck:
There was a bug in the assumptions of #1296, as it treated depth and height as
having a 1-1 relationship. This is only true with balanced trees. For example,
consider the following tree:

      A
     / \
    B   C
         \
          D

In this tree, B has height 0 and depth 1 while D has height 0 and depth 2. So
the graph will be laid out weirdly if we compute the depth of B and D by
subtracting their heights from the height of A.

There seems to be another bug in the implementation that can actually throw
B off of the graph, but I'm not sure exactly why that was happening.

This should address #1301.

### DIFF
--- a/luigi/static/visualiser/js/graph.js
+++ b/luigi/static/visualiser/js/graph.js
@@ -64,27 +64,18 @@ Graph = (function() {
     /* Compute the maximum depth of each node for layout purposes, returns the number
        of nodes at each depth level (for layout purposes) */
     function computeDepth(nodes, nodeIndex) {
-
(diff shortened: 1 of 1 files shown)


Tarrasch:
Thanks @daveFNbuck

daveFNbuck:
I now see what the second bug is. If you get the depths wrong, the second traversal gets them right and never sees the claimed height. So xOrder never gets set and we end up with a bunch of invalid operations that prevent the node from being placed properly. So it's all just the one bug.

PR #1306, opened 2015-10-13
Add copy function to HdfsTarget

kunminghouzz:
Allows copy function of underlying file system to be accessed
by HdfsTarget


Tarrasch:
Hmm... Tests would have been nice :)

Issue #1270, opened 2015-10-05
Unable to use empty parameters


demoray:
class A(luigi.Task):
   b = luigi.Parameter()
A(b="")

This generates the following exception:
luigi.server[23966] WARNING: Error saving Task history
Traceback (most recent call last):
  File "/tmp/luigi/local/lib/python2.7/site-packages/luigi/scheduler.py", line 1007, in _update_task_history
    self._task_history.task_started(task_id, host)
  File "/tmp/luigi/local/lib/python2.7/site-packages/luigi/db_task_history.py", line 94, in task_started
    task = self._get_task(task_id, status=RUNNING, host=worker_host)
  File "/tmp/luigi/local/lib/python2.7/site-packages/luigi/db_task_history.py", line 104, in _get_task
    task = self.tasks[task_id] = task_history.Task(task_id, status, host)
  File "/tmp/luigi/local/lib/python2.7/site-packages/luigi/task_history.py", line 38, in __init__
    self.task_family, self.parameters = task.id_to_name_and_params(task_id)
  File "/tmp/luigi/local/lib/python2.7/site-packages/luigi/task.py", line 64, in id_to_name_and_params
    return luigi.tools.parse_task.id_to_name_and_params(task_id)
  File "/tmp/luigi/local/lib/python2.7/site-packages/luigi/tools/parse_task.py", line 59, in id_to_name_and_params
    parsed = parser.parseString(task_id).asDict()
  File "/tmp/luigi/local/lib/python2.7/site-packages/pyparsing.py", line 1125, in parseString
    raise exc
ParseException: Expected ")" (at char 242), (line:1, col:243)

erikbern:
let's kill `id_to_name_and_params` I hate that function

stephenpascoe:
@erikbern I'm interested in trying to to get this fixed.  Is there somewhere that summarises the issue?  Where would be best to discuss a solution?

stephenpascoe:
Duplicate of #1171 and #1132.

erikbern:
We should just remove the function `id_to_name_and_params` and store all the parameters in the database

stephenpascoe:
So, are some parameters not stored in the database (task_history I assume)?

I also think there are other issues with task_id which makes using it anywhere as a key problematic.  Clearly it is not safe to parse if parameters contain characters like "," or "=".  You can't reconstruct the task_id from a dictionary of parameters because the serialisation order depends on the order parameters are defined in the class.

[...]

Issue #1171, opened 2015-08-27
Task history breaks if you specify an empty string as a parameter value


martinxsliu:
The task history interface currently uses the [id_to_name_and_params](https://github.com/spotify/luigi/blob/1596c06e75c65f09a0e824f03ccab36f8790525f/luigi/tools/parse_task.py) function to parse the task_id into a (task_family, {params}) tuple.

Unfortunately, the parsing patterns used is not robust to empty parameter values (e.g. task_id = 'Foo(bar=bar, baz=)' as the value token is omitted.

Does anyone have an easy fix for this, otherwise I'll try to submit a PR sometime soon?

Tarrasch:
No, there's no easy fix. It's just broken. :(

We would very happy to review a PR removing the use of that function. :)

erikbern:
ugh i've maintained for a long time that we need to kill the `id_to_name_and_params` function.

maybe i should just go ahead and do it. i hate it

martinxsliu:
Fixed by https://github.com/spotify/luigi/pull/1309

Issue #1132, opened 2015-08-13
WARNING: Error saving Task history


EncoreHQ:
When Luigi is saving the task history it just throws this error:

```
WARNING: Error saving Task history
Traceback (most recent call last):
  File "/mnt/anaconda/lib/python2.7/site-packages/luigi/scheduler.py", line 972, in _update_task_history
    self._task_history.task_finished(task_id, successful)
  File "/mnt/anaconda/lib/python2.7/site-packages/luigi/db_task_history.py", line 90, in task_finished
    task = self._get_task(task_id, status=event_name)
  File "/mnt/anaconda/lib/python2.7/site-packages/luigi/db_task_history.py", line 104, in _get_task
    task = self.tasks[task_id] = task_history.Task(task_id, status, host)
  File "/mnt/anaconda/lib/python2.7/site-packages/luigi/task_history.py", line 38, in __init__
    self.task_family, self.parameters = task.id_to_name_and_params(task_id)
  File "/mnt/anaconda/lib/python2.7/site-packages/luigi/task.py", line 64, in id_to_name_and_params
    return luigi.tools.parse_task.id_to_name_and_params(task_id)
  File "/mnt/anaconda/lib/python2.7/site-packages/luigi/tools/parse_task.py", line 58, in id_to_name_and_params
    parsed = parser.parseString(task_id).asDict()
  File "/mnt/anaconda/lib/python2.7/site-packages/pyparsing.py", line 1125, in parseString
    raise exc
ParseException: Expected ")" (at char 28), (line:1, col:29)
```

Anyone have a clue on what is going on?

erikbern:
you probably have a task with a parenthesis or something weird name

we *really* need to kill `id_to_name_and_params`. i've said it lots of times before. i should just go ahead and remove that function and all its references

mfcabrera:
This was apparently fixed by #1309 so closing

PR #1309, opened 2015-10-13
Remove id to name and params

stephenpascoe:
This PR removes the problematic function luigi.task.id_to_name_and_params().  It was being used extensively when the scheduler called TaskHistory methods.  Now these methods take a luigi.scheduler.Task() instance instead so that deserialisation of task_id is unnecessary.

I have added a test which would catch #1270, #1171 and #1132.

I have also optimised an SqlAlchemy query in DbTaskHistory().


erikbern:
This is great! The function has been deprecated long enough so I'm glad we're removing it. Awesome.

I think the test failure is just flakiness so I retriggered it.

luigi/task.py
```diff
@@ -58,12 +58,6 @@ class Task2(luigi.Task):
     Register._default_namespace = namespace
 
 
-def id_to_name_and_params(task_id):
```
Tarrasch: \o/

luigi/tools/parse_task.py
```diff
@@ -1,68 +0,0 @@
...
-# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
-# License for the specific language governing permissions and limitations under
-# the License.
-
-import warnings
-
-import pyparsing as pp
-
-
-def id_to_name_and_params(task_id):
```
Tarrasch: DIE DIE DIE DIE!!!!!!!!!!!!!!!!!!!!!!

Tarrasch:
LGTM. But it seems like the first 3 commits could be squashed to keep history to only logical changes.

stephenpascoe:
I squashed the commits and now Travis is failing again.  Looks like spurious failures.

Tarrasch:
Hmm...

selection_091

Well, it's not spurious. But something have made all tests fail. I just saw it in #1310 too. @ulzha, can you take a look? :)

erikbern:
the test breakages have something to do with elastic search so it's kind of weird

Tarrasch:
Since `id_to_name_and_params` will be shown in readthedocs (doh). I think this deserves merging before 2.0.

erikbern:
huh is it visible in the docs? if they are then we should remove that too

Tarrasch:
Every method that doesn't start with a `_` will be in the docs I think.

PR #1313, opened 2015-10-15
Rdbms post_copy addition

dlstadther:
Addition of post_copy method within luigi/contrib/rdbms.py.

This method allows for data cleansing, and other post-copy query functionality for rdbms-extended classes.


Tarrasch:
Needs tests, and somebody who knows this to review. :)

PR #1314, opened 2015-10-15
Redshift - Copy to temp table, prune table based on date, utilization of init_copy and post_copy

dlstadther:
Added support for redshift copy to temp table, prune table (delete some amount of records based on date), init copy, post copy (also added to rdbms.py)

Also includes the rdbms.py post_copy update found in PR #1313

### DIFF
--- a/luigi/contrib/rdbms.py
+++ b/luigi/contrib/rdbms.py
@@ -119,6 +119,15 @@ class CopyToTable(luigi.task.MixinNaiveBulkComplete, luigi.Task):
         if hasattr(self, "clear_table"):
             raise Exception("The clear_table attribute has been removed. Override init_copy instead!")
 
+    def post_copy(self, connection):
+        """
+        Override to perform custom queries.
+
+        Any code here will be formed in the same transaction as the main copy, just after copying data.
+        Example use cases include cleansing data in temp table prior to insertion into real table.
+        """
+        pass
+
     @abc.abstractmethod
     def copy(self, cursor, file):
         raise NotImplementedError("This method must be overridden")
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -103,6 +103,34 @@ class S3CopyToTable(rdbms.CopyToTable):
         """
         return ''
 
+    @property    
+    def prune_table(self):
+        """
+        Override to set equal to the name of the table which is to be pruned.
+        Intended to be used in conjunction with prune_column and prune_date
+        i.e. copy to temp table, prune production table to prune_column with a date greater than prune_date, then insert into production table from temp table
+        """
+        return None
+
+    @property
+    def prune_column(self):
+        """
+        Override to set equal to the column of the prune_table which is to be compared
+        Intended to be used in conjunction with prune_table and prune_date
+        i.e. copy to temp table, prune production table to prune_column with a date greater than prune_date, then insert into production table from temp table
+        """
+        return None
+
+    @property
+    def prune_date(self):
+        """
+        Override to set equal to the date by which prune_column is to be compared
+        Intended to be used in conjunction with prune_table and prune_column
+        i.e. copy to temp table, prune production table to prune_column with a date greater than prune_date, then insert into production table from temp table
+        """
+        return None
+
+    @property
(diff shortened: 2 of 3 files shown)


Tarrasch:
Can you either keep the PRs seperate or close #1313?

luigi/contrib/redshift.py
```diff
@@ -103,6 +103,15 @@ def copy_options(self):
         """
         return ''
 
+    def prune_table(self):
```
Tarrasch: Docstrings would be nice :)

Tarrasch:
It would be great if you can find somebody who uses redshift in luigi to review. Thanks :)

dlstadther:
I closed PR #1313 and will include updates here.

Luigi redshift users are quiet and unknown. Mortar Data uses some redshift functionality in their implementation of luigi, but they seem to have been inactive in the luigi community for awhile.

dlstadther:
As rdbms.py is a common module for which is to be implemented (not executed itself), I do not see how it needs its own tests. Postgres.py and Redshift.py both implement Rdbms.py; thus, their unit tests indirectly test rdbms.py.

test/contrib/redshift_test.py
```diff
@@ -44,10 +43,42 @@ class DummyS3CopyToTable(luigi.contrib.redshift.S3CopyToTable):
...
+        ('some_text', 'text'),
+        ('some_int', 'int'),
+    )
+
+    aws_access_key_id = AWS_ACCESS_KEY
+    aws_secret_access_key = AWS_SECRET_KEY
+    copy_options = ''
+    sql = ["insert into dummy_table select * from stage_dummy_table;"]
+    prune_date = 'current_date - 30'
+    prune_column = 'dumb_date'
```
Tarrasch: This looks weird to me. Is `prune_column` (as well as the other 2) a method or a property? I mean, is it `myobj.prune_column` or `myobj.prune_column()` that's supposed to be a string?

Tarrasch:
@dlstadther, oh ok. I didn't know that redshift inherited rdbms. If you think all code paths are tested. I'm fine with that :)

test/contrib/redshift_test.py
dlstadther: I intend them to be properties...optional properties. If I specify @property over them in redshift.py, will that allow them to be optional properties with 'None' as their default value?

dlstadther: @Tarrasch Do you have any suggestions on how to specify prune_* as optional properties such that they can be used like the test case?

Tarrasch: Can't you just do:

```
@property
def prune_table(self):
    return None
```

Or maybe even just `prune_table = None`? Hmm...

dlstadther: @Tarrasch If the `prune_*` variables should be marked as properties, shouldn't `table_type` and `table_attributes` also be marked as such?

moandcompany:
I'm currently using Luigi with Redshift, so I'll try these out when I get a chance.

luigi/contrib/rdbms.py
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 """
-A common module for postgres like databases, such as postgres or redshift
+A common module for posgres like databases, such as postgres or redshift
```
Tarrasch: huh? Introducing a spell error?

dlstadther: I'm not sure how that happened. I will fix and merge commits.

test/contrib/redshift_test.py
Tarrasch: Hmm. Good point, maybe it's better to stay consistent. Do something clever. :)

dlstadther:
+bump

(PRs are getting a little stagnate...I understand Luigi is undergoing a change in management, but the development must go on!)

Comments, questions, concerns, etc. are welcome.

dlstadther:
@erikbern Thanks for the quick action! Greatly appreciated!

erikbern:
yes no problem – I'm happy to merge most stuff that goes into contrib right away :) just forgot about it

let me check if other things are pending

PR #1315, opened 2015-10-15
Limit graph size

daveFNbuck:
Like with task lists, excessively large graphs can render the visualiser
unusable. To prevent this issue, we've added a max-graph-nodes parameter that
cuts off the graph after enough nodes are seen.

In order to achieve this, the graph recursion had to be re-written as a an
iterative breadth-first traversal. The inverse dependency graph function was
re-written to be able to call this same traversal, resulting in greatly improved
complexity. Rather than scanning the entire graph once per node to find its
reverse neighbors, we scan the whole graph exactly once and store the reverse
edges in a lookup table. This improves the complexity from number of nodes times
size of graph to just number of nodes.

When the scheduler sends a truncated graph, some nodes may have dependencies
that are not displayed. In order to not give a false impression to the end
user, these nodes are added back and marked as TRUNCATED status. These nodes
show that more dependencies exist and can be clicked on for more information
but mostly preserve a smaller graph.

In order to distinguish these nodes from UNKNOWN status nodes, I added UNKNOWN
and TRUNCATED to the legend. UNKNOWN remains black while TRUNCATED is given a
light purple.

example with max-graph-nodes set to 10:
image

I marked this as being part of the 2.0 release in the configuration documentation, but if that's not appropriate I can change it.

This change has been a hit with the Houzz team, as graph pages that previously hung forever have started showing something usable within 20 seconds (using a 1000 limit but having a root task that far exceeds it).

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -147,9 +147,19 @@ max-shown-tasks
   .. versionadded:: 1.0.20
 
   The maximum number of tasks returned in a task_list api call. This
-  will restrict the number of tasks shown in any section in the
+  will restrict the number of tasks shown in task lists in the
   visualiser. Small values can alleviate frozen browsers when there are
   too many done tasks. This defaults to 100000 (one hundred thousand).
+
+max-graph-nodes
+  .. versionadded:: 2.0.0
+
+  The maximum number of nodes returned by a dep_graph or
+  inverse_dep_graph api call. Small values can greatly speed up graph
+  display in the visualiser by limiting the number of nodes shown. Some
+  of the nodes that are not sent to the visualiser will still show up as
+  dependencies of nodes that were sent. These nodes are given TRUNCATED
+  status.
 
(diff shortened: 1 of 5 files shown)


daveFNbuck:
Note: truncated nodes are not shown in D3. I just realized I forgot to do that and I'm not sure how now. In my experience, graphs that need truncation aren't really usable in d3 anyway.

erikbern:
Cool – is there a way to expand those nodes?

daveFNbuck:
I figured out how to make this work for d3 too - I just added nodes before making the d3 or svg callbacks. I've done a force push to change the javascript commit.

daveFNbuck:
You can expand the nodes by clicking them. They're able to use the task id to link to the right graph.

daveFNbuck:
popover:
image

after clicking on truncated node:
image

in d3:
image

after clicking in d3:
image

PR #1316, opened 2015-10-16
Computes number of columns per graph level more accurately

daveFNbuck:
The number of columns per level was computed to be the minimum of the
number of columns allowed in the viewport and the number of items in the
row. This is an overestimate, however, due to the way that columns are
filled in. Suppose we have 6 columns and 7 items. We'll put two items in
each column, as ceil(7 / 6) = 2. Because there are only 7 items, we only
end up using 5 columns. Because we are laying out for 6, this means we
don't use enough horizontal space.

To fix this, we can either use a variable number of items per column or
adjust the layout to display the number of columns actually used. I opt
for the second choice here, as the code is simpler and it looks nicer
with the binary tree example graph I was using.

### DIFF
--- a/luigi/static/visualiser/js/graph.js
+++ b/luigi/static/visualiser/js/graph.js
@@ -110,8 +110,8 @@ Graph = (function() {
             return rowStartPosition(depth-1)+Math.max(rowHeight * nodeHeight + 100);
         }
         $.each(nodes, function(i, node) {
-            var levelCols = Math.min(rowSizes[node.depth], numCols)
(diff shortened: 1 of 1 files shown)


Tarrasch:
>  Because there are only 7 items, we only end up using 5 columns

Shouldn't it be "we only end up using 4 columns"?

Tarrasch:
Looks right to me, after staring at it for 5 minutes or so :)

daveFNbuck:
Yeah, the description should have said 4 columns.

PR #1318, opened 2015-10-18
Travis fixes (mostly ElasticSearch)

erikbern:
Noticed a lot of issues in Travis, trying this


miku:
Thanks for looking into this.

Tarrasch:
Thanks!!

erikbern:
interesting/scary that our tests have been running on python 3.4 only for the last month or so. luckily there were no regressions in python 2.7/3.3

Tarrasch:
@erikbern, nah, that's not true. It just says 3.4 in the label on Travis. It's `tox` that decides the python version, which is taken from the `.travis.yml` matrix section.

erikbern:
The `python --version` command returned 3.4 as well

Tarrasch:
That's expected, I think. When Travis says it's 3.4, it only means that the system python will be python 3.4. The first thing `tox` does is to install a python virtual env with a pinned python version.

Issue #1304, opened 2015-10-12
Unexpected behavior with command-line overrides of parameters with default values


daveFNbuck:
Using the following code example as test.py:
```
import luigi

class MyTask(luigi.Task):
    param1 = luigi.IntParameter()
    param2 = luigi.BoolParameter(default=False)

    def requires(self):
        if self.param1 > 0:
            yield MyTask(self.param1 - 1)

    def complete(self):
        return False

    def run(self):
        pass

if __name__ == '__main__':
    luigi.run()
```

if I run `python test.py MyTask --param1 1` I'll see the output I expect, with the following run summary:

```
===== Luigi Execution Summary =====

Scheduled 2 tasks of which:
* 1 ran successfully:
    - 1 MyTask(param1=0, param2=False)
* 1 failed:
    - 1 MyTask(param1=1, param2=False)

This progress looks :( because there were failed tasks

===== Luigi Execution Summary =====
```

but if I add `--param2` to set param2 to true in my root task, I get

```
===== Luigi Execution Summary =====

Scheduled 2 tasks of which:
* 1 ran successfully:
    - 1 MyTask(param1=0, param2=True)
* 1 failed:
    - 1 MyTask(param1=1, param2=True)

This progress looks :( because there were failed tasks

===== Luigi Execution Summary =====
```

which shows that I've actually changed param2 in _all_ tasks of the same class. This is pretty far from expected and normal behavior for member parameters. I usually only expect to be able to make global changes via parameters set in Config classes. This behavior seems to be new with #1182.

erikbern:
This is no good, we should fix

Tarrasch:
This change was intentional. One big win of #1182 was that global and local parameters were merged. There's no difference between `--MyTask-param2` and `--param2` any more. The former will always work (in case of conflicts) while `--param2` is shorter and more convenient, otherwise there's no difference.

I suppose that we could reintroduce the old behaviour without needing to rewrite everything, but then the parsing logic would start to become complex again. 

@daveFNbuck, Don't you like that you can override variables in tasks too? It have been convenient for me a few times. Currently there's no distinction between Config classes and Tasks, which is pretty nice from an implementers point of view as well. :)

Have there been any real world use cases that's affected (negatively) by this?

erikbern:
I see your point @Tarrasch but not sure if the new behavior is that much more logical. The old behavior wasn't super logical either tbh.

Let me think about this some more

daveFNbuck:
[...]

PR #1320, opened 2015-10-19
Distingiush local and global cli params [#1304]

Tarrasch:
Since the command line overhaul in #1182. One unexpected behavior
was introduced, namely that --x and --MyTask-x would mean the same
thing. As a consequence, `--x 123` would set parameter x to 123 not only
for the parsed root Task specified on the command line. But for all
parmeters x of that task family. Since even Dave from Houzz was confused
by this, we decided to revert that behavior in issue #1304.

While at it. I also made these changes:

 * I decided to do a quite big refactoring. Including finally
   removing the `is_bool` kwarg from the Parameter constructor.
 * I added a test for deps.py script, as I've already been close to
   breaking it twice.

### DIFF
--- a/luigi/cmdline_parser.py
+++ b/luigi/cmdline_parser.py
@@ -65,17 +65,17 @@ class CmdlineParser(object):
         # We have to parse again now. As the positionally first unrecognized
         # argument (the task) could be different.
         known_args, _ = self._build_parser().parse_known_args(args=cmdline_args)
-        self._task_name = known_args.task
-        parser = self._build_parser(active_tasks=self._active_tasks(),
+        root_task = known_args.task
+        parser = self._build_parser(root_task=root_task,
                                     help_all=known_args.core_help_all)
         self._possibly_exit_with_help(parser, known_args)
-        known_args = parser.parse_args(args=cmdline_args)
-        if not self._active_tasks():
+        if not root_task:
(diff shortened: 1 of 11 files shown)


Tarrasch:
@daveFNbuck, since you reported this bug. Would you like to review? Or at least take a look at the test cases added?

erikbern:
isn't the is_bool removal going to break some CLI invocations?

erikbern:
code tl;dr but otherwise lgtm

daveFNbuck:
Unit tests look like you've covered my case well. In test_local_takes_precedence, you run `MyTask --param 5 --MyTask-param 6` and check that the 5 takes precedence but there's no corresponding test that the 6 still works for non-root instances.

Tarrasch:
> isn't the is_bool removal going to break some CLI invocations?

It's going to break for people who've written code like this:

```
MyOwnBoolParameter(luigi.Parameter)
    def __init__(..):
         super(MyOwnBoolParameter, is_bool=True, ...)
```

In the code it said "only to be used by BoolParameter". But not in the documentation. I doubt many people have created their own parameter types that need  `store_true` aspect of bool semantics but still is something different from BoolParameter.

> but there's no corresponding test that the 6 still works for non-root instances.

Good point. I'll add.

PR #1326, opened 2015-10-20
Add __repr__() to Parameter for nicer docs

Tarrasch:
Currently, the documentation page for Range doesn't look nice. One
reason is that the parmeters are documented like this:

     start = <luigi.parameter.DateParameter object>

This patch makes it look like this:

     now = IntParameter (defaults to None): set to override current time. In seconds since epoch

Also see the screenshots in the upcoming PR.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -16,6 +16,36 @@ import sys
 import os
 import sphinx.environment
 from docutils.utils import get_source_line
+import luigi
+import luigi.parameter
+
+
+def parameter_repr(self):
+    """
+    When building documentation, we want Parameter objects to show their
+    description in a nice way
+    """
+    significance = 'Insignificant ' if not self.significant else ''
+    class_name = self.__class__.__name__
+    has_default = self._default != luigi.parameter._no_value
+    default = ' (defaults to {})'.format(self._default) if has_default else ''
+    description = (': ' + self.description if self.description else '')
+    return significance + class_name + default + description
+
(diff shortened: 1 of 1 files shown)


Tarrasch:
So

selection_094

Becomes

selection_095

demoray:
I do not care for this patch.  In my development process, I regularly use repr as a mechanism to enable debugging within REPL console, and this will seriously screw up my 

I have a patch in progress to modifies __repr__ to follow what repr is commonly used for: https://docs.python.org/2/library/functions.html#func-repr

Tarrasch:
Yea. I'm fine with nomerge on this. @demoray, do you happen to know what rules sphinx use to create the documentation? I was surprised that it used `__repr__` and not `__str__` as I've heard str is for supposed to be a human readable version (sensible for documentation).

Please share your patch where you define repr in the way it's supposed to be used.

Tarrasch:
@demoray, can you review again? Does it add value for you now?

demoray:
Looks ok.  I'll submit a pull request once I'm done with my __repr__ updates.

sisidra:
:+1:

ulzha:
Wow, surprised about involvement of __repr__ in this, but this is awesome.

Tarrasch:
Yea, I tried to google if there's anything else sphinx uses before doing `repr()`, but didn't find anything. I checked and it doesn't do `str()` either.

PR #1327, opened 2015-10-20
Delete the marker file (_$folder$) when removing a directory

QuentinFra:
When calling remove() on an S3Target pointing to a "directory", all files inside the directory are removed as expected. But when using Hadoop S3 Native FileSystem to store files in S3, each "directory" comes with a marker with the suffix _$folder$. This marker was not deleted by remove() causing bugs such as letting Hadoop think that the directory still exists when it does not. This patch fixes this.

### DIFF
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -145,6 +145,11 @@ class S3Client(FileSystem):
         delete_key_list = [
             k for k in s3_bucket.list(self._add_path_delimiter(key))]
 
+        # delete the directory marker file if it exists
+        s3_dir_with_suffix_key = s3_bucket.get_key(key + S3_DIRECTORY_MARKER_SUFFIX_0)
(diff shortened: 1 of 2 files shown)


Tarrasch:
Maybe add a test case? :)

QuentinFra:
@Tarrasch Just added a additional check in test_remove()

Tarrasch:
Merge the commits into one. They belong to the same logical sense.

It would also be great if you know somebody who knows S3 that can review this. Maybe you can look in the blame lines?

QuentinFra:
@Tarrasch @ddaniels888 Could you have a look at this please?

Tarrasch:
I'll merge once there's a +1 from somebody knowing s3.

ddaniels888:
:+1: looks good to me.

PR #1331, opened 2015-10-21
LUIGI_CONFIG_PATH Warning Message - File doesn't exist

dlstadther:
When LUIGI_CONFIG_PATH points to a file which does not exist, a warning message will be displayed.

### DIFF
--- a/luigi/configuration.py
+++ b/luigi/configuration.py
@@ -49,7 +49,11 @@ class LuigiConfigParser(ConfigParser):
         'luigi.cfg',
     ]
     if 'LUIGI_CONFIG_PATH' in os.environ:
-        _config_paths.append(os.environ['LUIGI_CONFIG_PATH'])
+        config_file = os.environ['LUIGI_CONFIG_PATH']
+        if not os.path.isfile(config_file):
+            warnings.warn("LUIGI_CONFIG_PATH points to a file which does not exist. Invalid file: {path}".format(path=config_file))
+        else:
+            _config_paths.append(config_file)
 
     @classmethod
     def add_config_path(cls, path):


luigi/configuration.py
```diff
@@ -49,7 +49,10 @@ class LuigiConfigParser(ConfigParser):
         'luigi.cfg',
     ]
     if 'LUIGI_CONFIG_PATH' in os.environ:
-        _config_paths.append(os.environ['LUIGI_CONFIG_PATH'])
+        if not os.path.isfile(os.environ['LUIGI_CONFIG_PATH']):
+            warnings.warn("LUIGI_CONFIG_PATH is not a valid path to configuration file. Invalid file: {path}".format(path=os.environ['LUIGI_CONFIG_PATH']))
```
Tarrasch: Should you say that the path is not valid? That sounds like the path-syntax is malformed. Isn't it better to simply say that the file doesn't exit?

luigi/configuration.py
```diff
@@ -49,7 +49,10 @@ class LuigiConfigParser(ConfigParser):
         'luigi.cfg',
     ]
     if 'LUIGI_CONFIG_PATH' in os.environ:
-        _config_paths.append(os.environ['LUIGI_CONFIG_PATH'])
+        if not os.path.isfile(os.environ['LUIGI_CONFIG_PATH']):
```
Tarrasch: Can you extract `os.environ['LUIGI_CONFIG_PATH']` to a variable?

Tarrasch:
Thanks for the PR. I just had some small comments :)

dlstadther:
@Tarrasch Do you have any additional comments regarding this PR?

Tarrasch:
I think you should merge the commits into one. They are all about the same logical change.

A test would be nice too of course. :)

dlstadther:
@Tarrasch I'll merge the commits when I get into work tomorrow.

Regarding a test, I can't seem to find any pre-existing tests for configuration.py. Am I being oblivious? Or do they not exist yet?

Thanks

Tarrasch:
It's probably tested through some code path[1]. But it would make sense to me to have a test file for it too.

[1]: https://codecov.io/github/stevepeak/luigi/luigi/configuration.py?ref=e4714696eee167a1811f7590f0d95417f4f17f9e

PR #1334, opened 2015-10-21
Keeps scheduler updated with running tasks during ping

daveFNbuck:
If the rpc fails to receive a get_work response, it will do an automatic retry
and invisibly return the results of the next call. After this point, the
scheduler will think that the worker is RUNNING the task from the first get_work
call while the worker has no idea that anything happened.

Without further communication, the scheduler will continue to think that the
worker is doing the task until it dies. In the worst case, the scheduler will
end up doing this to all of the leaf tasks for a given worker, which will then
be stuck forever requesting work that depends on tasks that are perputually
RUNNING. This case has happened to me in production.

In order to avoid this issue, this commit adds a list of currently running tasks
to some pings. This list is optional is can be left out of the majority of pings
to keep the pings small and allow for backward compatibility with older workers.
When the scheduler sees that a RUNNING task is not in the sent task list, it
marks that task as FAILED after two consecutive such pings. Two pings are
required to avoid race conditions when tasks have just started or finished.

In order to also handle a related case where the main worker thread dies but the
KeepAliveThread and subprocesses are still alive, the KeepAliveThread checks
that each task is still alive before including it in a ping. This way a long-
running task won't keep all other tasks marked as RUNNING in the scheduler after
the main worker thread stops updating the running task list.

The scheduler could potentially see that a worker is running tasks that are not
RUNNING in the scheduler. I'm not sure what to do about that, so I've left it as
a TODO for now.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -30,6 +30,7 @@ import logging
 import multiprocessing  # Note: this seems to have some stability issues: https://github.com/spotify/luigi/pull/438
 import os
 import signal
+import itertools
 
 try:
     import Queue
@@ -90,6 +91,7 @@ class TaskProcess(multiprocessing.Process):
         self.worker_id = worker_id
         self.result_queue = result_queue
         self.random_seed = random_seed
+        self._running_locally = False
         if task.worker_timeout is not None:
             worker_timeout = task.worker_timeout
         self.timeout_time = time.time() + worker_timeout if worker_timeout else None
@@ -116,6 +118,16 @@ class TaskProcess(multiprocessing.Process):
                 next_send = getpaths(requires)
             else:
                 return new_deps
+
+    def is_alive(self):
+        return self._running_locally or super(TaskProcess, self).is_alive()
+
+    def run_single_threaded(self):
+        self._running_locally = True
+        try:
+            self.run()
+        finally:
+            self._running_locally = False
 
     def run(self):
         logger.info('[pid %s] Worker %s running   %s', os.getpid(), self.worker_id, self.task.task_id)
@@ -268,6 +280,10 @@ class worker(Config):
 
     ping_interval = FloatParameter(default=1.0,
                                    config_path=dict(section='core', name='worker-ping-interval'))
+    task_ping_ratio = IntParameter(default=60,
+                                   description='Pings will include the running tasks once in this '
+                                               'many pings. Defaults to 60 for once a minute with '
+                                               'ping_interval of 1 hz')
     keep_alive = BoolParameter(default=False,
                                config_path=dict(section='core', name='worker-keep-alive'))
     count_uniques = BoolParameter(default=False,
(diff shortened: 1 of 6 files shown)


luigi/worker.py
```diff
@@ -268,6 +280,11 @@ class worker(Config):
 
     ping_interval = FloatParameter(default=1.0,
                                    config_path=dict(section='core', name='worker-ping-interval'))
+    task_ping_ratio = IntParameter(default=60,
+                                   config_path=dict(section='core', name='worker-task-ping-ratio'),
```
Tarrasch: remove `config_path`

Tarrasch: I mean. It's deprecated, we just have it for being able to keep backward compatibility

erikbern:
Isn't it possible to solve the original problem where `get_work` fails? Or is this similar to the Two Generals' Problem? https://en.wikipedia.org/wiki/Two_Generals%27_Problem

I'm looking forward to the Jepsen analysis of Luigi :)

daveFNbuck:
Yeah, I think it's a Two Generals' Problem. I guess we could include a list of running tasks in get_work rather than ping and have the scheduler re-send tasks that the worker doesn't know are supposed to be running.

There's also the problem where throwing any exception in the main worker thread while subprocesses are running will only kill that thread and result in the scheduler thinking all tasks are running until they all finish and the keep alive thread finally dies. This only happens to me with weird get_work failures right now (somehow returning None) but I don't see another generic way to handle this without killing the subprocesses.

The pinging solves both problems, although it does add extra delays by failing the tasks instead of marking them available for get_work again right away. I guess we could go straight to pending or at least keep it failed for less time.

luigi/worker.py
daveFNbuck: Done, but that moves task-ping-ratio to be the only item in the [worker] section in configuration.rst. Can we move all the  all and keep the older names under an older version number in readthedocs?

Tarrasch:
> There's also the problem where throwing any exception in the main worker thread while subprocesses are running will only kill that thread and result in the scheduler thinking all tasks are running until they all finish and the keep alive thread finally dies. This only happens to me with weird get_work failures right now (somehow returning None) but I don't see another generic way to handle this without killing the subprocesses.

I thought there was a `finally: keep_alive_thread.terminate()` wrapped around the worker code. If not, that should be enougth to solve this issue. No?

daveFNbuck:
> I thought there was a finally: keep_alive_thread.terminate() wrapped around the worker code. If not, that should be enough to solve this issue. No?

No, Worker.run isn't wrapped in anything. The problem with that approach is that the worker subprocess keep going. So you lose the protection of the scheduler guaranteeing that your task is only being run once at a time. We could fix that by also killing all subprocesses in Worker.stop, but I tend to prefer to let them finish.

Tarrasch:
In my mind, it's fine if a task gets scheduled twice under exceptional behaviour. I mean, that's why we have the whole `mv mydirectory-luigi-tmp-123456 mydirectory` dance in all hdfs jobs. I guess we've never explicitly decided on if we assume tasks to be atomic and idempotent and have all those nice properties.

daveFNbuck:
My tasks are atomic and idempotent as much as possible, but I'm not really comfortable with rogue tasks running without the scheduler's knowledge if it can be avoided. If a task is being run twice at the same time and the one that the scheduler's aware of finishes first, we might end up atomically overwriting the output of the job the scheduler knew about while a dependent is using it as input. This will probably just fail or kill a dependent mapreduce job, but can have other unexpected effects with different job types.

daveFNbuck:
Example: say I have a job that copies one file at a time from hdfs and processes it before fetching the next file. If it's processing the output of an MR, a second run that finishes while my job is running can overwrite the same number of files but with the contents shuffled around. If this happens between fetches, my processing script will end up processing some lines twice and others not at all.

daveFNbuck:
I think I see a nicer pair of solutions that separates these problems. I'll try to submit a new pair of pull requests later today.

daveFNbuck:
I think #1340 and #1341 make for cleaner solutions to these problems. Let me know what you guys think.

Issue #1335, opened 2015-10-21
Parameters named "task" no longer working on CLI


daveFNbuck:
Most of my Tasks have a parameter named task. This was fine until I recently grabbed the latest code and found that it's treating this flag as an override for the positional task argument. This will be very annoying for me to write out of my code. Is there a strong reason that it needs to be like this?

daveFNbuck:
If this is necessary, there should at least be an error when I try to name a parameter "task", as I would have at least been able to detect this in my unit tests before deploying the new luigi.

Tarrasch:
Thanks for reporting!!! I'll fix right away!

daveFNbuck:
Thanks for the quick fix!

PR #1336, opened 2015-10-22
cli: Allow for --task as a parameter name

Tarrasch:
Note that this does not make --task specify the task name as it did with
the old OptParser.

I didn't find a general solution, with this implementation, the name
--root-task still fails to parse. I could chose an arbitrary string, but
I figured it's kind of "good" to prohibit users from calling a parameter
--root-task, as it would be confusing that won't be the actual root
task.

This fixes spotify/luigi#1335.

### DIFF
--- a/luigi/cmdline_parser.py
+++ b/luigi/cmdline_parser.py
@@ -65,7 +65,7 @@ class CmdlineParser(object):
         # We have to parse again now. As the positionally first unrecognized
         # argument (the task) could be different.
         known_args, _ = self._build_parser().parse_known_args(args=cmdline_args)
-        root_task = known_args.task
+        root_task = known_args.root_task
         parser = self._build_parser(root_task=root_task,
                                     help_all=known_args.core_help_all)
         self._possibly_exit_with_help(parser, known_args)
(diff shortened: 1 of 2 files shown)


Tarrasch:
Can you review @daveFNbuck?

erikbern:
lgtm but i don't get it – is the problem a name clash between parameters?

Tarrasch:
The user-problem should be pretty obvious from the test case.

As for the technical problem: When doing `parser.add("task")`. It for some reason also clobbers with `parser.add("--task")`, even though `"task"` is positional and `"--task"` is passed by flag.

erikbern:
ok cool i think you could also use `dest` for this but it doesn't really matter

https://argparse.googlecode.com/svn/trunk/doc/add_argument.html#dest

Tarrasch:
Aha! The collision was maybe on the `dest` level. Nevermind. `root_task` is still a better name. :)

Tarrasch:
Still feel free to audit @daveFNbuck

daveFNbuck:
Looks fine to me.

PR #1341, opened 2015-10-22
Adds a current_tasks argument for retransmission in get_work

daveFNbuck:
Sometimes a worker will not receive the result of a get_work call from the
scheduler. This leaves the scheduler thinking that the worker is running tasks
that the worker is not running. In order to safely clear up the misunderstanding
we let the scheduler know which tasks the worker is running with each get_work
call. If the scheduler finds that the worker isn't aware of a task that it
should be running, the scheduler will send it to the worker.

In order to maintain backward compatibility and not have to rewrite every unit
test, leaving out current_tasks from the get_work call will be treated as
equivalent to sending every currently running task.

Note that this does not check that every task sent in the currently running
list is actually running.

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -169,10 +169,15 @@ class RemoteScheduler(Scheduler):
             'assistant': assistant,
         })
 
-    def get_work(self, worker, host=None, assistant=False):
+    def get_work(self, worker, host=None, assistant=False, current_tasks=None):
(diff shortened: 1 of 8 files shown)


Tarrasch:
> If the scheduler finds that the worker isn't aware of a task that it
should be running, the scheduler will send it to the worker.

This was a very clever solution! :)

PR #1342, opened 2015-10-23
docs: Unbreak readthedocs build

Tarrasch:
I apparently broke, see:

https://readthedocs.org/projects/luigi/builds/3418481/

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -16,36 +16,39 @@ import sys
 import os
 import sphinx.environment
 from docutils.utils import get_source_line
-import luigi
-import luigi.parameter
 
 
-def parameter_repr(self):
-    """
-    When building documentation, we want Parameter objects to show their
-    description in a nice way
-    """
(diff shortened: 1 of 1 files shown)


Tarrasch:
Seems to have unbroken it: https://readthedocs.org/projects/luigi/builds/3424040/

But seems like the parameters still are not looking as I intended them http://luigi.readthedocs.org/en/latest/api/luigi.tools.range.html?highlight=rangehourly#luigi.tools.range.RangeBase.of, relating to #1326.

Tarrasch:
Ok. I tried to fix this but give up.

I tried enabling the checkbox in readthedocs saying "Install your project inside a virtualenv using setup.py install". But then I got error logs like [1], and other people have had exact same issue [2].

[1]: https://readthedocs.org/projects/luigi/builds/3424094/
[2]: https://www.mail-archive.com/python-daemon-devel@lists.alioth.debian.org/msg00111.html

Tarrasch:
:D :D :D

selection_113

PR #1345, opened 2015-10-23
Retries get_work on null response

daveFNbuck:
Sometimes the scheduler gives a null response to get_work, resulting in an error
when the worker tries to treat the response as a dict and killing the worker.
This change will retry up to 3 times upon receiving a null response, allowing
the worker to live and continue working.

Note: this has a merge conflict with #1341 because they both modify get_work but it's easily resolved.

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -139,12 +139,15 @@ class RemoteScheduler(Scheduler):
             )
         return response
 
-    def _request(self, url, data, log_exceptions=True, attempts=3):
+    def _request(self, url, data, log_exceptions=True, attempts=3, allow_null=True):
         body = {'data': json.dumps(data)}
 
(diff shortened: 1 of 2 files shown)


daveFNbuck:
Note: this has been happening to me about 4 times a day since recently moving my scheduler to a different server from my workers.

erikbern:
LGTM but why does the scheduler return a null result? That sounds like the underlying issue

daveFNbuck:
I have no idea how that happens or whether it's a network issue or a scheduler issue. I don't think my logs go back far enough to check.

erikbern:
oops seems like there's a merge conflict here. any chance you can rebase?

daveFNbuck:
Done. Thanks for all the merges today!

PR #1349, opened 2015-10-26
Do not propagate luigi-interface logs

rouzwawi:
This causes duplicate output if the root logger is configured.

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -45,6 +45,7 @@ def setup_interface_logging(conf_file=None):
     if conf_file is None:
         logger = logging.getLogger('luigi-interface')
         logger.setLevel(logging.DEBUG)
+        logger.propagate = False
 
         stream_handler = logging.StreamHandler()
         stream_handler.setLevel(logging.DEBUG)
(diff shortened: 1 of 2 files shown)


Tarrasch:
Cool! Can you perhaps add a test case? I think it could be added to `cmdline_test.py` since it allows you to introspect stdout.

ulzha:
The "recommended construction" is `logging.getLogger(__name__)`, https://docs.python.org/2/library/logging.html#logger-objects - so tasks using logging.getLogger() are doing it wrong, let's perhaps fix them instead? They're disrupting the hierarchy (capturing logs of interface.py which orchestrates them in the first place). (And logging.getLogger('luigi-interface') is also doing it wrong...)

Tarrasch:
@rouzwawi, Please make sure to squash the commits before merging. I always tried to ensure we have a very clean commit log in luigi.

On a related note, avoid commits like https://github.com/spotify/luigi/commit/bb387a2baeaa400d71b34bc5b95380c3ee6eb286, and as far as possible, avoid pushing anything (even non-master*) to spotify/luigi, let everything come in through your personal fork.

*This view is a bit polluted right now: https://github.com/spotify/luigi/branches

PR #1352, opened 2015-10-26
Fixes two small bugs with scheduler auto-disables

daveFNbuck:
First, when the scheduler automatically disables a job, it becomes re-enabled after disable_persist seconds by changing the status from DISABLED to FAILED. This is a problem if the scheduler is set to disable after 1 failure, as it immediately gets disabled again, making the disable effectively permanent. This is fixed by not counting the status change from DISABLED to FAILED as a failure toward the number needed to automatically disable.

Second, when the scheduler disables a job, that prevents most other status changes from taking effect. Due to oversight, this included setting the job as permanently DISABLED. So if you tried to disable a job that the scheduler had already disabled, it would still become re-enabled again after disable_persist seconds. This fixes that bug by allowing workers to set tasks that the scheduler has already disabled to status DISABLED.

I've combined both of these because they're both very simple fixes and will cause merge conflicts with each other if done separately. Also I only found out about the first bug while writing simplified tests for the second bug and I'd like to be able to use those.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -405,7 +405,7 @@ class SimpleTaskState(object):
             elif task.scheduler_disable_time is not None:
                 return
 
-        if new_status == FAILED and task.can_disable():
+        if new_status == FAILED and task.can_disable() and task.status != DISABLED:
             task.add_failure()
             if task.has_excessive_failures():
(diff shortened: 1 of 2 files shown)


erikbern:
sgtm!

PR #1357, opened 2015-10-28
Properly handles errors in requires functions during scheduling

daveFNbuck:
We had a test that bugs in requires were properly caught and dealt with,
allowing scheduling to continue with other tasks. Unfortunately, this test
didn't really work properly because it assumed a depth-first scheduling but used
the breadth-first single-threaded scheduling. After adjusting the test to throw
an exception before the rest of the tasks were done scheduling, it failed.

In order to fix this, we wrap the call to deps in Worker._add to catch any
exceptions and log, e-mail, and not schedule the task or its dependencies. This
matches the behavior of errors in the complete function and prevents one bad
task from blocking an entire pipeline.


erikbern:
lgtm

PR #1358, opened 2015-10-28
Do not propagate luigi-interface logs

rouzwawi:
This causes duplicate output if the root logger is configured.

### DIFF
--- a/test/cmdline_test.py
+++ b/test/cmdline_test.py
@@ -84,6 +84,17 @@ class FooSubClass(FooBaseClass):
     pass
 
 
+class LoggingTask(luigi.Task):
+
+    def complete(self):
+        return False
+
+    def run(self):
+        import logging
+        logging.getLogger().addHandler(logging.StreamHandler())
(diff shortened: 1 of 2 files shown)


test/cmdline_test.py
```diff
@@ -84,6 +84,17 @@ class FooSubClass(FooBaseClass):
...
 
 
+class LoggingTask(luigi.Task):
+
+    def complete(self):
+        return False
+
+    def run(self):
+        import logging
+        logging.getLogger().addHandler(logging.StreamHandler())
```
ulzha: Can we have a comment here that this test is actually testing a malpractice, i.e. that https://docs.python.org/2/library/logging.html#logger-objects should be comprehended and `logging.getLogger(__name__)` should be used? Then +1.

ulzha:
Duplicate of https://github.com/spotify/luigi/pull/1349? -1, as commented in https://github.com/spotify/luigi/pull/1349#issuecomment-151829136

Issue #1360, opened 2015-10-29
PR  #1137 yields SIGUSR1 AttributeError on Windows


danielsack:
PR #1137 put in place `signal.SIGUSR1` which breaks luigi on Windows machines. Replacing all instances of `SIGUSR1` with `SIGTERM` appears to fix the issue.

Specifically, the error generated:
```
(env) PS C:\Users\...> python .\hello_world.py
python : 'ps' is not recognized as an internal or external command,
At line:1 char:1
+ python .\hello_world.py
+ ~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: ('ps' is not rec...ternal command,:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
operable program or batch file.
Traceback (most recent call last):
  File ".\hello_world.py", line 12, in <module>
    luigi.run(['examples.HelloWorldTask', '--workers', '1', '--local-scheduler'])
  File "C:\Users\...\env\lib\site-packages\luigi\interface.py", line 207, in run
    return _run(*args, **kwargs)['success']
  File "C:\Users\...\env\lib\site-packages\luigi\interface.py", line 235, in _run
    return _schedule_and_run([cp.get_task_obj()], worker_scheduler_factory)
  File "C:\Users\...\env\lib\site-packages\luigi\interface.py", line 185, in _schedule_and_run
    scheduler=sch, worker_processes=env_params.workers, assistant=env_params.assistant)
  File "C:\Users\...\env\lib\site-packages\luigi\interface.py", line 138, in create_worker
    scheduler=scheduler, worker_processes=worker_processes, assistant=assistant)
  File "C:\Users\...\env\lib\site-packages\luigi\worker.py", line 370, in __init__
    signal.signal(signal.SIGUSR1, self.handle_interrupt)
AttributeError: 'module' object has no attribute 'SIGUSR1'
```

erikbern:
can you send a PR

danimcgoo:
daniel do you want to do this? I'm happy to send a PR. Let me know.

danielsack:
@erikbern @danimcgoo - happy to; will submit in a min or two

erikbern:
@daveFNbuck you added the SIGUSR1 signal in 3800a986 – any issues changing it to SIGTERM?

danielsack:
@erikbern forgive my lack of familiarity with this whole process, but do I not have permission to commit branches?

[...]

PR #1361, opened 2015-10-29
Replace all instances of SIGUSR1 with SIGTERM; addresses #1360

danielsack:
PR #1137 put in place `signal.SIGUSR1` which breaks luigi on Windows machines. Replacing all instances of `SIGUSR1` with `SIGTERM` appears to fix the issue.


erikbern:
lgtm

erikbern:
maybe we should throw `getattr(signal, 'SIGTERM', signal.SIGUSR1)` or something

danielsack:
Fixed by #1363

PR #1172, opened 2015-08-28
pass pig parameters by file

toidi:
PigJobTask expose sensitive data to logs. For example:

```
INFO [pid 2464] Worker Worker(salt=466800829, workers=1, host=win91.dev.mail.ru, username=e.iskandarov, pid=2464) running   FalsePositive(java_home=/usr/lib/jvm/jre, secrets_conf=contrib/mario_secrets.yaml, site=mail.ru, max_number_of_files=180, mailru_input_dir=/data/mras/logs/dstat_falsepositive_snippets, mycom_input_dir=/data/mras/logs/dstatr_falsepositive_snippets)
INFO /usr/lib/pig/bin/pig -p mpop_user_jdbc=jdbc:mysql://mras-test2.mail.ru:3306/mPOP?user=mras&password=mras -p hdfs_input=/data/mras/logs/dstat_falsepositive_snippets/dstat_falsepositive_snippets-2014-09-20_00001 -p mras_jdbc=jdbc:mysql://mras-test2.mail.ru:3306/mras?useUnicode=true&characterEncoding=UTF-8&user=mras&password=mras -p site=mail.ru -p udf_path=/home/e.iskandarov/workspace/mario-jobs/mario_jobs/false_positive/rule_udf.py -p mpop_domain_jdbc=jdbc:mysql://mras-test2.mail.ru:3306/mPOP?user=mras&password=mras -p job_name=false_positive_processing -propertyFile pig_property_file -f /home/e.iskandarov/workspace/mario-jobs/mario_jobs/false_positive/false_positive.pig
```
As you can see there are mysql connection parameters. This is not desired behavior.

With my patch luigi build pig command line arguments using pig `-param_file` parameter like `-propertyFile` do. All parameters stay in file and do not visible in logs:

```
INFO [pid 2009] Worker Worker(salt=314780808, workers=1, host=win91.dev.mail.ru, username=e.iskandarov, pid=2009) running   FalsePositive(java_home=/usr/lib/jvm/jre, secrets_conf=contrib/mario_secrets.yaml, site=mail.ru, max_number_of_files=180, mailru_input_dir=/data/mras/logs/dstat_falsepositive_snippets, mycom_input_dir=/data/mras/logs/dstatr_falsepositive_snippets)
INFO /usr/lib/pig/bin/pig -param_file pig_param_file -propertyFile pig_property_file -f /home/e.iskandarov/workspace/mario-jobs/mario_jobs/false_positive/false_positive.pig
```

PR #1366, opened 2015-11-04
Temporary pig properties file is not flushed

alexeygrigorev:
Need to flush temporary files so the content is visible for pig processed (see #1172 and [my comment](https://github.com/spotify/luigi/pull/1172#issuecomment-153716963))


toidi:
@alexeygrigorev, thank you. It was my fault. I forgot to add flush.

Issue #1364, opened 2015-10-29
Dependency graph fails sometimes


Massou31:
hi,
With luigi 2.0.0, sometimes the task dependency graph fails to display with errors like:
```Couldn't find task DailyBiWorkflow(executor=all,%20date=2015-09-12,%20with_sftp_upload=False,%20with_reconciliation_analyse=False)```

There are no errors in luigi logs.
I think this may be caused by a failure in the urlencode but I have not investigated in it yet.

is there some parameters to luigi task that may cause this bug (to long strings, python objects, ...)?

Massou31:
Hi,
the issue is solved.

dlstadther:
If this has been resolved, could you close the issue? Thanks!

PR #1367, opened 2015-11-04
Fixes to visualizer for tasks with multi-line parameters

patricksmith:
When running tasks that have a string parameter that span multiple lines, there were a couple small issues with the visualizer (perhaps this is related to #1364). For example, this is a task that would cause issues:

```
$ cat multiline_config.cfg
[Example]
a_parameter: This is a parameter that spans
  multiple lines.

$ cat multiline_example.py
import time
import luigi


class Example(luigi.Task):

    a_parameter = luigi.Parameter()

    def output(self):
        []

    def run(self):
        time.sleep(30)

$ LUIGI_CONFIG_PATH=multiline_config.cfg luigi --module multiline_example Example
```

When visiting the index page, the task is not displayed and spinners are shown instead of the number of tasks in each category:

image

This is because of a regex using `.*`, which does not match across line breaks, when parsing the task ID. 71dd897 fixes this so the tasks and category numbers are properly displayed:

image

Now, the tasks shows up, but when clicking on a link to the dependency graph, the url is `http://localhost:8082/static/visualiser/index.html#Example(a_parameter=This is a parameter that spansmultiple lines.)`, which loses the newline character between "spans" and "multiple", causing the task to not be found:

image

9cb2913 fixes this by URI encoding the task id and decoding before retrieving the dependency graph:

image

### DIFF
--- a/luigi/static/visualiser/index.html
+++ b/luigi/static/visualiser/index.html
@@ -224,7 +224,7 @@
 
         <script type="text/template" name="actionsTemplate">
             <div class="span2">
-                <a href="#{{taskId}}" class="btn btn-info btn-xs" title="View graph" data-toggle="tooltip"><i class="fa fa-sitemap"></i></a>
+                <a href="#{{encodedTaskId}}" class="btn btn-info btn-xs" title="View graph" data-toggle="tooltip"><i class="fa fa-sitemap"></i></a>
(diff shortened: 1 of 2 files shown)


erikbern:
We shouldn't be doing this – we should just display the parameters straight from the source. The task ID is not a serialization mechanism and we already cleaned this up on the backend

I'm fine merging this just making a note that this approach is not sustainable

PR #1370, opened 2015-11-05
Notification.py Authenticating Gmail

dlstadther:
Luigi's notification.py previously would not successfully authenticate a gmail account. This PR adds `smtp.ehlo_or_helo_if_needed()` and `smtp.starttls()` prior to attempting to login to the email account.


dlstadther:
See [this post](http://stackoverflow.com/questions/10147455/trying-to-send-email-gmail-as-mail-provider-using-python) for specific error message one receives.

erikbern:
nice!

dlstadther:
@erikbern Thanks!

PR #1372, opened 2015-11-06
add self.path to S3Target

### LINKED ISSUE #1371: Add method to S3Target to get back path
With an `S3Target(path)`, we can load data into S3 at the specified path. It would be useful if we could also get back the path at which the target is pointing to so it can be used in subsequent tasks.

Correct me if I'm wrong, but I couldn't find any way to do this. I can work on a PR if this sounds reasonable.

matthewdu:
closes #1371


erikbern:
LGTM nice!

PR #1377, opened 2015-11-08
play ok with no req.

meyerson:
i needed to add this to get a my tasks to finish w/out dumb errors


Tarrasch:
You can open a new PR when you've got time to create a proper description and add a test case.

PR #1378, opened 2015-11-09
Abstract worker fork handlers (For workers with worker_processes > 1)

nkdhny:
We are using Crypto.Random in our setup, it must be reinited (call Crypto.Random.atfork()) each time after fork, in both forked process and in child process. `TaskProcess` already contains code block to do things like this one (see usages of `random_seed`), but client is unable to extend this code.
Fork handlers are introduced in this PR to give client ability to specify code to be executed after fork. Three types of fork handlers are introduced:

* ForkHandlerType.Master - to be called only in master process
* ForkHandlerType.Child to be called only in a child process
* ForkHandlerType.Both - to be called in both master and child process

Call to `random.seed` rewritten using this approach.

In our client code we use decorator `Worker.fork_handler` to specify fork handlers, for example we use following code to re-init pycrypto after the fork

```python
@worker.Worker.fork_handler()
def reinit_random_on_fork():
    logger.info("Reinit random for thread {}".format(os.getpid()))
    Random.atfork()
```

### DIFF
--- a/doc/parameters.rst
+++ b/doc/parameters.rst
@@ -2,7 +2,7 @@ Parameters
 ----------
 
 Parameters is the Luigi equivalent of creating a constructor for each Task.
-Luigi requires you to declare these parameters instantiating
+Luigi requires you to declare these parameters by instantiating
 :class:`~luigi.parameter.Parameter` objects on the class scope:
 
 .. code:: python
@@ -11,7 +11,7 @@ Luigi requires you to declare these parameters instantiating
         date = luigi.DateParameter(default=datetime.date.today())
         # ...
 
-By doing this, Luigi can do take care of all the boilerplate code that
+By doing this, Luigi can take care of all the boilerplate code that
 would normally be needed in the constructor.
 Internally, the DailyReport object can now be constructed by running
 ``DailyReport(datetime.date(2012, 5, 10))`` or just ``DailyReport()``.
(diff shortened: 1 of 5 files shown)


erikbern:
Why can't you just check the pid and reinitialize if the pid changes?

nkdhny:
1) 'atfork' should be called in both, child and master processes. Pid of
master process remains same.
2) what does 'changes' mean? I mean task process uses its own thread.
Probably, I can store pid of the worker in the constructor of my task, then
in 'run' call 'atfork' and store updated pid again. But how can I do the
same for worker? Moreover some complicated logic could be implemented in
'complete' or 'requires' etc. and maybe 'atfork' need to be called in those
methods too.

That's why I think your approach doesn't solve the problem.

Btw, why don't 'luigi' ask client to call 'random.seed' when pid changes,
imho we are trying to solve same problem.
Пн, 9 нояб. 2015 г. в 18:50, Erik Bernhardsson <notifications@github.com>:

erikbern:
I just mean something like this 

```python
class Random(object):
    def __init__(self):
        self._pid = None
    def random(self):
        if os.getpid() != self._pid:
            reinitialize_random()
            self._pid = os.getpid()
        return my_random()
```

Sorry for being skeptical but I don't see what problem this `atfork` thing really solves that can't be solved in other much simpler ways. Maybe if you can provide some more context I would understand the problem. The reason I'm hesitant to add more complexity is that worker.py is already very complex and hard to refactor

nkdhny:
Yes, I think you are right.

Issue #1379, opened 2015-11-09
DateIntervalParameter not ISO8601 interval?


wires:
Hello,

according to [the docs](https://luigi.readthedocs.org/en/latest/api/luigi.html?highlight=interval#luigi.DateIntervalParameter) `DateIntervalParameter` uses [ISO 8601 Time Interval notation](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals).

However, there are four ways to express a time interval (according to wikipedia):

1. Start and end, such as `2007-03-01T13:00:00Z/2008-05-11T15:30:00Z`
2. Start and duration, such as `2007-03-01T13:00:00Z/P1Y2M10DT2H30M`
3. Duration and end, such as `P1Y2M10DT2H30M/2008-05-11T15:30:00Z`
4. Duration only, such as `P1Y2M10DT2H30M`, with additional context information

Looking at [the tests for `date_interval`](https://github.com/spotify/luigi/blob/master/test/date_interval_test.py#L130) it seems this is not what is implemented:

- `2015-01-01-2015-01-30` is considered a valid time interval
- this should probably be `2015-01-01/2015-01-30`

erikbern:
Yeah it's not entirely true. We should fix the docs
* The four classes `Date`, `Week`, `Month`, and `Year` all implement ISO 8601 notation. Note that they do *not* implement *time interval* notation, just date notation, but then these are considered intervals (i.e. the year 2015 is considered the interval `[2015-01-01, 2016-01-01)`
* There is no ISO 8601 standard for *date intervals* afaik, just *time intervals*. So there's no obvious thing to do
* I chose the representation `YYYY-MM-DD-YYYY-MM-DD` for the `Custom` class, with a dash rather than a slash since dashes are unlikely to mess up directory structures. There's actually a comment for it in the constructor: https://github.com/spotify/luigi/blob/master/luigi/date_interval.py

I suggest we change the docs so it doesn't say that we use ISO 8601 time interval notation. The docs should say we support ISO 8601 date notation for date intervals

wires:
Nice! closing issue

PR #1380, opened 2015-11-09
Update docs for DateIntervalParameter

erikbern:
Clarifying #1379


Tarrasch:
Build was failing, but it seemed safe to merged based on the error messages. (we had some redshift failures last week)

PR #1382, opened 2015-11-09
Use DequeQueue instead of multiprocess.Queue if worker_process==1

dangarthwaite:
I changed DequeuQueue to ignore **kwargs.  Elsewhere in the code queue.get() is called with a timeout= argument.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -239,11 +239,9 @@ class DequeQueue(collections.deque):
     """
 
     def put(self, obj, block=None, timeout=None):
-        del block, timeout
         return self.append(obj)
 
     def get(self, block=None, timeout=None):
-        del block, timeout
         return self.pop()
 
 
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -238,10 +238,12 @@ class DequeQueue(collections.deque):
     deque wrapper implementing the Queue interface.
     """
 
-    def put(self, *args, **kwargs):
-        return self.append(*args)
+    def put(self, obj, block=None, timeout=None):
+        del block, timeout
+        return self.append(obj)
 
-    def get(self, **kwargs):
+    def get(self, block=None, timeout=None):
+        del block, timeout
         return self.pop()
 
 
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -238,8 +238,11 @@ class DequeQueue(collections.deque):
     deque wrapper implementing the Queue interface.
     """
 
-    put = collections.deque.append
-    get = collections.deque.pop
+    def put(self, *args, **kwargs):
+        return self.append(*args)
+
+    def get(self, **kwargs):
+        return self.pop()
 
 
 class AsyncCompletionException(Exception):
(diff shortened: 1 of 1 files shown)


daveFNbuck:
Looks good to me. Maybe add a test?

erikbern:
LGTM

need to fix the broken unit tests

luigi/worker.py
```diff
@@ -238,8 +238,11 @@ class DequeQueue(collections.deque):
     deque wrapper implementing the Queue interface.
     """
 
-    put = collections.deque.append
-    get = collections.deque.pop
+    def put(self, *args, **kwargs):
```
ulzha: Omit **kwargs? Better than including them to be silently dropped if passed in.

dangarthwaite: The point is to ignore 'timeout=N' that some parts of the code include in the call.

dangarthwaite: The one instance I can find of passing in any kwargs is worker.py line 746.  So I presume DequeQueue needs to handle timeouts?  Anyone want to volunteer for that?  This is preventing me from using luigi in Amazon Lambda.

daveFNbuck: It only needs to handle timeouts if it's using multiple processes. The worker will periodically check for new work while waiting for jobs to finish. If you're running with only one process, the timeout doesn't mean anything. Get and put will always return immediately. It's correct to throw out the timeout argument, but it would also be reasonable not to indiscriminately ignore any keyword argument and just match the signatures of the multiprocessing.Queue functions:

    put(obj[, block[, timeout]])
    get([block[, timeout]])

Just adding optional block and timeout arguments that you ignore looks more intentional and makes it more clear that you're not throwing out important arguments. If someone were to call DequeQueue().put(obj=x), your code would throw out the obj argument and result in an error.

dangarthwaite: Great advice, thank Dave.

The PR has been updated.

class DequeQueue(collections.deque):
    """
    deque wrapper implementing the Queue interface.
    """

    def put(self, obj, block=None, timeout=None):
        del block, timeout
        return self.append(obj)

    def get(self, block=None, timeout=None):
        del block, timeout
        return self.pop()

luigi/worker.py
```diff
@@ -238,8 +238,13 @@ class DequeQueue(collections.deque):
     deque wrapper implementing the Queue interface.
     """
 
-    put = collections.deque.append
-    get = collections.deque.pop
+    def put(self, obj, block=None, timeout=None):
+        del block, timeout
```
daveFNbuck: Why delete these variables? It should be sufficient to ignore them. Does this line do anything?

luigi/worker.py
```diff
@@ -238,8 +238,13 @@ class DequeQueue(collections.deque):
...
+        return self.append(obj)
+
+    def get(self, block=None, timeout=None):
+        del block, timeout
```
daveFNbuck: same

erikbern:
merging this, it's been sitting here for a while

i think all the existing unit tests will run this code

Issue #1381, opened 2015-11-09
ReadTheDocs - Latest (Incorrect Page titles)


dlstadther:
For http://luigi.readthedocs.org/en/latest/*.html, the page titles are "____ -- Luigi 1.0 Documentation".

Since Luigi 2.0 is the Latest, these pages should be titled " _____ -- Luigi 2.0 Documentation".

I can make this change if someone can point me in the correct direction for Luigi ReadTheDoc code.

Thanks,

dlstadther:
^ Same goes for http://luigi.readthedocs.org/en/stable

dlstadther:
Is [luigi/doc/conf.py](https://github.com/spotify/luigi/blob/master/doc/conf.py#L115-L117) the correct code to make this change?

erikbern:
great if you can make the change! not sure where it goes

Tarrasch:
Ah, so that's why Google results always say "Luigi 1.0 Documentation" even though it's not.

Thanks for fixing @mvj3!

PR #1384, opened 2015-11-10
Update luigi version number to lastest from setup.py automatically.

mvj3:
ref to https://github.com/spotify/luigi/issues/1381


erikbern:
hm isn't this just going to copy the version numbers in a circle?

mvj3:
I'm a little confused ...

`conf.py` is auto generated by run `sphinx-quickstart`, and it's used to be customized (I saw many modification git-history on this file). So it's depended on luigi package, but luigi package doesn't require this sphinx doc to the contrary .

dlstadther:
Historically (and currently) are Luigi moderators pushing tags to releases? Read the Docs mentions [here](http://read-the-docs.readthedocs.org/en/latest/versions.html) that they parse the VCS tags.

mvj3:
@dlstadther yes, tags or branches are only related to when to build the doc, but not about the doc content (including HTML title).

I found a good example from Django, take a look at http://django.readthedocs.org/en/stable/contents.html and https://github.com/django/django/blob/master/docs/conf.py#L75

doc/conf.py
```diff
@@ -105,16 +105,18 @@ def _warn_node(self, msg, node):
 
 # General information about the project.
 project = u'Luigi'
-copyright = u'2014, Erik Bernhardsson and Elias Freider'
+copyright = u'2015, Erik Bernhardsson and Elias Freider'
```
econchick: I think this should be `2014-2015` - I think you need to include the year it is initially copyrighted. And with that I'm pretty sure it is actually earlier than 2014. git history looks like it should be 2011.

econchick: FYI This may be a better way of doing the copyright, taken from another package I maintain:

```python
copyright = u'2014{0}, Lynn Root'.format(
    u'-{0}'.format(year) if year != 2014 else u""
)
```

econchick:
The changes make sense to me at least. I'd say fix the copyright header in order to merge.

erikbern:
i looked at this again and it lgtm but might be worth updating the copyright as @econchick pointed out

doc/conf.py
mvj3: @econchick yes, I agree with you. This is maybe the first commit that mentioned luigi.

```txt
SHA:	999630bd4043c0c12e096c9e930d42ba76d4907d
Author:	Erik Bernhardsson <erikbern@spotify.com>
Date:	Wed Nov 23 2011 07:55:31 GMT+0800 (CST)
Subject:	renamed builder3 to luigi in the code
Parent:	07dee441cac1f9bcf28851a08087fbc9e00b360a
```

`year` variable is not defined, and it seems there's no way to get the closest year.

Maybe getting it dynamically is a good way. e.g.

```python
copyright = u"2011-{}, Erik Bernhardsson and Elias Freider".format(datetime.datetime.now().year)
```

erikbern:
awesome thanks!

PR #1199, opened 2015-09-08
docs clarification of mapper, reducer, combiner functions format

stroykova:
Discussion here: #1123
Not sure, that assertion in generator functions call (mapper, reducer and combiner) and every line check  is good.
So I have added hint to docs. 

Maybe you have ideas how to implement assertion without overhead, I will add it later.

PR #1386, opened 2015-11-10
Map reduce format

stroykova:
Request: https://github.com/spotify/luigi/pull/1199
Was closed due to inactivity.

### DIFF
--- a/doc/example_top_artists.rst
+++ b/doc/example_top_artists.rst
@@ -140,7 +140,8 @@ here is how this could look like, instead of the class above.
 Note that :class:`luigi.contrib.hadoop.JobTask` doesn't require you to implement a
 :func:`~luigi.task.Task.run` method. Instead, you typically implement a
 :func:`~luigi.contrib.hadoop.JobTask.mapper` and
-:func:`~luigi.contrib.hadoop.JobTask.reducer` method.
+:func:`~luigi.contrib.hadoop.JobTask.reducer` method. *mapper* and *combiner* require
+yielding tuple of only two elements: key and value. Both key and value also may be a tuple.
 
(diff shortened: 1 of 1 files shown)


erikbern:
LGTM but mapper should be lower case

stroykova:
agree

Tarrasch:
Can you amend the three commits?

stroykova:
Amended.

Now I discovered that if reducer yields string, in the output we see letters joined by \t. Is it bug or feature?

Tarrasch:
> Now I discovered that if reducer yields string, in the output we see letters joined by \t. Is it bug or feature?

I do not understand what you mean. It's not any code I'm comfortable with.

Is this otherwise ready to be merged?

erikbern:
the key and value can be any type btw

stroykova:
@erikbern thank you, I will add this explanation.
Question is about reducer. Looks like it also is expected to yield tuple. Unlike mapper and combiner it can have any length.

This thing about reducer is something to decide in architecture. I do not think it is expected string to become '\t' joined letters in output.

Tarrasch:
Can you give an example of what you mean with '\t' joined?

erikbern:
the '\t' is needed for hadoop streaming because it expects tab separated key/values

Luigi uses repr on top of that to represent objects (sorry, I said pickle, but I meant repr)

stroykova:
I mean this:

```Python
def reducer(key, values):
    count = sum(1 for _ in values)
    yield key
```

Input:
("my_key", (1, 2, 3))

Output:
"m\ty\t_\tk\te\ty"

So I think I should also add info about reducer. Reducer is expected to yield tuples of any length.

PR #1388, opened 2015-11-11
Allows disabling of worker e-mails via luigi.cfg

daveFNbuck:
After a night of excessive failures, I found that my workers were dying
whenever a job failed and they weren't able to access the overloaded
smtp server. When I tried to find the configuration option to disable
emails and get things running again, I found that there wasn't one,
although error-email falsely claims to be ablo to do that in the
documentation.

I've updated the documentation for error-email to clarify that e-mails
will still be sent for jobs that have owners and added a "none" option
to email-type that disables email sends. I also cleaned up the logging
for why emails are disabled.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -115,14 +115,19 @@ email-sender
   Default value: luigi-client@<server_name>
 
 email-type
-  Type of e-mail to send. Valid values are "plain" and "html". When set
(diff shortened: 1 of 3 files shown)


erikbern:
thanks! assuming tests failed because of the redshift issue

daveFNbuck:
Probably, I guess we'll find out soon.

erikbern:
:)

PR #1390, opened 2015-11-11
Fix broken redshift test

daveFNbuck:
Note: This includes #1389 because some lines were both broken and violated pep8.

The redshift test asserts that the output is only called once in run,
when run clearly called it twice. I fixed that by storing the output in
a local variable rather than calling it again.

The redshift test asserted that RedshiftTarget is called with arguments
table_type and sql that it does not accept, so I removed those. There
was also a typo in the update_id argument.

The redshift test verified the existence check by asserting that it was
the only redshift call. I fixed that by changing the assertion to verify
that the existence check was made, as the run makes two other calls
afterward.

It seems that this test is incomplete, but it at least no longer fails.

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -103,7 +103,7 @@ class S3CopyToTable(rdbms.CopyToTable):
         """
         return ''
 
-    @property    
+    @property
     def prune_table(self):
         """
         Override to set equal to the name of the table which is to be pruned.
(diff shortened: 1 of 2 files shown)


erikbern:
thanks so much!

dlstadther:
@daveFNbuck Thank you for fixing these tests. I do not know what I'm doing when it comes to mock tests and so I had to use the existing tests for Redshift (which were sparse) to attempt something. Thank you.

daveFNbuck:
No problem. Python docs have a pretty good overview you should read if you're doing any serious Python development. https://docs.python.org/dev/library/unittest.mock.html

PR #1391, opened 2015-11-11
Add notification through Amazon SNS

jslovan:
We run luigi data processing in Amazon EMR and find more convenient to use SNS for notifications than setting up an e-mail.

I added support for SNS and replaced boto for boto3.  Also I dropped AWS credentials from SES config as boto3 takes creds from env variables and/or aws config files and/or EC2 instance profile.

Thanks for review.


erikbern:
looks good!

PR #1392, opened 2015-11-12
examples: pyspark_wc: Make parameter insignificant

Tarrasch:
The parameter is clearly not significant, as the output path does not depend on it.


DeaconDesperado:
:+1:

Issue #1394, opened 2015-11-12
tornado < 4 consistently makes requests to luigid fail with 500 Internal server error


nresare:
Using luigid with a version earlier than 4.0 makes a simple hello world fail with a `KeyError: 'worker'`and the stack trace below.

Steps to reproduce:
*  in a virtualenv pip install luigi, then force-downgrade tornado: `pip install tornado==3.1.1`
* start a luigid in one window
* run a trivial luigi Task in another.

Suggested solution: enforce at least tornado 4.0
```
2015-11-12 15:11:32,176 tornado.application[79210] ERROR: Uncaught exception GET /api/add_task (10.48.20.57)
HTTPRequest(protocol='http', host='lon6-luigid-edgedocker1:8082', method='GET', uri='/api/add_task', version='HTTP/1.1', remote_ip='10.48.20.57', headers={'Content-Length': '477', 'Accept-Encoding': 'gzip, deflate, compress', 'Content-Type': 'application/x-www-form-urlencoded', 'Host': 'lon6-luigid-edgedocker1:8082', 'Accept': '*/*', 'User-Agent': 'python-requests/2.2.1 CPython/2.7.6 Linux/4.1.10-boot2docker'})
Traceback (most recent call last):
  File "/home/noa/luigid-test/local/lib/python2.7/site-packages/tornado/web.py", line 1141, in _when_complete
    callback()
  File "/home/noa/luigid-test/local/lib/python2.7/site-packages/tornado/web.py", line 1162, in _execute_method
    self._when_complete(method(*self.path_args, **self.path_kwargs),
  File "/home/noa/luigid-test/local/lib/python2.7/site-packages/luigi/server.py", line 79, in get
    result = getattr(self._scheduler, method)(**arguments)
  File "/home/noa/luigid-test/local/lib/python2.7/site-packages/luigi/scheduler.py", line 605, in add_task
    worker_id = kwargs['worker']
KeyError: 'worker'
2015-11-12 15:11:32,177 tornado.access[79210] ERROR: 500 GET /api/add_task (10.48.20.57) 1.21ms
```

nresare:
Suggested solution: https://github.com/spotify/luigi/pull/1395

erikbern:
Is there some trivial way to fix the issue for older versions of Tornado?

nresare:
[...]

PR #1395, opened 2015-11-12
Enforce at least tornado version 4.0

nresare:
Otherwise luigid will fail every request with a KeyError. More
details here: https://github.com/spotify/luigi/issues/1394

It is slightly strange that the dependency is declared as 
<5 in it's current incarnation, when tornado 5 is not released.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ with open('README.rst') as fobj:
     long_description = readme_note + fobj.read()
 
 install_requires = [
-    'tornado<5.0',
+    'tornado>=4.0,<5',
     'python-daemon<3.0',
 ]
 


davemt:
having <5 even without the release having gone out makes sense, because the moment 5 is released environments will begin upgrading to a potentially non-backwards-compatible version

setup.py
```diff
@@ -37,7 +37,7 @@ def get_static_files(path):
     long_description = readme_note + fobj.read()
 
 install_requires = [
-    'tornado<5.0',
+    'tornado>=4.0',
```
davemt: you can write ```tornado>=4.0,<5.0```

Tarrasch:
So *thats* what have caused so much trouble for me!!!

Tarrasch:
Thanks @nresare! Great job!

Tarrasch:
Confirming that this is the issue that have plagued me all the time. Thanks again for the fix! :)

PR #1400, opened 2015-11-15
Add a serialize method to TaskParameter

nizox:
Without this method it is not possible to program a task requiring
another task with a TaskParameter. This commit also provides a test
case for this scenario.


Tarrasch:
Thanks! Very clean commit. :)

PR #1402, opened 2015-11-16
Add kwargs for boto s3 writing

jpiper:
This allows the use of extra arguments to the boto uploading functions, such as `headers`, `metadata`, `reduced_redundancy`, `encrypt_key`, and `policy`.

Allowing server side encryption is the biggest advantage here, as buckets which enforce SSE will not be writable unless the following is used.

        s3 = luigi.s3.S3Client()
        s3.put(localfile, remotefile, encrypt_key=True)

### DIFF
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -165,9 +165,11 @@ class S3Client(FileSystem):
 
         return s3_bucket.get_key(key)
 
-    def put(self, local_path, destination_s3_path):
+    def put(self, local_path, destination_s3_path, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


erikbern:
lgtm

i could see a problem if you want to use kwargs for other things (not `put`) in the future, but for now that's not a big deal

erikbern:
also would have been good to add tests so try to do that next time :)

jpiper:
Sorry - long day at the office! See #1403 for some basic tests :smiley:

PR #1405, opened 2015-11-17
PhantomJS tests for the visualiser javascript.

stephenpascoe:
These tests check that the visualiser displays the correct number of DONE/PENDING/FAILED/etc. tasks using PhantomJS.  The Python unittest starts PhantomJS in a subprocess whilst running the tornado IOLoop in-process.  I use an implementation of MergeSort to populate the scheduler with a medium-sized task-graph.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ env:
     - TOXENV=pypy-scheduler
     - TOXENV=py27-gcloud
     - TOXENV=py27-postgres
+    - TOXENV=visualiser
 
 sudo: false
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -18,6 +18,9 @@ then run one of the tox commands below.
     # You can also test particular files for even faster iterations
     tox -e py27-nonhdfs test/rpc_test.py
 
+    # The visualiser tests require phantomjs to be installed on your path
+    tox -e visualiser
+
     # And some of the others involve downloading and running Hadoop:
     tox -e py33-cdh
     tox -e py34-hdp
--- a/test/visualiser/__init__.py
+++ b/test/visualiser/__init__.py
@@ -1,0 +1,1 @@
+# Tests for visualiser javascript.
--- a/test/visualiser/phantomjs_test.js
+++ b/test/visualiser/phantomjs_test.js
@@ -1,0 +1,335 @@
+var page = require('webpage').create();
+var system = require('system');
+
+var tests = [];
+
+/*
+ * Parse command line to get Luigi scheduler URL
+ */
+if (system.args.length === 1) {
+    console.log('Usage: phantom_test.js <scheduler-url>');
+    phantom.exit();
+}
+
+var url = system.args[1];
+
+
+/*
+ * Minimal test framework
+ */
+function do_tests(page) {
+    var ok = true;
+    var retval;
+
+    tests.forEach(function (spec) {
+        var name = spec[0];
+        var test_func = spec[1];
+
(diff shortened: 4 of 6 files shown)


erikbern:
This is super cool!

We should probably set it up as a separate set of tests in Travis but other than that, looks awesome! I have been thinking about building something similar for a long time but didn't know where to start.

stephenpascoe:
OK, I'll refactor as a separate travis run.  Travis went OK on my fork but it looks like it's been hit by the KeepAliveThread problem mentioned in server_test.py.

econchick:
Hey @stephenpascoe ! I tried running this locally, and I can't get it to run. Both via `tox -e visualiser` and `nosetests -v --tests=test/visualiser`. I get the same FAIL issue, I think this is the relevant log:

```

/bin/sh: phantomjs: command not found
FAIL

FAIL: test (visualiser.visualiser_test.TestVisualiser)
  File "/Users/lynnroot/.pyenv/versions/luigi-27/lib/python2.7/site-packages/tornado/testing.py", line 132, in __call__
    result = self.orig_method(*args, **kwargs)
  File "/Users/lynnroot/Dev/spotify/luigi/test/visualiser/visualiser_test.py", line 63, in test
    assert status == 0
AssertionError:
-------------------- >> begin captured stdout << ---------------------
Server port is 51877
Starting phantomjs
Entering event loop
PhantomJS return status is 127
--------------------- >> end captured stdout << ----------------------
```

Also I'm not too familiar with PhantomJS so perhaps I'm missing a piece of setup.

erikbern:
I'm guessing it's installed by default on Travis but not locally.

I don't think we can enforce installing it so maybe we should disable it by default when running locally and enable it on Travis

econchick:
ah yes looking at this random [Travis job](https://travis-ci.org/spotify/luigi/jobs/91589646#L54) it looks like PhantomJS is installed. Which actually is surprising since - from what I can tell - it isn't declared in the `.travis.yml` as a dependency. But perhaps we get that "for free". 

@stephenpascoe I'd actually suggest having a script in `scripts/`, similar to `scripts/ci/setup_hadoop_env.sh` to download & install PhantomJS. @erikbern what are your thoughts on that?

erikbern:
if it's easy to install phantomjs then sure (i suspect unfortunately it's not)

econchick:
Hmm it does seem like it may be a bit complicated: [docs](http://phantomjs.org/build.html) - especially since it says it takes a long time to build, and it is dependent on the system (and simple linux binaries are not yet available). Perhaps thoroughly documenting how to run tests locally in `CONTRIBUTING.rst` or somewhere.

econchick:
Also after `brew install phantomjs` tests pass \o/

stephenpascoe:
Yes, PhantomJS is installed on Travis and is pretty easy to install on unix.  We could skip if phantomjs is not available but that risks missing the tests without realising it.

stephenpascoe:
Travis failed :( It passed in my Travis build (https://travis-ci.org/stephenpascoe/luigi) but there is a timeout in the test that might have been exceeded.  I'll look into making it less fragile.

econchick:
@stephenpascoe before we merge, how would you feel about adding a line or two to `CONTRIBUTING.rst` on needing to have PhantomJS installed?

Also mind squashing commits into one?

erikbern:
i think it's good as it is and i'm happy to merge but will give it another day or so

erikbern:
looks awesome!

https://travis-ci.org/spotify/luigi/jobs/92008229

PR #1406, opened 2015-11-17
bigquery: Enable range support by implementing bulk_complete.

mbruggmann:


erikbern:
LGTM but why did you have to do this using a mixin rather than a single chain implementation inheritance?

mbruggmann:
@erikbern We have multiple `ExternalBigqueryTask`s in our pipelines that are using this Mixin. We could also push that one upstream if it sounds useful?

igorbernstein:
As it stands, this will break for lookbacks > 50...by default bigquery sets maxResults to 50 for listing tables. Whats worse is that you can't do a reverse chronological sort or random access pagination (ie. give me the last 50 table names that whose name is greater than endsong_cleaned_20151116). You can temporarily workaround the issue by setting maxResults to 1000, but that will start breaking after 1000 partitions.

Another way you can deal with the issue is by doing a metadata query on the dataset:
``` SELECT table_id FROM <dataset_name>.__TABLES__ WHERE table_id BETWEEN "<startdate>" AND "<enddate>" ```

See here for relevant limits:
https://cloud.google.com/bigquery/docs/data#paging

erikbern:
@mbruggmann what i mean is why can't you put it on the bigquery base class?

is the problem that the ExternalBigQueryTask have to inherit from ExternalTask? I've been planning to make external tasks not a class but rather a property on the task class for a while now.

mbruggmann:
@erikbern Exactly. We could change this once `ExternalTask` is no more. Deal?

erikbern:
yeah that sounds fair :) will try to refactor that before Luigi derails into a land of mixins :)

PR #1421, opened 2015-11-23
Remove redundant swap logic

DeaconDesperado:
This logic should no longer be necessary:  It was a result of some problems we were having with ingestion on the bigquery connector that have now be verified to work.

@Tarrasch sorry for the delay with this but should be all set now :)

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -19,7 +19,6 @@ import collections
 import logging
 import luigi.target
 import time
-import random
 
 logger = logging.getLogger('luigi-interface')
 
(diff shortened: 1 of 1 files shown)


erikbern:
Is it still atomic?

DeaconDesperado:
@erikbern [Thankfully, it is!](https://cloud.google.com/bigquery/docs/reference/v2/jobs#configuration.copy.writeDisposition)  The problem we were seeing was in unrelated java code that relied on partial loads per mapper in hadoop followed by a final table merge, so the intention here was to hide the table as it was loaded with disposition APPEND to avoid querying partial data.

The loads as performed by luigi via GCS are fully atomic.

erikbern:
excellent

Tarrasch:
Finally! Great job! :D

Just please amend the commits so the history looks clean. :)

DeaconDesperado:
@Tarrasch Rebased

Tarrasch:
Cool. Lets merge once tests pass! :)

mbruggmann:
:+1: Also triggered a rebuild, tests are green now.

PR #1426, opened 2015-11-24
Add username parameter for Hadoop tasks

rtharper:
This is a proposed change to add a user option for hadoop jobs, which then uses the HADOOP_USER_NAME env variable to run a job as the specified user.

This is mostly a pull request for testing and to elicit review comments

### DIFF
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -396,7 +396,7 @@ class HadoopJobRunner(JobRunner):
     def __init__(self, streaming_jar, modules=None, streaming_args=None,
                  libjars=None, libjars_in_hdfs=None, jobconfs=None,
                  input_format=None, output_format=None,
-                 end_job_with_atomic_move_dir=True):
+                 end_job_with_atomic_move_dir=True, username=None):
         def get(x, default):
             return x is not None and x or default
         self.streaming_jar = streaming_jar
@@ -409,6 +409,7 @@ class HadoopJobRunner(JobRunner):
         self.output_format = output_format
         self.end_job_with_atomic_move_dir = end_job_with_atomic_move_dir
         self.tmp_dir = False
+        self.username = username;
 
     def run_job(self, job, tracking_url_callback=None):
         packages = [luigi] + self.modules + job.extra_modules() + list(_attached_packages)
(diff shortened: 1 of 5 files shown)


luigi/contrib/hive.py
```diff
@@ -342,15 +347,19 @@ def run_job(self, job, tracking_url_callback=None):
...
             hiverc = job.hiverc()
             if hiverc:
                 if isinstance(hiverc, str):
                     hiverc = [hiverc]
+
                 for rcfile in hiverc:
                     arglist += ['-i', rcfile]
             if job.hiveconfs():
-                for k, v in six.iteritems(job.hiveconfs()):
+                for k, v in six. iteritems(job.hiveconfs()):
```
econchick: is there a space between `six.` and `iteritems` ?

luigi/contrib/hive.py
```diff
@@ -342,15 +347,19 @@ def run_job(self, job, tracking_url_callback=None):
                 query = query.encode('utf8')
             f.write(query)
             f.flush()
-            arglist = [load_hive_cmd(), '-f', f.name]
+            username_cmd = ""
+            if self.username:
+                username_cmd = "HADOOP_USER_NAME=" + self.username
+            arglist = [username_cmd] + [load_hive_cmd(), '-f', f.name]
```
Tarrasch: Does this really work? I always thought the environment variable setting in `MY_VAR=my_val echo hi` was done by the shell.

rtharper: No, you are correct, this does not properly set the environment variable.
This has to be done in the call to popen, so this will need to be
refactored, unfortunately.

2015-11-25 5:29 GMT+01:00 Arash Rouhani <notifications@github.com>:

luigi/contrib/hadoop.py
```diff
@@ -523,7 +524,10 @@ def run_job(self, job, tracking_url_callback=None):
 
         job.dump(self.tmp_dir)
 
-        run_and_track_hadoop_job(arglist, tracking_url_callback=tracking_url_callback)
+        job_env = os.environ.copy()
+        if self.username:
+            job_env["HADOOP_USER_NAME"] = self.username
```
econchick: Creating a copy of the env variables and updating the `job_env` won't actually update the environment itself:

```python
>>> import os
>>> job_env = os.environ.copy()
>>> job_env["HADOOP_USER_NAME"] = "bar"
>>> os.environ.get("HADOOP_USER_NAME", "not set")
'not set'
```

You'll need to update it directly:

```python
>>> os.environ["HADOOP_USER_NAME"] = "bar"
>>> os.environ.get("HADOOP_USER_NAME", "not set")
'bar'
```

Same with the other files below that try to set the env var.

erikbern: the whole environment dict is passed to `run_and_track_hadoop_job` so it looks like it's fine

econchick: Ah definitely missed that. Let me just go grab some more coffee...

erikbern:
Can you rebase and squash your commits? And fix unittests?

Otherwise LGTM

erikbern:
great if you can fix the flake8 issues so we can merge!

Tarrasch:
Closing because of being inactive for a month (almost). Please feel free to reopen whenever you have time to work on it. :)

PR #1427, opened 2015-11-26
Honor keyboardinterrupt during scheduling

Tarrasch:
This is a series of commits about making luigi to quit when you ask it too with Ctrl+C. The individual commits and their contents are explaning the changes in detail.

Since @graingert and @daveFNbuck have touched the related codes before, would you guys mind reviewing?

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -226,10 +226,7 @@ class SingleProcessPool(object):
     def apply_async(self, function, args):
         return function(*args)
 
-    def close(self):
-        pass
-
-    def join(self):
+    def terminate(self):
         pass
 
 
@@ -534,8 +531,22 @@ class Worker(object):
             task.trigger_event(Event.BROKEN_TASK, task, ex)
             self._email_unexpected_error(task, formatted_traceback)
         finally:
-            pool.close()
-            pool.join()
+            # We must not "pool.join()" here, as the pool proccesses may have
+            # deadlocked (since this main tread is not .get()-ing from the
+            # queue anymore).
+            # http://stackoverflow.com/a/27492515/621449
+            try:
+                # However, as some users have disabled garbage collection (see
+                # spotify/luigi#1089), we want to explicitly try to kill off
+                # any remaining processes right away.
+                pool.terminate()
+            except (OSError, EOFError):
+                # If this fails however, we still want to propagate the outer
+                # exception and not re raise!
+                #
+                # The enumerated exception classes in the "except" clause is
+                # the union of exceptions when I ran this 100 times.
+                pass
         return self.add_succeeded
 
     def _add(self, task, is_complete):
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -267,7 +267,7 @@ def check_complete(task, out_queue):
     logger.debug("Checking if %s is complete", task)
     try:
         is_complete = task.complete()
-    except BaseException:
+    except Exception:
         is_complete = TracebackWrapper(traceback.format_exc())
     out_queue.put((task, is_complete))
 
--- a/test/retcodes_test.py
+++ b/test/retcodes_test.py
@@ -108,11 +108,3 @@ class RetcodesTest(LuigiTestCase):
 
         self.run_and_expect('RequiringTask --retcode-task-failed 4 --retcode-missing-data 5', 5)
         self.run_and_expect('RequiringTask --retcode-task-failed 7 --retcode-missing-data 6', 7)
-
-    def test_keyboard_interrupts(self):
-
-        class KeyboardInterruptTask(luigi.Task):
-            def run(self):
-                raise KeyboardInterrupt()
-
-        self.assertRaises(KeyboardInterrupt, luigi_run, ['KeyboardInterruptTask', '--local-scheduler', '--no-lock'])
--- a/test/worker_test.py
+++ b/test/worker_test.py
@@ -25,7 +25,8 @@ import tempfile
 import threading
 import time
 import psutil
-from helpers import unittest, with_config, skipOnTravis
+import multiprocessing
+from helpers import unittest, with_config, skipOnTravis, LuigiTestCase
 
 import luigi.notifications
 import luigi.worker
@@ -35,6 +36,7 @@ from luigi.mock import MockTarget, MockFileSystem
 from luigi.scheduler import CentralPlannerScheduler
 from luigi.worker import Worker
 from luigi import six
+from luigi.cmdline import luigi_run
 
 luigi.notifications.DEBUG = True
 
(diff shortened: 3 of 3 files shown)


Tarrasch:
@graingert, I was thinking about #1089 that you authored. Did you have a use case where garbage collection was turned off? I think the `finally` clause should simply just be a `finally: pass` if we did not need to `join()` or `terminate()` the child processes right away. It would also remove all the things related to python 2 vs 3 from this patch. So please remind me again why we need to either wait or terminate the spawned processes.

graingert:
@Tarrasch this is even without garbage collection, I find running under ipython can hold these processes open indefinitely

graingert:
the GC pause code is just to get the test to fail on Travis where Python's GC seems to be more aggressive than on my local machine

Tarrasch:
Ok fair enough. :)

luigi/worker.py
```diff
@@ -534,8 +531,22 @@ def add(self, task, multiprocess=False):
...
             self._email_unexpected_error(task, formatted_traceback)
         finally:
-            pool.close()
-            pool.join()
+            # We must not "pool.join()" here, as the pool proccesses may have
+            # deadlocked (since this main tread is not .get()-ing from the
+            # queue anymore).
+            # http://stackoverflow.com/a/27492515/621449
+            try:
+                # However, as some users have disabled garbage collection (see
```
graingert: this is technically not correct, we want to terminate the process because we don't need it running at this point in the program.

There are various reasons for the GC-auto-terminator to not terminate the process, the GC being paused is one of them

Tarrasch: Cool. I'll fix it :)

Tarrasch:
Doh. I get similar kind of errors when I try this in production as when Travis gets. God damn it... Luckily I believe I can reproduce the errors...

Tarrasch:
Hmm, seems like this is easy to reproduce just changing

    pool = multiprocessing.Pool()

into

    pool = multiprocessing.Pool(processes=30)

And the default value is the number of cores on your machine.

erikbern:
this is excellent!

daveFNbuck:
Thanks! I'd given up on trying to cancel scheduling with ctrl+c lately.

Tarrasch:
Closing this. I've opened #1429 which is the last two commits and I'll keep working on the first commit (that breaks the tests) in a seperate PR.

Tarrasch:
Thanks for joining in @pb-cdunn. I really appreciate an extra pair of eyes on this very non-trivial code.

Sure. We could do a `join(timeout)`. I want luigi too feel responsive though. So maybe only 2 seconds for timeout?

---

As for the deadlock I don't understand any of the solutions you are proposing. We are already using `Manager.Queue()`. I believe we use it because `multiprocessing.Queue` can't be passed as an argument to `pool.apply()`. What property of the Manager do you think is relevant here?

I think doing `Queue.put(timeout=x)` in a loop would increase the complexity. Why is it important that they eventually put something in the queue? When the main thread is aborting we do not care what ends up in the queue. No?

Which process are you suggesting should call `Queue.close()`/`Queue.join_thread()`? Reading the documentation for [close](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.Queue.close), it doesn’t make sense for the main thread to call it because it won't do `put()` operations anyway.

---

I'm really confused and you seem to have experience with python multiprocessing. Would you mind taking a stab at a correct implementation? I have already added a test case which I can't make pass. Do you want to take a look @pb-cdunn? Thanks! :)

pb-cdunn:
Yes, everybody is confused by Python's **multiprocessing** module. For example, even if you do everything right, you'll still see some ugly random stack-traces on Ctrl-C if you have un-utilized processes in your process pool, which makes people scratch their heads. But it's important to remember that these are processes, not threads. At my previous company, we used this module extensively, but they set it up before I got there and made many of the same mistakes. At my current company, we use threads instead, in something we call [**pypeflow**](https://github.com/PacificBiosciences/pypeFLOW/wiki/Example), which is in many ways similar to **Luigi**. We looked into switching to Luigi and liked a lot of what we saw, but unfortunately we would need thousands of sub-processes, which are too expensive. (We need direct access to each qsub job.)

Anyway, **Luigi** uses processes, so let's concentrate on that. When working with processes, you need to be aware that you cannot kill them directly. Well, you can send a signal, so they're easier to kill than threads, but there is a higher principle: It's important to let users debug your program. So it's important to get stack-traces on Ctrl-C. So it's important for every sub-process to be non-blocking for long enough to handle KeyboardInterrupt. That's the reason to have `Queue.put()` in a loop. Yes, it's slightly more complex, but it's "nice".

However, I don't even know why you are having a problem. I think you should simplify first to discover exactly what is "deadlocking".  I suspect that you are trapping KeyboardInterrupt in workers, so their queues are not being closed automatically (during garbage-collection).

[This gist](https://gist.github.com/pb-cdunn/d8345c36fc572d24fe10) is a toy example. Try un-commenting various parts, different Queue types, etc. I don't yet know how to replicate your behavior. This toy always works.

(The reason for Queue.close() in the main thread is only because Queue.join_thread() requires it. But you don't necessarily need either at all.)

Tarrasch:
@pb-cdunn, I believe to know the reason for the deadlocking already. As it says in the [patch](https://github.com/spotify/luigi/pull/1427/files#diff-f1cd3288d80ac94f428311dbefa70c7fR1169): Pool-processes are stuck on `put()` since nobody is reading on the other end. And main thread doesn't quit because it waits for the pool-processes to join (exit).

Then I noticed that it is still "getting stuck" on Ctrl+C despite changing `pool.join()` to `pool.terminate()`, so my theory might very well be wrong.

@pb-cdunn, do you think you will have time to look if you can make the test case pass?

pb-cdunn:
Hmmm. That code (your `honor_...` branch of your Tarrasch repo) passes for me on MacOSX. (I uncommented the skip line.) However, it shows an error "No such file or directory", so maybe I am not running the test right via nose. (Tox sets up an environment, which I'm trying to replicate. I'm avoiding tox so I don't have to wait for *every* test to run.)

What is the problem you are seeing? Your fix seems to avoid trapping KI (by trapping Ex instead of BaseEx) and then to avoid `out_queue.put()` on KI, so I don't see how the deadlock can be related to the put which is now skipped. Is KI happening *during* `out_queue.put()`? If so, it seems like your test does not exercise that.

Try this in your test: Put a long sleep (but not too long, else you'll have trouble killing the procs, as I mention below -- maybe 5seconds) in front of queue.get(). Then, run only the failing test. (I have no idea how to do that with tox, but you can do it with `nosetests --no-skip test/worker_test.py:KeyboardInterruptBehaviorTest.test_propagating_with_par_scheduling`. With that, Ctrl-C should still work! But it doesn't. Instead, the sub-processes keep running after nose has quit, another worker dying every few seconds, which is very annoying. This looks like an architectural problem, or at least a mistake in how you are using nose. I don't have time to dive deep into your code. But you could probably ameliorate the problem by using timeouts for all blocking calls, in a loop. (Both put and get, I think.) As a general principle, nothing should block forever.

If the problem is only for users who have disabled GC, maybe use a try/finally somewhere to close the queue in all workers. But that won't solve the problem I've (maybe) found by sleeping ahead of get().

I think you can force the problem to go away by sending a specific signal to sub-procs. SIGKILL, I guess.

Btw, I am skeptical of your `queue_size = 1`, but any problem there is unrelated.

Btw, on Mac, test_get_info (lock_test.LockTest) fails at line 56 in test_get_info, assertEqual(cmd, 'yes ...')

PR #1428, opened 2015-11-26
Make luigi client work without tornado

Tarrasch:
If I download luigi without installing tornado, I still would like to
try it out using say the local scheduler or a scheduler that's running
somewhere else. Or as in my case, having a different python version
that has tornado running luigid.

This means that you now can just download the luigi source code and get
started quite quickly by just running:

    PYTHONPATH=. python3 -m examples.foo

or

    PYTHONPATH=. python3 -m luigi --module examples.foo examples.Foo --local-scheduler

### DIFF
--- a/luigi/cmdline.py
+++ b/luigi/cmdline.py
@@ -3,10 +3,6 @@ import argparse
 import logging
 import sys
 
-import luigi.interface
-import luigi.server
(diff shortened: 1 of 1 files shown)


erikbern:
great!

PR #1431, opened 2015-11-27
Make the Worker a context manager

Tarrasch:
This way, we will shut down the KeepAliveThread even when there is an
error during the scheduling phase.  And context managers is a more
pythonic construct than the invented stop() method. This was also a now
removed TODO in the code.

I also put the side effect of starting the KeepAliveThread into the
__enter__ block. I found it surprising that constructing a Worker()
would start a new thread.

A lot of test cases had to be rewritten. While at it, I cleaned up some
of the tests and also added many forgotten cleanups. Hopefully tests are
more stable from now on.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -375,10 +375,6 @@ class Worker(object):
         except AttributeError:
             pass
 
-        self._keep_alive_thread = KeepAliveThread(self._scheduler, self._id, self._config.ping_interval)
-        self._keep_alive_thread.daemon = True
-        self._keep_alive_thread.start()
-
         # Keep info about what tasks are running (could be in other processes)
         if worker_processes == 1:
             self._task_result_queue = DequeQueue()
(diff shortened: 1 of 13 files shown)


luigi/worker.py
```diff
@@ -407,25 +403,25 @@ def _add_task(self, *args, **kwargs):
...
-
-        TODO: also kill all currently running tasks
+        Start the KeepAliveThread.
+        """
+        self._keep_alive_thread = KeepAliveThread(self._scheduler, self._id, self._config.ping_interval)
+        self._keep_alive_thread.daemon = True
+        self._keep_alive_thread.start()
+        return self
 
-        TODO (maybe): Worker should be/have a context manager to enforce calling this
```
Tarrasch: Consider it done @freider ;)

test/worker_test.py
```diff
@@ -990,9 +968,12 @@ def run(self):
 class AssistantTest(unittest.TestCase):
     def setUp(self):
         self.sch = CentralPlannerScheduler(retry_delay=100, remove_delay=1000, worker_disconnect_delay=10)
-        self.w = Worker(scheduler=self.sch, worker_id='X')
+        self.w = Worker(scheduler=self.sch, worker_id='X').__enter__()
```
erikbern: why do you need to call the `__enter__` method here

erikbern: I see why now. There must be some workaround though

erikbern:
this looks nice but couldn't it just be a destructor method?

i'm also not too excited about the explicit calls to enter/exit – but can't think of another solution

test/worker_test.py
Tarrasch: I thought so too. Somebody asked the same question and unfortunately there doesn't seem to be a better way really. http://stackoverflow.com/questions/8416208/in-python-is-there-a-good-idiom-for-using-context-managers-in-setup-teardown

Note that there still are multiple tests where I do use it as a context manager.

Tarrasch:
I generally like context managers. :)

Context managers are better than hoping that the GC will call __del__ as some point in time. And [this SO-answer](http://stackoverflow.com/a/865272) thinks so too. :)

erikbern:
lol I just googled how to use context managers and found this: https://gist.github.com/cjerdonek/2984537

Seems like you were there first :) How do you feel about using that approach instead? I can amend the PR if you want

Tarrasch:
Haha, did I say that? :D

Yea, you can do that. It's also suggested in the Stack Overflow thread, but it has only a moderate amount of upvotes. That said, I would of course be happy if you amended commits too. :)

erikbern:
let me fix in a separate pr

PR #1432, opened 2015-11-29
bigquery: Split the testcases into unit and integration tests.

mbruggmann:
The unit tests can now be run without having local GCP credentials. This means we can enable them on all PR builds as well.

ping @DeaconDesperado


erikbern:
LGTM!

PR #1435, opened 2015-11-30
Have snakebite HDFS client fallback to HADOOP_USER_NAME from env

nresare:
It turns out that the HADOOP_USER_NAME environment variable propagates nicely to the `hadoop jar` and `hdfs dfs` invocations that luigi triggers but the same is not true for the in proc snakebite client.

The attached patch rectifies this, provided that effective_user is not explicitly set in configuration.

### DIFF
--- a/luigi/contrib/hdfs/config.py
+++ b/luigi/contrib/hdfs/config.py
@@ -36,7 +36,11 @@ except ImportError:
 
 class hdfs(luigi.Config):
     client_version = luigi.IntParameter(default=None)
-    effective_user = luigi.Parameter(default=None)
+    effective_user = luigi.Parameter(
+        default=os.getenv('HADOOP_USER_NAME'),
+        description="Optionally specifies the effective user for snakebite. "
+                    "If not set the environment variable HADOOP_USER_NAME is "
+                    "used, else USER")
     snakebite_autoconfig = luigi.BoolParameter(default=False)
     namenode_host = luigi.Parameter(default=None)
     namenode_port = luigi.IntParameter(default=None)


mbruggmann:
:+1:

erikbern:
lgtm although if config is set then I think we should set the env variables.

i.e. if you don't want to use environment variables you shouldnt' have to

Tarrasch:
Can you add a parameter description saying that the default value is not a constant, it is what is in your `$HADOOP_USER_NAME`?

erikbern:
also would prefer if we don't fetch this on startup but when needed (i.e. pull from env dynamically)

otherwise i think it will try to pull it when building documentation etc

but no biggie

Tarrasch:
Ah true. It could be put in the code where it's actually used, or there could be a `def get_effective_user():` helper function

nresare:
I have tried to have the default of effective user to be set to something other than None in the documentation generation without luck (i.e. `HADOOP_USER_NAME=some_name tox -e docs`). Because of this I think it makes sense to keep `getenv` in the Parameter constructor call. I did update the documentation though to reflect behaviour.

nresare:
(As the snakebite client is lazily constructed but kept between usages, the semantics of getting the environment variable after the fact would not change but setting default in config.py is a more straight forward approach IMHO) 

ping @Tarrasch @erikbern for re-review.

Tarrasch:
I'm not sure I understand. Did you add any documentation? The patch seems to only contain a one line code. :)

nresare:
@Tarrasch I did push a new ref but it did not contain any differences compared to the old one, with the actual documentation update resting safely on my laptop. Silly me. This time I actually looked at the change in the github UI :)

Tarrasch:
LGTM.

PR #1436, opened 2015-12-01
Set initgroups to False in DaemonContext

xbello:
luigid cannot be initialized because python-daemon uses os.initgroups() which requires root permissions.

Setting initgroups to False python-daemon uses setgid instead.

### DIFF
--- a/luigi/process.py
+++ b/luigi/process.py
@@ -100,7 +100,8 @@ def daemonize(cmd, pidfile=None, logdir=None, api_port=8082, address=None, unix_
     ctx = daemon.DaemonContext(
         stdout=stdout_proxy,
         stderr=stderr_proxy,
-        working_directory='.'
+        working_directory='.',
+        initgroups=False,
     )
 
     with ctx:


Tarrasch:
There seem to be an intermittent failure. I restarted and it's green now.

@xbello, is this for a particular platform or something? I'm pretty sure I've been able to run luigid when not being root.

xbello:
I don't know. I've just installed it in a new virtualenv and it didn't work
with:

luigid --background --pid ./luigi-server.pid --log ./luigi-server.log

With error:

File
"/home/xbello/Projects/Test_1/lib/python3.4/site-packages/daemon/daemon.py",
line 635, in change_process_owner
    os.initgroups(username, gid)
daemon.daemon.DaemonOSEnvironmentError: Unable to change process owner
([Errno 1] Operation not permitted)

After adding that line everything worked as expected. Pip freeze:

backports-abc==0.4
docutils==0.12
lockfile==0.12.2
luigi==2.0.0
python-daemon==2.1.0
sarge==0.1.4
SQLAlchemy==1.0.9
tornado==4.3

python --version

Python 3.4.1

2015-12-01 10:33 GMT+01:00 Arash Rouhani <notifications@github.com>:

Tarrasch:
Well it's green, so LGTM. I'll let someone else upvote before we merge.

xbello:
This is the change in python-daemon that breaks Luigi:

https://alioth.debian.org/scm/browser.php?group_id=100328

xbello:
This is the change in python-daemon, 4 days ago, that broke my luigid:

https://alioth.debian.org/scm/loggerhead/python-daemon/python-daemon.devel/revision/350/daemon/daemon.py?remember=350

Tarrasch:
Oh, so you mean that a recent upstream change have made this break?

xbello:
I guess so.

erikbern:
won't this break for everyone not having the latest version of daemon? (where initgroups is not a proper argument)

xbello:
Yes, it would break with older versions.

Tarrasch:
@xbello, maybe we should put an upper version bound on the python-daemon package?

PR #1438, opened 2015-12-02
Integrate webhdfs with rest of hdfs module

Tarrasch:
With this patch you can use normal HdfsTargets together with webhdfs.
The old WebHdfsTarget still exists but is deprecated.

See the documentation for the [hdfs] section for enabling webhdfs.

This will change the behavior of the existing implementation a bit.
The issue was that it was using [hdfs]namenode_port for the webhdfs
port. But the namenode port that for example snakebite and
hadoopcli uses is different from the webhdfs port.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -332,9 +332,11 @@ Parameters controlling the use of snakebite to speed up hdfs queries.
 
 client
   Client to use for most hadoop commands. Options are "snakebite",
-  "snakebite_with_hadoopcli_fallback", and "hadoopcli". Snakebite is
-  much faster, so use of it is encouraged. Using snakebite requires it
-  to be installed separately on the machine. Defaults to "hadoopcli".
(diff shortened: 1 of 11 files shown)


Tarrasch:
What do you think about this @danielfrg? This is kind of what I wanted in #507 [here](https://github.com/spotify/luigi/pull/504#issuecomment-59653048). I had totally forgotten about it up until now ^^

danielfrg:
That is really nice and its going to simplify some of my jobs ;)
Definitely +1

Tarrasch:
Whoups. I meant to refer to #504, not 507.

@danielfrg I'm happy that this can be useful for you too. And thanks for the implementation starting point you had created, it made it much easier for me to create this patch :)

Do anyone else want to review? Otherwise I'll merge.

Tarrasch:
I've been using this in production and it seems to be working. It would be super great if you could start using this as well @danielfrg and let me know how you like it.

PR #1440, opened 2015-12-02
Range helpers: Honor param_name when inferring fs

Tarrasch:
Range* allows you to specify the name of the parameter that is the date
the range is about. Another orthogonal feature is that you don't need to
provide a bulk_complete for the most common file system use case.

However, when using both these features doesn't work. This patch fixes
that.

### DIFF
--- a/luigi/tools/range.py
+++ b/luigi/tools/range.py
@@ -82,6 +82,8 @@ class RangeBase(luigi.WrapperTask):
     datetime parameter class, e.g. DateParameter, DateHourParameter, ..., and
     different parameter naming, e.g. days_back/forward, hours_back/forward,
     ..., as well as different documentation wording, for good user experience.)
+
+    Subclasses will need to use the ``of`` parameter when overriding methods.
     """
     # TODO lift the single parameter constraint by passing unknown parameters through WrapperTask?
     of = luigi.TaskParameter(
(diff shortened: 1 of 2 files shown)


Tarrasch:
@ulzha @joeshaw, can anyone of you review please? :)

luigi/tools/range.py
```diff
@@ -516,7 +516,7 @@ def missing_datetimes(self, task_cls, finite_datetimes):
         except NotImplementedError:
             return infer_bulk_complete_from_fs(
                 finite_datetimes,
-                lambda d: task_cls(self.datetime_to_parameter(d)),
```
joeshaw: this change removes the use of `task_cls` as an argument to `missing_datetimes()`.  So it seems to me that either this patch should remove that argument (but I'm not sure what side effects that might have... when is `self.of_cls` ever different from `task_cls`?), or that instead of dropping `task_cls` as an argument to `_instantiate_task_cls()` this should call `self._instantiate_task_cls(task_cls, self.datetime_to_parameter(d))` instead.

In other words, we either need `task_cls` or we don't, and if we don't we shouldn't pass it around and should just use `self.of_cls` instead.

Tarrasch:
I made a pretty big change to correct for the flaw that you found @joeshaw. How does it look now? Please see the amendment to the commit message too. :)

Also ping @ulzha

joeshaw:
LGTM :+1: but definitely could use another pair of eyes as I am still new to the codebase.

PR #1441, opened 2015-12-02
Don't print nested stack traces for run() failures

Tarrasch:
In python3, stack traces seem to report if the raised exception happened
during another exception handling. It will then say something like
"exception happned during handling of above exception". This is nice,
but just cluttery in the case of our tracking_url_callback hack.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -102,11 +102,14 @@ class TaskProcess(multiprocessing.Process):
         self.timeout_time = time.time() + worker_timeout if worker_timeout else None
 
     def _run_get_new_deps(self):
+        run_again = False
         try:
             task_gen = self.task.run(tracking_url_callback=self.tracking_url_callback)
         except TypeError as ex:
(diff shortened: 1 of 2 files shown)


luigi/worker.py
```diff
@@ -102,11 +102,14 @@ def __init__(self, task, worker_id, result_queue, random_seed=False, worker_time
         self.timeout_time = time.time() + worker_timeout if worker_timeout else None
 
     def _run_get_new_deps(self):
+        run_again = False
         try:
             task_gen = self.task.run(tracking_url_callback=self.tracking_url_callback)
         except TypeError as ex:
             if 'unexpected keyword argument' not in getattr(ex, 'message', ex.args[0]):
```
erikbern: what is this if statement – it's pretty horrible

Tarrasch: It was the best way we were able to implement #1298.

I could send a PR later that adds comments around the code if it helps.

Tarrasch:
Does this look sane to you @erikbern?

PR #1443, opened 2015-12-02
Added initgroups kwarg if python-daemon version is over 2.1.0

xbello:
python-daemon in its upgrade to 2.1.0 included a new kwarg that broke the daemonization.

This commit checks its version and ads the "initgroup" kwarg if needed. Tested manually with python-daemon 2.0.0 and 2.1.0 versions.

### DIFF
--- a/luigi/process.py
+++ b/luigi/process.py
@@ -77,6 +77,8 @@ def _server_already_running(pidfile):
 
 def daemonize(cmd, pidfile=None, logdir=None, api_port=8082, address=None, unix_socket=None):
     import daemon
+    from daemon._metadata import get_distribution_version_info as gdvi
+    from distutils.version import StrictVersion
 
     logdir = logdir or "/var/log/luigi"
     if not os.path.exists(logdir):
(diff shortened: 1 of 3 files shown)


Tarrasch:
Cool. But can you squash the commits?

Also, did you think about just setting a version constraint for python-daemon as we discussed in #1436?

xbello:
Do you mean forcing the user to install a python-daemon under 2.1.0?

Tarrasch:
Yea. Just throwing the idea out there. But if you think it's not feasible lets go for this solution. :)

xbello:
Whatever you feel its correct. The 2.1.0 version of python-daemon only included this change. I saw another projects in the same spot (https://github.com/ThiefMaster/maildump/issues/17) and they opted for the version bump.

luigi/process.py
```diff
@@ -97,11 +99,16 @@ def daemonize(cmd, pidfile=None, logdir=None, api_port=8082, address=None, unix_
...
-        stderr=stderr_proxy,
-        working_directory='.'
-    )
+    daemon_kwargs = {
+        "stdout": stdout_proxy,
+        "stderr": stderr_proxy,
+        "working_directory": '.',
+    }
+
+    if (StrictVersion(gdvi().get("version") or "0.0.0") >= StrictVersion("2.1.0")):
```
erikbern: minor thing (not blocking) – the `get` method takes a default as a second argument, so you could do `.get("version", "0.0.0")`

also not blocking – comparing strings is generally not correct but works in this case. For instance "1.10" should be larger than "1.2"

erikbern: i also would prefer it if you just apply "it's easier to ask for forgiveness than permission" and just did two calls – try first with the argument, if it throws an `ArgumentError` then try without

xbello: I thought that using StrictVersion from distutils to compare versions was the way to go.

>>> from distutils.version import StrictVersion
>>> StrictVersion("1.10") > StrictVersion("1.2")
True

erikbern: i'm dumb, sorry not sure what i was thinking. you are totally right

Issue #1312, opened 2015-10-15
Proposal to change task_id serialisation


stephenpascoe:
## Problem summary

luigi.Task.task_id is currently used for 2 purposes within the codebase: a unique identifier for the task and a short summary of the task and its significant parameters.  IMO these two purposes are in conflict and we should separate them out.  The main problems I've encountered boil down to these points:

 - task_id cannot be reliably deserialised to get parameter values.  We know we shouldn't use it as such, but it is sometimes tempting.

 - task_id cannot be reliably serialised from a dictionary of parameters.  It depends on the order of parameter declarations in the task class.  Therefore we can't create it if we don't have the class (such as in the scheduler, javascript or external tools).  We always need to pass around both the task_id and the parameters.  This duplicates all parameter values.

 - DbTaskHistory does not store task_id at all.  It stores parameters in a table which looses ordering information.  We cannot reliably find a task in DbTaskHistory given a task_id.  Even if we know the parameters, we need a query with 1 join per parameter, which can be inefficient with many parameters with similar values.

## Proposal

 1. Change task_id to be based on the task_family and a hash of the parameter dictionary.  This would still give some human-readable information but would hide the parameter values.  E.g. "<family>#<hash>" where <hash> is the md5 digest of a sorted json serialisation of the parameter dictionary.

 2. Keep str(task) as a human readable summary of the task

 3. Review references to task_id in the worker, scheduler and visualiser code and use task_id when we want an identifier and str(task) when we want a summary.

 4. Add the new task_id as a column in the DbTaskHistory database

I expect this would be disruptive to people's downstream tools.  It would need evaluating and probably a major version change.  If I get the impression this change would be accepted, I am willing to do the implementation, test it and run it in production for a while before it is merged into master.

ulzha:
Sounds good!

For correct hashing one might need canonicalization on codepoint escaping too in JSON strings, not only the sorting of JSON keys. (Googled http://wiki.laptop.org/go/Canonical_JSON as an example, but it seems to be outright specifying non-JSON, by allowing control characters...?!)

ulzha:
I always wondered if it might even be preferable to deprecate the ambiguous summary format completely. Instead always have str(task) be of the form `<Python class name>(<significant parameters as canonicalized JSON>)`, with some conventions about DateParameter etc. string representations...

erikbern:
[...]

PR #1444, opened 2015-12-02
Task id hashing

stephenpascoe:
Implementing semi-opaque, hashed task_ids as discussed at #1312.

This is the version I will be testing on our infrastructure in the next few weeks.  This PR is intended to give my proposal better visibility whilst I test.

The PR replaces task_ids with a value which is reliably unique but you can't extract the full parameter values from it.  The actual algorithm is `family_pval1_pval2_pval3_hash` where:
 - family: The task_family (a.k.a task_name)
 - pval\*: A **truncated** serialisation of the first 3 parameter values, sorted by parameter name
 - hash: A md5hash of the canonical JSON serialisation of the family and parameters, truncated to 10 characters.

`str(task)` will return the traditional serialisation.

The db_task_history database schema is changed to include tash_id in the tasks table.  No further use is made of this column at this stage.  A migration script is supplied to add the column to an existing database.  This has been tested on MySQL only.

All tests are updated to pass by replacing hard-coded task_ids with reference to `Task.task_id`.


erikbern:
this looks great!

luigi/task.py
```diff
@@ -253,14 +255,20 @@ def __init__(self, *args, **kwargs):
...
+        # sorted by parameter name and a md5hash of the family/parameters as a cananocalised json.
+        TASK_ID_INCLUDE_PARAMS = 3
+        TASK_ID_TRUNCATE_PARAMS = 16
+        TASK_ID_TRUNCATE_HASH = 10
+
+        params = self.to_str_params(only_significant=True)
+        param_str = json.dumps(params, separators=(',', ':'), sort_keys=True)
+        param_hash = hashlib.md5(param_str.encode('utf-8')).hexdigest()
+
+        param_summary = '_'.join(p[:TASK_ID_TRUNCATE_PARAMS]
```
erikbern: you should probably remove non-alphanumeric characters from `param_summary`

luigi/task.py
```diff
@@ -253,14 +255,20 @@ def __init__(self, *args, **kwargs):
...
+                                 for p in (params[p] for p in sorted(params)[:TASK_ID_INCLUDE_PARAMS]))
+        self.task_id = '{}_{}_{}'.format(self.task_family, param_summary, param_hash[:TASK_ID_TRUNCATE_HASH])
```
erikbern: so this doesn't include a `#` but some other code implies we're using a `#` (eg your change to luigi/static/visualiser/index.html)

luigi/tools/migrate.py
```diff
@@ -0,0 +1,133 @@
+"""
```
erikbern: Does this have to be run manually?

stephenpascoe: Yes, I can make an automatic check on luigid startup.  The options are either migrate automatically or exit with message to run luigi-migrate.

luigi/task.py
stephenpascoe: I've removed references in `index.html` and `sge.py`.  Shout if you know of any more.

luigi/task.py
stephenpascoe: I've removed refs in `index.html` and `sge.py`.  Shout if you see any more.

luigi/task.py
stephenpascoe: Done, all non-alphanumeric (+ "_") converted to "_".

luigi/tools/migrate.py
erikbern: i think according to semver we need a major revision bump right? at least a minor one seems warranted.

erikbern:
this is great – would be great if some other people can weigh in. @Tarrasch @econchick @freider ?

luigi/tools/migrate.py
stephenpascoe: Yes, at least a minor version change.

stephenpascoe:
It's worth saying that we don't use Hadoop or any of the other integrations so they won't get tested by me.  Further input would be good.

erikbern:
hadoop has tests in travis so should be fine

luigi/task.py
```diff
@@ -324,7 +337,19 @@ def __hash__(self):
...
     def __repr__(self):
-        return self.task_id
+        params = self.get_params()
+        param_values = self.get_param_values(params, [], self.param_kwargs)
+
+        # Build up task id
+        task_id_parts = []
+        param_objs = dict(params)
+        for param_name, param_value in param_values:
+            if param_objs[param_name].significant:
```
Tarrasch: Now, since we're not sending this to the scheduler any more. Maybe we should include the insignificant parameters too?

(What I'm suggesting is to remove this one line)

luigi/task.py
```diff
@@ -324,7 +337,19 @@ def __hash__(self):
...
+                task_id_parts.append('%s=%s' % (param_name, param_objs[param_name].serialize(param_value)))
```
Tarrasch: I believe this line should now be

```
task_id_parts.append('%s=%r' % (param_name, param_value)) 
```

Note the second is %r not %s. Because According to this [SO answer](http://stackoverflow.com/a/2626364/621449), one wants to be close to the `eval(repr(x)) == x` property.

luigi/task.py
```diff
@@ -324,7 +337,19 @@ def __hash__(self):
         return self.__hash
 
     def __repr__(self):
-        return self.task_id
+        params = self.get_params()
+        param_values = self.get_param_values(params, [], self.param_kwargs)
+
+        # Build up task id
+        task_id_parts = []
```
Tarrasch: Rename this to something like `repr_parts`. :)

luigi/task.py
```diff
@@ -324,7 +337,19 @@ def __hash__(self):
         return self.__hash
 
     def __repr__(self):
```
Tarrasch: Perhaps a docstring:

```
Build a task representation, for example "MyTask(param1=5, param2='5')"
```

test/task_serialize_test.py
```diff
@@ -0,0 +1,113 @@
+"""
+We want to test that task_id is consistent when generated form:
```
Tarrasch: generated *from*

Legally I suppose this file needs one of those license header thingies. :)

test/task_serialize_test.py
```diff
@@ -0,0 +1,113 @@
...
+
+ 1. A real task instance
+ 2. The task_family and a dictionary of parameter values (as strings)
+ 3. A json representation of #2
+
+We use the hypothesis package to do property-based tests.
```
Tarrasch: Cool! :D :D 

I love the functional programming influence luigi is getting more and more :)

Tarrasch:
Looks good. :)

I know that @ulzha is usually saying wise things when it comes to design. Do you have any thoughts on this Uldis?

Tarrasch:
@stephenpascoe You should not need to worry about other modules breaking as everything in luigi is (usually) tested by Travis. :)

sisidra:
Sorry for being late for discussion, but I would have loved to see URIfication for task id serialization. Like
`taskId?param=value&...` and for hash you could `taskId#hash` as mentioned before.
Everybody cries when sees self-invented formats like `taskId(parm=value)`... :(

luigi/task.py
stephenpascoe: I'm undecided.  Maybe people will want to keep repr() short by marking configuration parameters as insignificant?  I don't use this feature so I don't really have an opinion.

erikbern: `__repr__` is only used to assist debugging right? yeah in that case let's include all params

erikbern:
@sisidra my concern using '#' in the task id is now you need to uri encode it whereas if it's only [A-Za-z0-9_] then you don't need to worry

stephenpascoe:
@sisidra, `taskId(parm=value)` will be gone except for `str(task)`.  task_id will no longer be a serialization in the sense that it doesn't contain all the information, only the family and some param-value prefixes.  

For serialization we have the JSON {'family': ..., 'params': ...} returned by the API.  You could URIfy that if you wanted, with correct escaping.

luigi/task.py
stephenpascoe: I've tried this but the results are inconsistent in the tests.  I'm seeing a mixture of `param='1'` and `param=1` for `IntParameter`s.  I'm not sure what's going on but I think keeping `str(task)` the same for backward compatibility is the pragmatic choice.

stephenpascoe:
This branch has been running in production for a week without issue.  A couple of tests are failing because of a syntax error in the HTTPretty mocking library which has not been fixed in pypi.

AFAIK the only outstanding issue is the version number.  I suggest 2.1.1 would be suitable, considering how recently we moved to 2.0.

erikbern:
sgtm. but users have to run a migration script right?

stephenpascoe:
Migration script has been removed.  The schema is now automatically upgraded.

erikbern:
that's great. cool. let's merge this!

erikbern:
Will merge on a successful build.

Then let's have it sit in master for say two weeks before we publish to PyPI. I'm still a bit scared :)

stephenpascoe:
HTTPretty fix requested at https://github.com/gabrielfalcao/HTTPretty/issues/278

erikbern:
let's just set the version of httpretty in tox.ini to the latest working version

Tarrasch:
I think the commit history is a bit too dirty for merge. In particular no reverting commits or "fix pep8" commits should be in the commit log. I believe. Having those makes reverting commits much harder for us maintainers.

Other than that. Looks good.

stephenpascoe:
Squashed some commits.

stephenpascoe:
All tests passing.  Ready to merge?

erikbern:
LET'S MERGE

erikbern:
let's see if this causes any trouble

stephenpascoe:
Great. :+1:

Tarrasch:
Wow, this finally happened. Amazing execution of this @stephenpascoe!

Tarrasch:
@stephenpascoe, do you think this will cause breakage for this use case?

https://github.com/spotify/luigi/blob/154c283e9ff322c619635eb97719684da2f1184c/luigi/contrib/rdbms.py#L98-L102

I mean people who upgrade notice that all their `SomeDatabaseTarget().exists()` methods will suddenly return `False`.

stephenpascoe:
@Tarrasch, yes, it looks like it.  

You could migrate the database by locating each row via `update_id=str(Task)` and updating it to `self.task_id`.  However, because we removed the `only_significant` parameter from `to_str_params()` that wouldn't work if there are insignificant parameters.

Tarrasch:
Hmm...

I suppose the renaming of tables also isn't easy either since each db renames tables in different ways.

stephenpascoe:
Migration tools like alembic can do it so I guess we could work it out for anything using sqlalchemy.  Do you think it is enough to move the table to a backup and start again?

Tarrasch:
Not sure of exactly what you mean with "move the table to a backup and start again".

But happy with any solution that strikes a reasonable work-for-maintainers/work-for-users balance.

erikbern:
in retrospect that mechanism wasn't great to start with, we shouldn't have had it in the code.

where is `update_id` actually used?

Tarrasch:
Unfortunately I believe it's quite widely used. All CopyToTable jobs would be affected.

Tarrasch:
> in retrospect that mechanism wasn't great to start with, we shouldn't have had it in the code.

I can totally see what tempted the authors to write that, it's so convenient and short! :)

stephenpascoe:
`luigi.contrib.rdbms.CopyToTable` is inherited by 2 task classes:
 - `luigi.postgres.CopyToTable` 
 - `luigi.contrib.redshift.S3CopyToTable`

There is also `luigi.contrib.sqla.CopyToTable` which also puts `self.task_id` in a database.

erikbern:
a dumb temporary fix would be to put back the old task_id code into `luigi.deprecated.old_task_id` or something like that. just buys us some time though

erikbern:
is there some automatic way of converting an old task_id into a new style task_id? in that case we can try to migrate the marker tables

stephenpascoe:
We can convert old_task_id to new_task_id in cases where we don't run into the original deserialisation problems (param values containing [ "'=] etc. )  We can convert using an instantiated Task instance if we roll-back the change to `to_str_params()`.

Tarrasch:
Hmm. I wonder if we can just not care about fixing backward compatibility for the database-target issue. Since luigi is a bit makeish, I think all data will be generated where-ever needed/requested in the dependency chain.

I mean, surely people have added parameters and then had all their database uploads being done again without them really caring. The database ingestions are in many cases small compared to the data crunched to produce the ingestion data.

Tarrasch:
Again thanks for this contribution @stephenpascoe. But I noticed a minor niceness glitch in the visualiser that I think is because of this PR:

Before:

Before

After:

selection_001

I mean that the parameters are formatted like a python dict now and not in the slightly nicer `(k1=v1, k2=v2)` format. Can this be fixed? :)

stephenpascoe:
Yes, I was aware of this. I can fix it. 

--
Stephen Pascoe from iPhone

Tarrasch:
Thanks for verifying! Kudos for fixing :)

PR #1445, opened 2015-12-02
contrib.ecs: check the exit code

gregwebs:
@jfeala this is untested both in the production sense and the unit test sense. But I believe it is necessary.


erikbern:
not sure whta this does but lgtm

erikbern:
can you fix the flake8 issues?

gregwebs:
I am not using this code. I am providing this as a benefit for @jfeala. Perhaps it should be a github issue instead? Feel free to close this.

jfeala:
Thanks again, Greg. I'll add tests and submit a new PR soon, adding @gregwebs as author if possible. Ok to close

PR #1447, opened 2015-12-02
Allow scheduler to listen on unix socket normally

### LINKED ISSUE #1437: Scheduler only hosts on unix socket when run in the background
Support for hosting the central scheduler on a unix socket was added, which is nice, but the scheduler ignores the `--unix-socket` argument from the command line when `--background` is not also supplied.  

This will work properly, and the scheduler will listen on the provided unix socket:
```
luigid --unix-socket /path/to/socket --background
```

With this command, the scheduler will still listen on the default port (8082):
```
luigid --unix-socket /path/to/socket
```

Fixing this would be a simple matter of passing the `unix_socket` argument onto the call to `server.run` in the case where the server is not daemonized, but was there a reason this functionality was left out in the first place?  If so, it probably ought to be in the documentation; as is, reading it gives me the impression that I should be able to tell the scheduler to listen on a unix socket regardless of whether it's running in the background.

jasonnance:
Previously, the scheduler would only listen on a supplied unix
socket when run in the background.  When run normally, it would
ignore the given unix socket and serve on the default port.  This
is fixed by passing the given unix_socket to server.run when
the scheduler is run normally.  Fixes #1437

### DIFF
--- a/luigi/cmdline.py
+++ b/luigi/cmdline.py
@@ -41,4 +41,4 @@ def luigid(argv=sys.argv[1:]):
                                 filename=os.path.join(opts.logdir, "luigi-server.log"))
         else:
             logging.basicConfig(level=logging.INFO, format=luigi.process.get_log_format())
-        luigi.server.run(api_port=opts.port, address=opts.address)
+        luigi.server.run(api_port=opts.port, address=opts.address, unix_socket=opts.unix_socket)


erikbern:
lgtm

Tarrasch:
Thanks!

PR #1449, opened 2015-12-08
Fix issue with complete external tasks having missing deps

davemt:
There are certain scenarios where an external task has actual deps, such as when a task has been made external with `externalize()` method, or if someone wants to simply subclass then set `run = NotImplemented`.

At runtime if the output exists for the external task but not for its upstream dependencies, this should be OK (task success). However this case was broken because we failed in the missing-dependencies check
before ever checking `task.compete()`. There's no reason to ever even check deps.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -161,7 +161,12 @@ class TaskProcess(multiprocessing.Process):
                 # External task
                 # TODO(erikbern): We should check for task completeness after non-external tasks too!
                 # This will resolve #814 and make things a lot more consistent
-                status = DONE if self.task.complete() else FAILED
+                if self.task.complete():
+                    status = DONE
+                else:
+                    status = FAILED
+                    expl = 'Task is an external data dependency ' \
+                        'and data does not exist (yet?).'
             else:
                 new_deps = self._run_get_new_deps()
                 status = DONE if not new_deps else PENDING
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -144,11 +144,15 @@ class TaskProcess(multiprocessing.Process):
         missing = []
         new_deps = []
         try:
-            # Verify that all the tasks are fulfilled!
-            missing = [dep.task_id for dep in self.task.deps() if not dep.complete()]
-            if missing:
-                deps = 'dependency' if len(missing) == 1 else 'dependencies'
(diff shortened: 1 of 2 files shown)


Tarrasch:
I think I understand what this does and it does look good. :)

But I wonder if a better solution is not just to make `externalize()` also make `requires()` return `None`. Or do you like it that your external tasks show that they have dependencies in the UI?

erikbern:
let me look at this later, but one thing i've been planning for a long time is changing how external task works. the `if self.task.run != NotImplemented` seems a bit cargo cult-ish because we're checking for an artifact of something we really want to detect. never been super happy with it

davemt:
I don't care about external tasks showing dependencies in UI (but actually it might be nice to have some indicator of external-ness there in general). But I do know that there are existing use cases in our organization where `run = NotImplemented` is used to create an external version of a non-external task at class level. This is obviously stemming from the historical definition of an external task being just `run = NotImplemented`.  So this fix works for those extra use (mis-use?) cases.

If there existed a better way of defining is-external, that would definitely make things cleaner. I could even imagine wanting to toggle this dynamically from CLI args or config.

Tarrasch:
I think this can be merged. @erikbern, opinions?

erikbern:
Was looking at the code again. I'm a bit confused. Why is the task even scheduled in the first place if it's already complete?

davemt:
The external task is incomplete when scheduled. Between time it is added to scheduler and time it is handed out in in get_work, the task becomes complete (external input is created). Or, between retries w/ retry-external-tasks.

We have task graphs where these time gaps can be significant.

erikbern:
oh... cool. got it. I guess in that case we should just do a completeness check in the TaskProcess and abandon if it's already run. This could be done regardless of whether it's external. Would that be doable?

davemt:
Are you saying to do the completeness check early in `TaskProcess.run` and return immediately if the task is already complete? If so, I think we would also need to report the complete state to scheduler to guarantee that downstream task(s) would be handed out in get_work. Whatever task originally did the work to complete (assuming it was even a luigi task) may not have same task id or be in scheduler still to fulfill the deps-complete check.

Let me know if you meant something different.

erikbern:
sorry this has been sitting here. i have some refactoring ideas but will probably never have time and i don't want it to hold up this PR

PR #1451, opened 2015-12-09
Checks if SMTP server supports 'starttls' before attemping to call sm…

sklarsa:
Checks if SMTP server supports 'starttls' before attemping to call smtp.starttls()


erikbern:
lgtm although i hope no one is expecting TLS and wants it enforced

sklarsa:
Would it be worth adding a force-TLS to the config?

Tarrasch:
Hi @sklarsa, thanks for this contribution. I will not merge this because of the unclean commit history. But I've got good news. I needed this exact patch myself so I did all the work for you in #1464. I hope you don't mind. :)

Closing in favor of #1464.

PR #1452, opened 2015-12-09
allow binary opens

bmaggard:


Tarrasch:
@bmaggard Can you clarify what this does, why you need to do this etc? In what circumstances do one run into problems (python version? OS? etc.)

bmaggard:
Python 3.  Linux.  I have Tasks that utilize the wonderful [petl](https://github.com/alimanfoo/petl) package, which utilizes the 'b' flag when it opens files (why?).  Apart from the code affected by this patch, Luigi does not seem to care, and beyond not allowing 'append' or 'update' operations, should not (re:(https://groups.google.com/d/msg/luigi-user/qnY64bRDiBw/OyWwRNXMCiYJ)) care, how I want to open files.  I realize now that this hasty patch is deficient in that allowing 'b' has been achieved, but now adding '+' is also allowable ('w+').   How about:

`if mode[:1] == 'w' and not '+' in mode:`

instead? (works for me, but I don't 'python 2', so I'm not sure of the implications for the general populace)

erikbern:
sgtm (`and '+' not in mode` might be more pythonic)

PR #1453, opened 2015-12-09
Fixed: Broken Redshift Table Creation

ybroze:
Table creation is broken in `contrib.redshift`.

The definition of `table_type` and `table_attributes` had been changed to include a `@property` decorator in October, but their use in `create_table` was still expecting methods.

This fix assumes we want properties (as doing so makes it easier for subclasses to simply fill in static values).

Unit test included.

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -216,10 +216,10 @@ class S3CopyToTable(rdbms.CopyToTable):
             query = ("CREATE {type} TABLE "
                      "{table} ({coldefs}) "
                      "{table_attributes}").format(
-                type=self.table_type(),
+                type=self.table_type,
                 table=self.table,
                 coldefs=coldefs,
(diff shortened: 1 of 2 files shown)


erikbern:
lgtm but will let someone who actually uses redshift weigh in too

Tarrasch:
@ybroze, Can you blame the file to see the authors? Then you can @ summon them here in this PR.

ybroze:
@dlstadther, @ddaniels888, @steenzout, @rantav, how does this look?

dlstadther:
lgtm. I apologize for not returning to correct these issues in a timely fashion. I was the cause of the broken tests. Thanks for fixing @ybroze .

ybroze:
@dlstadther, no problemo!

Tarrasch:
Can this be merged @ybroze?

PR #1455, opened 2015-12-10
fix for 'allow binary opens'

bmaggard:
This commit prevents 'append' and 'update' modes, and is a fix to the previous commit, which was intended to allow 'binary' mode.

### DIFF
--- a/luigi/file.py
+++ b/luigi/file.py
@@ -112,7 +112,7 @@ class LocalTarget(FileSystemTarget):
                 pass
 
     def open(self, mode='r'):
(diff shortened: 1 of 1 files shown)


Tarrasch:
Can you please squash these commits? We try to keep the commit history clean. :)

And please add a description. Luigi has tons of modules where even the most active maintainers have no idea what is doing what.

And of course thanks for contributing with your patches! :)

Tarrasch:
Works seem to continue in #1456. This itself was a fix of #1452.

PR #1460, opened 2015-12-16
log scheduler updates with multi workers as DEBUG

adamgreenhall:
when running luigi with multiple workers, this changes the logging so that updates from the scheduler saying that there isn't any more work (because one worker is taking longer) get logged as ``DEBUG`` rather than ``INFO``. This gets rid of a lot of logging noise (several messages every few seconds) if you have your logger set to ``INFO``.


erikbern:
lgtm

Issue #1273, opened 2015-10-06
Parameters aren't validated when passed via constructor


mikekap:
``` python
class Task(luigi.Task):
    p = luigi.IntParameter()

print Task(p='hi').p  # 'hi'
```

While the above is a trivial case of "something will probably break down the line," this can introduce subtle bugs when e.g. a date parameter is passed to a datetime parameter (or vice-versa).

My suggestion would be to add some sort of validation, and possibly the ability to tweak the value (similar to parse) to the Parameter interface. The latter would be useful when implementing an e.g. "every N hours" parameter that automatically floors to the nearest hour.

erikbern:
Yeah that sounds reasonable. Would you mind putting together a PR?

Tarrasch:
Sounds great. Though beware that a couple of (bad) tests would break. Also I'm not sure what to do with the `None` issue.

dlstadther:
Resolved

PR #1461, opened 2015-12-17
Add normalize(v) to Parameter to do pre-processing on arguments passed to Task

mikekap:
This can be used for validation as well, but for the sake of back-compat this just introduces the functionality. This starts fixing issue #1273.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -202,7 +202,7 @@ class Parameter(object):
         if value == _no_value:
             raise MissingParameterException("No default specified")
         else:
-            return value
+            return self.normalize(value)
 
     def parse(self, x):
         """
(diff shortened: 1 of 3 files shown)


Tarrasch:
LGTM.

In addition to doing normalizing of bools, one could maybe also normalize all date parameters. For instance we could normalize so we don't store `2015-May-20 13:37` but always store `2015-May-20 00:00`. Is this how you envision it @mikekap?

This is changing the core of luigi. Perhaps @erikbern want to review as well?

mikekap:
Yes, definitely how I envision it. I wanted to do chrono params that in a separate PR (see https://github.com/spotify/luigi/compare/master...vine:datetime_normalize?expand=1). I ran into a slightly larger refactoring with those. I could also add validation for int/float parameters in this one though.

Also any ideas on why the CI is failing? I don't think I broke the visualizer with this change, but I could never know for sure... Unfortunately phantomjs is broken on the latest osx so it'd be nice if someone can root cause that failure without me having to go down that rabbit hole :X.

Tarrasch:
Let's keep this PR to only this change. No need to implement all the `normalize()`. Method yet. Oh, your `datetime_normalize` branch looks amazing!

Sorry about Travis failing, it was an intermittent failure. Since I've quit Spotify I have not prioritized to fix these regressions.

I just want to give @erikbern a chance to review before we Merge. This looks perfect to me otherwise.

Tarrasch:
Thanks @mikekap

PR #1393, opened 2015-11-12
Fixed: Rdbms and Redshift Properties

dlstadther:
Changed Rdbms' `update_id` to a `@property`. Appropriate changes were made to Redshift.py and Postgres.py. Also changed Redshift's `do_truncate_table` and `queries` to properties. Finally, fixed references to properties as `self.<property>` rather than their previous method forms `self.<property_method>()`.

PR #1463, opened 2015-12-18
Fixed: Rdbms and Redshift Properties

dlstadther:
Same as PR #1393 b/c I screwed up the rebase...

Changed Rdbms' update_id to a @property. Appropriate changes were made to Redshift.py and Postgres.py. Also changed Redshift's do_truncate_table and queries to properties. Finally, fixed references to properties as self.<property> rather than their previous method forms self.<property_method>().


dlstadther:
Link back to PR #1444 involving update_id

Tarrasch:
LGTM

PR #1465, opened 2015-12-21
bigquery: New task to create views.

mbruggmann:
This is handy to keep a set of views in sync with BigQuery (e.g. from a CD pipeline).

Example task:
```python
import luigi.contrib.bigquery as bigquery

class BQView(bigquery.BigqueryCreateViewTask):
    view='''SELECT * FROM [dataset.table] LIMIT 10'''

    def output(self):
        return bigquery.BigqueryTarget('project_id', 'dataset', 'view')
```

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -216,6 +216,59 @@ class BigqueryClient(object):
 
             response = request.execute()
 
+    def get_view(self, table):
+        """Returns the SQL query for a view, or None if it doesn't exist or is not a view.
+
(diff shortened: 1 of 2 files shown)


Tarrasch:
Cool!

Can you find some bigquery user to review this too?

mbruggmann:
Maybe @DeaconDesperado could have a look?

DeaconDesperado:
Looking now!

DeaconDesperado:
Nice patch, :+1:

Issue #1467, opened 2015-12-23
Memory leak when reading from huge gzipped S3Target


jpiper:
I have a 80GB `.csv.gz` file on S3 that I'm trying to use an in input, e.g.

`luigi.s3.S3Target(S3_PATH, format=luigi.format.Gzip)`

When I try to iterate through it during my Task's `run()` method, even when passing the data to nowhere, my memory gets eaten up completely!

```python
infile = self.input().open('r')
for i in infile:
     pass
```

I'm on boto 2.38.0 and luigi 2.0.1

deuxpi:
I'm going to guess that line 79 of luigi/format.py is a major suspect because I don't expect boto to provide a `fileno()` attribute on the key. That reads the whole data in memory to be copied to a temporary file. A simple fix would be to read from the `input_pipe` by chunks, just like boto would do for us if it were a plain file.

Tarrasch:
Wow, thanks @deuxpi !!

jpiper:
Wow, talk about a fast turnaround! I just can't believe that I'm the person person to encounter this issue :)

PR #1468, opened 2015-12-23
Format classes will not read whole files in memory

deuxpi:
If an input pipe did not provide a file descriptor via a fileno() method (such as a S3 key object), the whole input data was read in memory to be written to a temporary file. To avoid using too much memory and improve efficiency, the data is copied immediately by chunks of the size of the file system blocks. This does not affect most cases of Format classes which pass a file descriptor to the subprocess.

Fixes issue #1467.

Below is a small test task that will trigger the bug with a huge data set. Without the fix, be prepared to have it use all memory it is allowed to get. Otherwise, don't forget to kill it before you download the whole 1.7 GB. 

```python
from luigi import ExternalTask, Parameter, Task
from luigi.format import Gzip
from luigi.s3 import S3Target


class GzippedS3PathTask(ExternalTask):
    path = Parameter()

    def output(self):
        return S3Target(self.path, format=Gzip)


class LargeS3Task(Task):
    path = 's3://3kricegenome/9311/IRIS_313-15896.snp.vcf.gz'

    def requires(self):
        return GzippedS3PathTask(self.path)

    def run(self):
        infile = self.input().open('r')
        for i in infile:
            pass
```

All tests are passing except one which is probably related to available locales on my system.

### DIFF
--- a/luigi/format.py
+++ b/luigi/format.py
@@ -76,7 +76,11 @@ class InputPipeProcessWrapper(object):
                 self._original_input = False
                 f = tempfile.NamedTemporaryFile('wb', prefix='luigi-process_tmp', delete=False)
                 self._tmp_file = f.name
-                f.write(input_pipe.read())
+                while True:
+                    chunk = input_pipe.read(io.DEFAULT_BUFFER_SIZE)
+                    if not chunk:
+                        break
+                    f.write(chunk)
                 input_pipe.close()
                 f.close()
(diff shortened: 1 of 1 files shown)


erikbern:
lgtm!

Tarrasch:
Travis doesn't seem to build... closing and reopening

Tarrasch:
Yay it worked!

We IT engineers are so great! :D

Issue #1344, opened 2015-10-23
Update S3 client to use Boto3


jfeala:
AWS is moving from `boto` to `boto3` for its Python SDK. Boto3 includes high-level wrappers for simple uploads and downloads from local files with [S3Transfer](https://boto3.readthedocs.org/en/latest/reference/customizations/s3.html#module-boto3.s3.transfer). Besides the simpler interface, in my own experience I've seen these advantages with Boto3
- More stable downloads for big files
- Built-in support for multipart uploads
- Encryption with [KMS](https://aws.amazon.com/kms/)

Happy to contribute this when I have some time over the next few months, otherwise I thought I'd put it out there for others to pick up

themalkolm:
Just curious. Do you have any problems with current `boto` implementation?

jfeala:
Those three bullet points are the problems I have with `boto`. The deal-breaker for me is that you can't do encryption at rest via `boto` using the Key Management Service (KMS), but I also get some instability with big file downloads (occasional timeouts without restart), and it's inconvenient that multipart uploads are not included and have to be patched on (as it is implemented in the Luigi S3 client). 

Boto3 solved these problems for us and so now we currently only use the Luigi S3 client for its S3Target.exists() and do the file transfers manually. We'd like to get back to using the Luigi client for its convenience and atomic transfers, once it's updated.

vlcinsky:
There are more reasons to move to `boto3`

- boto is not going to be maintained - number of not resolved issues is growing
- AWS gives priority to boto3
- there are long term issues (e.g. problems to work with bucket names having dots in it)

See discussion at boto issue tracker  [`boto3` is not a replacement of `boto` (yet?) #3306](https://github.com/boto/boto/issues/3306), which, contrary to the issue title shows, that **boto3 is really the way to go**.

brianbruggeman:
`boto` is [no longer maintained](https://codegangsta.io/blog/2014/05/19/my-thoughts-on-martini/).  I'm not sure this should really be a discussion point: Luigi should move to [`boto3`](https://boto3.readthedocs.io/en/latest/).

burdiyan:
Can't use boto because of this issue https://github.com/boto/boto/issues/2741

Just can't talk to frankfurt buckets. So sad :( Will have to implement custom boto3 implementation, hope it won't be too tricky and I'll be able to submit the PR for this.

rhaarm:
[...]

PR #1469, opened 2015-12-24
Enable multipart S3 Copy

jpiper:
`S3Client.copy()` doesn't currently support multipart, so files larger than 5GB are unable to be copied from S3 to S3 without pulling down locally and reuploading. I'm opening this PR that fixes that so I can get an automated build :) Ideally there should be some automated tests here - it doesn't look like the `copy()` method is tested at all at the moment. If I get some time over the weekend I'll add some tests here.

On a separate note, we will probably get faster performance for this if we multithread it and issue a single multipart copy instruction per thread (this is what `awscli` does). I will also have a stab at implementing this at some point, but at least the basic functionality is here already.

### DIFF
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -338,8 +338,8 @@ class S3Client(FileSystem):
         # As the S3 copy command is completely server side, there is no issue with issuing a single
         # API call per part, however, this may in theory cause issues on systems with low ulimits for
         # number of threads when copying really large files, e.g. with a ~100GB file this will open ~1500
-        # threads.
-        pool = ThreadPool(processes=num_parts)
+        # threads. We take the max of this and 1 as if we're copying an empty file we will have  `num_part == 0`
+        pool = ThreadPool(processes=max(1, num_parts))
 
         mp = None
         try:
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -281,7 +281,7 @@ class S3Client(FileSystem):
 
         return contents
 
-    def copy(self, source_path, destination_path, **kwargs):
+    def copy(self, source_path, destination_path, part_size=67108864, **kwargs):
         """
         Copy an object from one S3 location to another.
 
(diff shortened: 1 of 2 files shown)


luigi/s3.py
```diff
@@ -281,7 +281,7 @@ def get_as_string(self, s3_path):
 
         return contents
 
-    def copy(self, source_path, destination_path, **kwargs):
+    def copy(self, source_path, destination_path, part_size=67108864, **kwargs):
```
erikbern: any explanation to this magic constant? :)

luigi/s3.py
```diff
@@ -291,16 +291,70 @@ def copy(self, source_path, destination_path, **kwargs):
...
+        """
+        (src_bucket, src_key) = self._path_to_bucket_and_key(source_path)
+        (dst_bucket, dst_key) = self._path_to_bucket_and_key(destination_path)
+
+        dest_bucket = self.s3.get_bucket(dst_bucket, validate=True)
+        source_bucket = self.s3.get_bucket(src_bucket, validate=True)
+
+        source_size = source_bucket.lookup(src_key).size
+
+        num_parts = \
```
erikbern: i think this is just ``(source_size + part_size - 1) // part_size``

luigi/s3.py
```diff
@@ -291,16 +291,70 @@ def copy(self, source_path, destination_path, **kwargs):
...
+                part_start = cur_pos
+                cur_pos += part_size
+                part_end = min(cur_pos - 1, source_size - 1)
+                part_num = i + 1
+
+                mp.copy_part_from_key(src_bucket, src_key, part_num, part_start, part_end)
+                logger.info('Copying part %s/%s to %s', part_num, num_parts, destination_path)
+            # finish the copy, making the file available in S3
+            mp.complete_upload()
+        except BaseException:
```
erikbern: could also use a `finally` handler here instead of re-raising the exception

erikbern:
looks great (see some stylistic non-blocking comments though)

luigi/s3.py
jpiper: Yeah, this can probably be removed from here and moved into a var within the function.

luigi/s3.py
jpiper: I just copied the way `num_parts` is calculated from the `put_multipart` method, but yes, this looks correct. I'll patch it into `num_parts` as well and hopefully the tests succeed :)

luigi/s3.py
jpiper: I'm not sure what you're getting at here - I was just following the style of the other S3 uploading methods in the same script

jpiper:
Ok, I've addressed the comments and there's another enhancement here - as S3 copies are just API calls, there's no reason for them  to happen synchronously. This uses a ThreadPool to concurrently request all parts data to be copied. From my testing, for copying a `12GiB` S3 object this results in a speedup from about 12 minutes to 1 minute!

It will probably be worth performing a similar procedure for the `s3_put_multipart` command, which I will do in a separate PR.

jpiper:
It appears the build is failing on external dependencies -

```Could not find a version that satisfies the requirement mysql-connector-python<3.0 (from versions: )
No matching distribution found for mysql-connector-python<3.0```

**Comment by jpiper** [comment]
I changed my mind, here's the patch for parallelised `put_multipart` as well.

**Comment by Tarrasch** [comment]
Hmm, can you get rid of the merge commit from this PR? It's good that you have kept the logical changes in separate commits. :)

Sorry about the failing build. I don't have time to fix it now though. And I guess the Spotify maintainers are off during Christmas ...

**Comment by jpiper** [comment]
I changed my mind again. Let's not conflate this PR with other things. I wasn't happy with the code for the multithreaded put, and I want to test it some more anyway.

**Comment by Tarrasch** [comment]
Yea sounds good. I prefer many small easy-to-review patches over few overhaul PRs :)

**Comment by jpiper** [comment]
Visualizer build failed... How odd

**Comment by Tarrasch** [comment]
Try retriggering. It could be spurious failure. I haven't had time to fix/disable the broken tests.

File: luigi/s3.py [hunk 4]
```diff
@@ -284,23 +283,82 @@ def get_as_string(self, s3_path):
...
-
         :param kwargs: Keyword arguments are passed to the boto function `copy_key`
         """
         (src_bucket, src_key) = self._path_to_bucket_and_key(source_path)
         (dst_bucket, dst_key) = self._path_to_bucket_and_key(destination_path)
 
         s3_bucket = self.s3.get_bucket(dst_bucket, validate=True)
+        source_bucket = self.s3.get_bucket(src_bucket, validate=True)
+
+        multipart_size = 67108864
```
Tarrasch: Wait. This magic constant is (1) Copied twice in the code and (2) completely undocumented. Both are issues needing to be fixed.

Maybe this commit https://github.com/spotify/luigi/commit/7c089a91ffe070478c5814762c1ce6cc77aa8c8a can be inspiring? :)

jpiper: Yes, this irks me too - this magic constant is present in the `put_multipart` method, and I just copied it from there. I have no idea why this was chosen as the `multipart_size` for that, either.

jpiper: I can add this as a comment in the commit as per your example, or try and dig back through the repo to see why this magic constant exists in the `put_multipart` method in the first place?

Tarrasch: I think it's okay to extract it out and add a TODO and honestly admitting that we don't know what it comes from. But it would be good to do a git-blame and @-ping here in this PR

erikbern:
Can you fix the magic constant and I'm happy to merge this

jpiper:
I'll try and get around to it this week :)

erikbern:
ping... great if we can merge this

jpiper:
I've added some more comments, and made it clear that one of these is the threshold before copying in multiple parts, and one is the size of the parts that it's splitting the copy, and I'm not sure why these sizes were choses.

The original committer (@ddaniels888) set the multipart threshold and multipart block size to 64MB [here](https://github.com/humanlongevity/luigi/commit/6579f755c8316c1a8a0a5124b31de4530c79f87c#diff-68ac377fbe87ada8c1af3a743e141c9dR162). The s3 cli takes a slightly different approach, and sets the default thresholds to 8MB (see [here](http://docs.aws.amazon.com/cli/latest/topic/s3-config.html#multipart-threshold), but I've been consistent with 64MB (why fix it if it ain't broke?).

jpiper:
Please don't merge yet, incoming: unit tests

erikbern:
excellent!

jpiper:
So I wrote the tests and then spent a few hours trying to work out why I was getting failures. Turn's out I discovered a [bug in moto](https://github.com/spulec/moto/issues/526) which means it doesn't handle multipart copy correctly. 

We've been using this S3 multipart copy fork in production here at HLI for a few weeks, and haven't had any issues, but you'll probably have to try it out for yourself if you want to see it in action considering moto isn't working for this.

I've used unittest decorators to skip them and explain why they're being skipped for the meantime.

jpiper:
I'm seeing no movement on my moto bug report - how do you want to proceed here?

erikbern:
does the moto bug currently break anything in luigi?

what do you recommend that we do?

jpiper:
At the moment, I've skipped the unit tests that will fail because of the moto bug [see here](https://github.com/humanlongevity/luigi/commit/e1aaa128d32179ce15090b67969acb090c8bc775#diff-0fb5575a1647ac0e607e96b40a667702R406) and [here](https://github.com/humanlongevity/luigi/commit/e1aaa128d32179ce15090b67969acb090c8bc775#diff-0fb5575a1647ac0e607e96b40a667702R414)

But otherwise everything is fine. I would say we merge anyway, considering they're not moving on that issue at all.

jfeala:
Hi all, just caught this thread and want to give you a heads up that Boto3 has built-in support for multipart copy. Although last I checked it didn't support bucket-to-bucket multipart transfer, which it looks like you have here. This looks like a perfectly fine implementation but boto should probably be upgraded eventually, regardless.

I mentioned this in https://github.com/spotify/luigi/issues/1344 but I haven't had a chance to work on it yet, sorry!

PR #1471, opened 2015-12-29
notifications: More emails with proper coloring

Tarrasch:
This commits contains the following changes:

 * Make luigi send emails for task timeouts
 * Make luigi send emails for when task process dies
 * When sending normal run() failures, do proper pygments coloring
 * Remove code duplication
 * Clean tests and tests what previously was not working

As an effect of properly rendering stack traces with pygments, the
"expl" field in the worker-internal result_queue will not contain
json encoded fields anymore. We instead encode the field just before
submitting it to the scheduler. I believe this should make sense.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -169,7 +169,7 @@ class TaskProcess(multiprocessing.Process):
             elif status == DONE:
                 self.task.trigger_event(
                     Event.PROCESSING_TIME, self.task, time.time() - t0)
-                expl = json.dumps(self.task.on_success())
+                expl = self.task.on_success()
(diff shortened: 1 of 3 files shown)


Tarrasch:
Hi @dhruvg, as I wrote in the last paragraph I reverted the json encoding that you introduced earlier in this patch. What do you think. Does it make sense to only json encode something just before you submit it to the scheduler and keep it in it's raw format internally?

Tarrasch:
Screenshot:

selection_002


The first email is not exactly how it originally looked like but it totally lacked color. Now it should have colors.

Tarrasch:
I guess reviewing manpower is low during the Christmas holidays. So I self-approve for now.

PR #1473, opened 2015-12-30
Change the way chrono parameters (date/datetime) are handled.

mikekap:
- Split date & datetime - the similarities only work due to duck typing; dates & datetimes are different objects. It becomes much easier to reason about the implementation when you know whether the value is a date or a datetime.
 - Add interval & start to both date & datetime params to handle the "every N minutes/hours/days/months/years" case.
   - These are clamped in normalize() so even if you pass e.g. 3PM to an every 2 hours task, the actual value of the parameter will become 2PM.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -32,7 +32,6 @@ from luigi import task_register
 from luigi import six
 
 from luigi import configuration
-from datetime import timedelta
 from luigi.cmdline_parser import CmdlineParser
 
 _no_value = object()
@@ -240,8 +239,7 @@ class Parameter(object):
         """
         return x  # default impl
 
-    @classmethod
-    def next_in_enumeration(_cls, _value):
+    def next_in_enumeration(self, _value):
         """
         If your Parameter type has an enumerable ordering of values. You can
         choose to override this method. This method is used by the
@@ -269,11 +267,19 @@ class Parameter(object):
         return "store"
 
 
+_UNIX_EPOCH = datetime.datetime.utcfromtimestamp(0)
+
+
 class _DateParameterBase(Parameter):
     """
-    Base class Parameter for dates. Code reuse is made possible since all date
-    parameters are serialized in the same way.
+    Base class Parameter for date (not datetime).
     """
+
+    def __init__(self, interval=1, start=None, **kwargs):
+        super(_DateParameterBase, self).__init__(**kwargs)
+        self.interval = interval
+        self.start = start if start is not None else _UNIX_EPOCH.date()
+
     @abc.abstractproperty
     def date_format(self):
         """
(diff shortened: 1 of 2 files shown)


luigi/parameter.py
```diff
@@ -311,13 +311,19 @@ class DateParameter(_DateParameterBase):
...
-        return datetime.datetime.strptime(s, self.date_format).date()
+    def next_in_enumeration(self, value):
+        return value + datetime.timedelta(days=self.interval)
+
+    def normalize(self, value):
+        if value is None:
+            return None
+
+        if isinstance(value, datetime.datetime):
+            value = value.date()
```
Tarrasch: ah yes. finally!

This means that we cna pass datetime's and not only date's as constructors to {Date,Month,Year}Parameter and it would automatically do the right thing. Am I correct?

Tarrasch:
Anyway this looks excellent to me.

luigi/parameter.py
mikekap: Yep, I think so. It sort of worked out via duck-typing before (i.e. datetime has .year/.month/etc), but now there's a far smaller chance of introducing bugs by accidentally using a date as a datetime (and not noticing cause you always passed in a datetime)

luigi/parameter.py
```diff
@@ -331,9 +337,35 @@ class MonthParameter(DateParameter):
...
-        return None
+    def _add_months(self, date, months):
+        """
+        Add ``months`` months to ``date``.
+
+        Unfortunately we can't use timedeltas to add months because timedelta counts in days
+        and there's no foolproof way to add N months in days without counting the number of
+        days per month.
+        """
+        year = date.year + (months // 12)
```
erikbern: this logic seems complicated. pretty sure you can do it using something like this

`year = date.year + (months-1)//12`
`month = date.month + (months-1)%12 + 1`

erikbern:
haha i don't even recognize the original code so lgtm

what happened to next/prev? why is it called next_in_enumeration now?

luigi/parameter.py
mikekap: Hmm I don't think that works with something like 12/2015 + 2 == 2/2016 ?

Separately, the month < 1 logic is for correctness in the value < self.start case (which seems like a bad idea, but I'd rather it not fail :D).

erikbern: i think it would work (haven't tested it though)
```
>>> (12 + 2 - 1)//12
1
>>> (12 + 2 - 1)%12 + 1
2
```

mikekap: Oh I see what you mean - cool. Done.

Tarrasch:
Next/prev? As explained in the docs, we use `next_in_enumeration` for generating the awesome execution summary.

Issue #1476, opened 2015-12-31
basic task example from docs fails


ryantuck:
I believe the basic `Task.run` code snippet from the docs doesn't work properly. I put the code snippet in a file named `test.py`:

```
class TaskA(luigi.Task):
    def output(self):
        return luigi.LocalTarget('xyz')

class FlipLinesBackwards(luigi.Task):
    def requires(self):
        return TaskA()

    def output(self):
        return luigi.LocalTarget('abc')

    def run(self):
        f = self.input().open('r') # this will return a file stream that reads from "xyz"
        g = self.output().open('w')
        for line in f:
            g.write('%s\n', ''.join(reversed(line.strip().split())))
        g.close() # needed because files are atomic
```

On running `luigi --module test FlipLinesBackwards`, I get the following error:

```
Runtime error:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/luigi/worker.py", line 151, in run
    raise RuntimeError('Unfulfilled %s at run time: %s' % (deps, ', '.join(missing)))
RuntimeError: Unfulfilled dependency at run time: TaskA()
```

The webserver indicates TaskA ran successfully. Is this code snippet just incomplete? Am I missing critical command-line arguments? Is there an incredibly bare-bones example anywhere of simply having one task require another?

Tarrasch:
Yes. The example is wrong. Where did you grab it from? :)

It should be

```
class TaskA(luigi.ExternalTask):
    def output(self):
        return luigi.LocalTarget('xyz')
```

Alternatively, `TaskA` should have a `run()` method that creates the `"xyz"` file. But it's bad that luigi tries to run the task even though a run method was never provided.

Tarrasch:
If you could have time to fix the example (send a PR). It would be appreciated :)

Tarrasch:
Oh, and the reason that it "once worked for someone" is probably because that someone already had the file `"xyz"` manually first.

ryantuck:
cool! pr submitted.

Tarrasch:
Thanks for your doc improvement. Next person will hopefully get a better experience :)

PR #1477, opened 2015-12-31
fix code snippet example for Task.run doc section

ryantuck:
related to the following issue: https://github.com/spotify/luigi/issues/1476

original code snippet didn't work out-of-the-box. my example works and i think helps to highlight how input/output works between tasks using `LocalTarget`.


Tarrasch:
Yes. This looks so much better than the previous example. Thank you! :)

PR #1479, opened 2015-12-31
add EnumParameter

jimjh:
This adds parameter for enums, which has been a built-in module since python 3.4.

Example:

```python
class Models(enum.IntEnum):
  Honda = 1

class MyTask(luigi.Task):
  my_param = luigi.EnumParameter(enum=Models)
```

```bash
$ MyTask --my-param Honda
```

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -667,9 +667,22 @@ class EnumParameter(Parameter):
     """
     A parameter whose value is an :class:`~enum.Enum`.
 
-    .. code:: python
+    In the task definition, use
 
-        my_param = EnumParameter(enum=MyEnum)
+    .. code-block:: python
+
+        class Models(enum.IntEnum):
+          Honda = 1
+
+        class MyTask(luigi.Task):
+          my_param = luigi.EnumParameter(enum=Models)
+
+    At the command line, use,
+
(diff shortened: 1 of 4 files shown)


Tarrasch:
I wanted EXACTLY this last week! +100

luigi/parameter.py
```diff
@@ -661,3 +661,28 @@ def serialize(self, cls):
...
         return cls.task_family
+
+
+class EnumParameter(Parameter):
+    """
+    A parameter whose value is an :class:`~enum.Enum`.
+
+    .. code:: python
+
+        my_param = EnumParameter(enum=MyEnum)
```
Tarrasch: Can you just copy in the documentation you added in the PR description (how it integrates with CLI) into here as well? TaskParameter already has similar documentation you can copy paste from if you're unsure about rst syntax.

jimjh:
Done! Updated docstring for EnumParameter.

erikbern:
thanks!

Tarrasch:
@jimjh, Do you also experience problems with the execution summary crashing because of Enums being uncomparable? It's a bug I knew from the start with the implementation, I can take a stab at fixing it...

PR #1480, opened 2016-01-04
Add move operations into the fs target interface

Tarrasch:
Also, make so that `rename_dont_move` reports success/failure by raising
exceptions instead of returning a bool value. As it's kind of
unexpected and unpythonic to return True/False based on success or not.

### DIFF
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -304,7 +304,13 @@ class GCSClient(luigi.target.FileSystem):
                 body={}).execute()
             _wait_for_consistency(lambda: self._obj_exists(dest_bucket, dest_obj))
 
-    def rename(self, source_path, destination_path):
+    def rename(self, *args, **kwargs):
+        """
+        Alias for ``move()``
+        """
+        self.move(*args, **kwargs)
+
+    def move(self, source_path, destination_path):
         """
         Rename/move an object from one GCS location to another.
         """
--- a/luigi/contrib/hdfs/abstract_client.py
+++ b/luigi/contrib/hdfs/abstract_client.py
@@ -22,7 +22,6 @@ Module containing abstract class about hdfs clients.
 import abc
 from luigi import six
 import luigi.target
-import warnings
 
 
 @six.add_metaclass(abc.ABCMeta)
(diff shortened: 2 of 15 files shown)


Tarrasch:
The reason I'm doing this is that I want to later add a `FileSystemTarget.temporary_path()` context manger. Working like this:

```
with self.output().temporary_path() as self.temp_output_path:
    self.do_stuff()
```

And `rename_dont_move` is kind of required for implementing that. I currently have found myself manually implementing the rename-only-after-success pattern so many times now that I want a built in solution in luigi.

I think this shouldn't be considered as affecting core luigi, because the fs stuff should really go into contrib.

erikbern:
Luigi's file operations should be atomic by default, so I don't see why this is necessary. Can you explain what's a situation where you need this?

Tarrasch:
Say you're using spark/crunch/hive/etc.. You want to write to file `my/file`. Typical strategy is to write to `my/file-tmp-luigi-123456`. But then comes the question: How do you move/rename the file to `my/file`?. There are two wrong ways and one right away afaik.

 * plain `mv` is wrong, you can end up with nested folders
 * mkdir+raise_if_exists followed by `mv my/file-tmg-luigi-123456/* my/file/` is wrong because files are moved one by one and not atomically
 * You need a rename-dont-move file operation primitive. You can not build it using existing primitives. But there is a default implementation that's pretty close though as I've described in this PR.

erikbern:
Ok got it I think. hadoop.py has its own mechanism for doing so, so we
should prob factor it out to this mechanism instead. It's only atomic if
you open a stream to it, not if you use its path and have some other
mechanism of writing to it

erikbern:
you could potentially even make this a method on the target class and even `LocalTarget` could benefit from it

https://github.com/spotify/luigi/blob/master/luigi/file.py#L97

Tarrasch:
Well. This PR is just preparing for a follow up PR where I'll do the interesting stuff including refactoring away much of the code duplication.

Anyway this PR is not entirely done I realize. There are failing tests + that I want to add a few more tests as I realized this broke webhdfs for instance ...

Tarrasch:
Ok. The new commit message contains more information. Lets' pray that the tests pass :)

Tarrasch:
Doh, tests are still failing. This re-factoring had bigger than I expected. But I think it's healthy in that it increases code coverage for all the clients type (and none seem to have implemented `move()`).

Tarrasch:
Yay. Finally passing.

Does this patch look good to you @erikbern? Can this be merged?

Tarrasch:
@ulzha @erikbern @econchick @DeaconDesperado christmas is over start reviewing! :christmas_tree:  :santa: :snowman: :snowflake: :D

PR #1482, opened 2016-01-05
Make webhdfs tests for for cdh and recent hdp

Tarrasch:
See commit message.


Tarrasch:
I changed this commit a bit. Solving one problem at a time ...

Tarrasch:
Merging, it only affects tests so...

PR #1484, opened 2016-01-05
Improved: S3 Copy (files, folders, buckets), optionally based on modified date

dlstadther:
Based on how copy() and listdir() were written, if a bucket was passed the `_path_to_bucket_and_key(path)` method would determine that `src_bucket` = the bucket name, but the `src_key` = an empty string. Thus, when that `src_key` is passed to `_add_path_delimiter(key)`, a `/` is returned. This makes the `src_prefix + key` = `bucket//`. (i.e. Input: `s3://this.is.a.bucket/` -> src_prefix+key: `this.is.a.bucket//`). I have added a check in the `_add_path_delimiter(key)` to return the key as is if it is an empty string.

Also, when listing a folder within a bucket, the first key was always an empty string. This empty string causes `copy_key(...)` to fail. So I have included a check to be sure that if the key is an empty string or forward slash, it will not be copied.

Optional parameters have been added to `copy()`, `listdir()`, and `list()` to allow for start and/or end times to be defined (none, one, or both may be defined).

Multithreaded copy managed to list 1.12 million files, compare last_modified dates, and copy 32,159 files in 10 minutes 40 seconds. A maximum of 100 threads were specified, but the number of active threads peaked at 91. This test was also completed on my local machine. Execution on an EC2 instance would perform even better.

### DIFF
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -27,6 +27,7 @@ import itertools
 import logging
 import os
 import os.path
+
 import sys
 import threading
 import time
@@ -38,6 +39,7 @@ try:
 except ImportError:
     from urllib.parse import urlsplit
 import warnings
+
 try:
     from ConfigParser import NoSectionError
 except ImportError:
@@ -49,7 +51,8 @@ from luigi.six.moves import range
 from luigi import configuration
 from luigi.format import get_default_format
 from luigi.parameter import Parameter
-from luigi.target import FileAlreadyExists, FileSystem, FileSystemException, FileSystemTarget, AtomicLocalFile, MissingParentDirectory
+from luigi.target import FileAlreadyExists, FileSystem, FileSystemException, FileSystemTarget, AtomicLocalFile, \
+    MissingParentDirectory
 from luigi.task import ExternalTask
 
 logger = logging.getLogger('luigi-interface')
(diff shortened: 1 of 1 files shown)


dlstadther:
I see that tests fail. But I am a bit confused as to why the tests fail, but the code works for copying buckets...

dlstadther:
From what I can tell, my code works for entire buckets whereas the previous code works for folders or specific files within a bucket.

dlstadther:
The failing check is due to a conditional statement formatted for clarity. If someone has a better suggestion for formatting that will pass the check, I will gladly edit that conditional statement.

dlstadther:
I have found that this serial copy is terribly slow for large buckets. I will work to make the copy process multi-threaded.

dlstadther:
Note: PR #1469 will cause conflicts with this PR. One will need to be merged first and the other will need to adapt accordingly.

dlstadther:
It looks like it'll be easier for me to resolve his conflicts into my code than from him to resolve mine.

Feel free to merge copy_multipart first and then I will update multithreading accordingly.

dlstadther:
Since the copy multipart PR has now been merged, I will work to merge its changes into this PR (as time allows).

dlstadther:
This should be good to go once someone verifies.

PR #1486, opened 2016-01-05
Move s3 boto import to avoid un-needed warnings

### LINKED ISSUE #1485: "missing boto" warning for hadoop import
Currently when the `luigi.contrib.hadoop` module is imported we get a WARNING about s3 being imported without boto module available  (also manifests as "No handler for logger" errors if hadoop imported outside the task-running interface context).

I assume plenty of folks are using hadoop without s3, so we should get rid of the warning somehow.

davemt:
Rather than warning, we can just fail if we try to actually use an S3
target without boto.  This allows for use cases where s3 module is
imported but an S3 target is not actually used.

Fixes #1485

### DIFF
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -16,7 +16,8 @@
 #
 """
 Implementation of Simple Storage Service support.
-:py:class:`S3Target` is a subclass of the Target class to support S3 file system operations
+:py:class:`S3Target` is a subclass of the Target class to support S3 file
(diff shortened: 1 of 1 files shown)


davemt:
Failure is on `visualiser.visualiser_test.TestVisualiser` from `TOXENV=visualizer`; assuming it is an unrelated failure.

Tarrasch:
Thanks!! :)

PR #1489, opened 2016-01-07
Added a config option to disable TLS in SMTP entirely.

palicand:
I've implemented a config option which allows to disable TLS encryption in SMTP, if the user desires so. If it's not specified, it defaults to True, so it doesn't change the behaviour for those, who don't have it set.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -215,6 +215,9 @@ smtp_port
 smtp_ssl
   If true, connects to smtp through SSL. Defaults to false.
 
+smtp_tls
+  If true, connects to smtp through TLS. Defaults to true.
+
 smtp_timeout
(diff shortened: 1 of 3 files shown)


erikbern:
LGTM

Tarrasch:
There is a small problem. Eventually these options will be converted to the slightly more typed `luigi.Paramaters` and for `BoolParameter` we always expect `False` to be the default. Can you change it so it's perhaps called `smtp_without_tls` or `smpt_no_tls` and set the default value to `False`?

If possible, perhaps you could have a fresh test so that you both test the `True` and the `False code paths?

palicand:
Sure, I'll change it.

Tarrasch:
Thanks @palicand :)

erikbern:
great

PR #1490, opened 2016-01-07
allow for use of BigQuery's flatten results flag

blacker:
In keeping with BigQuery's default behavior, this defaults to true.

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -466,6 +466,12 @@ class BigqueryRunQueryTask(MixinBigqueryBulkComplete, luigi.Task):
         return CreateDisposition.CREATE_IF_NEEDED
 
     @property
+    def flatten_results(self):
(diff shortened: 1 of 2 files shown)


erikbern:
LGTM

mbruggmann:
Would be nice to verify this in `bigquery_test.py`, but in terms of functionality :+1:

erikbern:
is it easy to add a test?

blacker:
It's easy to verify that the parameter defaults to what we expect. Testing that the parameter actually produces the expected behavior in BigQuery would be more involved, requiring authorization with a BQ service account, and is probably overkill here. Does the former seem like enough?

erikbern:
yeah it's great. thanks!

mikekap:
Would anyone mind if I change the default back to False for backwards compat?

mbruggmann:
@blacker Cool, thanks for adding the tests!

Issue #1496, opened 2016-01-09
Disable Task History when using --local-scheduler


riga:
As you suggested in the docs, we use --local-scheduler only for testing purposes. Each member of our group works on his/her own computer (we have remote targets, so this setup works just great). Our central scheduler records the task history, as configured by our common luigi config file.

Now, the problem is that the local scheduler also tries to open the task history db file, that is defined for the central scheduler. This file does not necessarily exist. As you may guess, we don't want to maintain two separate config files ;)

It would be extremely helpful is there was an option to disable the task history for the local scheduler.
Any chance that a PR might be accepted ?

Tarrasch:
> Now, the problem is that the local scheduler also tries to open the task history db file, that is defined for the central scheduler.

This sounds like a bug. Of course we welcome a PR. :)

I'm not sure how popular the Task History is. I know Spotify kinda wrote their own internal system and I haven't ever tried the task history db myself. Do you and your colleagues like it?

riga:
Well, we enabled it but we never *had to* use it. We like to think of it as a lightweight provenance backup, mainly for good conscience ;)

erikbern:
imo would be awesome if someone could actually give the task history a lot more love. it could be incredibly useful

riga:
Definitely.

We use luigi for our analyses of CERN data. Currently, we are in the development phase, but I think when we start to transition into "production", task history will become quite important.

erikbern:
That's great! Would love it if it can become a more integrated part of Luigi

Also cool that you are using Luigi at CERN :sunglasses: :sparkles:

Tarrasch:
This is fixed so I close.

Haha yea it's totally awesome you guys use it at CERN. I can tell people I'm part of finding the higgs boson ^^

PR #1497, opened 2016-01-09
Disable task history for local scheduler.

riga:
Disables the task history in ``_WorkerSchedulerFactory.create_local_scheduler``.
Fixes issue #1496.


Tarrasch:
Well that was easy ^^. But maybe it's worthwhile to add a test case so it doesn’t break suddenly. :)

erikbern:
thanks!

Tarrasch:
Great!

PR #1500, opened 2016-01-12
Assistant workers never retry failed tasks

casey-green:
Assistant workers never retry failed tasks that are eligible for retry.

I noticed this with `ExternalTask`s, but same goes for regular task failures, given the correct retry configurations.

@erikbern I mentioned this to you on the mailing list.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -567,7 +567,8 @@ class CentralPlannerScheduler(Scheduler):
 
         for task in self._state.get_active_tasks():
             self._state.fail_dead_worker_task(task, self._config, assistant_ids)
-            if task.id not in necessary_tasks and self._state.prune(task, self._config):
+            removed = self._state.prune(task, self._config)
+            if removed and task.id not in necessary_tasks:
                 remove_tasks.append(task.id)
 
         self._state.inactivate_tasks(remove_tasks)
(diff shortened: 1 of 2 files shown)


casey-green:
Closing this for now.

luigi/scheduler.py
```diff
@@ -567,7 +567,8 @@ def prune(self):
 
         for task in self._state.get_active_tasks():
             self._state.fail_dead_worker_task(task, self._config, assistant_ids)
-            if task.id not in necessary_tasks and self._state.prune(task, self._config):
+            removed = self._state.prune(task, self._config)
+            if removed and task.id not in necessary_tasks:
```
Tarrasch: Ah, so the change is that you want the `prune()` side effect before doing the `task.id not in necessary_tasks` membership check?

Stupid non-Haskell programming languages :p

daveFNbuck: just switching the order of the and should work, but maybe this is more clear

casey-green: Exactly @Tarrasch. I figured it would be safer (and more clear) to break out the method call in case someone comes along and adds another condition that would prevent the method from being called, i.e.

```python
if some_other_false_condition and self._state.prune(task, self._config) and task.id not in necessary_tasks:
   # ...
```

Tarrasch: I think it's very good that you broke it up. Nice thinking there.

Tarrasch:
Though before I can merge this you have to merge your commits into one, because they belong to the same logical change.

casey-green:
Ah, sorry about that - squashed commits into one.

Tarrasch:
Thanks for this contribution!

PR #1501, opened 2016-01-12
Stop worker's run loop if the keep-alive thread is stopped

casey-green:
This is an issue for workers where assistant and keep_alive are True, and a worker's run method is executed in a different thread (which seems to be the best way to programmatically use assistant workers).  For example, this would have to be forcibly killed:

```python
scheduler = CentralPlannerScheduler()
w = worker.Worker(scheduler, worker_processes=2, assistant=True, keep_alive=True)
t = Thread(target=lambda: w.run())
with w:
    t.start()
    # submit some tasks, wait for them to finish, do some other stuff, yadda, yadda
t.join() # without this change this will not exit cleanly and will have to be killed.
```

Also, if the worker's keep-alive thread is stopped, and it is no longer reporting to the scheduler, I think it should stop asking for work.  But, let me know if you think otherwise.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -813,6 +813,7 @@ class Worker(object):
         Returns true if a worker should stay alive given.
 
         If worker-keep-alive is not set, this will always return false.
+        If the keep-alive thread has been stopped, this returns false.
         For an assistant, it will always return the value of worker-keep-alive.
         Otherwise, it will return true for nonzero n_pending_tasks.
 
@@ -821,10 +822,28 @@ class Worker(object):
         """
         if not self._config.keep_alive:
             return False
+        if not self._is_alive():
+            return False
         elif self._assistant:
             return True
         else:
             return n_pending_tasks and (n_unique_pending or not self._config.count_uniques)
+
+    def _is_alive(self):
+        """
+        Returns True if this worker is alive.
+
+        Returns False if the keep-alive thread has been stopped.
+        Returns False if the worker has not been initialized using a with statement.
+        Otherwise, this will return true.
+        """
+        try:
+            return self._keep_alive_thread.is_alive()
+        except AttributeError:
+            # self._keep_alive_thread is initialized in the __enter__ method, so this exception will occur if
+            # worker.run is called outside of a with block.
+            logger.error('Worker not fully initialized. You probably called worker.run() outside a "with" block')
+            return False
 
     def handle_interrupt(self, signum, _):
         """
--- a/test/worker_test.py
+++ b/test/worker_test.py
@@ -1039,6 +1039,17 @@ class AssistantTest(unittest.TestCase):
         self.assertTrue(self.assistant.run())
         self.assertEqual(list(self.sch.task_list('DONE', '').keys()), [task.task_id])
 
+    @mock.patch('luigi.worker.KeepAliveThread')
+    def test_dead_assistant(self, keep_alive_fn):
+        self.assistant = Worker(scheduler=self.sch, worker_id='Y', assistant=True, keep_alive=True)
+        task = Dummy2Task('A')
+        keep_alive_thread = keep_alive_fn.return_value
+        keep_alive_thread.is_alive.side_effect = lambda: not task.complete()
+        with self.assistant:
+            self.assistant.add(task)
+            self.assistant.run()
+        self.assertEqual(list(self.sch.task_list('DONE', '').keys()), [task.task_id])
+
 
 class ForkBombTask(luigi.Task):
     depth = luigi.IntParameter()


luigi/worker.py
```diff
@@ -821,11 +822,27 @@ def _keep_alive(self, n_pending_tasks, n_unique_pending):
...
+        Returns True if this worker is alive.
+
+        Returns False if the keep-alive thread has been stopped.
+        Returns False if the worker has not been initialized using a with statement.
+        Otherwise, this will return true.
+        """
+        try:
+            return self._keep_alive_thread.isAlive()
+        except AttributeError:
+            logger.error('Worker not fully initialized. You are probably called run outside a "with" block')
```
Tarrasch: Maybe change `run` to `worker.run()` for clarity?

luigi/worker.py
```diff
@@ -821,11 +822,27 @@ def _keep_alive(self, n_pending_tasks, n_unique_pending):
...
+    def _is_alive(self):
+        """
+        Returns True if this worker is alive.
+
+        Returns False if the keep-alive thread has been stopped.
+        Returns False if the worker has not been initialized using a with statement.
+        Otherwise, this will return true.
+        """
+        try:
+            return self._keep_alive_thread.isAlive()
```
Tarrasch: Shouldn't you prefer the [`is_alive()` spelling](https://docs.python.org/2/library/threading.html#threading.Thread.is_alive)?

luigi/worker.py
Tarrasch: Wait. When and why does an AttributeError occur? Can you `# inline comments` inside the except block?

Tarrasch:
In what cases can the keep alive thread be stopped? It's private to the Worker class so users shouldn't normally shut it down themselves.

luigi/worker.py
erikbern: i think it's part of the Thread class. it's one of the old python modules before casing was standardized

Tarrasch:
Also. Please (a) squash these commits as they belong to the same logical change and (b) associate your email with your GitHub account please.

Thank you for this PR. Once I understand it and tests passes this probably can be merged. =)

luigi/worker.py
casey-green: I'll update the log msg and add a comment.

An `AttributeError` can occur if this is called outside a `with` block because the `_keep_alive_thread` is initialized in the `__enter__` method.

casey-green:
The keep-alive thread is stopped in the `__exit__` method. I'm trying to use workers in a more programatic way (perhaps this is ill-advised), sort of like this:

```python
scheduler = CentralPlannerScheduler()
w = worker.Worker(scheduler, worker_processes=2, assistant=True, keep_alive=True)
t = Thread(target=lambda: w.run()) # start run loop in a separate thread because it blocks
with w:
    t.start()
    # do application stuff, add tasks asynchronously, etc.
# at this point, the keep-alive thread has been shut down, but
# the worker.run loop will continue and t.join() will hang until you kill the process
t.join() 
```

The more I think about this, the less convinced I am that this is the best approach :)

casey-green:
Squashed the commits.  The CI failure looks unrelated to this PR, or am I wrong about that?

Tarrasch:
Wait a minute.

```
# time is t=0.000
with w:
    #  1. start keep alive thread, takes 1ms
    t.start() # 2. start a thread, takes 1ms
    # about to exit context, takes 1ms to run __exit__()
# Time is now t=0.003
t.join()
```

Are my comments correct? Also, we [say in worker.py docs](https://github.com/spotify/luigi/blob/6a8ceb02e10d8c562d5ff2d4a15517df7fe51b06/luigi/worker.py#L26-L28) that these things of the worker are private to luigi. You may only use `luigi.interface` module. 

Thanks for the amends and fixes. But lets make sure we understand what this change does (or rather the behaviour it encourages) before we merge.

casey-green:
Your comments are correct. However (without this change), `t.join()` won't return because `worker.run()` will continue to run inside thread `t` even though the worker has exited the context, it's keep-alive thread has been stopped, and it's tasks have been terminated (another side effect of `__exit__`).  Note this is only true for workers where both `assistant` and `keep_alive` are true.

I know I'm not _supposed_ to use workers directly :) ...but run/build in `luigi.interface` has the limitation that it can only be run inside the main execution thread [because of this line](https://github.com/spotify/luigi/blob/master/luigi/worker.py#L374). Which, for my use case, doesn't really fly.

I understand if this change requires some more thought - let me know and I can close it out :)

Tarrasch:
I'm against this change. Because, as we both seem to agree on, it only encourages misuse and adds no extra functionality to the allowed `luigi.interface` API that is approved to use.

However. I understand your use case better now, and I see the technical reasons for your issues. Thanks for pointing out the exact line. And here is the [relevant docs](https://docs.python.org/3/library/signal.html#signal.signal). I'll be more than happy to accept a patch that makes the Worker only conditionally install the `SIGUSR1` handler. What do you think about that?

casey-green:
Yea, I agree with you, I'll close this out.  Thanks for reviewing it!

I would definitely like the ability to conditionally install the SIGUSR1.  I'll start working on that and throw up a PR, probably over the weekend.

Thanks @Tarrasch!

Tarrasch:
Thank you for understanding! Looking forward to your next PR. :)

PR #1502, opened 2016-01-13
Make to call post_copy in postgres.CopyToTable.copy

woojing:
It seems just missing. so I filling.


Tarrasch:
Somebody knowing postgres should review this :)

dlstadther:
LGTM.

I am the one who included the post_copy for rdbms and redshift. I also set up redshift to accept an optional property `queries` such that the [post_copy method](https://github.com/spotify/luigi/blob/master/luigi/contrib/redshift.py#L314-L320) was already implemented. However, it is up to you whether you want to do the same/similar within `postgres.py` or if you want to leave it up to the user to implement their own `post_copy()`.

PR #1503, opened 2016-01-13
tests: Disable visualiser tests

Tarrasch:
Caused so much maintainer pain due to intermittent failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,8 @@ env:
     - TOXENV=pypy-scheduler
     - TOXENV=py27-gcloud
     - TOXENV=py27-postgres
-    - TOXENV=visualiser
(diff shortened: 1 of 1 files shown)


Tarrasch:
I'm merging until somebody has time to fix this ... :-/

erikbern:
sounds like this is the best option right now unfortunately :)

would love to make the visualizer tests working though

Tarrasch:
I think this is an example where it failed sporadically: https://travis-ci.org/spotify/luigi/jobs/102026785

I found that example only by going back a few PRs. Though it didn't seem that frequent, but often enough to annoy.

ping @QuentinBrosse in #2193

QuentinBrosse:
I have tried to reproduce the bug many times without success.. :/

PR #1505, opened 2016-01-13
remove unnecessary double escaping in raw string

woojing:
Escaping does not working for `r` prefixed raw string.
But those codes are use both of them. escaping and raw string.
As a result, null field that use `\N` is make error when it loaded.

### DIFF
--- a/luigi/postgres.py
+++ b/luigi/postgres.py
@@ -258,7 +258,7 @@ class CopyToTable(rdbms.CopyToTable):
         Default behaviour is to escape special characters and identify any self.null_values.
         """
         if value in self.null_values:
-            return r'\\N'
+            return r'\N'
         else:
             return default_escape(six.text_type(value))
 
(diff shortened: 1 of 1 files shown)


Tarrasch:
It seems like the tests are actually failing. Could you look at the tests and do `git blame` on them? The current behaviour maybe is intentional. You could also find some reviewers by doing git blame :)

woojing:
Understand this code writer's intension after test code review.
This code writer expect `copy` method always called by `run` method and assume `run` method is not overrided usually.
Anyway this is by design. Sorry for waste of time :cry:

Tarrasch:
Don't worry @woojing. I'm grateful for this PR even if it's not merged. :)

It would be cool if you could add some `# comments` or `''' documentation '''` into the code, so next user doesn't get confused =)

woojing:
Sounds great. I am going to do that soon :smile:

PR #1506, opened 2016-01-13
Unpack parameter tuples on MixinNaiveBulkComplete

thejens:
Change-Id: I8ce61b8736de87c77a5d3250f72a803bfd5aaf49

PR #1508, opened 2016-01-13
Unpack parameter tuples on MixinNaiveBulkComplete

thejens:
re-make of https://github.com/spotify/luigi/pull/1506 due to some git issues

"parameter_tuples" was not handling an iterable of tuples.
Updated to work on iterable of tuples or lists as well as iterable of dict
Intention is to keep original behavior if "parameter_tuples" contains single parameters instead of tuples
"

### DIFF
--- a/test/task_bulk_complete_test.py
+++ b/test/task_bulk_complete_test.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2012-2016 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from helpers import unittest
+from luigi import Task
+from luigi import Parameter
+from luigi.task import MixinNaiveBulkComplete
+
+COMPLETE_TASKS = ["A", "B", "C"]
+
+
+class MockTask(MixinNaiveBulkComplete, Task):
+    param_a = Parameter()
+    param_b = Parameter(default="Not Mandatory")
+
+    def complete(self):
+        return self.param_a in COMPLETE_TASKS
+
+
+class MixinNaiveBulkCompleteTest(unittest.TestCase):
+    """
+        Test that the MixinNaiveBulkComplete can handle
+        input as
+         - iterable of parameters (for single param tasks)
+         - iterable of parameter tuples (for multi param tasks)
+         - iterable of parameter dicts (for multi param tasks)
+    """
+    def test_single_arg_list(self):
+        single_arg_list = ["A", "B", "x"]
+        expected_single_arg_list = set(
+            [p for p in single_arg_list if p in COMPLETE_TASKS]
+        )
+        self.assertEqual(
+            expected_single_arg_list,
+            set(MockTask.bulk_complete(single_arg_list))
+        )
+
+    def test_multiple_arg_tuple(self):
+        multiple_arg_tuple = (("A", "1"), ("B", "2"), ("X", "3"), ("C", "2"))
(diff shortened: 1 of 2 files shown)


erikbern:
lgtm – is it good to merge?

test/task_bulk_complete_test.py
```diff
@@ -0,0 +1,73 @@
...
+from luigi import Task
+from luigi import Parameter
+from luigi.task import MixinNaiveBulkComplete
+
+COMPLETE_TASKS = ["A", "B", "C"]
+
+
+class MockTask(MixinNaiveBulkComplete, Task):
+    param_a = Parameter()
+    param_b = Parameter("Not Mandatory")
```
Tarrasch: Parameters should not accept positional arguments as far as I've ever seen.

test/task_bulk_complete_test.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2012-2015 Spotify AB
```
Tarrasch: Feel free to change to your company, and set 2016 (I think that is how licenses should work, but I'm not sure :))

Tarrasch:
Look OK. Some remarks in addition to the minor comments:

 1. Why do you need this? Range tools won't understand this anyway right?
 2. (nitpick) can you remove the "Change-id" clutter from your commit message?
 3. Please also associate your email with your github account.

Thanks :)

test/task_bulk_complete_test.py
thejens: :) is already my company

Tarrasch: :+1:

test/task_bulk_complete_test.py
thejens: Allright, I could change to only initializing single-parameter tasks and kwargs unpacking dicts. Thought it was supported as the parameter on bulk_complete seem to be named "parameter_tuples". 

Is the reason for this that the order of parameters based on initiation order is not guaranteed to be consistent?

thejens:
##lgtm – is it good to merge?
Depends if you think Arash's comments is a blocker, and if so what actions you think I should take.

##Why do you need this? Range tools won't understand this anyway right?
- No RangeTools seem to be built with single (datetime) param tasks in mind. I wanted to schedule multi-param tasks (not using RangeTools btw, just re-using the bulk_complete implementations) and assumed the "parameter_tuples" param on bulk_complete could handle tuples of parameters, but turns out it should have been called parameter_tuple with the current implementation. Technically though I think this could be used to improve RangeTools to handle arbitrary parameters passed to scheduled tasks

##(nitpick) can you remove the "Change-id" clutter from your commit message?
Guess I did, but am not that advanced on Git, so if changing the top message in this thread wasn't it then please point me in the right direction

##Please also associate your email with your github account.
Not sure how to do that, will see if I can find some guide

Tarrasch:
>  Not sure how to do that, will see if I can find some guide 

Looks like you figured it out. Thanks for doing it, helps us keep track of the actual number of contributors :)

>  Guess I did, but am not that advanced on Git, so if changing the top message in this thread wasn't it then please point me in the right direction

Not really a biggie, but since you seem curious about it :)

```
rm .git/hooks/post-commit-something-something
git commit --amend # then remove it and save
```

then force push :)

> Why do you need this? Range tools won't understand this anyway right?

Thanks. That explanation makes total sense. I just needed to have it confirmed. =)


I (as chief maintainer) would appreciate it if you could fix the nitpicks. But if you absolutely don't have time I will rather merge than abandon. The tests look really good! :)

test/task_bulk_complete_test.py
Tarrasch: Nono, I just meant  change the line to `Parameter(default="Not Mandatory"). Because it's more readable over plain `Parameter("Not Mandatory")`. :)

> Is the reason for this that the order of parameters based on initiation order is not guaranteed to be consistent?

It's guaranteed to be consistent. You have understood it correctly I think, sorry if my previous comment was confusing!

thejens: Fixed

test/task_bulk_complete_test.py
thejens: changed to 2016

thejens:
@Tarrasch believe I cleaned things up a bit, so should be ready to merge.

Tarrasch:
Thank you so much. It looks very clean and good now! :)

PR #1509, opened 2016-01-14
Provide the ability to conditionally install the SIGUSR1 hook on workers

casey-green:
The reason for this is that we can only install signal hooks inside the main execution thread. So this change provides the ability to skip this if we are programmatically launching jobs/creating workers outside the main execution thread.

@Tarrasch - as discussed in my last PR :)

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -267,6 +267,19 @@ worker-wait-jitter
   Default: 5.0
 
 
+[worker]
+
+no_install_shutdown_handler
+  By default, workers will stop requesting new work and finish running
+  pending tasks after receiving a `SIGUSR1` signal. This provides a hook
+  for gracefully shutting down workers that are in the process of running
+  (potentially expensive) tasks. If set to true, Luigi will NOT install
+  this shutdown hook on workers. Note this hook does not work on Windows
+  operating systems, or when jobs are launched outside the main execution
+  thread.
+  Defaults to false.
+
+
 [elasticsearch]
 ---------------
 
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -309,6 +309,9 @@ class worker(Config):
                                          config_path=dict(section='core', name='retry-external-tasks'),
                                          description='If true, incomplete external tasks will be '
                                          'retested for completion while Luigi is running.')
+    no_install_shutdown_handler = BoolParameter(default=False,
+                                                description='If true, the SIGUSR1 shutdown handler will'
+                                                'NOT be install on the worker')
 
 
 class KeepAliveThread(threading.Thread):
@@ -379,10 +382,13 @@ class Worker(object):
         self.run_succeeded = True
         self.unfulfilled_counts = collections.defaultdict(int)
 
-        try:
-            signal.signal(signal.SIGUSR1, self.handle_interrupt)
-        except AttributeError:
-            pass
+        # note that ``signal.signal(signal.SIGUSR1, fn)`` only works inside the main execution thread, which is why we
+        # provide the ability to conditionally install the hook.
+        if not self._config.no_install_shutdown_handler:
+            try:
+                signal.signal(signal.SIGUSR1, self.handle_interrupt)
+            except AttributeError:
+                pass
 
         # Keep info about what tasks are running (could be in other processes)
         if worker_processes == 1:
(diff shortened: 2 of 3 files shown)


doc/configuration.rst
```diff
@@ -266,6 +266,14 @@ worker-wait-jitter
...
   Default: 5.0
 
+worker-install-shutdown-handler
+  If true, workers will stop requesting new work and finish running pending
+  tasks after receiving a `SIGUSR1` signal. This provides a hook for
+  gracefully shutting down workers that are in the process of running
+  (potentially expensive) tasks. Note this hook does not work on Windows
+  operating systems, or with jobs launched outside the main execution thread.
+  Defaults to true.
+
```
Tarrasch: Put this into [worker] section, and call it only "install-shutdown-handler" (naturally)

Tarrasch: Actually call it "no_install_shutdown_handler". "no" because False should be default for all bool parameters and underscores because that's the new convention. (docs lag behind a lot, sorry about that!)

luigi/worker.py
```diff
@@ -300,6 +300,10 @@ class worker(Config):
                                          config_path=dict(section='core', name='retry-external-tasks'),
                                          description='If true, incomplete external tasks will be '
                                          'retested for completion while Luigi is running.')
+    install_shutdown_handler = BoolParameter(default=True,
+                                             config_path=dict(section='core', name='worker-install-shutdown-handler'),
```
Tarrasch: Remove this line. `config_path` should actually be called `old_deprecated_config_path` so to say.

luigi/worker.py
```diff
@@ -300,6 +300,10 @@ class worker(Config):
                                          config_path=dict(section='core', name='retry-external-tasks'),
                                          description='If true, incomplete external tasks will be '
                                          'retested for completion while Luigi is running.')
+    install_shutdown_handler = BoolParameter(default=True,
```
Tarrasch: As said in precious comments, make the default False. :)

luigi/worker.py
```diff
@@ -370,10 +374,13 @@ def __init__(self, scheduler=None, worker_id=None, worker_processes=1, assistant
         self.run_succeeded = True
         self.unfulfilled_counts = collections.defaultdict(int)
 
-        try:
-            signal.signal(signal.SIGUSR1, self.handle_interrupt)
-        except AttributeError:
-            pass
+        # note that ``signal.signal(signal.SIGUSR1, fn)`` only works inside the main execution thread, which is why we
+        # provide the ability to conditionally install the hook.
```
Tarrasch: Thanks for the comment. Makes it readable for the next reader. :)

Tarrasch:
Overall everything is correct. Just some changes need to be done :)

erikbern:
LGTM on the stuff @Tarrasch pointed out

doc/configuration.rst
```diff
@@ -267,6 +267,19 @@ worker-wait-jitter
   Default: 5.0
 
 
+[worker]
```
casey-green: I can move the rest of the `worker-*` configs into this section in another PR if you'd like!

casey-green:
Thanks for reviewing - responded to all comments!

casey-green:
hmm.. sorry, let me figure out why these tests are failing.

doc/configuration.rst
Tarrasch: Would love to. Thanks

Tarrasch:
I think it's related to fragile test I added in a  pretty recent PR #1471. I relied on a config path in a stringly typed way in one of the tests: https://github.com/spotify/luigi/pull/1471/files#diff-f1cd3288d80ac94f428311dbefa70c7fR817

Maybe that helps. Thanks for looking into it!

casey-green:
@Tarrasch - are you good to merge this now that the tests are fixed? Also, there is https://github.com/spotify/luigi/pull/1517 which just updates the worker configs in `doc/configuration.rst` - might be easier to merge that first and then I can rebase and fix the conflicts in `doc/configuration.rst`.

Tarrasch:
Thanks you for rebasing this change.

As you requestd I merged #1517 first.

Tarrasch:
Apparently git(hub) claims it can merge this without rebasing. So I'm just gonna do it. Assuming the mergetool does the right thing (as always)

Tarrasch:
Thanks for all your awesome patches lately @casey-green!!

PR #1510, opened 2016-01-14
Sort task families in sidebar alphabetically

harpaj:
The task families in the sidebar of the visualiser were previously unsorted, which made it pretty difficult to find something in there. This commit sorts them alphabetically.


erikbern:
lgtm

Tarrasch:
Is this tested to some reasonable extent @harpaj? Is so lets merge. :)

harpaj:
We had it running for a few days, it doesn't throw any warnings/errors in the JS console and it appears to work:
Sorted taskbar
If you can think of another way of testing it, please let me know - otherwise I think it should be fine to merge it.

Tarrasch:
Thanks. That seems production-ready enough ^^

PR #1511, opened 2016-01-14
Allow deps.py tool to print tasks with multiple outputs

lebolo:
Previously, tasks with multiple outputs would print "to be determined" because the return value of a task's `output()` would be an _Iterable_ instead of a _Target_.

Now, if the output is iterable, we loop through and print each one.

### DIFF
--- a/luigi/tools/deps.py
+++ b/luigi/tools/deps.py
@@ -46,6 +46,7 @@ from luigi.task import flatten
 from luigi import parameter
 import sys
 from luigi.cmdline_parser import CmdlineParser
+import collections
 
 
 def get_task_requires(task):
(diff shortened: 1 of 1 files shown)


lebolo:
I'm seeing errors in Python 3 tests due to the old `print` **statement**. However, I'm not sure if _deps.py_ ever passed these tests (since it used the `print` statement before this PR).

Should I add `from __future__ import print_function` and use the `print` **function** here?

erikbern:
it's really weird that it used to work before this PR... either way why don't you convert to the python 3 way?

lebolo:
Okay, I converted to Py3 and also removed the unicode strings (e.g. `u"blah"`) since I didn't see this behavior in any other luigi code. I left this as 2 commits. Let me know if you prefer I squash or submit another PR for the `print` function changes.

erikbern:
This looks good – I would have preferred a test, but it's fine since it's a separate tool!

erikbern:
(and also there were no tests before you added this anyway)

Tarrasch:
There is this minimal test: https://github.com/spotify/luigi/blob/43c744b611b8b9c90ac4a8d9d67ae6bd29cddd49/test/cmdline_test.py#L257-L265

However, I suppose that test is flawed, because `python` binary always refer to python2 not python3.

lebolo:
I'll give it a shot and see if I can increase the coverage (no promises =)). Not sure if I'll be able to do this any time soon, though. So I can open another PR for that test if you want to merge this in as is.

PR #1513, opened 2016-01-15
README: Add CoC

nailor:


Tarrasch:
Thanks, two thoughts:

 * I would prefer to reword it like: "When unsure about how to behave, consult this code of conduct". I don't think one needs to actually read whole page before contributing to this project.
 * This should be in CONTRIBUTING.md (people are also more likely to see it right before sending their PR then :))

nailor:
Wording is actually the same we use for other @spotify projects, would prefer it to be the same here too. And we do expect people to honor it. 

Moving to CONTRIBUTING.md makes sense, shall do that

erikbern:
LGTM

Tarrasch:
> Wording is actually the same we use for other @spotify projects

Ok. I don't like the wording but my liking for following conventions outweighs. :)

OK. I read the through the entire CoC and I like it.

PR #1514, opened 2016-01-15
fix .travis.yml errors reported by http://lint.travis-ci.org/

benjixx:
Fixes the following reported errors:
- value for cache section is empty, dropping
- in cache section: unexpected key $HOME/.pip-cache, dropping
- specified jdk, but setting is not relevant for python

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: python
 
-jdk: oraclejdk7
-
 services:
   - elasticsearch
   - mysql
@@ -45,7 +43,8 @@ matrix:
 sudo: false
 
 cache:
-  - $HOME/.pip-cache
+  directories:
+    - $HOME/.pip-cache
 
 install:
   - pip install 'tox<3.0'


erikbern:
cool thanks!

benjixx:
Hmm, any idea why this fails? I cannot easily spot the actual error at the moment.

benjixx:
What first came to my mind was that it may have to do something with the Java version, but according to the [build details](https://travis-ci.org/spotify/luigi/jobs/102632420#L63) this seems to be still oraclejdk7.

Tarrasch:
I thought we need to set a java version, as the minicluster and many of the hadoop tests rely on Java being installed and it's environment variables being set.

I really like PRs that are cleaning things up. Good job. :)

benjixx:
@erikbern @Tarrasch Having another fresh look at the failing build, error seems to be in:
[Test that we do not send error emails on the failures of external tasks](https://travis-ci.org/spotify/luigi/jobs/102632420#L1309)

Looking at [other recent PRs](https://travis-ci.org/spotify/luigi/pull_requests) this seems to be common and not related to this PR in particular.

Tarrasch:
Eh. sorry about that @benjixx! Maybe I can fix that build error tomorrow. I don't know how/why it suddenly started to fail!

Anyway as you correctly explained. It's not part of this PR. I'm merging this. :)

PR #1515, opened 2016-01-15
put temp file into temporary directory and make it relative

### LINKED ISSUE #1368: FTP RemoteTarget tries to create local temp file in root
luigi.contrib.ftp.RemoteTarget open('r') tries to create a local temporary file based on the remote file path with a random suffix but using the entire path of the remote file.

In my case, the remote file is in the root folder of the FTP server so it tries to create the same file locally with '-luigi-tmp-%09d' as a suffix.  Because I don't have write permission to my root folder it fails.

a86c6f7964:
fixes #1368

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -29,6 +29,7 @@ import datetime
 import ftplib
 import os
 import random
+import tempfile
 import io
 
 import luigi
@@ -258,7 +259,8 @@ class RemoteTarget(luigi.target.FileSystemTarget):
             return self.format.pipe_writer(AtomicFtpFile(self._fs, self.path))
 
         elif mode == 'r':
-            self.__tmp_path = self.path + '-luigi-tmp-%09d' % random.randrange(0, 1e10)
+            temp_dir = os.path.join(tempfile.gettempdir(), 'luigi-contrib-ftp')
+            self.__tmp_path = temp_dir + '/' + self.path.lstrip('/') + '-luigi-tmp-%09d' % random.randrange(0, 1e10)
             # download file to local
             self._fs.get(self.path, self.__tmp_path)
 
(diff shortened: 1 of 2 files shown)


erikbern:
great. btw I have some ideas on how to refactor the temp file stuff, but will do separately

Tarrasch:
This definitely needs a test case documenting the behaviour. I'm intending to refactor and generalize  this eventually, and I don't want this to break then =)

a86c6f7964:
I assumed there is already a test case making sure that a file is written.

Tarrasch:
The purpose of the test would be to make sure that if the implementation change is reverted, the test fails.

a86c6f7964:
I fixed the existing test case to work, but it is an integration test with an ftp server (not my favorite). Then I added a line to make sure that the file can be opened and read properly. I do not think that the actual location that the file is put should be considered part of the test, but since the file is at an absolute location the test cannot pass since most machines do not let you write into /test/ directory location. (open to suggestions on a better way to test this)

erikbern:
Can you resolve the flake8 errors?
```
./luigi/contrib/ftp.py:118:30: E261 at least two spaces before inline comment
./luigi/contrib/ftp.py:124:24: E261 at least two spaces before inline comment
./test/_test_ftp.py:191:66: E261 at least two spaces before inline comment
```

a86c6f7964:
okay.

erikbern:
you put the space in the wrong place :) should be before the # not after :)

erikbern:
fyi you can run the tests locally by running `tox -e flake8`

a86c6f7964:
oops. sorry about that. thanks for the tip. they look good now.

erikbern:
thanks. will merge on travis!

PR #1517, opened 2016-01-17
Update configurations doc to reflect new configuration convention for workers

casey-green:
I have another PR (https://github.com/spotify/luigi/pull/1509) that creates the `[worker]` configuration section, figured I might as well move the rest of the configs in there.


Tarrasch:
Thanks!

PR #1520, opened 2016-01-21
Add ExternalProgramTask for running external programs

ehdr:
To be used for running any external application or script as a luigi Task. The code is basically a generalization of that in `luigi.contrib.spark`, and we've found it pretty useful. Is this something you would want to include in luigi? Feedback appreciated!

A trivial (and useless) example:

    class MvTask(luigi.contrib.external_program.ExternalProgramTask):
        def program_args(self):
            return ['mv', self.input().path, self.output().path]

        def input(self):
            return luigi.LocalTarget('source')

        def output(self):
            return luigi.LocalTarget('target')

If this is a reasonable idea, other task types could be "rebased" on top of this one. I've done it for `SparkSubmitTask`, and I can push that separately.

Also add an `ExternalPythonProgramTask` that allows optionally setting a virtualenv and `PYTHONPATH`.

(It's called "external _program_" (vs. "app" or "process" or "shell script"...) simply because the docs for [`subprocess.Popen`][1] refer to it as a "[child] program")

[1]: https://docs.python.org/3/library/subprocess.html#subprocess.Popen


Tarrasch:
This looks good. And I can definitely see this being merged. I wanted something like this a couple of weeks back!

In my case however, I realized there's an external library called [plumbum](https://plumbum.readthedocs.org/en/latest/) which is so much better than dealing with the clunky `subprocess` library. I'm really interested to hear from you @ehdr if you've considered to use plumbum inside the `run()` of normal `luigi.Task`s. How does it relate to ExternalProgramTask?

luigi/contrib/external_program.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    virtualenv = luigi.Parameter(
+        default=None,
+        description='path to the virtualenv directory to use. It should point to '
+                    'the directory containing the ``bin/activate`` file used for '
+                    'enabling the virtualenv.')
+    extra_pythonpath = luigi.Parameter(
+        default=None,
+        description='extend the search path for modules by prepending this '
+                    'value to the ``PYTHONPATH`` environment variable.')
```
Tarrasch: You must set `positional=False` for these two parameters. Because this is a typical base task.

luigi/contrib/external_program.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        Subclass this method to control environment variables for the program
+
+        :return: dict mapping environment variable names to values
+        """
+        env = os.environ.copy()
+        return env
+
+    def _clean_output_file(self, file):
+        file.seek(0)
```
Tarrasch: it  looks like file is a reserved keyword? Doesn't pep8 complain?

luigi/contrib/external_program.py
```diff
@@ -0,0 +1,159 @@
...
+
+        try:
+            with ExternalProgramRunContext(proc):
+                proc.wait()
+
+            stdout = self._clean_output_file(tmp_stdout)
+            stderr = self._clean_output_file(tmp_stderr)
+
+            if stdout:
+                logger.info('Program stdout:\n{}'.format(stdout))
```
Tarrasch: So you will not see the output of the program until it finishes?

Tarrasch:
Also big thanks for putting this in contrib and adding test cases already :)

luigi/contrib/external_program.py
ehdr: :blush:

erikbern: it's just a globally defined variable that refers to the `file` class, so one of those things that you *can* override, but generally shouldn't :)

luigi/contrib/external_program.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def _clean_output_file(self, file):
+        file.seek(0)
+        return ''.join(map(lambda s: s.decode('utf-8'), file.readlines()))
+
+    def run(self):
+        args = list(map(str, self.program_args()))
+
+        logger.info('Running command: %s', ' '.join(args))
+        tmp_stdout, tmp_stderr = tempfile.TemporaryFile(), tempfile.TemporaryFile()
```
erikbern: note for myself: another example of temporary files used in Luigi. we should merge all uses :)

erikbern:
Can you add a test that doesn't mock `Popen`? You can run some dummy thing like just `cp` an empty file or `touch` or something silly

luigi/contrib/external_program.py
ehdr: Actually, it seems it's not a [reserved word](https://docs.python.org/3/reference/lexical_analysis.html#keywords). And pep8 doesn't complain about it. But I'll change it anyway. :)

luigi/contrib/external_program.py
ehdr: True. This behavior is "inherited" from [luigi.contrib.spark][1] (which in turn I think inherited it from [luigi.contrib.hadoop][2]), and I didn't consider it, but this could perhaps be improved if we move to plumbum?

[1]: https://github.com/spotify/luigi/blob/master/luigi/contrib/spark.py#L241
[2]: https://github.com/spotify/luigi/blob/master/luigi/contrib/hadoop.py#L280

luigi/contrib/external_program.py
ehdr: :+1: Done.

erikbern:
Nice PR! I've thought about building something like this for a while.

ehdr:
Thanks! :) I wasn't aware of plumbum, but it looks very nice indeed! I'll have a closer look at it tonight. If you approve of the public interface ('program_args' and 'program_environment') then maybe moving to plumbum could be a next-step refactoring?

ehdr:
Thanks @erikbern! I'll look at adding a non-mocked test tonight.

erikbern:
Since you already built it using subprocess I think it's totally fine to merge as-is. I've also looked at plumbum a bit and it seems nice.

luigi/contrib/external_program.py
Tarrasch: @erikbern  Does this count? I thought we only cared to re-factor cases where you write to a temporary path and then move the file to `self.output().path`.

     tmp_stdout, tmp_stderr = tempfile.TemporaryFile(), tempfile.TemporaryFile() 

Looks very sane to me, code corresponds to intent I feel.

Tarrasch:
@ehdr I'm not suggesting you to rewrite. I mostly just want that the documentation for this module also fairly advertises plumbum, and perhaps guides the luigi user when to use plumbum and when to use this helper task. If that makes sense. :)

luigi/contrib/external_program.py
erikbern: oh wait you are right, these return file handles, not "physical" files

ehdr:
Thanks both for the comments. :) I added a simple non-mocked test, that touches a file and then asserts that it exists. Caveat: I'm not sure if this would work on for example Windows?

erikbern:
great! i don't think the tests run on windows anyway :)

Tarrasch:
Cool. Can this be (1) rebased so we see that the tests pass and (if you like) (2) have some module-level docs at least mentioning plumbum as an alternative?

My intuition is that if you have one task which does doing a lot of program calls to things like `cat`, `cut`, `grep`, you want plumbum. But if you only call one external program, or want to create a base class that mainly calls another program (spark, hadoop, pig, etc.) you want ExternalProgramTask.

ehdr:
@Tarrasch Will do! I see now what you mean, sorry for being slow to catch on. :) I agree about your intuition, but since I don't have experience of plumbum I'm not in a good position to compare. Generally I'd guess the tricky parts are wrapping the command in robust handling of command failures (non-zero exit codes), `KeyboardInterrupt`s and logging of output? I guess you get some of that "for free" from plumbum?

I'll push a rebased version later, and you prefer it squashed as well, right? Will also add to the docs. Where would such module level documentation typically go?

ehdr:
@Tarrasch Ok, rebased and CI looks good. Also added module documentation - does it capture your view? Remains to squash, and maybe add a copyright notice snippet to the files?

erikbern:
let's merge

Tarrasch:
Oh yes you're right. These files should have a copyright notice. I don't know why but I've been told so. @nresare maybe knows?

Yes I think you should squash the commits while at it. The history is not so interesting in this case.

Thank you very much @ehdr :)

erikbern:
let's just merge. i think the commit history is fine, and i can add copyright headers. need to reduce all the round trips back and forth :)

Tarrasch:
Sure. Joy of contributing factor should be high and of round trips are making it less fun I suppose. Good work @ehdr!

ehdr:
Nice! :) Thanks @Tarrasch and @erikbern !

PR #1522, opened 2016-01-23
Fix test: do not send an email if a task is external and retry-able

casey-green:
Fixes this test:

```python
@with_config(dict(worker=dict(retry_external_tasks='true')))
@email_patch
def test_external_task_retries(self, emails):
        """
        Test that we do not send error emails on the failures of external tasks
        """
        class A(luigi.ExternalTask):
            pass

        a = A()
        luigi.build([a], workers=2, local_scheduler=True)
        self.assertEqual(emails, [])
```

This is holding up one of my other PRs :)

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -766,8 +766,9 @@ class Worker(object):
                 # Not a running task. Probably already removed.
                 # Maybe it yielded something?
 
-            if status == FAILED and expl:
-                # If no expl, it is because of a retry-external-task failure.
+            # external task if run not implemented, retry-able if config option is enabled.
+            external_task_retryable = task.run == NotImplemented and self._config.retry_external_tasks
+            if status == FAILED and not external_task_retryable:
                 self._email_task_failure(task, expl)
 
             new_deps = []


Tarrasch:
Ah, goes to show how important tests can be, and how tricky version control is! Thanks @casey-green!. I'm merging right away, thanks for fixing the build!

erikbern:
this is quite horrible but i don't blame you. we need to refactor this. let me look at it

Tarrasch:
@erikbern, yea these things are unfortunate but we have to live it I think. We'll at least be gray-haired before CI tools can safely say "Yes, you can merge this to 'master' and the build won't fail, despite that there have been changes to 'master' since I ran this patch".

To clarify, I suspect that the create/merge order of these two "contradictory" PRs https://github.com/spotify/luigi/pull/1471 and https://github.com/spotify/luigi/pull/1449 caused the build to be red.

erikbern:
I think the thing that bothers me is that we're using FAILED to determine that some external task is missing.

How do you feel about a new state "INCOMPLETE" that happens any time a task.complete() returns False? That would apply to both external tasks, but also non-external tasks that simply never created any output

Tarrasch:
> I think the thing that bothers me is that we're using FAILED to determine that some external task is missing.

Yea that bothers me as too. I like your idea of INCOMPLETE. Maybe we could kill the not so user-friendly flag `--retry-external-tasks`. The reason it's not so user friendly is that you often need to mix it with `--worker-keep-alive`.

I wish I had more time to work on this ... :-/

erikbern:
cool will think some more about it. also wish i had more time :)

worker.py is a cluterfuck right now, so maybe i'll spend some time cleaning it up!

PR #1524, opened 2016-01-25
added copyright headers to some files

erikbern:


Tarrasch:
Shouldn't it be 2016? :)

What ever. Looks fine to merge to me.

erikbern:
not sure if we should update all the other files too

Tarrasch:
I think the commit message "2015-2016" is too confusing to make this PR merge-able. Can you squash the commits instead?

erikbern:
just did it

but i really think we should be more lax about rebasing. it's not that big of a deal

Tarrasch:
Thanks for doing it. :)

It's a difficult topic with many pros and cons. I really just wished GitHub would let me squash commits in the web UI in a convenient quick way ...

PR #1525, opened 2016-01-25
Rebase SparkSubmitTask onto ExternalProgramTask

ehdr:
As a continuation of #1520, this is an example of how `SparkSubmitTask` could be "rebased" onto `ExternalProgramTask`, since most of the code is shared between them. Not sure if this is a good way to go?

The public interface of `SparkSubmitTask` remains as-is.

However, there will be subtle changes to the output to `stdout` and logs (e.g. 'Program failed[...]' with this patch vs. 'Spark job failed[...]' before). Also it will raise a `ExternalProgramRunError` on execution errors instead of a `SparkJobError` as before. Is this considered acceptable deviations from backwards compatibility?

~~Finally, it will now log `stdout` and `stderr` whenever it's (and only when it's) non-empty, even if the process exits successfully (return code 0). Previously, `stderr` was only logged if the return code was non-zero.~~ [changed back to preserve the old behavior in a [later commit](https://github.com/ehdr/luigi/commit/a0417c64fd5da4faa3ad6c2eb7cada47a6711acd#diff-7245b212453781765bd8571c3cc95fc2R53), thanks to comment from @deuxpi]

### DIFF
--- a/luigi/contrib/spark.py
+++ b/luigi/contrib/spark.py
@@ -17,8 +17,6 @@
 
 import logging
 import os
-import signal
-import subprocess
 import sys
 import tempfile
 import shutil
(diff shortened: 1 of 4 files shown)


deuxpi:
The `stdout` from a Spark job can be terribly verbose, so it's nice to have that silences in the case where there were no errors. I personally would like that behavior to stay available somehow, even if I have to add a flag parameter to each of our task classes.

luigi/contrib/spark.py
```diff
@@ -189,12 +146,18 @@ def hadoop_conf_dir(self):
         return configuration.get_config().get("spark", "hadoop-conf-dir", None)
 
     def get_environment(self):
-        env = os.environ.copy()
+        env = super(SparkSubmitTask, self).program_environment()
```
erikbern: not a big fan of calling methods on the parent class. should only be doable for constructors

ehdr:
@deuxpi Good idea. But [`stdout` was always logged](https://github.com/spotify/luigi/pull/1525/files#diff-7245b212453781765bd8571c3cc95fc2L241) even before this patch, so that does not change. The only change is for `stderr`.

It could be a nice new feature though, to add an option to turn it off?

luigi/contrib/spark.py
ehdr: Sure! It's just doing `os.environ.copy()` anyway. I'll change it.

deuxpi:
@ehdr Ack! Sorry, I meant `stderr`. An option would be great, I remember at least one time where I was actually interested by the output of Spark, but most of the time I'm clearly not.

Tarrasch:
> However, there will be subtle changes to the output to stdout and logs (e.g. 'Program failed[...]' with this patch vs. 'Spark job failed[...]' before). Also it will raise a ExternalProgramRunError on execution errors instead of a SparkJobError as before. Is this considered acceptable deviations from backwards compatibility?

Yes I think this is acceptable. :)

Tarrasch:
I like this patch. I'm +1 the idea. @ehdr, When you think you've fixed all the comments from this  PR, let us know and we'll merge.

ehdr:
I think I got everything now @Tarrasch @deuxpi @erikbern. Not super elegant... but please comment! :)

Tarrasch:
Great! All improvements are welcome and most code in this code base is not really elegant ^^

PR #1527, opened 2016-01-27
Add RedshiftUnloadTask class

chenzhan:
To export data from Redshift to s3. This is normally done by calling
a SQL command "UNLOAD". However it requires s3 credentials and some
other parameters. This class serves as a template for these tasks.

First commit here :-)
Thanks!


Tarrasch:
Thanks @chenzhan! I'm happy to welcome your first commit! There seems to be some styling issues that Travis doesn't like about your code. See [this](https://github.com/spotify/luigi/blob/master/CONTRIBUTING.rst#running-the-tests) for how to test things locally. And you also need to add tests to your code. Others might later come and "improve" the code and we don't want any breaking behaviour to go unnoticed. :)

ramnathmedikonda:
One quick follow up question with regards to RedshiftUnloadTask class which I overlooked initially.

Any specific reason why the output is a RedshiftTarget. I guess it should be S3Target instead, right? Because as per my understanding, Luigi checks for the target in output before it runs the task and in this case it should be a S3Target.

Please correct me if I am wrong.

chenzhan:
As I understand it, output in Luigi is really the "target" of your operation. In this case (UNLOAD to s3), there are actually two targets: Redshift (the target for your query) and S3 (the target for your data). However, I chose to use Redshift as the output because it is what you would interact with (run a query), while you don't interact S3 at all.

PR #1533, opened 2016-02-02
Allows the usage of Hive Metastore

rizzatti:
The usage of Hive Metastore greatly speeds up table and partition lookup in luigi.
Although the MetastoreClient was already implemented, the code always looked for one of two alternatives of the hive command, which is slow.

This PR's settings are only active if
```
[hive]
release: metastore
```
is set

### DIFF
--- a/luigi/contrib/hive.py
+++ b/luigi/contrib/hive.py
@@ -183,7 +183,7 @@ class MetastoreClient(HiveClient):
                 try:
                     partition_str = self.partition_spec(partition)
                     thrift_table = client.get_partition_by_name(database, table, partition_str)
-                except NoSuchObjectException as e:
+                except hive_metastore.ttypes.NoSuchObjectException as e:
                     return ''
             else:
                 thrift_table = client.get_table(database, table)
--- a/luigi/contrib/hive.py
+++ b/luigi/contrib/hive.py
@@ -180,8 +180,11 @@ class MetastoreClient(HiveClient):
     def table_location(self, table, database='default', partition=None):
         with HiveThriftContext() as client:
             if partition is not None:
-                partition_str = self.partition_spec(partition)
-                thrift_table = client.get_partition_by_name(database, table, partition_str)
+                try:
+                    partition_str = self.partition_spec(partition)
+                    thrift_table = client.get_partition_by_name(database, table, partition_str)
+                except NoSuchObjectException as e:
+                    return ''
             else:
                 thrift_table = client.get_table(database, table)
             return thrift_table.sd.location
--- a/luigi/contrib/hive.py
+++ b/luigi/contrib/hive.py
@@ -26,7 +26,6 @@ from luigi import six
 
 import luigi
 import luigi.contrib.hadoop
-import hive_metastore
 from luigi.target import FileAlreadyExists, FileSystemTarget
 from luigi.task import flatten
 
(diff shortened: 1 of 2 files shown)


luigi/contrib/hive.py
```diff
@@ -180,8 +180,12 @@ class MetastoreClient(HiveClient):
     def table_location(self, table, database='default', partition=None):
         with HiveThriftContext() as client:
             if partition is not None:
-                partition_str = self.partition_spec(partition)
-                thrift_table = client.get_partition_by_name(database, table, partition_str)
+                try:
+                    import hive_metastore.ttypes
+                    partition_str = self.partition_spec(partition)
+                    thrift_table = client.get_partition_by_name(database, table, partition_str)
+                except hive_metastore.ttypes.NoSuchObjectException as e:
```
erikbern: can you fix this flake8 issue?
`./luigi/contrib/hive.py:187:71: F841 local variable 'e' is assigned to but never used`

(btw you can run `tox -e flake8` locally to test)

rizzatti: Sure, I can fix that.
@erikbern This PR requires either the Hive python libs in the pythonpath or hive-thrift-py package installed. Do you have any idea on how to handle this?

erikbern: I think you should just leave it as it is. Any dependency in luigi.contrib shouldn't be in requirements.txt. Optionally throw a warning if the right package isn't installed (and you should probably add a note that installing hive-thrift-py is highly recommended)

rizzatti: Ok, I can do that.
But that's the reason Travis tests are failing as well.

erikbern: Where do you see that? When I look at the Travis tests, all I see is the error above
`./luigi/contrib/hive.py:187:71: F841 local variable 'e' is assigned to but never used`

erikbern: You can also add test time dependencies in tox.ini:

https://github.com/spotify/luigi/blob/master/tox.ini

luigi/contrib/hive.py
```diff
@@ -244,8 +248,11 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
 
 def get_default_client():
-    if get_hive_syntax() == "apache":
+    syntax = get_hive_syntax()
+    if syntax == "apache":
         return ApacheHiveCommandClient()
+    elif syntax == "metastore":
```
davemt: Can we introduce a new configuration that is `[hive]client` set as either `hivecli` or `thrift` (defaults to `hivecli`)?

This is using `[hive]release` which was intended to be the version of hive/hadoop (which translated to client differences for standard CLI hive). It is confusing to have a choice which is not a version. Also there may be some future thing that depends on us knowing the version outside of picking client.

davemt: `thrift` could also be `metastore` instead if you think thatmakes more sense

erikbern: sure

luigi/contrib/hive.py
rizzatti: My mistake. You're right. Previously there was an error, but my last commit had fixed it. I thought it still had something to do with the libraries, but if I had checked travis sooner, and seen it was this simple, I would have fixed it earlier.

rizzatti:
thanks!

PR #1536, opened 2016-02-06
make saving data for the execution summary optional

demoray:
When a worker executes a large number of tasks, this has the same impact as a
memory leak.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -313,6 +313,8 @@ class worker(Config):
     no_install_shutdown_handler = BoolParameter(default=False,
                                                 description='If true, the SIGUSR1 shutdown handler will'
                                                 'NOT be install on the worker')
+    save_summary_data = BoolParameter(default=True, description='If true, save data required for '
+                                      'execution summary')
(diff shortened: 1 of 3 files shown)


demoray:
I included the unit test as is, as I am unsure why the with_config statement is not working as I expect.  If I set the config by hand immediately prior to self.summary via the following, then the test works as expected.

        self.worker._config.save_summary_data = False

erikbern:
lgtm

luigi/worker.py
```diff
@@ -313,6 +313,8 @@ class worker(Config):
     no_install_shutdown_handler = BoolParameter(default=False,
                                                 description='If true, the SIGUSR1 shutdown handler will'
                                                 'NOT be install on the worker')
+    save_summary_data = BoolParameter(default=True, description='If true, save data required for '
```
Tarrasch: `default=True` shouldn't be used for BoolParameters iirc.

Tarrasch:
Did you drop the ball on this?

PR #1537, opened 2016-02-08
Redshift Update - init_copy

dlstadther:
Independently check the 3 possible init_copy paths, allowing multiple of them to execute, if applicable.


DeaconDesperado:
Trying to familiarize myself with the RedShift code path in order to be able to better review, but this change LGTM :+1:

PR #1541, opened 2016-02-09
Keep track of num failed tasks.

dhruvg:
Add bookkeeping for number of failed tasks when getting work from the scheduler and workers.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -261,6 +261,9 @@ class Worker(object):
         else:
             return state.get_pending_tasks()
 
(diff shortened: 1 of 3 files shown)


erikbern:
LGTM but will this return the global number of failed tasks since last restart of the scheduler?

dhruvg:
Good catch, my intention was to get the number of failed tasks for the given worker. I have updated the PR.

dhruvg:
Closing in favor of https://github.com/spotify/luigi/pull/1543 which squashes the commits.

PR #1543, opened 2016-02-10
Keep track of num failed tasks for a worker.

dhruvg:
Add bookkeeping for number of failed tasks when getting work from the scheduler and workers.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -261,6 +261,13 @@ class Worker(object):
         else:
             return state.get_pending_tasks()
 
+    def get_failed_tasks(self):
+        return six.moves.filter(lambda task: task.status == FAILED, self.tasks)
+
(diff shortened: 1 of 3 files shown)


dhruvg:
@erikbern Any idea what is up with the builds? Doesn't seem to be a code issue.

dhruvg:
@erikbern My hunch is that it is impacting the performance of that test case since it is called `get_work` thousands of times and we are now calculating num failed tasks on each request.

dhruvg:
@erikbern bump?

erikbern:
anyway this was closed? i guess it's been sitting here for a long time, not sure why

Tarrasch:
After a quick glance of the code I don't understand it. It just threads along the variable but it never seems to print or use it for anything. Anyway still looks interesting. :)

PR #1530, opened 2016-01-28
Improve getpcmd

KyleJamesWalker:
Some versions of busybox's ps do not include the paramaters being used (`-o`). For example this function failed on Alpine linux.

Tested on base docker images of Ubuntu, and Apline.

PR #1547, opened 2016-02-14
Fix of process lock on Windows

ngr-t:
The current version of
luigi.lock.getpcmd does not work properly on Windows, because Windows does not have ps command which is used to get command in the function. As a result, locking functionality does not work on Windows.
Fix the function to use wmic command instead of ps when luigi is run on Windows.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -335,3 +335,8 @@ texinfo_documents = [
 
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #texinfo_no_detailmenu = False
+
+# Some regression introduced
+# https://github.com/sphinx-doc/sphinx/issues/2330
+# https://github.com/spotify/luigi/pull/1555
+highlight_language = "python"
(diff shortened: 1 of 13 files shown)


erikbern:
After merging #1530 this has a merge conflict – would appreciate if you can rebase and resubmit! Thanks!

ngr-t:
Thank you. I've done rebase and push (there seems to be conflicts, though...)
I didn't confirmed all tox tests pass, but "$ tox -e docs" passed.

erikbern:
I think something went wrong when you rebased. Let me know if you can fix it. Alternatively if it's all messed up let me know and I can create a separate PR

ngr-t:
I've fixed it.

erikbern:
Something looks completely messed up. I think it's easier if you just restart from master and apply the change manually

ngr-t:
Sorry.
I'll do as you suggested.

PR #1555, opened 2016-02-21
Add OpenerTarget Support.

KyleJamesWalker:
Allow a single luigi target to open multiple file system types.  This can allow for easier local testing, unit test, and greater flexibility of your luigi tasks.

Example:
``` python
import luigi

from luigi.contrib.opener import OpenerTarget

class ConfigurableTask(luigi.Task):
    output_file = luigi.parameter(default='s3://foo/bar/baz.txt')

    def run(self):
        payload = {'foo': 'bar'}
        with self.output().open('w') as fp:
            json.dump(payload, fp)

    def output(self):
        return OpenerTarget(self.output_file)

example1 = ConfigurableTask(output_file='/tmp/local_file.txt')
example2 = ConfigurableTask(output_file='mock://tmp/memory_file.txt')

{% endhighlight %}
```

I've only add three target types so far to make sure this pattern is accepted into the code base (and this is all I needed to run our pipeline in all environments needed).

### DIFF
--- a/luigi/contrib/opener.py
+++ b/luigi/contrib/opener.py
@@ -1,0 +1,270 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2012-2015 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""OpenerTarget support, allows easier testing and configuration by abstracting
+out the LocalTarget, S3Target, and MockTarget types.
+
+Example:
+
+.. code-block:: python
+
+    from luigi.contrib.opener import OpenerTarget
+
+    OpenerTarget('/local/path.txt')
+    OpenerTarget('s3://zefr/remote/path.txt')
+
+"""
+
+import json
+
+from luigi.file import LocalTarget
+from luigi.mock import MockTarget
+from luigi.s3 import S3Target
+from luigi.target import FileSystemException
+from six.moves.urllib.parse import urlsplit, parse_qs
+
+__all__ = ['OpenerError',
+           'NoOpenerError',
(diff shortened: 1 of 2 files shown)


KyleJamesWalker:
Not sure what's going on with the error, seems to be the same error in #1547, #1543, #1538, and others.

```
Warning, treated as error:
./luigi/doc/api/luigi.rst:11: WARNING: Could not lex literal_block as "python3". Highlighting skipped.
```

KyleJamesWalker:
Note this is fixed if you edit `./luigi/.tox/docs/lib/python2.7/site-packages/sphinx/config.py` and change `highlight_language = ('python3', 'env'),` to `highlight_language = ('python', 'env'),`

Maybe a config was changed, or the sphinx version isn't pinned and it had a breaking change added?

luigi/contrib/opener.py
```diff
@@ -0,0 +1,270 @@
...
+        raise NotImplemented("get_target must be overridden")
+
+
+class MockOpener(Opener):
+
+    """Mock target opener, works like LocalTarget but files are all in
+    memory.
+
+    example:
+    * mock://foo/bar.txt
```
erikbern: minor thing but shouldn't it be just "mock:foo/bar.txt"? double slash usually indicates a hostname

not blocking

luigi/contrib/opener.py
```diff
@@ -0,0 +1,270 @@
...
+        return MockTarget(full_path, **query)
+
+
+class LocalOpener(Opener):
+
+    """Local filesystem opener, works with any valid system path. This
+    is the default opener and will be used if you don't indicate which opener.
+
+    examples:
+    * file://relative/foo/bar/baz.txt (opens a relative file)
```
erikbern: same here as mock – seems like it should be `file:relative/foo/bar/baz.txt`

KyleJamesWalker:
I set it up to use a URI-like syntax (inspired by the syntax of http://commons.apache.org/vfs/filesystems.html).

erikbern:
that's great – thanks for making sure we adhere to standards!

Tarrasch:
@sisidra didn't Spotify want something like this once?

sisidra:
@Tarrasch not sure. We are in JVM land and Luigi is only orchestration layer. Anyway thanks for pinging ;)

Tarrasch:
I think luigi only for orchestration is the case for many luigi players these days. I also don't use luigi MR today based on my experience from Spotify. Anyway I thought Spotify previously had migration issues when it came to HDFS --> GFS migration. And that this Opener factory could smooth things. :)

PR #1557, opened 2016-02-23
Fix of process locking functionality on Windows

ngr-t:
This is the re-submission of PR #1547.
I restarted from the master branch due to the merge problem.

Below is the original PR message:
---
The current version of
luigi.lock.getpcmd does not work properly on Windows, because Windows does not have ps command which is used to get command in the function. As a result, locking functionality does not work on Windows.
Fix the function to use wmic command instead of ps when luigi is run on Windows.

### DIFF
--- a/luigi/lock.py
+++ b/luigi/lock.py
@@ -33,14 +33,23 @@ def getpcmd(pid):
 
     :param pid:
     """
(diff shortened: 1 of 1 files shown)


erikbern:
thanks

PR #1558, opened 2016-02-23
Add support to IAM roles in AWS instances

eramirem:
The retrieval order for S3 credentials will be:
1. kwargs
2. luigi configuration
3. IAM role credentials via instance metadata

### DIFF
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -85,9 +85,14 @@ class S3Client(FileSystem):
         for key in ['aws_access_key_id', 'aws_secret_access_key']:
             if key in options:
                 options.pop(key)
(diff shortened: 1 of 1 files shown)


deuxpi:
We are currently using the S3Target class with IAM roles by omitting the `aws_access_key_id` and `aws_secret_access_key` parameters. The only place where that doesn't quite work for now is when we're using the Redshift task classes, which require that the access key and the secret key be explicitly provided.

erikbern:
looks good, can you rebase?

eramirem:
I just checked and the IAM role is used indeed by just explicitly omitting key parameters. Should I close this?

erikbern:
Sure if the PR doesn't add any functionality let's just close it. But thanks anyway

PR #1560, opened 2016-02-24
Fix: trim email address

### LINKED ISSUE #1559: send_email doesn't trim recipient's email address
The send_email function simply split email address list by comma, and this leads to email address like " john.doe@example.com" (a space before the address).

The problem line is [here](https://github.com/spotify/luigi/blob/master/luigi/notifications.py#L260)

Suggest fix is:
```python
for r in recipients:
    recipients_tmp.extend(r.split(',').strip())
```

qiao-meng-zefr:
Fix #1559

### DIFF
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -257,7 +257,7 @@ def send_email(subject, message, sender, recipients, image_png=None):
     # separated in luigi.cfg
     recipients_tmp = []
     for r in recipients:
-        recipients_tmp.extend(r.split(','))
+        recipients_tmp.extend(r.split(',').strip())
 
     # Replace original recipients with the clean list
     recipients = recipients_tmp


luigi/notifications.py
```diff
@@ -257,7 +257,7 @@ def send_email(subject, message, sender, recipients, image_png=None):
     # separated in luigi.cfg
     recipients_tmp = []
     for r in recipients:
-        recipients_tmp.extend(r.split(','))
+        recipients_tmp.extend(r.split(',').strip())
```
erikbern: but `r.split(',')` returns a list right? seems like it should be `recipients_tmp.extend([a.strip() for a in r.split(',')])`

qiao-meng-zefr:
Oops :-)

qiao-meng-zefr:
Sent a new PR

PR #1562, opened 2016-02-24
Adding optional API authentication

bwall:
I felt uneasy using the Luigi central scheduler while anyone could connect in and potentially tamper with job data or potentially worse (while it should be run on a private network, if an attacker broke the perimeter, they may have a field day with such an unauthenticated service).  As a quick resolution for security focused individuals such as myself, I have added a very simple API key system, where accepted keys are loaded by default from /etc/luigi/api_keys.lst (configurable path) and the worker API key is set in the worker configuration.  If there are 0 available API keys or they fail to load, the API authentication is disabled on the server.

There is no length/complexity limitations set on API keys aside from they can not start or end with white space (white space is stripped during loading of keys), and they are transported by the LAUTH HTTP header.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -181,6 +181,9 @@ rpc-connect-timeout
   Number of seconds to wait before timing out when making an API call.
   Defaults to 10.0
 
+rpc-api-key
+  API key to authenticate with central scheduler
+
 smtp_host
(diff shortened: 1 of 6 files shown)


bwall:
Temporarily pulling this PR until I can fix an issue where this negatively impacts the web viewer.

PR #1563, opened 2016-02-24
Shows last update time instead of start time in visualizer

daveFNbuck:
The time field in the visualiser datatable not very helpful, as I don't really
need to know when a job was scheduled. I usually know that, as I control
scheduling. Instead, I want to know when the job finished or failed or was
disabled. In order to make the time column more useful, this sets the time as
the last time the status was changed. The old start_time field is kept in the
task_list api response for backward compatibility with any tools that might make
that api call.


erikbern:
cool

PR #1565, opened 2016-02-25
Allows package binaries to be passed to job runner

daveFNbuck:
I've recently started using Twitter's Pants Build System to produce binaries for
my pipelines for ease of deployment. This is incompatible with how the streaming
jobs work, but luckily the pre-compiled binary already contains everything I
need to run my streaming jobs.

This allows a package_binary to be specified in a streaming job task, which then
gets used in place of both mrrunner.py and packages.tar. To use this, I updated
my runner script to act like mrrunner.py when the first argument passed is map,
reduce, or combiner by adding the following lines at its start:

    if len(sys.argv) > 1 and sys.argv[1] in ('map', 'reduce', 'combiner'):
        return luigi.mrrunner.main(sys.argv)

Then I just need to add the package_binary to my streaming job base class. This
also has the benefit of removing my need to maintain an extensive extra_modules
list because they're already included in my pre-built pex file.

### DIFF
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -450,9 +450,11 @@ class HadoopJobRunner(JobRunner):
         # build arguments
         config = configuration.get_config()
         python_executable = config.get('hadoop', 'python-executable', 'python')
-        map_cmd = '{0} mrrunner.py map'.format(python_executable)
-        cmb_cmd = '{0} mrrunner.py combiner'.format(python_executable)
-        red_cmd = '{0} mrrunner.py reduce'.format(python_executable)
+        runner_arg = 'mrrunner.pex' if job.package_binary is not None else 'mrrunner.py'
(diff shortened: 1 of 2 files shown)


erikbern:
is it possible/easy to add a test for this?

daveFNbuck:
Probably. I'll add some tomorrow.

On Wed, Feb 24, 2016 at 7:01 PM, Erik Bernhardsson <notifications@github.com

erikbern:
ok would be nice but don't do it if it's nontrivial

daveFNbuck:
Added a couple test cases, caught a nasty bug that broke the original functionality. The minicluster tests caught it too, but I rarely run those on my laptop.

erikbern:
great that you caught a nasty bug! :tada:

daveFNbuck:
Updated to not pass arguments, in accordance with how we're now handling tracking url callbacks. This should be good to merge now.

Tarrasch:
Thanks! :)

PR #1566, opened 2016-02-25
Adds a resources tab to the visualiser

daveFNbuck:
This tab allows a quick view of resource utilization in the visualiser. It also
shows which jobs are running for each resource, but this information is
collapsed by default. This is very helpful when jobs for debugging/tracking when
jobs for a particular resource are getting backed up.

image

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -205,6 +205,9 @@ class RemoteScheduler(Scheduler):
     def worker_list(self):
         return self._request('/api/worker_list', {})
 
+    def resource_list(self):
+        return self._request('/api/resource_list', {})
+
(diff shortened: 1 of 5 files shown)


erikbern:
nice!

Tarrasch:
Ah, very nice. This handles decimal values too right?

daveFNbuck:
I'm not sure, I've never tried decimal values for resources. Does that work for you? Wouldn't it be easier/safer to scale everything up to integers to avoid floating point errors?

If this doesn't handle decimal values, it should be an easy fix. Let me know.

Tarrasch:
Ok I'll test this out and let you know :)

PR #1567, opened 2016-02-25
Fallback process command name in lock.getpcmd

Lucas-C:


luigi/lock.py
```diff
@@ -50,6 +50,8 @@ def getpcmd(pid):
                 spid, scmd = line.strip().split(' ', 1)
                 if int(spid) == int(pid):
                     return scmd
+    # Fallback instead of None, for e.g. Cygwin where -o is an "unknown option" for the ps command:
```
erikbern: is there a better way to detect this rather than just always assuming it's cygwin if everything else fails?

i think you can check if `sys.platform == 'cygwin'`

Lucas-C: I was more inclined to a a generic fallback, as enumerating the platforms just won't work as I experienced recently :
https://github.com/pybuilder/pybuilder/issues/328

Else this code should just stick to the POSIX `ps` argument `-f` to get a full output and do some parsing, and not rely on `-o`

Lucas-C:
I've been bitten by this again today.
Could this be merged please ?

Lucas-C:
Thanks !

PR #1569, opened 2016-02-26
Improves display of unknown tasks

daveFNbuck:
Rather than showing UNKNOWN as the family for an unknown task id in the
visualiser graph, we can usually successfully infer the family from the task id.
This is far preferable to just showing UNKNOWN, as it provides useful
information while looking at dependency graphs. Including the full task id in
the params dict allows for a little more information to be gleaned.

before:
image

after:
image

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -31,6 +31,7 @@ import functools
 import itertools
 import logging
 import os
+import re
 import time
 
 from luigi import six
@@ -74,6 +75,8 @@ STATUS_TO_UPSTREAM_MAP = {
     PENDING: UPSTREAM_MISSING_INPUT,
     DISABLED: UPSTREAM_DISABLED,
 }
+
+TASK_FAMILY_RE = re.compile(r'([^(_]+)[(_]')
 
 
 class scheduler(Config):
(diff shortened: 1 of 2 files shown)


daveFNbuck:
This seems not to be working for me in production now, looking into it...

erikbern:
not sure why we get unknown tasks in the first place. is it only during scheduling?

daveFNbuck:
It's a pretty common pattern for me. I schedule job A, which depends on job B. Both jobs get scheduled and complete, but by cron job keeps scheduling A. After the initial worker dies, job B falls out of the scheduler. When we continue to schedule job A as DONE, we stop updating its dependencies so the scheduler remembers that it depends on the now unknown B.

Tarrasch:
Looks nice! But are we sure we want to start parsing Task Ids again? Didn't we introduce the new id scheme in a way so that nobody would ever be tempted to parse it? @erikbern @ulzha @stephenpascoe

stephenpascoe:
I think parsing the task_family is pretty safe.  Since it's a Python identifier there are no parsing gotcha's and it's clearly the first part of the id.

erikbern:
i would rather fix the underlying issue that tasks disappear

but i have like no time so let's just merge this

PR #1570, opened 2016-02-26
Optionally hides DONE nodes in dependency graph

daveFNbuck:
Often when viewing the dependency graph, I'm trying to figure out why a
particular task hasn't run yet. This can be difficult due to the presence of
hundreds or even thousands of DONE tasks cluttering up the view. In order to aid
debugging, this adds a "Hide Done" checkbox to the dependency graph view in the
visualiser. When checked, the scheduler will ignore DONE tasks (other than the
root) when doing graph traversal (including inverse graph traversal).

image

vs

image

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -189,11 +189,12 @@ class RemoteScheduler(Scheduler):
     def graph(self):
         return self._request('/api/graph', {})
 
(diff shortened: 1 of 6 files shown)


Tarrasch:
LGTM

Tarrasch:
I'm testing this in production now also.

PR #1572, opened 2016-02-29
Added optional timeout to ftp connection. Default set to 60 seconds.

dlstadther:
luigi.contrib.ftp previous didn't not explicitly set connection timeout. [Python docs](https://docs.python.org/2/library/ftplib.html#ftplib.FTP.connect) specify timeout to be "the global default timeout setting" when no timeout is specified. This PR allows explicit timeout, set at 60 seconds by default.


Tarrasch:
LGTM

PR #1574, opened 2016-03-01
Add DictParameter

joemeszaros:
# What is this PR?

This adds parameter for dictionaries.

## Example 

Create a simple task with a single `DictParameter`:

```python
class Task(luigi.Task):
    tags = luigi.DictParameter()

    def run(self):
        environment = self.tags['environment']
        print('I am running in {}'.format(environment))
```

Configure your `tags` parameter with a JSON string and execute your task:

```DictTask --tags '{"environment":"prod","role":"hadoop"}'```

## Implementation details

`DictParameter` parses a `dict` from a JSON string using the stanard JSON library. It uses an  immutable wrapper around dictionaries (`frozendict`), because immutability is required for two reason:
- Execution summary crashes in case of unhashable parameters.  Sample traceback is attached.
- Tasks with unhashable parameters are not cached

[Execution summary crash on unhashable parameters](https://github.com/spotify/luigi/files/152737/execution_summary_unhashable.txt)

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -23,6 +23,8 @@ See :ref:`Parameter` for more info on how to define parameters.
 import abc
 import datetime
 import warnings
+import json
+
 try:
     from ConfigParser import NoOptionError, NoSectionError
 except ImportError:
@@ -30,9 +32,10 @@ except ImportError:
 
 from luigi import task_register
 from luigi import six
-
+from frozendict import frozendict
 from luigi import configuration
 from luigi.cmdline_parser import CmdlineParser
+
 
 _no_value = object()
 
@@ -699,3 +702,18 @@ class EnumParameter(Parameter):
 
     def serialize(self, e):
         return e.name
+
+
+class DictParameter(Parameter):
+    """
+    Parameter whose value is a ``dict``.
+    """
+
+    def parse(self, s):
+        """
+        Parses a ``dict`` from a JSON string using standard JSON library.
+        """
+        mutable_dict = json.loads(s)
+        immutable_dict = frozendict(mutable_dict)
+        frozendict.__repr__ = mutable_dict.__repr__
+        return immutable_dict
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -25,8 +25,7 @@ import datetime
 import warnings
 import json
 from json import JSONEncoder
-from collections import OrderedDict
-import collections
+from collections import OrderedDict, Mapping
 import operator
 import functools
 
@@ -708,7 +707,7 @@ class EnumParameter(Parameter):
         return e.name
 
 
-class FrozenOrderedDict(collections.Mapping):
+class FrozenOrderedDict(Mapping):
     """
     It is an immutable wrapper around ordered dictionaries that implements the complete :py:class:`collections.Mapping`
     interface. It can be used as a drop-in replacement for dictionaries where immutability and ordering are desired.
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -750,16 +750,24 @@ class DictParameter(Parameter):
     .. code-block:: python
 
         class MyTask(luigi.Task):
-          dict_param = luigi.DictParameter()
+          tags = luigi.DictParameter()
 
             def run(self):
-                value = self.dict_param[key]
+                logging.info("Find server with role: %s", self.tags['role'])
+                server = aws.ec2.find_my_resource(self.tags)
 
-    At the command line, use,
+
+    At the command line, use
 
     .. code-block:: console
 
-        $ luigi --module my_tasks MyTask --dict-param <JSON string>
+        $ luigi --module my_tasks MyTask --tags <JSON string>
+
+    Simple example with two tags:
+
+    .. code-block:: console
+
+        $ luigi --module my_tasks MyTask --tags '{"role": "web", "env": "staging"}'
 
     It can be used to define dynamic parameters, when you do not know the exact list of your parameters (e.g. list of
     tags, that are dynamically constructed outside Luigi), or you have a complex parameter containing logically related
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by Tarrasch** [comment]
This looks cool! But I'm sceptical to this (I haven't looked at the code yet)

So far I think all parameters have a bidirectional pair of parse-serialize. This parameter type would need to be an ordereddict for that to work right?

The other thing that worries me is that if it's not better in most cases (but of course not all) to not use DictParameters but have the at least spelling-checked regular parameters, like `NotDictTask --environment prod --role hadoop`.

What do you think?

**Comment by joemeszaros** [comment]
Thanks for your feedback! 

We use `DictParameter` in our `Luigi` data flow tasks to define tags dynamically, which are then used to find the right AWS resource with the given tags. Unfortunately we do not know the exact list of these tags, because they differ for the same type of resource in different environments and scenarios. As an example we add extra tags to our resources (e.g. EMR-Hadoop cluster) in our automated data flow test environment, which must be passed to `Luigi` as well. Because the list of tags is dynamic, we are unable to create a separated parameter (like role and environment) for each and every entry in the dictionary.

An other useful case if you have logically related parameters and you do not want to flood Luigi with a bunch of parameters. My first example is the parameters of a database connection, or multiple database connections. Instead of passing user, pw, host, port, dbname [user2, pw2,...] as a single parameter, you can pass them as a single unit. It is a relatively simple example, it should be more complex examples out there :-)

**Comment by fy0** [comment]
Very useful. But how to confirm the order of  keyword won't be changed?
If hash function changed some day, old parameter and new parameter will be not equal ....
@joemeszaros

**Comment by Tarrasch** [comment]
@joemeszaros - Yea. There are reasons when this is appropriate. Can you summarize your when-to-use guide in a docstring in the code? Next person looking at `DictParameter` on readthedocs will appreciate it.

But what about the issue of serialization and parsing? We can't have it be the case that TaskIds changes because the traversal order for dictionaries are undefined. Do you understand my concern?

**Comment by joemeszaros** [comment]
Please check the latest commit. I added when-to-use guide docstring and changed `DictParameter` to use an ordered and immutable dictionary, called `FrozenOrderedDict`. Therefore the previous `frozendict` dependency was removed.

I completely understand that the string representation of the parameter must be the same, because task id contains the hash of parameters' JSON representation. Because the traversal order of standard dict is undefined, it is not guaranteed.

**Comment by Tarrasch** [comment]
The implementation looks good. I wish we could put new parameter types in their own file. But that could/should be a separate PR.

This PR however still lacks one important thing, tests. :)

**Comment by joemeszaros** [comment]
:-)

@Tarrasch Test cases implemented

**Comment by joemeszaros** [comment]
@Tarrasch Travis CI TOXENV=docs step failed because `tox` used an updated version of `Sphinx` for doc generation, which dropped error messages for files not part of my commits. It is a global problem, affecting the master branch as well. See my new PR fixing this issue:  #1581. So the CI did not fail due to the changes of this PR, rather than an updated library in the CI process.

**Comment by joemeszaros** [comment]
@Tarrasch I rebased the commits to be up-to-date with the changes in #1581 and avoid documentation generation step failure in Travis CI.

**Comment by Tarrasch** [comment]
Before fully approving and merging this. I need to double check in what sense this is an *ordered* dict.

If I do `--mydictparam {"a": "1", "b": "2"}` vs  `--mydictparam {"b": "2", "a": "1"}`, will the two values be different or parsed to the same value?

File: luigi/parameter.py [hunk 1]
```diff
@@ -699,3 +706,87 @@ def parse(self, s):
...
+          dict_param = luigi.DictParameter()
+
+            def run(self):
+                value = self.dict_param[key]
+
+    At the command line, use,
+
+    .. code-block:: console
+
+        $ luigi --module my_tasks MyTask --dict-param <JSON string>
```
Tarrasch: I think just saying `<JSON string>` is not clear for the average reader (at least not for me). Can you give an example too like `{"a": "1"}`, and also `{123: "456"}` if that were to be a valid value. You can keep this line but add another identical under it with `<JSON string>` replaced with an example value.

Tarrasch: I think just saying `<JSON string>` is not clear for the average reader (at least not for me). Can you give an example too like `{"a": "1"}`, and also `{123: "456"}` if that were to be a valid value. You can keep this line but add another identical under it with `<JSON string>` replaced with an example value.

Tarrasch:
Oh and sorry for slow and partial responses on this PR. Hope you understand. :)

joemeszaros:
Regarding ordering: Python [OrderedDict](https://docs.python.org/2/library/collections.html#collections.OrderedDict) remembers the order that keys were first inserted and does not sort the entries based on the key. It means that they will be different. I added an extra test case to highlight the fact, that DictParam ouputs the same JSON as it parsed.

I also updated DictParam docstring with a verbose example.

setup.py
```diff
@@ -38,7 +38,7 @@ def get_static_files(path):
 
 install_requires = [
     'tornado>=4.0,<5',
-    'python-daemon<3.0',
+    'python-daemon<3.0'
```
Tarrasch: In the future, do not remove these trailing commas as they have the purpose of making append-diffs easier to read. :)

Tarrasch:
Yea this looks great with me. Somebody can review within 24 hours otherwise I'll merge. :)

joemeszaros:
Cool, thanks

luigi/parameter.py
```diff
@@ -23,6 +23,13 @@
 import abc
 import datetime
 import warnings
+import json
+from json import JSONEncoder
+from collections import OrderedDict
+import collections
```
jpiper: Maybe just one import line - `from collections import OrderedDict, Mapping` ?

joemeszaros:
@jpiper Import line updated

Tarrasch:
I'll better merge this before it reaches beyond perfection. :)

PR #1577, opened 2016-03-03
Moves task id computation into a separate function in task.py

daveFNbuck:
This allows re-use of the task id logic if necessary in other places. This is
necessary because task ids need to be consistent everywhere for the same task
if we want consistency in the scheduler.

This is used in #1538


daveFNbuck:
@Tarrasch as requested, I've split this off from #1538

Tarrasch:
Thanks for splitting this off :)

PR #1579, opened 2016-03-04
Speeds up get_work with a number of small tweaks

davemt:
Some profiling of `get_work` under different scheduler-state scenarios
showed some obvious areas for improvement without major algo changes.

The biggest contributor was in `_upstream_status` where the task_stack was
being re-allocated rather than appended-to (`_upstream_status` is used in
other places as well like `task_list`).

A remaining area for optimization is the sorting of pending tasks that
happens for each call; this would be a bigger change.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -719,8 +719,8 @@ class CentralPlannerScheduler(Scheduler):
     def _used_resources(self):
         used_resources = collections.defaultdict(int)
         if self._resources is not None:
-            for task in self._state.get_active_tasks():
-                if task.status == RUNNING and task.resources:
+            for task in self._state.get_active_tasks(status=RUNNING):
+                if task.resources:
                     for resource, amount in six.iteritems(task.resources):
                         used_resources[resource] += amount
         return used_resources
@@ -799,7 +799,6 @@ class CentralPlannerScheduler(Scheduler):
         tasks.sort(key=self._rank, reverse=True)
 
         for task in tasks:
-            upstream_status = self._upstream_status(task.id, upstream_table)
             in_workers = (assistant and getattr(task, 'runnable', bool(task.workers))) or worker_id in task.workers
             if task.status == RUNNING and in_workers:
                 # Return a list of currently running tasks to the client,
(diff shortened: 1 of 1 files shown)


luigi/scheduler.py
```diff
@@ -869,13 +870,13 @@ def _upstream_status(self, task_id, upstream_status_table):
...
-                if self._state.has_task(dep_id):
-                    dep = self._state.get_task(dep_id)
+                dep = self._state.get_task(dep_id)
+                if dep:
                     if dep.status == DONE:
                         continue
                     if dep_id not in upstream_status_table:
                         if dep.status == PENDING and dep.deps:
-                            task_stack = task_stack + [dep_id] + list(dep.deps)
+                            task_stack += [dep_id] + list(dep.deps)
```
davemt: This was the most impactful line

Tarrasch:
LGTM. This is cool! Can you also show some numbers?

davemt:
Sure, I did capture timing results across each of the small changes made (they were split out originally in this branch: https://github.com/compete/luigi/tree/topics/speedup_get_work_testing). This shows the impact of each change on the slowest 5% of requests (of a few hundred tested):

| change | speedup |
|--------|---------|
| 0-master-master | 1 |
| 1-append-no-copy-badd1d5 | 3.255 |
| 2-call-get_task-less-274b6d9 | 3.516 |
| 3-kwarg-filter-get_active_tasks-076c31f | 3.514 |
| 4-fewer-calls-to-upstream_status-d26429f | 6.263 |
| 5-fewer-list-allocs-upstream_status-c42e219 | 6.529 |

The changes most impacted the slowest case requests -- the absolute slowest cases became about 10x faster and on average across all requests it was 2.7x. The requests tested covered a variety of different scenarios, but mostly focused on worst-case.

Also here's a gist showing a before/after of just how bad that single task_stack list alloc line was: https://gist.github.com/davemt/2b4fdd711baef3e6557b

davemt:
The change from 2 -> 3 was minor for the worst 5%, other factors contributed so much more that my machine's background load had more of an effect than the change. On the average across all requests it was an improvement of a few %.

erikbern:
great!!

Tarrasch:
Wow. This is truly great numbers!

Issue #1573, opened 2016-02-29
Execution summary crashes on a date parameter being None


deuxpi:
There seem to be some issues with the execution summary that I'm just starting to investigate. We're  always lagging master by a few days and this crash happened in an installation from commit dd37d973cfef2a1f93aad816c17a6e834705eb46.

This is the end of the log from a run of `luigi --local-scheduler --module tasks DailyFlow` of a wrapper task that takes a DateParameter, but uses its default parameter value which is yesterday's date. The task is pretty much:

```python
class DailyFlow(luigi.WrapperTask):
    date = luigi.DateParameter(default=datetime.datetime.utcfromtimestamp(time.time()).date() - datetime.timedelta(days=1))

    def requires(self):
        yield BuildAllDimensions()
        yield BuildAggregations(date=self.date)
```

Under the `BuildAllDimensions` task, there is at least one sub-task that has a DateParameter member that is set to `None` (which in our case makes sense and means to use the latest available data).

```
2016-02-28 23:04:05,815 DEBUG [luigi-interface] Asking scheduler for work...
2016-02-28 23:04:05,816 DEBUG [luigi-interface] Done
2016-02-28 23:04:05,816 DEBUG [luigi-interface] There are no more tasks to run at this time
2016-02-28 23:04:05,832 INFO  [luigi-interface] Worker Worker(salt=531340983, workers=1, host=***, username=azkaban, pid=31882, sudo_user=ubuntu) was stopped. Shutting down Keep-Alive thread
2016-02-28 23:04:05,839 ERROR [luigi-interface] Uncaught exception in luigi
Traceback (most recent call last):
  File "/srv/azkaban/azkaban-executor-2.5.0/executions/859/local/lib/python2.7/site-packages/luigi/retcodes.py", line 61, in run_with_retcodes
    worker = luigi.interface._run(argv)['worker']
  File "/srv/azkaban/azkaban-executor-2.5.0/executions/859/local/lib/python2.7/site-packages/luigi/interface.py", line 238, in _run
    return _schedule_and_run([cp.get_task_obj()], worker_scheduler_factory)
  File "/srv/azkaban/azkaban-executor-2.5.0/executions/859/local/lib/python2.7/site-packages/luigi/interface.py", line 198, in _schedule_and_run
    logger.info(execution_summary.summary(worker))
[...]

PR #1580, opened 2016-03-04
Prevent execution summary to fail on parameters set to None

deuxpi:
In some cases where a task parameter type that can usually be enumerated is set to some value such as None, the execution summary will fail while trying to group sequences of tasks. This change adds exception handling to prevent some tasks to make the execution summary fail.

This should fix issue #1573 . It's a change that currently allows us to continue running from the master branch. However, our practice of using None as a default value for DateParameter may simply be wrong and in that case I would recommend not to merge this branch.

### DIFF
--- a/luigi/execution_summary.py
+++ b/luigi/execution_summary.py
@@ -187,7 +187,11 @@ def _ranging_attributes(attributes, param_class):
     """
     Checks if there is a continuous range
     """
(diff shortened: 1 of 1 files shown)


Tarrasch:
Yea. You are not really meant to have a non Date-value as the default value of a DateParameter.

deuxpi:
Closing because this is only a problem with the way we use Luigi.

PR #1581, opened 2016-03-08
Fix documentation generation

joemeszaros:
# What is this PR?

Running `tox -e docs` crashes with the following error for all branches:

```
Warning, treated as error:
git/luigi/doc/api/luigi.rst:11: WARNING: Could not lex literal_block as "python". Highlighting skipped.

ERROR: InvocationError: 'git/luigi/.tox/docs/bin/sphinx-build -W -b html -d git/luigi/.tox/docs/tmp/doctrees doc doc/_build/html'
```

The problem is that `tox` installs the latest version (`1.4b1`) of Sphinx with `pip`, which is not so permissive with highlights without language specification. The default `highlight_language` is `python` in the Sphinx config file, therefore Sphinx tries to parse every code highlights without language specification using the python lexer. In some cases it stumbles upon with non-python code blocks and drops error messages breaking the doc generation.

Previous Sphinx version (`v1.4a1`) does not generate any error message.

# Solution

There are 3 possible solution:
- Use the older version of Sphinx, and leave the documentation untouched = close our eyes and postpone the problem
- Set the default `highlight_language` to `none` and leave the documentation untouched = close our eyes and postpone the problem 
- Resolve the problem and fix the documentation itself.

# Changes

This PR solves the problem with an updated documentation. Documentation changes:
- Specify language for each and every highlight block
- Use ini lexer for luigid config blocks
- Use prompt in console blocks everywhere

### DIFF
--- a/doc/central_scheduler.rst
+++ b/doc/central_scheduler.rst
@@ -25,9 +25,9 @@ The luigid server
 
 To run the server as a daemon run:
 
-::
+.. code-block:: console
 
-    luigid --background --pidfile <PATH_TO_PIDFILE> --logdir <PATH_TO_LOGDIR> --state-path <PATH_TO_STATEFILE>
+    $ luigid --background --pidfile <PATH_TO_PIDFILE> --logdir <PATH_TO_LOGDIR> --state-path <PATH_TO_STATEFILE>
 
 Note that this requires ``python-daemon``.
 By default, the server starts on AF_INET and AF_INET6 port ``8082``
@@ -59,7 +59,8 @@ When starting up,
 ``luigid`` will create all the necessary tables using `create_all
 <http://docs.sqlalchemy.org/en/rel_0_9/core/metadata.html#sqlalchemy.schema.MetaData.create_all>`_.
 
-Example configuration::
+Example configuration
+.. code:: ini
 
     [scheduler]
     record_task_history = True
--- a/doc/command_line.rst
+++ b/doc/command_line.rst
@@ -20,12 +20,12 @@ that will be installed with the pip package.
 
 Should be run like this
 
-.. code:: console
+.. code-block:: console
 
-        luigi --module my_module MyTask --x 123 --y 456 --local-scheduler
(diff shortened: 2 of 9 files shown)


erikbern:
great! will merge on passing build

joemeszaros:
Hmm,only `py34-nonhdfs` failed. Could you please take a quick look, why it failed? Here is the [log](https://travis-ci.org/spotify/luigi/jobs/114542187)

erikbern:
You should have access to everything on Travis too. I.e. there's not much I can do except re-running the test, which I just did. Maybe it's some flakiness

joemeszaros:
It seems that it was a flakiness

Tarrasch:
Wait. Why did you change configs from being on the form of `key: value` into `key=value`? Will the python configuration module even recognize it?

joemeszaros:
@Tarrasch `ConfigParser` in [Python 2]https://docs.python.org/2.7/library/configparser.html and [Python 3](https://docs.python.org/2.7/library/configparser.html) is an extended [INI parser](https://docs.python.org/3.5/library/configparser.html#supported-ini-file-structure), that accepts key-value entries in the form of `name: value` in addition to usual `key=value` INI format. So to answer your question, yes, python config module will recognize it, and now the configuration sections are in a more standard and ordinary form. 

In addition to this you need to write a custom Pygments lexer and configure Sphinx to use it, if you would like to apply highlights for config sections with `key:value` entries, because  there is no extended INI lexer in [Pygments](http://pygments.org), that allows `key:value` format, only a standard [INI lexer](http://pygments.org/docs/lexers/).

I thought writing and configuring a custom lexer adds extra complexity, compared to the applied changes, where each config section is in usual INI format and highlighted with a standard INI lexer.

Tarrasch:
Thanks for clarifying @joemeszaros. I think you did the correct thing for sure. If there's one way that's more preferred we should do that. It's a bit unfortunate that the `key:value` syntax is supported at all, as it seemed to have tricked our community for years to believe that it's the "correct" way. Thanks for making us confirm to the standard!

PR #1583, opened 2016-03-09
Refactor s3 to s3 copying

### LINKED ISSUE #1582: S3 multipart directory copy not working
The method call [here](https://github.com/spotify/luigi/blame/master/luigi/s3.py#L361) is using an incorrect method signature which means when trying to copy a folder of files with files larger than 64mb, the copy will fail. Unfortunately because of the `moto` bug that prevents the s3 multipart_copy test from working, this isn't covered in the test suite.

Also, we have two sets of threading methodologies here, where the `multipart_copy` doesn't use the same thread pool as the the copy method.

I'll try and consolidate both of these methods into one `copy` method that share a common thread pool.

jpiper:
I've simplified the threading model here, and I’ve increased the default number of simultaneous copy calls to the AWS S3 API to 100.

This should also fix #1582

I'm anticipating this might break the [s3 copy tests I put in](https://github.com/humanlongevity/luigi/commit/e1aaa128d32179ce15090b67969acb090c8bc775) due to https://github.com/spulec/moto/issues/526 - let's see.

### DIFF
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -28,11 +28,8 @@ import logging
 import os
 import os.path
 
-import sys
-import threading
 import time
 from multiprocessing.pool import ThreadPool
-from threading import Thread, BoundedSemaphore
 
 try:
     from urlparse import urlsplit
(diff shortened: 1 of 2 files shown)


jpiper:
I'll refactor the s3 copy testing to deal with the changes tomorrow. Let me know if there's anything else that you'd like changing.

luigi/s3.py
```diff
@@ -285,9 +282,9 @@ def get_as_string(self, s3_path):
 
         return contents
 
-    def copy(self, source_path, destination_path, threads=3, start_time=None, end_time=None, **kwargs):
+    def copy(self, source_path, destination_path, threads=100, start_time=None, end_time=None, **kwargs):
         """
-        Copy an object from one S3 location to another.
+        Copy object(s) from one S3 location to another. Works for individual files or
```
jpiper: I'll need to fix this docstring.

erikbern:
this looks good to me. consider also using multiprocessing if there's threading issues – i remember boto has some issues and we ended up just using multiple processes at Spotify

jpiper:
Thanks for the tip!

I was hoping I didn't need to resort to a multiprocessing pool as I worry it'll increase the overhead due to the serialization that needs to be done to interact between the processes. The threads aren't really doing any work (just waiting for AWS API to return a success) so there's no issues with the GIL here.

I've been testing this patch on AWS proper with large parquets (400 partitions with a few hundred megs for each partition) and it's been copying everything in about 1 minute using 100 threads, compared to 10 mins when using `aws s3 cp`, which appears to use 14 on my machine.

The downside to my implementation is that any errors in the copying are only raised after the entire thread pool has been joined, rather than an error copying a single file aborting the entire operation immediately. One solution to this is to add a callback to the `apply_async` that processes the return code as each thread complete and then immediately halts the entire copy. I might also just switch to using `concurrent.futures` module if I have more time :)

dlstadther:
I appreciate the refactoring and optimization here! That was my first real (non-academic) experience implementing anything multithreaded. Thanks again!

erikbern:
afaik threading.Pool and multiprocessing.Pool has identical interfaces so it should be somewhat easy to swap out one for another if there are any issues.

the problem with multiprocessing is that forking can be very memory consuming if it happens after a long chain of work has already been done – memory is copy-on-write but with Python's GC etc my experience is you end up consuming a ton of memory. so if threading works then that's probably preferable

jpiper:
Hmm that's odd - the copy tests passed but the build then failed on something else completely unrelated - `INETLuigidDaemonServerTest`. I've made some tweaks to address it, but If I still don't have any luck soon then I'll try a `concurrent.futures` implementation to see if that helps.

I also want to add in a s3 copy test for a directory of files before we merge this.

jpiper:
Provided the build goes through, I think we should be good to go here. Unfortunately I still can't test the multipart copy as the bug in moto still exists, but I can confirm from empirical testing that it's working in the field properly.

Tarrasch:
Didn't look into this, but it seem reviewed enough for a contrib module change. Great job! :)

PR #1584, opened 2016-03-09
Add enbrite.ly to companies that use Luigi

joemeszaros:
enbrite.ly is using luigi for more than 1 year and we are happy with it :-)


joemeszaros:
thanks

PR #1586, opened 2016-03-09
Improved Salesforce Reliability

dlstadther:
Made use of python TempFile rather than keeping (potentially) large lists of lists of OrderedDicts in memory.

Moved parse_results() and _parse_results() to be global functions. Call these functions on each page of salesforce results within query_all(). Seek to beginning of tempfile and copytoobj() then close tempfile (thus deleting it and cleaning up memory).

Regex now matches case insensitive when extracting field names from soql query.

Recursive get_all_responses() inner function has been replaced by the iterative equivalent (deep recursion was causing a segmentation fault).

### DIFF
--- a/luigi/contrib/salesforce.py
+++ b/luigi/contrib/salesforce.py
@@ -21,6 +21,9 @@ import xml.etree.ElementTree as ET
 from collections import OrderedDict
 import re
 import csv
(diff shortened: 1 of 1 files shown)


dlstadther:
Broken.......

PR #1587, opened 2016-03-10
Update spark_als.py

oluies:
Should be opened as "self.output().open('w')"

https://groups.google.com/forum/#!topic/luigi-user/tfpe67ChFco


Tarrasch:
Did you test that it works? If so lets merge.

oluies:
still having other problems, it the first time I am running luigi ! 

Can you point me to an example /etc/luigi/client.cfg ?

PR #1588, opened 2016-03-10
correct example spark_als

oluies:


Tarrasch:
Were you able to run it this time? :)

Thanks for doing this. Clearly that example file was poorly maintained. This is very appreciated.

oluies:
Yes, no idea if the output is correct, but at least it does run

Tarrasch:
Well. That's a clear improvement at least. Thanks for contributing. :)

PR #1593, opened 2016-03-14
Fix incorrect usage of Redshift/PostgresQuery 'query' property.

dlstadther:
'query' was defined as a property within rdbms.py, but referenced and used (within postgres.py and postgres_test.py) as if it were a method.

I believe it is more appropriate for query to be a property than a method.

Using PostgresQuery or RedshiftQuery is as easy as ...

```python
from luigi import configuration
from luigi.contrib import redshift

class QueryExample(redshift.RedshiftQuery):
    query = luigi.Parameter()
    table = luigi.Parameter()

    host = config.get('redshift', 'host')
    database = config.get('redshift', 'database')
    user = config.get('redshift', 'user')
    password = config.get('redshift', 'password')
    aws_access_key_id = config.get('s3', 'aws_access_key_id')
    aws_secret_access_key = config.get('s3', 'aws_secret_access_key')
```

### DIFF
--- a/luigi/postgres.py
+++ b/luigi/postgres.py
@@ -362,7 +362,7 @@ class PostgresQuery(rdbms.Query):
     def run(self):
         connection = self.output().connect()
         cursor = connection.cursor()
-        sql = self.query()
+        sql = self.query
 
         logger.info('Executing query from task: {name}'.format(name=self.__class__))
         cursor.execute(sql)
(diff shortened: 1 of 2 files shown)


erikbern:
just a minor note: you get syntax highlighting for free if you put "python" after the three backticks in Github's markdown (i edited your comment and added that)

dlstadther:
@erikbern Thanks for the tip!

dlstadther:
I may be mistaken, but the two tests that fail don't appear to be anything on my end of things. Please correct me if i'm wrong.

Tarrasch:
Yes this looks sane and the failures surely looked spurious.

PR #1596, opened 2016-03-15
attempt to give luigi s3 module assume role support

Fsero:


erikbern:
no idea what this does so great if someone else can weigh in :)

Fsero:
@erikbern basically the current way to give credentials to luigi to write in S3 is using access and secret key, this method doesn't leverage other options more suited to production environments like IAM ROLES, this patch introduces two options in the configuration aws_role_arn and aws_role_name which is the needed information to assume the role.

The assume the role part means ask for ephemeral access and secret keys (this is why i need to introduce the token variable which identifies the transaction, more info [here](http://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html)).

Because of this i needed to change the boto call from the connect_to_s3 wrapper function to the real object, this of course doesnt break the way it was before but leverage people to use credentials obtained via a role which is really convenient if you want to maintain a service.

In our setup, luigi server and luigi workers ec2 instance are tied to a role which the minimum permissions and can assume other roles which give them the ability to read _SUCCESS files in the right path.

If you need more info or feel that the patch miss something im happy to complete it, but i would be happier if this gets merged so i don't need to maintain a fork :)

Tarrasch:
@Fsero, Can you blame the files you're touching and at-mention them here in this PR? It's good for everyone if somebody who knows s3 reviews this.

Fsero:
@ddaniels888 @steenzout @dlstadther @gpoulin @themalkolm @jpiper @tommyengstrom you commited something in this file in the past, maybe you can help

vicentgodella:
I think this is very useful for production environments. IMHO, using a role it's better than key and secret.

When you say aws_role_name, I guess you are saying aws_role_session_name, right?

Fsero:
@vicentgodella yes i thought this was obvious and i prefer short names, im ok changing that, thanks fior the feedback

luigi/s3.py
```diff
@@ -77,22 +77,41 @@ def __init__(self, aws_access_key_id=None, aws_secret_access_key=None,
...
         options = self._get_s3_config()
         options.update(kwargs)
         # Removing key args would break backwards compability
-        if not aws_access_key_id:
-            aws_access_key_id = options.get('aws_access_key_id')
-        if not aws_secret_access_key:
-            aws_secret_access_key = options.get('aws_secret_access_key')
-        for key in ['aws_access_key_id', 'aws_secret_access_key']:
+        role_arn = options.get('aws_role_arn')
+        role_session_name = options.get('aws_role_session_name')
```
dlstadther: If I don't include `aws_role_arn` and/or `aws_role_session_name` within my config, am I correct that line 86 and 87 will throw errors (something to the effect of 'variable doesn't exist in luigi.cfg')?

Fsero: i think it would return the default value which in this case is None, am i wrong?

dlstadther:
I'm fine with this change so long as it doesn't break/warn/etc. users who wish to use only token/pass from luigi.cfg.

I can't say whether it works. So if you can say that your code works for your use-case as well as for the old usage, then lgtm.

Fsero:
@dlstadther pass the tests, and i tried both cases and it works.

Fsero:
@erikbern Wdyt? can we merge?

luigi/s3.py
larsga: Those two lines will return `None`.

erikbern:
doesn't seem to pass unit tests – can you take a look?

Fsero:
Sure! but the only errors i see seem unrelated to my change,

check https://travis-ci.org/spotify/luigi/jobs/116349379#L9983 

and 
`
Failed connecting to remote scheduler 'http://zorg.com'
  File "/home/travis/build/spotify/luigi/luigi/rpc.py", line 127, in _fetch
  File "/opt/python/3.3.5/lib/python3.3/unittest/mock.py", line 846, in __call__
    return _mock_self._mock_call(*args, **kwargs)
  File "/opt/python/3.3.5/lib/python3.3/unittest/mock.py", line 906, in _mock_call
    raise result

`

so forcing a rebuild and see how it goes

erikbern:
ok weird. i already rerun the build a few times before i asked.

let's hope it works now!

Fsero:
@erikbern fixed the test! turns out i forgot to close the sts mock in the tests and this affected other tests! please merge :)

erikbern:
great!

PR #1598, opened 2016-03-16
Support inverted arguments for BoolParameter

tdyas:
Add two command-line arguments for BoolParameter's instead of one, specifically one to set it to `True` (e.g, `--foo`) and another (`--no-foo`) to set it to `False`. This is necessary because `BoolParameter` currently uses `argparse`'s "store_true" action, which does not take a parameter. This change allows `BoolParameter`'s to have a default of `True` and still be settable to `False`.

For cases where an inverted argument does not make sense, the caller should set `omit_inverted_argument=True` on the `BoolParameter`. (For example, this has been done for the `--help` and `--help-all` options, as `--no-help` is of no help at all.)

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -188,7 +188,7 @@ class Parameter(object):
         if cp_parser:
             dest = self._parser_global_dest(param_name, task_name)
             found = getattr(cp_parser.known_args, dest, None)
-            yield (self._parse_or_no_value(found), None)
+            yield (self.parse(found) if found is not None else _no_value, None)
         yield (self._get_value_from_config(task_name, param_name), None)
         yield (self._get_value_from_config(task_name, param_name.replace('_', '-')),
                'Configuration [{}] {} (with dashes) should be avoided. Please use underscores.'.format(
@@ -258,27 +258,17 @@ class Parameter(object):
         """
         return None
 
-    def _parse_or_no_value(self, x):
-        if not x:
-            return _no_value
-        else:
-            return self.parse(x)
-
     @staticmethod
     def _parser_global_dest(param_name, task_name):
         return task_name + '_' + param_name
 
-    @staticmethod
-    def _parser_action():
-        return "store"
-
     def _parser_add_arguments(self, parser, flag_name, dest, help):
         """
-        Updates the given parser with an argument to set this Parameter.        
+        Updates the given parser with an argument to set this Parameter.
         """
-        parser.add_argument(flag_name,
+        parser.add_argument('--{}'.format(flag_name),
                             dest=dest,
-                            action=self._parser_action(),
+                            action='store',
                             help=help)
 
 
@@ -525,10 +515,14 @@ class BoolParameter(Parameter):
     """
     A Parameter whose value is a ``bool``. This parameter have an implicit
     default value of ``False``.
+
+    :param bool omit_inverted_argument: specify ``True`` if the parameter should not have
+                                        the inverted argument --no-NAME created. Default: ``False``.
     """
 
-    def __init__(self, *args, **kwargs):
-        super(BoolParameter, self).__init__(*args, **kwargs)
+    def __init__(self, omit_inverted_argument=False, **kwargs):
+        super(BoolParameter, self).__init__(**kwargs)
+        self._omit_inverted_argument = omit_inverted_argument
         if self._default == _no_value:
             self._default = False
 
@@ -542,9 +536,18 @@ class BoolParameter(Parameter):
         # coerce anything truthy to True
         return bool(value) if value is not None else None
 
-    @staticmethod
-    def _parser_action():
-        return 'store_true'
+    def _parser_add_arguments(self, parser, flag_name, dest, help):
+        parser.add_argument('--{}'.format(flag_name),
+                            dest=dest,
+                            action='store_true',
+                            default=None,
+                            help=help)
+        if not self._omit_inverted_argument:
+            parser.add_argument('--no-{}'.format(flag_name),
+                                dest=dest,
+                                action='store_false',
+                                default=None,
+                                help=help)
 
 
 class BooleanParameter(BoolParameter):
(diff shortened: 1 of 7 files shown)


andy256:
This is a great idea!

Tarrasch:
This implementation looks nice and clean. Just some minor inline comments:

luigi/parameter.py
```diff
@@ -537,6 +550,19 @@ def normalize(self, value):
     def _parser_action():
```
Tarrasch: Can this be deleted now? Is `_parser_action()` still used anywhere?

Tarrasch:
The reason I like this patch is that I've by now repeatedly told other PR senders that "No, you're not allowed to have True as a default for your BoolParameter". But with this patch we could allow it from now on.

luigi/parameter.py
tdyas: Seems like it can be since the only place it is overridden is in BoolParameter which does not even call `_parser_action` any more in `_parser_add_arguments`.

Tarrasch: Ok. Please delete the dead code then :)

Tarrasch:
This is a reasonably big change (not in code size), but in what we communicate are good practices to the luigi users. I remember @erikbern saying some time that BoolParameters should always be false by default. But I don't know if he meant that as a "that's good design in general" or "due to current luigi limitations that's the only implementation that works out".

tdyas:
Understood. I agree that existing community usages may have a part to play in whether this PR should be accepted or not.

As an anecdotal data point, at my employer (Foursquare), we use Luigi and I have seen our developers confused on multiple occasions by not really being able to set `default=True` on BoolParameter. (The confusion stemming from the BoolParameter` behaving differently from every other `Parameter` subclass in Luigi.)

tdyas:
And this PR is an attempt to improve the API to remove that source of confusion.

a1k0n:
I agree with @tdyas on this, as someone who used Luigi heavily while working with @erikbern . It doesn't break the existing semantics, and it does exactly what I had assumed it would have done when I first started using Luigi: adds a --no-xyz argument, which is a de-facto standard for GNU tools and elsewhere.

The only reason `default=True` was a bad practice is because it lacked this implementation, and we had to do dumb hacks like create argument variables named `no_this` and `disable_that`.

Tarrasch:
Thank you both @tdyas and @a1k0n for sharing your design thoughts, they sound sane to me. I'll just give @erikbern until tomorrow to chime in. Otherwise I'll merge this tomorrow once you fix the build.

erikbern:
i'm fine with this

tdyas:
Adding the --no-xyz arguments seems to confuse argparse's `parse_known_args` method. I'll have to dig into why argparse is confused.

```
runtests.py: error: ambiguous option: --n could match --no-scheduler-record-task-history, --no-lock, --no-worker-no-install-shutdown-handler, --no-no-lock, --no-RangeBase-reverse, --no-worker-count-uniques, --no-scheduler-prune-on-get-work, --no-RangeDaily-reverse, --no-local-scheduler, --no-worker-retry-external-tasks, --no-RangeHourlyBase-reverse, --no-take-lock, --no-RangeDailyBase-reverse, --no-scheduler-prune-done-tasks, --no-RangeHourly-reverse, --no-parallel-scheduling, --no-TestNotificationsTask-raise-in-complete, --no-worker-keep-alive, --no-assistant
```

erikbern:
i'm guessing no argument can be a prefix of another

tdyas:
So I dug in further, and it seems like a bug in argparse. The error occurs during the call to `parse_known_args` in `CmdlineParser.__init__`. That method is trying to determine the `root_task` to use. argparse should really not throw an error about an ambiguous option when invoked via parse_known_args. It should just mark the option in question as unknown.

I can get the failing test to pass by renaming the `n` option in SomeTask in cmdline_test.py to something else (e.g., `z`).

This still brings up an issue about whether this will break Luigi for anybody who has a task with a "n" parameter.

tdyas:
And this issue is known to the Python developers: https://bugs.python.org/issue19814

And apparently only warrants a warning in the documentation.

Tarrasch:
Please let me know if you need any help. Otherwise I'm patiently waiting for something that builds. :)

tdyas:
As it turns out, the test I added wasn't actually testing whether a `BoolParameter` with `default=True` could be set to `False` by option. This uncovered some places where checks for truthy values had to actually check for `None`. 

I also merged all of the other commits into the commit adding the inverted argument support.

tdyas:
There was also a test failure locally for me with `test/file_test.py` which seems unrelated to my change since the tests in that file don't deal with parameters. It could be an issue with my local setup so will wait for the Travis build to see if that is the case.

tdyas:
CI tests passed.

Tarrasch:
I'm still a bit worried about we suddenly dropping support for parameters with names like `n` and I'm not fond of you renaming a local variable to an all caps name (`n` --> `N`). Would it not just be much easier if you only had the `--no-boolparam` option only if `default=True`? Alternatively you could another option `add_negative_cli_flag` with a value `False` as default.

Tarrasch:
Yea I understand this is entirely the fault of some stupid bug in the argparse module (I've also felt the pain). But I'm afraid people are going to start to say their stuff broke and this would need to be reverted in haste if we break compatibility.

erikbern:
apparently python 3.5 added a flag to argparse `allow_abbrev=False` that turns off the prefix matching. too bad it wasn't earlier

the correct way would probably be to first change luigi's behavior so that prefix flags throw a deprecation warning for a few months and then an exception

btw reminds me – maybe it's time to publish a new version of luigi to pypi?

tdyas:
So the solution for now it seems is to just add a `--no-FOO` option for `BoolParameter`'s with `default=True`?

Tarrasch:
Yea. I would be more comfortable that way actually @tdyas

dlstadther:
Resolved by #2427

PR #1600, opened 2016-03-17
Salesforce - Improve Matching and Remove Unnecessary Complication.

dlstadther:
Don't make every response an ordereddict. The order of the dictionary fields don't matter here.What matters is that the list of dicts is in the appropriate order if one specifies an order by within their SOQL. Since lists maintain order, this is good."

Rather than checking if the value is a string, check that it's not a data structure (by checking just for basestring, we miss ints and floats).


dlstadther:
Sorry for not catching this is previous salesforce PRs. Salesforce is a pain to work with and my previous SOQLs only returned strings.

PR #1601, opened 2016-03-18
Luigi tasks for google dataproc

constantijn:
Hi All,

I did  project with PySpark and Luigi and google Dataproc.
I wrote some Luigi tasks to work with Dataproc and I figured i'd share them, so here's the pull request :)


Tarrasch:
Wow, looks really good! Some more documentation on how to use the base class would be good but I'm fine with merging as-is. You can append more docs later if you have time.

luigi/contrib/dataproc.py
```diff
@@ -0,0 +1,208 @@
...
+            cluster_status = self._get_cluster_status()
+            status = cluster_status['status']['state']
+            logger.info("Creating new dataproc cluster: {} status: {}".format(self.dataproc_cluster_name, status))
+            if status == 'RUNNING':
+                break
+            if status == 'ERROR':
+                raise Exception(cluster_status['status']['details'])
+
+
+class DeleteDataprocClusterTask(_DataprocBaseTask):
```
Tarrasch: Wait. How is deletion supposed to be used?? I think of luigi a bit like GNU Make. In what kind of chain of tasks do deletion appear?

luigi/contrib/dataproc.py
```diff
@@ -0,0 +1,208 @@
...
+    return _dataproc_client
+
+
+def set_dataproc_client(client):
+    global _dataproc_client
+    _dataproc_client = client
+
+
+class _DataprocBaseTask(luigi.Task):
+    gcloud_project_id = luigi.Parameter(significant=False)
```
Tarrasch: I think also want `positional=False`.

luigi/contrib/dataproc.py
```diff
@@ -0,0 +1,208 @@
...
+    dataproc_cluster_name = luigi.Parameter(significant=False)
+    dataproc_region = luigi.Parameter(default="global", significant=False)
+
+    dataproc_client = get_dataproc_client()
+
+
+class DataprocTask(_DataprocBaseTask):
+    """ Base task for running jobs in Dataproc """
```
Tarrasch: Actually you should add docs on how to use this. I'm not really sure what this is it for even. Is it spark only?

luigi/contrib/dataproc.py
constantijn: It allows you to create chains that create a cluster, run a big daily job and then throw away the cluster. (With the actual input and output data for the job in gs://... instead of hdfs://...)

Then fire off that chain once a day with a curl in a cron job (or however you like) and you can run your big daily job once a day while keeping the costs of running the cluster to a minimum.

luigi/contrib/dataproc.py
constantijn: For now you can take a peek at the test to see how it's used, but i agree it needs some docs.

It's not spark only, submit_job lets you submit whatever you want. The spark and pyspark jobs are convenience wrappers around submit job.

I can add convenience wrappers for hadoop, hive and pig also by wrapping the API for submititng those jobs, however i dont actually use those tools myself

luigi/contrib/dataproc.py
Tarrasch: But I don't see this tasks place anywhere in a dependency tree. Like who would require this task and what tasks does it require normally?

Can this whole bring-up bring-down be done in a context manager?

luigi/contrib/dataproc.py
Tarrasch: The docs should at least say that "Currently XYZ are not implemented, but we believe here is the natural place for it.". Because anyone browsing the docs online would probably be confused.

luigi/contrib/dataproc.py
constantijn: You can extend the delete task to require a DataprocTask that does your calculation, then have that task require a cluster creation task.

What ends up running is: CreateCluster -> DataprocTask (do work on said cluster) -> Delete cluster.

However Since your dataproc task will need to be implemented to contain your own job that you actually want to run on the cluster I can't hard code that type of dependency chain in here.

constantijn:
How's this?
Added more documentation and split out the different kinds of Dataproc jobs into separate Luigi tasks so that people can just run basic tasks without extending anything.

Let me know if you think this is the way to go and i'll add the non-spark Dataproc job types.

Tarrasch:
Yea. This is good.

Only add non-spark job types if you want and have time and all that. :)

PR #1602, opened 2016-03-18
Fix ftp exists(...)

dlstadther:
Fixed _ftp_exists(...) method. Previously called nlst(path) on full path including filename and extension. This only worked if the file existed.

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -127,10 +127,14 @@ class RemoteFileSystem(luigi.target.FileSystem):
         return exists
 
     def _ftp_exists(self, path, mtime):
+        path_parts = path.split('/')
+        path = '/'.join(path_parts[:-1])
+        fn = path_parts[-1]
+
         files = self.conn.nlst(path)
 
         exists = False
-        if files:
+        if fn in files:
             if mtime:
                 mdtm = self.conn.sendcmd('MDTM ' + path)
                 modified = datetime.datetime.strptime(mdtm[4:], "%Y%m%d%H%M%S")


erikbern:
lgtm

luigi/contrib/ftp.py
```diff
@@ -127,10 +127,14 @@ def _sftp_exists(self, path, mtime):
         return exists
 
     def _ftp_exists(self, path, mtime):
+        path_parts = path.split('/')
+        path = '/'.join(path_parts[:-1])
+        fn = path_parts[-1]
```
Tarrasch: Does this assignment make sense?? You seem to override it right afterwards in the loop no?

dlstadther: I'm not sure I understand you. There is no loop within this method (at least not directly). Line 137 is checking to see if the filename exists in the list of files which was the result of `nlist(path)`.

Tarrasch: Ah, I read `if fn in files` as `for fn in files`. Sorry, still morning over here ^^

andreabedini:
hello! this patch seems to break ftp for me. The problem is that `nlist(path)` returns a list of paths starting with `path`. E.g.

```python
In [40]: conn.nlst()
Out[40]: ['public', 'pub', 'ietf', 'apnic', 'uploads', 'welcome.msg']

In [41]: conn.nlst('pub/stats')
Out[41]:
['pub/stats/lacnic',
 'pub/stats/ripe-ncc',
 'pub/stats/iana',
 'pub/stats/afrinic',
 'pub/stats/apnic',
 'pub/stats/arin']
```

so `if fn in files` will always be false since `fn` is only the filename.

Also, `conn.sendcmd('MDTM ' + path)` will fail if path is a not a regular file.

PR #1603, opened 2016-03-22
Fix pruning in the scheduler

Tarrasch:
See individual commits. :)


Tarrasch:
And know I'm running the "assistants bugfix" commit in production. Let's see if everything is still up and running tomorrow. ^^

The loading time have gone down from around 20 seconds to subseconds.

erikbern:
LGTM

Tarrasch:
Finally I feel that the assistants works on the small scale that I'm currently using luigi. Hopefully others dares to try this feature out too soon so we can make it fully stable.

PR #1487, opened 2016-01-07
execution summary: Don't crash on uncomparables

Tarrasch:
This has become increasingly important since spotify/luigi#1479 where
EnumParameters have been introduced (which are normally not sortable)

PR #1606, opened 2016-03-24
EnumParameter docs: Use Enum not IntEnum

Tarrasch:
I think at some point IntEnum was preferable as you wouldn't run into the bug fixed by #1487, but that's not relevant any more. Let's change this example to avoid creating a misconception that you need to use IntEnums.


Tarrasch:
@jimjh, is there any other reason why you choose IntEnum in the docs before?

jimjh:
No reason! It was on a whim.

Issue #1608, opened 2016-03-25
Parallel scheduling can't handle externalized dependencies


daveFNbuck:
If you do something like
```
def requires(self):
    return luigi.task.externalize(SomeTask())
```
in a Task and then schedule it with the `--parallel-scheduling`, the scheduling will hang forever. The cause is that externalize sets `task.run = NotImplemented` and `NotImplemented` is not serializable so it can't be put in the multiprocessing queue.

My instinct is to fix this by just setting `task.run = None` or some other picklable value and treating that like NotImplemented in worker, but I'm afraid this will have weird side effects and leave the bug open for people implementing their own externalize functionality. Any other ideas?

Also, there's no indication that anything bad happened when this bug is triggered. We should find a way to catch this type of error and kill scheduling when it happens.

erikbern:
sgtm – we could just change `NotImplemented` to `None`

daveFNbuck:
If we just change NotImplemented to None, won't that potentially break user code?

erikbern:
I think it's fine. We should probably check both cases in worker.py just in case someone has old code. But off the top of my head I don't see why it would break any code

metra:
Would it be possible to log any pickling exceptions to make it obvious that there was an error queuing up the task?

PR #1609, opened 2016-03-25
Sets task.run to None instead of NotImplemented in ExternalTasks

daveFNbuck:
Using NotImplemented prevents externalized tasks from being pickled, preventing
the use of parallel scheduling due to pickle being the serializer used in
multiprocessing. This uses the picklable None instead, resolving #1608

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -79,6 +79,10 @@ fork_lock = threading.Lock()
 _WAIT_INTERVAL_EPS = 0.00001
 
 
+def _is_external(task):
+    return task.run is None or task.run == NotImplemented
+
+
 class TaskException(Exception):
     pass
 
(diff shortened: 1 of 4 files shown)


luigi/worker.py
```diff
@@ -157,7 +161,7 @@ def run(self):
             t0 = time.time()
             status = None
 
-            if self.task.run == NotImplemented:
+            if is_external(self.task.run):
```
Tarrasch: This looks wrong? No?

luigi/worker.py
```diff
@@ -79,6 +79,10 @@
 _WAIT_INTERVAL_EPS = 0.00001
 
 
+def is_external(task):
```
Tarrasch: Maybe make it _private. Did you also consider putting it in task.py? No strong opinions though.

Tarrasch:
This makes senses to me.

luigi/worker.py
daveFNbuck: Yeah, should be self.task. Fixing.

luigi/worker.py
daveFNbuck: Done

Tarrasch:
Thank you :)

PR #1612, opened 2016-03-29
Luigi return codes

fabriziodemaria:
If completeness verification for a dependency fails with an exception, the corresponding task for the dependency is scheduled with status UNKNOWN.
Downstream tasks' status would be: "upstream_unverifiable_dependency".
Luigi return code can be set for this particular case.

EDIT: PR extended to cover other cases other than complete() check.


test/execution_summary_test.py
```diff
@@ -95,6 +95,53 @@ def requires(self):
...
+        self.assertFalse(d['still_pending_ext'])
+        summary = self.summary()
+        expected = ['',
+                    '===== Luigi Execution Summary =====',
+                    '',
+                    'Scheduled 2 tasks of which:',
+                    '* 1 completeness verification failed:',
+                    '    - 1 Bar()',
+                    '* 1 were left pending, among these:',
+                    '''    * 1 had dependencies whose completeness couldn't be verified:''',
```
Tarrasch: Did you consider using `"strings"` instead of `'''strings'''`?

doc/configuration.rst
```diff
@@ -492,6 +493,9 @@ missing_data
   For when an :py:class:`~luigi.task.ExternalTask` is not complete, and this
   caused the worker to give up.  As an alternative to fiddling with this, see
   the [worker] keep_alive option.
+complete_failed
+  For signaling that the verification of the completeness of a dependency
+  failed. Typically because some exception have been raised.
```
Tarrasch: Maybe just `For exceptions in the complete() method`? It doesn't matter if it's a dependency of any task, does it?

Also, how does this relate to "unhandled_exception" which already mentions  ``complete()`` method?

luigi/execution_summary.py
```diff
@@ -349,7 +361,10 @@ def _summary_format(set_tasks, worker):
             str_output += "    - {0} ran {1} tasks\n".format(ext_worker, len(task_dict))
             count += 1
         str_output += '\n'
-    if num_all_tasks == len(set_tasks["already_done"]) + len(set_tasks["still_pending_ext"]) + len(set_tasks["still_pending_not_ext"]):
+    if num_all_tasks == sum([len(set_tasks["already_done"]),
+                             len(set_tasks["unknown_completeness"]),
+                             len(set_tasks["still_pending_ext"]),
+                             len(set_tasks["still_pending_not_ext"])]):
```
Tarrasch: :+1:

Tarrasch:
Code looks good. But I want to grind on the docs a bit. :)

In particular it feels like this PR uses the term "completeness verification". That's a sensible name but we already use the term "complete() check" everywhere else. Can you change to use the latter?

oh and also, hi Fabrizio :)

fabriziodemaria:
Hello to you Arash :)

Prior to this PR, the exceptions in the `complete()` method are swallowed in Luigi and the final exit code is 0 (not the one defined for "unhandled_exception"). This PR is intended to fix that, but adding a specific return code for "complete_failed" would indeed reduce the scope of "unhandled_exception". Would that be ok?

doc/configuration.rst
```diff
@@ -481,6 +481,7 @@ We recommend that you copy this set of exit codes to your ``luigi.cfg`` file:
   # They are in increasing level of severity (for most applications)
   already_running=10
   missing_data=20
+  complete_failed=20
```
Tarrasch: Perhaps use a different number? I suggest 41. 41 because it's somewhat similar to "complete_failed" (has number 40).

Tarrasch:
> adding a specific return code for "complete_failed" would indeed reduce the scope of "unhandled_exception". Would that be ok?

Absolutely. If the docs were incorrect you should update it. Thanks for catching this. :)

I also notice that there is no test for `failure in complete()` in https://github.com/spotify/luigi/blob/master/test/retcodes_test.py. Can you add a test case there? Also, in the tests in this PR as of now, there are still no test cases checking the return code right?

luigi/worker.py
```diff
@@ -473,11 +473,11 @@ def _validate_task(self, task):
             raise TaskException('Task of class %s not initialized. Did you override __init__ and forget to call super(...).__init__?' % task.__class__.__name__)
 
     def _log_complete_error(self, task, tb):
-        log_msg = "Will not schedule {task} or any dependencies due to error in complete() method:\n{tb}".format(task=task, tb=tb)
+        log_msg = "Will not run {task} or any dependencies due to error in complete() method:\n{tb}".format(task=task, tb=tb)
         logger.warning(log_msg)
 
     def _log_dependency_error(self, task, tb):
-        log_msg = "Will not schedule {task} or any dependencies due to error in deps() method:\n{tb}".format(task=task, tb=tb)
+        log_msg = "Will not run {task} or any dependencies due to error in deps() method:\n{tb}".format(task=task, tb=tb)
```
Tarrasch: Just checking: You did this to use more user-understandable terminology over luigi-accurate terminology. Right?

test/retcodes_test.py
```diff
@@ -77,6 +77,50 @@ def new_func(*args, **kwargs):
...
+
+        self.run_and_expect('RequiringTask', 4)
+
+    def test_task_limit(self):
+        class TaskB(luigi.Task):
+            def complete(self):
+                return False
+
+        class TaskA(luigi.Task):
+            def requires(sefl):
```
Tarrasch: self*

Tarrasch:
The new changes to this PR are very interesting. Let's make sure we get it right. As I see, this is the first time the Worker will tell that a task has UNKNOWN status to the scheduler. Previously the scheduler would consider any dependency it has no info about (because it's been pruned) as UNKNOWN. We now introduce another way for a task to end up in UNKNOWN state.

Have you thought of the implications of this? Does the output make sense in the scheduler? I'm guessing it'll be much better with this change, because for instance when task-limit is reached, the scheduler task graph UI will hint "ya, something is wrong".

I just want make sure that you @fabriziodemaria have thought of the consequences of this patch. But it overall looks good to me.

luigi/execution_summary.py
```diff
@@ -248,12 +255,14 @@ def _get_comments(group_tasks):
...
     "completed",
     "failed",
+    "unknown",
     "still_pending",
     "still_pending_ext",
     "run_by_other_worker",
     "upstream_failure",
     "upstream_missing_dependency",
     "upstream_run_by_other_worker",
+    "upstream_unverifiable_dependency",
```
Tarrasch: This is called "unknown"/"upstream_unverifiable_dependency". That pair doesn't really match. Maybe rename to "scheduling_error"/"upstream_scheduling_error"?

luigi/worker.py
ulzha: As far as I understand, it does schedule the task now, with state UNKNOWN, so it's accurate.

Tarrasch: Ah. I see. Yes. Now it's both simple language and still technically accurate. Great!

fabriziodemaria:
> I just want make sure that you @fabriziodemaria have thought of the consequences of this patch. But it overall still looks good to me.

@Tarrasch I will investigate more the implications of these changes in the following weeks (I don't have much time to work on this at the moment). In the meantime, this code will run on testing pipelines to see if anything breaks.

luigi/retcodes.py
```diff
@@ -30,19 +30,29 @@ class retcode(luigi.Config):
...
     task_failed = IntParameter(default=0,
                                description='''For when a task's run() method fails.''',
                                )
+
+    # default value inconsistent with doc/configuration.rst for backwards compatibility reasons
     already_running = IntParameter(default=0,
                                    description='For both local --lock and luigid "lock"',
                                    )
 
+    scheduling_error = IntParameter(default=35,
```
ulzha: Ah, wait, if we want backwards compatibility, then this default should be 0 like the other inconsistent ones.

test/scheduler_visualisation_test.py
```diff
@@ -292,7 +292,7 @@ def test_dep_graph_missing_deps(self):
 
         fail = dep_graph[BadReqTask(succeed=False).task_id]
         self.assertEqual(fail[u'name'], 'BadReqTask')
-        self.assertEqual(fail[u'params'], {'task_id': BadReqTask(succeed=False).task_id})
+        self.assertEqual(fail[u'params'], {'succeed': 'False'})
```
fabriziodemaria: I think it is good to have information about the passed parameters for UNKNOWN tasks in the dependency graph. However, if a task gets status UNKNOWN for reasons different from the ones managed in this PR, then the code from https://github.com/spotify/luigi/pull/1569 is most likely executed, and the _task_id_ would be displayed in the dependency graph instead of the afore-mentioned parameters: when this is the case, If we want a consistent behaviour I would suggest not to display anything else apart from the task name. Such case should be covered with a new test as well?

Tarrasch: I think we should display as much as we can, even if then the two kinds of UNKNOWN will give an inconsistent amount of information when you mouse-hover over them.

Tarrasch:
@ulzha and @fabriziodemaria - about your commit:

> work around erroring (order sensitive) test 

Have you tried using the LuigiTestCase mechanism? Alternatively setting the `task_namespace = __name__` in the class definitions? Yea we should introduce automatic `__name__` namespacing. :)

ulzha:
LGTM.

Tarrasch:
I would like to audit (post-merge review) this change set a bit:

 * How was this tested?
 * How does thing look in the scheduler dependency graph if you run into the execution limit?
 * This looks a good improvement increasing the quality of luigi. What real-world use-case did you guys have to do this?

ulzha:
1. Just with the included tests, essentially.

2. Do you mean the worker-task-limit? image

3. To plug Luigi into a scheduler that is more success-aware than Cron, and communicate success via return code. "Success" being defined brutally as meaning that the "root task" is complete.

Tarrasch:
Thanks! This is clearer now. A cron replacement would be really cool! :)

Issue #1607, opened 2016-03-24
Dynamic Requires incorrectly treats all luigi.Parameter as strings


dlstadther:
As noted by the title, dynamic requirement Parameters are treated as strings (where the required class is expected luigi.Parameter()). _DateParameter_, _IntParameter_, etc are treated appropriately.

This issue was found when dynamically requiring an arbitrary number of _redshift.S3CopyToTable_ statements to temporary tables. _Columns_ and _queries_ are passed as lists, but errors were being throw because these lists were being interpreted as strings.

I have written the following luigi code to note this error (I have noted the actual output in both cases - normal requires, as well as dynamic requires):

```python
import datetime

import luigi
from luigi import mock

__author__ = "Dillon Stadther"
__date__ = '2016-03-24'


class TestTask(luigi.Task):
    date = luigi.DateParameter(default=datetime.date(2016, 03, 24))

    tmp_path = luigi.Parameter(config_path=dict(section='path', name='tmp_path'))
    table_path = luigi.Parameter(config_path=dict(section='path', name='table_path'))

    #def requires(self):
    #    yield TestClass(
    #        date=self.date,                # prints "2016-03-24" and "2016-03-17"
    #        string_in='Hello World',       # prints "Hello World" and "H"
    #        list_in=['Hello', 'World'],    # prints "('Hello', 'World')" and "Hello"
    #        tuple_in=('foo', 'bar'),       # prints "('foo', 'bar')" and "foo"
    #        int_in=10                      # prints "10" and "20"
    #    )

    def output(self):
        return mock.MockTarget('test_requires')

    def run(self):
        yield TestClass(
            date=self.date,                 # prints "2016-03-24" and "2016-03-17"
            string_in='Hello World',        # prints "Hello World" and "H"
            list_in=['Hello', 'World'],     # prints "('Hello', 'World')" and "("
            tuple_in=('foo', 'bar'),        # prints "('foo', 'bar')" and "("
            int_in=10                       # prints "10" and "20"
        )
        self.output().open('w').close()


class TestClass(luigi.Task):
    date = luigi.DateParameter()

    string_in = luigi.Parameter(default='')
    list_in = luigi.Parameter(default=[])
    tuple_in = luigi.Parameter(default=())
    int_in = luigi.IntParameter(default=0)

    tmp_path = luigi.Parameter(config_path=dict(section='path', name='tmp_path'))

    def output(self):
        return mock.MockTarget('test_out')

    def run(self):
        print(self.date)
[...]

PR #1614, opened 2016-03-29
Added ListParameter and TupleParameter.

dlstadther:
Dynamic requires serializes all `Parameter`. The default `serialize(...)` method casts everything to strings. Lists and tuples need to remain lists and tuples when serialized and parsed.

This is added as a solution to Issue #1607

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -28,6 +28,7 @@ from json import JSONEncoder
 from collections import OrderedDict, Mapping
 import operator
 import functools
+from ast import literal_eval
 
 try:
     from ConfigParser import NoOptionError, NoSectionError
@@ -798,3 +799,105 @@ class DictParameter(Parameter):
 
     def serialize(self, x):
         return json.dumps(x, cls=DictParameter.DictParamEncoder)
+
+
+class ListParameter(Parameter):
+    """
+    Parameter whose value is a ``list``.
+
+    In the task definition, use
+
+    .. code-block:: python
+
+        class MyTask(luigi.Task):
+          grades = luigi.ListParameter()
+
+            def run(self):
+                sum = 0
+                for element in self.grades:
+                    sum += element
+                avg = sum / len(self.grades)
+
+
+    At the command line, use
+
+    .. code-block:: console
+
+        $ luigi --module my_tasks MyTask --grades <JSON string>
+
+    Simple example with two grades:
+
+    .. code-block:: console
+
+        $ luigi --module my_tasks MyTask --grades '[100,70]'
+    """
+    def parse(self, x):
+        """
+        Parse an individual value from the input.
+
+        :param str x: the value to parse.
+        :return: the parsed value.
+        """
+        return list(json.loads(x))
+
+    def serialize(self, x):
+        """
+        Opposite of :py:meth:`parse`.
+
+        Converts the value ``x`` to a string.
+
+        :param x: the value to serialize.
+        """
+        return json.dumps(x)
+
+
+class TupleParameter(Parameter):
+    """
+    Parameter whose value is a ``tuple`` or ``tuple`` of tuples.
+
+    In the task definition, use
+
+    .. code-block:: python
+
+        class MyTask(luigi.Task):
+          book_locations = luigi.TupleParameter()
+
+            def run(self):
+                for location in self.book_locations:
+                    print("Go to page %d, line %d" % (location[0], location[1]))
+
+
+    At the command line, use
+
+    .. code-block:: console
+
+        $ luigi --module my_tasks MyTask --book_locations <JSON string>
+
+    Simple example with two grades:
+
+    .. code-block:: console
+
+        $ luigi --module my_tasks MyTask --book_locations '((12,3),(4,15),(52,1))'
+    """
+
+    def parse(self, x):
+        """
+        Parse an individual value from the input.
+
+        :param str x: the value to parse.
+        :return: the parsed value.
+        """
+        try:
+            return tuple(tuple(x) for x in json.loads(x))
+        except ValueError:
+            return literal_eval(x)  # if this causes an error, let that error be raised.
+
+    def serialize(self, x):
+        """
+        Opposite of :py:meth:`parse`.
+
+        Converts the value ``x`` to a string.
+
+        :param x: the value to serialize.
+        """
+        return json.dumps(x)
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -887,8 +887,20 @@ class TupleParameter(Parameter):
         :param str x: the value to parse.
         :return: the parsed value.
         """
+        # Since the result of json.dumps(tuple) differs from a tuple string, we must handle either case.
+        # A tuple string may come from a config file or from cli execution.
+
+        # t = ((1, 2), (3, 4))
+        # t_str = '((1,2),(3,4))'
+        # t_json_str = json.dumps(t)
+        # t_json_str == '[[1, 2], [3, 4]]'
+        # json.loads(t_json_str) == t
+        # json.loads(t_str) == ValueError: No JSON object could be decoded
+
+        # Therefore, if json.loads(x) returns a ValueError, try ast.literal_eval(x).
+        # ast.literal_eval(t_str) == t
         try:
-            return tuple(tuple(x) for x in json.loads(x))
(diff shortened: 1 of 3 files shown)


erikbern:
Great!

Is it easy for you to add your previously breaking example as a unit test?

dlstadther:
I've made my (first) attempt at writing a python test. I used the enum tests as a basis. If it sucks, I'll try again.

test/parameter_test.py
```diff
@@ -261,6 +275,16 @@ def test_enum_param_invalid(self):
     def test_enum_param_missing(self):
         self.assertRaises(ParameterException, lambda: luigi.parameter.EnumParameter())
 
+    def test_list_parameter(self):
+        a = luigi.ListParameter()
+        b_list = [1, 2, 3]
+        self.assertEqual(FooList.a_list, a.serialize(b_list).parse(b_list))
```
erikbern: I think this should say
`self.assertEqual(FooList.a_list, a.parse(a.serialize(b_list)))`

test/parameter_test.py
```diff
@@ -261,6 +275,16 @@ def test_enum_param_invalid(self):
...
+
+    def test_tuple_parameter(self):
+        a = luigi.TupleParameter()
+        b_tuple = ((1, 2), (3, 4))
+        self.assertEqual(FooTuple.a_tuple, a.serialize(b_tuple).parse(b_tuple))
```
erikbern: same here

erikbern:
LGTM is it easy for you to add the test with dynamic requirements too? (not blocking)

test/parameter_test.py
dlstadther: My bad. That's what I intended, but clearly not what I wrote. Fixed.

test/parameter_test.py
```diff
@@ -261,6 +275,16 @@ def test_enum_param_invalid(self):
     def test_enum_param_missing(self):
         self.assertRaises(ParameterException, lambda: luigi.parameter.EnumParameter())
 
+    def test_list_parameter(self):
+        a = luigi.ListParameter()
+        b_list = [1, 2, 3]
+        self.assertEqual(FooList.a_list, a.parse(a.serialize(b_list)))
```
davemt: I think you are trying to test that when you take a python value and ser-de it it ends up at the same python value, no? So you want something like 

`assertEqual(b_list, a.parse(a.serialize(b_list))`

davemt: Also have you tried running tests locally? It helps speed up the test dev process. You can even run just specific tests from just one module, see the CONTRIBUTING doc for `tox` commands

dlstadther: Yes, that is what i'm trying to test. Writing tests is still new to me. Thanks! I'll make that correction.

dlstadther:
I have some issues here. I'll resume work on it tomorrow.

luigi/parameter.py
```diff
@@ -798,3 +798,34 @@ def parse(self, s):
 
     def serialize(self, x):
         return json.dumps(x, cls=DictParameter.DictParamEncoder)
+
+
+class ListParameter(Parameter):
```
Tarrasch: Can you add docs? In particular how do you specify elements on CLI/config file. I for one don't know how output of `json.loads([1,2])` looks like.

test/parameter_test.py
```diff
@@ -261,6 +275,16 @@ def test_enum_param_invalid(self):
     def test_enum_param_missing(self):
         self.assertRaises(ParameterException, lambda: luigi.parameter.EnumParameter())
 
+    def test_list_parameter(self):
```
Tarrasch: Good that you have tests. Can you also have a test where you test the command line parsing or reading from config?

Tarrasch:
Yea. This is the correct way to resurrectt the functionality that got lost from #1133

dlstadther:
Can anyone explain to me why Luigi always prints lists as tuples when accessing the value of the parameter directly?

```python
class ListFoo(luigi.Task):
    my_list = luigi.ListParameter()

    def run(self):
        pass

class ParameterTest(LuigiTestCase):
    def test_list(self):
        a = luigi.ListParameter()
        b_list = [1, 2, 3]
        c = ListFoo(my_list=b_list)
        self.assertEqual(c.my_list, a.parse(a.serialize(b_list)))
```
Returns:
```
AssertionError: (1, 2, 3) != [1, 2, 3]
```

erikbern:
probably because json doesn't have a tuple type so it converts it into a list instead

erikbern:
you need to cast it back to a tuple

dlstadther:
I'm confused why i should cast a ListParameter to a tuple

luigi/parameter.py
dlstadther: Sure can. 

`json.dumps([1,2])` would return `'[1,2]'`

`json.loads('[1,2]')` would return `[1,2]`

test/parameter_test.py
dlstadther: I'll work on it

I have written config tests. But am struggling with the cli tests. I'll be redoing these commits and they will be forthcoming.

dlstadther:
I believe all i have left now is testing 1) cli usage and 2) dynamic requires. I'll recommit something by the end of my work day today.

luigi/parameter.py
```diff
@@ -798,3 +799,105 @@ def parse(self, s):
...
+            def run(self):
+                for location in self.book_locations:
+                    print("Go to page %d, line %d" % (location[0], location[1]))
+
+
+    At the command line, use
+
+    .. code-block:: console
+
+        $ luigi --module my_tasks MyTask --book_locations <JSON string>
```
Tarrasch: I'm pretty sure it's book-locations right?

luigi/parameter.py
```diff
@@ -798,3 +799,105 @@ def parse(self, s):
...
+        """
+        Parse an individual value from the input.
+
+        :param str x: the value to parse.
+        :return: the parsed value.
+        """
+        try:
+            return tuple(tuple(x) for x in json.loads(x))
+        except ValueError:
+            return literal_eval(x)  # if this causes an error, let that error be raised.
```
Tarrasch: Can you add a `#`-comment of when this  code path is supposed to happen?

test/parameter_test.py
```diff
@@ -678,6 +702,14 @@ def complete(self):
         self.assertTrue(self.run_locally_split('mynamespace.A --mynamespace.A-p 200 --expected 200'))
         self.assertFalse(self.run_locally_split('mynamespace.A --A-p 200 --expected 200'))
```
Tarrasch: I also would like a test with parsing a value on the command line, like this line. Can you add that please? :)

Tarrasch:
Sorry for asking for even more tests, but this is a pretty core functionality. The docs look great by the way! :)

luigi/parameter.py
dlstadther: Is this a luigi/python style choice that I'm not aware?

test/parameter_test.py
dlstadther: I can get the two assertTrue statements to work for ListParameter, but the `self.assertFalse(self.run_locally_split('mynamespace.A --A-p [1,2,3] --expected [1,2,3]'))` returns `SystemExit: 2` , ending the stacktrace with argparse.py causing the exit. statement.

dlstadther: Likewise for TupleParameter

dlstadther:
I'll wait to rebase with updated master after I satisfy @erikbern and @Tarrasch 's requested additions and modifications.

dlstadther:
Cli tests keep failing locally, but error message isn't helpful. Checking if travis-ci will throw the same error.

test/parameter_test.py
Tarrasch: It's okay. You can skip the `assertFalse` test. I think that that test verifies that `--A-p` is invalid syntax because it's either `--p` or `--mynamespace.A-p`.

dlstadther: I've removed those two lines.

luigi/parameter.py
dlstadther: @Tarrasch any explanation of this?

erikbern:
lgtm

dlstadther:
If @Tarrasch approves, are we good to merge?

PR #1621, opened 2016-03-31
Task status messages

riga:
This PR adds status messages to tasks which are also visible on the scheduler GUI.

Examples:

task list

worker list

Status messages are meant to change during the run method in an [opportunistic way](https://github.com/riga/luigi/blob/statusTexts/luigi/task.py#L515). Especially for long-running non-Hadoop tasks, the ability to read those messages directly in the scheduler GUI is quite helpful (at least for us). Internally, message changes are propagated to the scheduler via a ``_status_message_callback`` which is - in contrast to ``tracking_url_callback`` - not passed to the run method, but set by the ``TaskProcess``.

Usage example:
```python
class MyTask(luigi.Task):
    ...
    def run(self):
        for i in range(100):
            # do some hard work here
            if i % 10 == 0:
                self.set_status_message("Progress: %s / 100" % i)
```

I know that you don't like PR's that affect core code (which is reasonable =) ), but imho this feature is both lightweight *and* really helpful.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -289,7 +289,7 @@ class Task(object):
         self.task_id = task_id_str(self.task_family, self.to_str_params(only_significant=True))
         self.__hash = hash(self.task_id)
 
-        self._status_message = None
+        self.status_message = None
         self._status_message_callback = None
 
     def initialized(self):
@@ -507,14 +507,9 @@ class Task(object):
         Default behavior is to send an None value"""
         pass
 
-    @property
-    def status_message(self):
-        return self._status_message
-
-    @status_message.setter
-    def status_message(self, message):
-        if message != self._status_message:
-            self._status_message = message
+    def set_status_message(self, message):
+        if message != self.status_message:
+            self.status_message = message
             if hasattr(self._status_message_callback, "__call__"):
                 self._status_message_callback(message)
 
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -289,6 +289,9 @@ class Task(object):
         self.task_id = task_id_str(self.task_family, self.to_str_params(only_significant=True))
         self.__hash = hash(self.task_id)
 
+        self._status_message = None
+        self._status_message_callback = None
+
     def initialized(self):
         """
         Returns ``True`` if the Task is initialized and ``False`` otherwise.
@@ -503,6 +506,17 @@ class Task(object):
 
         Default behavior is to send an None value"""
         pass
+
+    @property
+    def status_message(self):
+        return self._status_message
+
+    @status_message.setter
+    def status_message(self, message):
+        if message != self._status_message:
+            self._status_message = message
+            if hasattr(self._status_message_callback, "__call__"):
+                self._status_message_callback(message)
 
 
 class MixinNaiveBulkComplete(object):
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -508,6 +508,11 @@ class Task(object):
         pass
 
     def set_status_message(self, message):
+        """
+        Sets the status message of the task to message. If the value actually changed w.r.t the
+        current value, the _status_message_callback is invoked which propagates the change down to
+        the scheduler.
+        """
         if message != self.status_message:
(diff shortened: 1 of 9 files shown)


erikbern:
this is quite nice!

i'm not a big fan of the property though – can you just make it a method on the Task class? this also makes it easier in case we need to extend it later with more arguments etc

riga:
Sure! I also updated the example above.

erikbern:
Nice! Some more things
* Please rebase on latest master – seems like there's some conflict
* Documentation plz
* Unit tests plz

otherwise great!

riga:
Is the doc string of Task.set_status_message sufficient?

erikbern:
yes looks good. let's merge!

luigi/task.py
```diff
@@ -504,6 +507,17 @@ def on_success(self):
...
 
+    def set_status_message(self, message):
+        """
+        Sets the status message of the task to message. If the value actually changed w.r.t the
+        current value, the _status_message_callback is invoked which propagates the change down to
+        the scheduler.
+        """
+        if message != self.status_message:
+            self.status_message = message
+            if hasattr(self._status_message_callback, "__call__"):
```
Tarrasch: and `else` at least print a warning?

luigi/task.py
```diff
@@ -289,6 +289,9 @@ def __init__(self, *args, **kwargs):
         self.task_id = task_id_str(self.task_family, self.to_str_params(only_significant=True))
         self.__hash = hash(self.task_id)
 
+        self.status_message = None
+        self._status_message_callback = None
```
Tarrasch: I don't get this pair. Shouldn't `status_message` be prefixed with `_` and not the callback? The user is meant to write a callback method, not tamper with the internal  `status_message` variable that's used as a cache.

In fact, why not just remove `status_message`. Are you sure you need the only-post-if-changed logic?

Tarrasch:
This need more docs before we can merge. I think it would be good (and enough) to copy paste your example in this PR as pydocs for `set_status_message` method.

luigi/worker.py
```diff
@@ -94,13 +94,14 @@ class TaskProcess(multiprocessing.Process):
...
     def __init__(self, task, worker_id, result_queue, random_seed=False, worker_timeout=0,
-                 tracking_url_callback=None):
+                 tracking_url_callback=None, status_message_callback=None):
         super(TaskProcess, self).__init__()
         self.task = task
         self.worker_id = worker_id
         self.result_queue = result_queue
         self.random_seed = random_seed
         self.tracking_url_callback = tracking_url_callback
+        self.task._status_message_callback = status_message_callback
```
Tarrasch: Hmm. I wish we had done this trick for `tracking_url_callback` too...

Nice trick! :)

luigi/task.py
Tarrasch: Nevermind the first paragraph. User is not overriding `_status_message_callback`. :)

Tarrasch:
Commit history is also a bit garbled and contains commits from other authors. Can you fix?

luigi/task.py
riga: As the scheduler receives a request each time the status is changed/set, I think the ``status_message`` cache is required to prevent user errors, i.e., the possibility that the scheduler is flooded with too many requests.

Tarrasch: Well, you mean if somebody would write code like

```
for i in range(100000):
     self.set_status("hello")
```

that "caching" would cover that. But what about the slightly different:


```
for i in range(100000):
     self.set_status("hello " + i)
```

I think it would be good for stability if we checked that there were no more than 105 (or something) status_updates per run invocation. And I don't think this arbitrary check is the way to go. I don't want anyone starting to rely on it and it suddenly gets re-factored away.

riga:
I tried to fix the commit history but I messed it up :/
I created a new PR #1625 which fixes the history. Sorry for the inconvenience.

Tarrasch:
No problem. Thanks for interlinking the PRs!

PR #1623, opened 2016-03-31
Fix typo in scheduler logging

joemeszaros:
# What is this PR?

I faced with the following log message in Luigi server log : 
`2016-03-31 06:44:27,172 luigi.server[2331] INFO: No prior state file exists at /var/lib/luigi-server/state.pickle. Starting with clean slate`

which has a typo at the end

# Solution

Fix typo in scheduler and rename `slate` to `state`.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -318,7 +318,7 @@ class SimpleTaskState(object):
                 with open(self._state_path, 'rb') as fobj:
                     state = pickle.load(fobj)
             except BaseException:
-                logger.exception("Error when loading state. Starting from clean slate.")
+                logger.exception("Error when loading state. Starting from clean state.")
                 return
 
             self.set_state(state)
@@ -326,7 +326,7 @@ class SimpleTaskState(object):
             for task in six.itervalues(self._tasks):
                 self._status_tasks[task.status][task.id] = task
         else:
-            logger.info("No prior state file exists at %s. Starting with clean slate", self._state_path)
+            logger.info("No prior state file exists at %s. Starting with clean state", self._state_path)
 
     def get_active_tasks(self, status=None):
         if status:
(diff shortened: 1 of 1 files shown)


dlstadther:
I'm not sure if that's actually a typo. To me _clean slate_ and _clean state_ convey the same meaning.

But if _clean **state**_ is more explicit and no one has an opposition, then lgtm.

erikbern:
yeah not sure if it's a typo

Tarrasch:
+1. "Clean slate"  is probably unnecessary complicated English (I've learned about importance of simple language recently). But I'm not if "clean state" is clear.  What about "empty state"?

joemeszaros:
Renamed to `empty state`. For non-native-English users it could be easier to understand, but without this micro modification Luigi is still a great tool, so I can live with the original log message too.

Tarrasch:
My opinion is that micro-modifications are very welcome, because I think spelling errors and improper English is a sign of inactivity/weakness of open source projects. Thanks for this patch @joemeszaros. I'll (or somebody) will merge this once build passes. Good job! :)

dlstadther:
LGTM and all tests passed :+1:

erikbern:
Yeah but this time I feel like we actually *introduced* a typo. Even Google corrects "clean state" to "clean slate"

image

ulzha:
No, Google's suggestion isn't based on meaning as much as on typo distance, right?

erikbern:
yes and i would consider "clean state" a mild typo

but it's fine, i'm just being grumpy.

erikbern:
also i'm clearly bikeshedding here so will stop :)

dlstadther:
Or we could actually elaborate on what is meant by _clean slate/state_ since we seem to disagree on what is and isn't a typo. Specifically, what is meant here when luigi prints `2016-03-31 06:44:27,172 luigi.server[2331] INFO: No prior state file exists at /var/lib/luigi-server/state.pickle. Starting with clean slate`?

But as has been mentioned, it's probably not the biggest deal.

Tarrasch:
@erikbern, but this PR changed it to "empty state", not "clean state". :)

PR #1624, opened 2016-03-31
Warn user if Parameter parses something other than a string.

dlstadther:
See within thread for Issue #1607

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -230,6 +230,8 @@ class Parameter(object):
 
         :param x: the value to serialize.
         """
+        if not isinstance(x, six.string_types) and self.__class__ == Parameter:
+            warnings.warn("Parameter {0} is not of type string.".format(str(x)))
         return str(x)
 
     def normalize(self, x):
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -86,7 +86,7 @@ class core(task.Config):
         description='Port of remote scheduler api process',
         config_path=dict(section='core', name='default-scheduler-port'))
     scheduler_url = parameter.Parameter(
-        default=None,
+        default='',
         description='Full path to remote scheduler',
         config_path=dict(section='core', name='default-scheduler-url'),
     )
@@ -106,10 +106,10 @@ class core(task.Config):
         default=1,
         description='Maximum number of parallel tasks to run')
     logging_conf_file = parameter.Parameter(
-        default=None,
+        default='',
         description='Configuration file for logging')
     module = parameter.Parameter(
-        default=None,
+        default='',
         description='Used for dynamic loading of modules',
         always_in_help=True)
     parallel_scheduling = parameter.BoolParameter(
@@ -158,7 +158,7 @@ def _schedule_and_run(tasks, worker_scheduler_factory=None, override_defaults=No
     # search for logging configuration path first on the command line, then
     # in the application config file
     logging_conf = env_params.logging_conf_file
-    if logging_conf is not None and not os.path.exists(logging_conf):
+    if logging_conf != '' and not os.path.exists(logging_conf):
         raise Exception(
             "Error: Unable to locate specified logging configuration file!"
         )
@@ -175,7 +175,7 @@ def _schedule_and_run(tasks, worker_scheduler_factory=None, override_defaults=No
     if env_params.local_scheduler:
         sch = worker_scheduler_factory.create_local_scheduler()
     else:
-        if env_params.scheduler_url is not None:
+        if env_params.scheduler_url != '':
             url = env_params.scheduler_url
         else:
             url = 'http://{host}:{port:d}/'.format(
--- a/test/cmdline_test.py
+++ b/test/cmdline_test.py
@@ -133,9 +133,9 @@ class CmdlineTest(unittest.TestCase):
     @mock.patch("luigi.interface.setup_interface_logging")
     def test_cmdline_logger(self, setup_mock, warn):
         with mock.patch("luigi.interface.core") as env_params:
-            env_params.return_value.logging_conf_file = None
+            env_params.return_value.logging_conf_file = ''
             luigi.run(['SomeTask', '--n', '7', '--local-scheduler', '--no-lock'])
-            self.assertEqual([mock.call(None)], setup_mock.call_args_list)
+            self.assertEqual([mock.call('')], setup_mock.call_args_list)
 
         with mock.patch("luigi.configuration.get_config") as getconf:
             getconf.return_value.get.side_effect = ConfigParser.NoOptionError(section='foo', option='bar')


Tarrasch:
But this is going to give so many warnings to so many users!!!



... and I like it :sunglasses:!

luigi/parameter.py
```diff
@@ -219,6 +219,8 @@ def parse(self, x):
         :param str x: the value to parse.
         :return: the parsed value.
         """
+        if not isinstance(x, basestring) or not isinstance(x, str):
```
dlstadther: My mac is not playing nice installing python 3.5 and setting up a venv, but does anyone know if this line will falsely evaluate that a string (unicode or ascii) is not an instance of basestring for python 3+?

This does seem to work correctly for python 2.7.

dlstadther: Or apparently all the python 3 tests will fail because they don't know what `basestring` is.

How should one go about this condition? I need `if not isinstance(x, basestring)` for python2 and `if not isinstance(x, str)` for python3.

dlstadther: I can only think of the following:

```python
def parse(self, x):
     try:  # try python2 syntax
          if not isinstance(x, basestring):
               warnings.warn(...)
      except:  # if py2 syntax fails, try py3
          if not isinstance(x, str):
               warnings.warn(...)
      return x
```

But this repeats `warnings.warn(...)` :disappointed:

dlstadther:
Opted for the duplicate warning message since i'm trying to get comments and our work schedules appear to complete opposites (yay timezones).

I'm very open to better suggestions.

Tarrasch:
Have you checked in six.py? In particular `six.string_types` looks relevant.

Tarrasch:
Once you've checked the six.py module, let me know when you think this is merge-worthy.

dlstadther:
I have not. But now that I know to look there, I will and will notify you when complete. Thanks

dlstadther:
I don't want to change how python 2 handles `str` or `unicode` nor python 3 `str`. So unless someone else believes i'm mistaken, I don't believe any change is necessary within six.py.

luigi/parameter.py
```diff
@@ -219,7 +219,14 @@ def parse(self, x):
...
         :return: the parsed value.
         """
-        return x  # default impl
+        try:
+            if not isinstance(x, basestring):   # python 2
+                warnings.warn("Parameter %r is not of type string." % x)
+        except:
+            if not isinstance(x, str):          # python 3
+                warnings.warn("Parameter %r is not of type string." % x)
+        finally:
```
dlstadther: However, i'm open to suggestion on how to clean this up and not have the `warnings.warn(...)` message written twice.

The only other thing i can think of to only write the message once is:
```python
warn = False
try:
    if not isinstance(x, basestring):
        warn = True
except:
    if not isinstance(x, str):
        warn = True
finally:
    if warn:
        warnings.warn("Parameter %r is not of type string." % x)
    return x
```

erikbern:
https://pythonhosted.org/six/#six.string_types

you can just do `if isinstance(value, six.string_Types)`

dlstadther:
Oh ok! I didn't realize that's how six.py was supposed to be used. Thanks @erikbern !

Tarrasch:
I was about to merge this, until it I just realized that I don't understand this patch. Can you reiterate in this PR why you are doing this? I thought we wanted to check that nobody passes a non-string to Parameter.

dlstadther:
**tl;dr**: `Parameter` should really be named `StrParameter` since it de/serializes strings; however, when using normal requirements this can be disregarded without unintended consequences, but misuse of `Parameter` within dynamic requirements will result in data treated as strings.

Through using dynamic requires, I found that anything passed as a `Parameter` was treated as a string (which is why I added PR #1614 ). 

Parameters take one of two paths within luigi execution: 1. dynamic requires 2. normal requires.

1. @davemt found that

> dynamic dependencies have `load_task` called on them before they are added to worker's task cache whereas regular `requires()` do not. This requires that a parameter value can be properly serialized/deserialized.

So since the `Parameter` is first serialized, it is cast as a string and then parsed (as a string). Thus no warning message.

2. However, during normal requires, there is no guaranteed serialization/deserialization. So you can (as I have historically) incorrectly pass any type of structure through as a `Parameter` (i.e. Lists, Tuples, etc.). 
So since `Parameter` assumes strings (as seen by its `serialize()` method), @erikbern suggested

> It might be good to do a warnings.warn in case serialize returns a non-string

dlstadther:
@Tarrasch , any comments or questions?

Tarrasch:
I stared at this for a while and I still don't understand this patch.

When could `parse()` ever be called with a non-string? That would only happen on fatal framework (luigi) errors. No?

Why not just put the check in `serialize` where it intuitively makes sense? That would occur on actual user errors.

dlstadther:
So @Tarrasch is correct, the string check within`parse` won't warn the user. However, we cannot just simply move that string check into `serialize`.

The problem with including the warn within `serialize` is that other Parameters inherit (and don't override) that method. e.g. `FloatParameter`, `IntParameter`, and `BoolParmeter`.

If we add the string check within `Parameter.serialize`, we will need to override the `serialize` method for int, float, and bool.

Tarrasch:
> If we add the string check within Parameter.serialize, we will need to override the serialize method for int, float, and bool.

This is not correct I think. There is an easier way, just check if the current class is Parameter or not.

```python
if not isinstance(x, six.string_types) and self.class == Parameter:
```

dlstadther:
Learn something new everyday. Thanks! I'll add that tomorrow morning.

Tarrasch:
Thanks. Did you test to see if this works as expected? If so let's merge :)

dlstadther:
Yes, I tested using the following code:

```python
import datetime

import luigi
from luigi import mock

__author__ = "Dillon Stadther"
__date__ = '2016-03-24'


class TestTask(luigi.Task):
    date = luigi.DateParameter(default=datetime.date(2016, 03, 24))

    def requires(self):
        yield TestClass(
            date=self.date,                # prints "2016-03-24" and "2016-03-17"
            string_in='Hello World',       # prints "Hello World" and "H"
            str1_in=['Bye'],
            list_in=['Hello', 'World'],    # prints "('Hello', 'World')" and "Hello"
            tuple_in=('foo', 'bar'),       # prints "('foo', 'bar')" and "foo"
            int_in=10                      # prints "10" and "20"
        )

    def output(self):
        return mock.MockTarget('test_requires')

    def run(self):
        yield TestClass(
            date=self.date,                 # prints "2016-03-24" and "2016-03-17"
            string_in='Hello World',        # prints "Hello World" and "H"
            str1_in=['Bye', 'Doge'],
            list_in=['Hello', 'World'],     # prints "('Hello', 'World')" and "("
            tuple_in=('foo', 'bar'),        # prints "('foo', 'bar')" and "("
            int_in=10                       # prints "10" and "20"
        )
        self.output().open('w').close()


class TestClass(luigi.Task):
    date = luigi.DateParameter()

    string_in = luigi.Parameter(default='')
    str1_in = luigi.Parameter(default='')
    list_in = luigi.ListParameter(default=[])
    tuple_in = luigi.TupleParameter(default=())
    int_in = luigi.IntParameter(default=0)

    def output(self):
        return mock.MockTarget('test_out')

    def run(self):
        print(self.date)
        print(self.date + datetime.timedelta(days=-7))  # should print the date 7 days ago
        print(self.string_in)
        print(self.string_in[0])        # should print first character
        print(self.str1_in)
        print(self.str1_in[0])  # should print first character
        print(self.list_in)
        print(self.list_in[0])          # should print first element
        print(self.tuple_in)
        print(self.tuple_in[0])         # should print first element
        print(self.int_in)
        print(self.int_in * 2)          # should print double the int

        self.output().open('w').close()


if __name__ == "__main__":
    luigi.run(['TestTask', '--workers', '1'])
```

When `str1_in` is serialized (for both normal requires and dynamic requires), the following warning messages are presented:

```
/Users/dstadther/venv-luigi/lib/python2.7/site-packages/luigi-2.1.1-py2.7.egg/luigi/parameter.py:240: UserWarning: Parameter ('Bye',) is not of type string.
  warnings.warn("Parameter {0} is not of type string.".format(str(x)))
/Users/dstadther/venv-luigi/lib/python2.7/site-packages/luigi-2.1.1-py2.7.egg/luigi/parameter.py:240: UserWarning: Parameter ('Bye', 'Doge') is not of type string.
  warnings.warn("Parameter {0} is not of type string.".format(str(x)))
```

erikbern:
bonus points if you want to add a unit test for this. but i'm fine mergin as-is

dlstadther:
@erikbern Bonus points have been attempted

erikbern:
fantastic!

test/parameter_test.py
```diff
@@ -70,6 +71,34 @@ def run(self):
...
+    def requires(self):
+        # normal requires
+        yield BlankTask(str_param=self.str_param, non_str_param=self.non_str_param)
+
+    def output(self):
+        return MockTarget("dumb_output")
+
+    def run(self):
+        # dynamic requires
+        yield DumbTask2(str_param=self.str_param, non_str_param=self.non_str_param)
```
erikbern: is DumbTask2 defined? Flake8 is complaining about it

test/parameter_test.py
```diff
@@ -285,6 +314,14 @@ def test_tuple_serialize_parse(self):
         b_tuple = ((1, 2), (3, 4))
         self.assertEqual(b_tuple, a.parse(a.serialize(b_tuple)))
 
+    def test_requires_warn_serialize(self):
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            NonStrParamTask()
```
erikbern: what's this statement useful for?

test/parameter_test.py
```diff
@@ -285,6 +314,14 @@ def test_tuple_serialize_parse(self):
...
+            assert len(w) == 1
+            assert issubclass(w[-1].category, Warning)
+            assert "is not of type string." in str(w[-1].message)
```
erikbern: you should use unittest assertions here – self.assertEquals etc

erikbern:
for some reason the unit tests are failing in 2.7 and 3.3 – not sure if it's because we get 0 warnings or 2+

test/parameter_test.py
dlstadther: I changed the name of the class and forgot to update all calls to it.

dlstadther:
This is weird...

I have corrected your in-line comments and reran locally. I get a :+1: for `tox -e py27-nonhdfs test/parameter_test.py`, but a :-1: for `tox -e py27-nonhdfs`. The difference in failure is with asserting that there is 1 warning.

Any idea why this may be occurring?

erikbern:
Not sure. If you can't figure it out, no problem. The test isn't super important right now, we can just remove the last commit

dlstadther:
I'm going to mess with this for a bit and i'll report back by tomorrow.

dlstadther:
I'm trying to understand the process of execution for `tox -e py27-nonhdfs`.

So tox is install on my local system (specifically within my venv). Tox reads the tox.ini file and creates a virtual testing environment based on the .ini file. It then runs the commands that are listed - which appears to be (optionally setup hadoop environment), print python version, measure the coverage of tests within the project by executing `test/runtests.py, combine coverage data, and then doing something with codecov.

[`test/runtests.py`](https://github.com/spotify/luigi/blob/master/test/runtests.py) appears to...
1. catch warnings using the `default` flag (according to [py `warnings` docs](https://docs.python.org/2/library/warnings.html#the-warnings-filter)...`default` will "print the first occurrence of matching warnings for each location where the warning is issued")  
2. ignore the printing of `UserWarning`  
3. execute `nose.main()` (...which does what exactly???) (If someone could explain this, that'd be much appreciated!)  

Somehow this executes all files within `test/*`, but I'm unsure how.

I'm also unsure how specifying a filename after the environment (`tox -e py27-nonhdfs test/parameter_test.py`) changes this execution.

My concern is that the environment for all tests is somehow different from that for a single test. If possible, I'd feel most comfortable getting a successful test prior to merging (in the case that something isn't quite right and spams everyone with false positive warning messages).

erikbern:
i'm not sure if there's an easy way to run a single test using tox. unfortunately i generally have pretty bad experience trying to use tox.

i'm fine if you just want to remove the last commit from this PR. we don't have to unit test warnings :)

dlstadther:
Yeah, tox documentation isn't very clear (and yet is at least semi-active - latest release < 6 months ago).

I'll remove the tests.

I just ran some other jobs using this PR and am getting a warning saying that `None is not of type string` although i'm not passing None as any Parameter. So i'm not comfortable merging this yet.

Tarrasch:
@dlstadther, Oh I guess that's because there's part of luigi that itself has "None" has default value. You should fix those cases too before we merge this. Otherwise users will get confused by unavoidable warnings.

erikbern:
ok that makes sense. yeah since warnings.warn only triggers the first time, it's probably because it was already run somewhere else. good catch, @Tarrasch

dlstadther:
@Tarrasch I was getting mildly suspicious of such. Do you know off the top of your head where this "part of luigi" might be location? As time allows today, I'll do some searching.

erikbern:
it should be easy to find if you just replace the warning with an exception and run the tests :)

dlstadther:
The culprit: a param named `scheduler_url` with a value of `None`. It is included in the list of other Luigi system Parameters which are returned by `dir(cls)` in `Task.py` `get_params(cls)` [L210](https://github.com/spotify/luigi/blob/master/luigi/task.py#L210).

``` python
@classmethod
    def get_params(cls):
        """
        Returns all of the Parameters for this Task.
        """
        # We want to do this here and not at class instantiation, or else there is no room to extend classes dynamically
        params = []
        for param_name in dir(cls):
            param_obj = getattr(cls, param_name)
            if not isinstance(param_obj, Parameter):
                continue

            params.append((param_name, param_obj))

        # The order the parameters are created matters. See Parameter class
        params.sort(key=lambda t: t[1]._counter)
        return params
```

Based on Luigi docs, `scheduler_url` should default to `http://localhost:8082` (unless I'm understanding this incorrectly), but execution shows that `scheduler_url` (at this point in execution) is set to `None`.

erikbern:
there must be a reason we set them to None. not sure

dlstadther:
If it's valid to have optional Luigi system Parameters as `None`, but not for User-Defined Parameters to be `None`, Can there just be an additional check that the `Parameter` is not a string and not `None` before it warns the user?

dlstadther:
`scheduler_url` looks to be set to `None` within interface.py so that it can be dynamically created via `scheduler_host` and `scheduler_port`.

dlstadther:
Scratch that last statement.

The same thing happens with `module`, but that appears to be involved with CLI parsing (or at least that's what `always_in_help=True` makes be believe).

dlstadther:
```python
 if not isinstance(x, six.string_types) and self.__class__ == Parameter and x is not None:
      warnings.warn("Parameter {0} is not of type string.".format(str(x)))
```

Does not warn against these Luigi system parameters, but also does not warn against user-defined Parameters that are defined as `None`. Are there any problems with this solution?

Tarrasch:
@dlstadther, can you just change those `default=None` to `default=''` then? If there is any `param is None` checks just change them to `!= ''`. Should be doable right?

dlstadther:
I can definitely do that for `scheduler_url`. I'm unsure what the significance of `module=None`. So I can, but there may be unknown side-effects if the CLI parser's help (or whatever `always_in_help` insinuates) expects a potential `None` value.

Tarrasch:
I don't think there's any "gotcha" for  the `module` parameter. Don't mind `always_in_help`. There's both tests and human review to catch any mistakes. :)

dlstadther:
Tests (luigi unit tests and my jobs) ran successfully and as expected locally. Hopefully the same will reign true for Travis-CI.

erikbern:
looks like your unit test disappeared?

dlstadther:
yeah, you told me to remove them since I was receiving different results depending on whether I ran just the `parameter_test.py` or the entire nonhdfs suite of tests.

dlstadther:
I can spend more time on testing if necessary

erikbern:
i thought you fixed the issue causing the unit tests to break? i think if you add back the unit tests they will pass right?

anyway, it's fine, let's merge this!

dlstadther:
I'll the test one more shot. Even if it's post-merge.

dlstadther:
For the life of me, I still cannot figure out why the below unit test passes for `tox -e py27-nonhdfs test/parameter_test.py`, but fails when the entire suite of nonhdfs tests are executed.

```python
    def test_requires_warn_serialize(self):
        class NonStrParamTask(luigi.Task):
            str_param = luigi.Parameter(default='Hello World')
            non_str_param = luigi.Parameter(default=1)

            def output(self):
                return MockTarget("dumb_output")

            def run(self):
                self.output().open('w').close()

        import warnings
        with warnings.catch_warnings(record=True) as w:
            warnings.simplefilter("always")
            NonStrParamTask(str_param="Hello World", non_str_param=1)
            self.assertEqual(len(w), 1)
            self.assertTrue(issubclass(w[-1].category, UserWarning))
            self.assertTrue("is not of type string." in str(w[-1].message))
```
```
FAIL: test_requires_warn_serialize (parameter_test.ParameterTest)
  File "/Users/dstadther/venv-l-test/luigi/test/parameter_test.py", line 304, in test_requires_warn_serialize
    self.assertEqual(len(w), 1)
AssertionError: 0 != 1
```

erikbern:
that's really weird. maybe something else triggers the warning earlier?

ulzha:
Do we not want people to use `None`, as a parameter default value, in general? I've treated `default=None` as a fitting way to specify that a parameter is optional (can have a value or have no value).

Tarrasch:
@ulzha I found it a bit unmathematical and it causes bugs all over. In particular string value `None` and `'None'` both serializes to `'None'` and that only de-serializes to `'None'`. The `inverse_of(serialize) == parse` invariant is broken.

In most real world cases, there's probably a suitable default value that's a valid set-member of the type of the parameter (typically values in Python corresponding to `None` like `''`, `0`, `[]` etc).

PR #1625, opened 2016-04-01
[fixed] Task status messages

riga:
> This PR is a fixed version of PR #1621 which has a faulty commit history.

This PR adds status messages to tasks which are also visible on the scheduler GUI.

Examples:

task list

worker list

Status messages are meant to change during the run method in an [opportunistic way](https://github.com/riga/luigi/blob/statusTexts/luigi/task.py#L515). Especially for long-running non-Hadoop tasks, the ability to read those messages directly in the scheduler GUI is quite helpful (at least for us). Internally, message changes are propagated to the scheduler via a ``_status_message_callback`` which is - in contrast to ``tracking_url_callback`` - not passed to the run method, but set by the ``TaskProcess``.

Usage example:
```python
class MyTask(luigi.Task):
    ...
    def run(self):
        for i in range(100):
            # do some hard work here
            if i % 10 == 0:
                self.set_status_message("Progress: %s / 100" % i)
```

I know that you don't like PR's that affect core code (which is reasonable =) ), but imho this feature is both lightweight *and* really helpful.

### DIFF
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -149,6 +149,25 @@ This can be useful if you have many dependencies:
             g = [y.open('r') for y in self.input()['b']]
 
 
+.. _Task.set_status_message:
+
+Task.set_status_message
+~~~~~~~~~~~~~~~~~~~~~~~
+
+For long-running tasks it is convenient to see status messages not only on the command line or in
+your logs but also in the GUI of the central scheduler. For this purpose you can use
+:func:`~luigi.task.Task.set_status_message`:
+
+.. code:: python
+
+    class MyTask(luigi.Task):
+        def run(self):
+            for i in range(100):
+                # do some hard work here
+                if i % 10 == 0:
+                    self.set_status_message("Progress: %d / 100" % i)
+
+
 Dynamic dependencies
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -179,21 +179,26 @@ For an example of a workflow using dynamic dependencies, see
 `examples/dynamic_requirements.py <https://github.com/spotify/luigi/blob/master/examples/dynamic_requirements.py>`_.
 
 
-Task Status messages
+Task status tracking
 ~~~~~~~~~~~~~~~~~~~~
 
-For long-running tasks it is convenient to see status messages not only on the command line or in
-your logs but also in the GUI of the central scheduler. For this purpose you can use the
-``status_message_callback`` which is passed opportunistically to Task.run_:
+For long-running or remote tasks it is convenient to see extended status information not only on 
+the command line or in your logs but also in the GUI of the central scheduler. Luigi implements
+dynamic status messages and tracking urls which may point to an external monitoring system. You 
+can set this information using callbacks within Task.run_:
 
 .. code:: python
 
     class MyTask(luigi.Task):
-        def run(self, status_message_callback):
+        def run(self):
+            # set a tracking url
+            self.set_tracking_url("http://...")
+
+            # set status messages during the workload
             for i in range(100):
                 # do some hard work here
                 if i % 10 == 0:
-                    status_message_callback("Progress: %d / 100" % i)
+                    self.set_status_message("Progress: %d / 100" % i)
 
 
 .. _Events:
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -149,25 +149,6 @@ This can be useful if you have many dependencies:
             g = [y.open('r') for y in self.input()['b']]
 
 
-.. _Task.set_status_message:
-
-Task.set_status_message
-~~~~~~~~~~~~~~~~~~~~~~~
-
-For long-running tasks it is convenient to see status messages not only on the command line or in
-your logs but also in the GUI of the central scheduler. For this purpose you can use
-:func:`~luigi.task.Task.set_status_message`:
-
-.. code:: python
-
-    class MyTask(luigi.Task):
-        def run(self):
-            for i in range(100):
-                # do some hard work here
-                if i % 10 == 0:
-                    self.set_status_message("Progress: %d / 100" % i)
-
-
 Dynamic dependencies
 ~~~~~~~~~~~~~~~~~~~~
 
@@ -196,6 +177,24 @@ In other words, you should make sure your Task.run_ method is idempotent.
 
 For an example of a workflow using dynamic dependencies, see
 `examples/dynamic_requirements.py <https://github.com/spotify/luigi/blob/master/examples/dynamic_requirements.py>`_.
+
+
+Task Status messages
+~~~~~~~~~~~~~~~~~~~~
+
+For long-running tasks it is convenient to see status messages not only on the command line or in
+your logs but also in the GUI of the central scheduler. For this purpose you can use the
+``status_message_callback`` which is passed opportunistically to Task.run_:
+
+.. code:: python
+
+    class MyTask(luigi.Task):
+        def run(self, status_message_callback):
+            for i in range(100):
+                # do some hard work here
+                if i % 10 == 0:
+                    status_message_callback("Progress: %d / 100" % i)
+
 
 .. _Events:
 
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -163,7 +163,7 @@ class Task(object):
 
     def __init__(self, task_id, status, deps, resources=None, priority=0, family='', module=None,
                  params=None, disable_failures=None, disable_window=None, disable_hard_timeout=None,
-                 tracking_url=None):
+                 tracking_url=None, status_message=None):
         self.id = task_id
         self.stakeholders = set()  # workers ids that are somehow related to this task (i.e. don't prune while any of these workers are still active)
         self.workers = set()  # workers ids that can perform task - task is 'BROKEN' if none of these workers are active
@@ -188,6 +188,7 @@ class Task(object):
         self.disable_hard_timeout = disable_hard_timeout
         self.failures = Failures(disable_window)
         self.tracking_url = tracking_url
+        self.status_message = status_message
         self.scheduler_disable_time = None
         self.runnable = False
 
@@ -870,6 +871,7 @@ class CentralPlannerScheduler(Scheduler):
             'priority': task.priority,
             'resources': task.resources,
             'tracking_url': getattr(task, "tracking_url", None),
+            'status_message': task.status_message
         }
         if task.status == DISABLED:
             ret['re_enable_able'] = task.scheduler_disable_time is not None
@@ -1095,6 +1097,18 @@ class CentralPlannerScheduler(Scheduler):
             return {"taskId": task_id, "error": task.expl, 'displayName': task.pretty_id}
         else:
             return {"taskId": task_id, "error": ""}
+
+    def set_task_status_message(self, task_id, status_message):
+        if self._state.has_task(task_id):
+            task = self._state.get_task(task_id)
+            task.status_message = status_message
+
+    def get_task_status_message(self, task_id):
+        if self._state.has_task(task_id):
+            task = self._state.get_task(task_id)
+            return {"taskId": task_id, "statusMessage": task.status_message}
+        else:
+            return {"taskId": task_id, "statusMessage": ""}
 
     def _update_task_history(self, task, status, host=None):
         try:
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -289,6 +289,9 @@ class Task(object):
         self.task_id = task_id_str(self.task_family, self.to_str_params(only_significant=True))
         self.__hash = hash(self.task_id)
 
+        self.set_tracking_url = None
+        self.set_status_message = None
+
     def initialized(self):
         """
         Returns ``True`` if the Task is initialized and ``False`` otherwise.
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -510,6 +510,8 @@ class Task(object):
         """
         Sets the status message of the task to message, i.e., invokes _status_message_callback if it
         is a callable. This propagates the message down to the scheduler.
+
+        See :ref:`Task.set_status_message`
         """
         if hasattr(self._status_message_callback, "__call__"):
             self._status_message_callback(message)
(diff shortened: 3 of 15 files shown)


riga:
Todo's from discussion in #1621:
- Remove ``status_message`` cache variable
- Add pydocs

luigi/task.py
```diff
@@ -504,6 +506,14 @@ def on_success(self):
         Default behavior is to send an None value"""
         pass
 
+    def set_status_message(self, message):
+        """
+        Sets the status message of the task to message, i.e., invokes _status_message_callback if it
+        is a callable. This propagates the message down to the scheduler.
```
Tarrasch: In the docs here, can you add a reference the ` _Task.set_status_message:` section? I think there are a few example in this file you can copy paste.

Tarrasch:
Cool. This looks totally merge-worthy. Feel free to fix the minor doc-comment.

riga:
Done ;)

Tarrasch:
Nice!

erikbern:
can you resolve the failing tests? https://travis-ci.org/spotify/luigi/jobs/120027335

riga:
Yep, working on it

riga:
Ok, this is a tough one.

``update_status_message`` is created at worker-level because it needs information on the scheduler. It is then passed to the task process where it is stored as ``_status_message_callback`` in the task itself. This is the reason why it can't be pickled.

This is not the case with ``tracking_url_callback`` since it's not saved in the task, but rather passed to its run method in ``TaskProcess._run_get_new_deps``. I afraid that the status message callback should use the same mechanism here.

Here is a suggestion for changes in [worker.py](https://github.com/spotify/luigi/blob/d7464bd5e6bc24c0f2881109e0a8722a0e42951c/luigi/worker.py) that would cleanup this mechanism in order to cope for other callbacks (like update_status_message) as well (and would solve the "problem" that @Tarrasch mentioned [here](https://github.com/spotify/luigi/pull/1621#discussion-diff-58156810)):

```python
from inspect import getargspec

class TaskProcess(multiprocessing.Process):

    def __init__(self, <current kw/args except tracking_url_callback>, run_callbacks=None):
        ...
        if run_callbacks is None:
            run_callbacks = {}
        self.run_callbacks = run_callbacks

def _run_get_new_deps(self):
    run_spec = getargspec(self.task.run)
    if run_spec.keywords:
        run_kwargs = self.run_callbacks.copy()
    else:
        run_kwargs = {key: cb for key, cb in self.run_callbacks.items() if key in run_sepc.args}

    task_gen = self.task.run(**run_kwargs)

    if not isinstance(task_gen, types.GeneratorType):
        return None
    ...
```

This way, the code in [``Worker._create_task_process``](https://github.com/spotify/luigi/blob/d7464bd5e6bc24c0f2881109e0a8722a0e42951c/luigi/worker.py#L721) becomes:

```python
def _create_task_process(self, task):
    ...
    run_callbacks = {
        "tracking_url_callback": update_tracking_url,
        "status_message_callback": update_status_message
    }

    return TaskProcess(
        task, self._id, self._task_result_queue,
        random_seed=bool(self.worker_processes > 1),
        worker_timeout=self._config.timeout,
        run_callbacks=run_callbacks
    )
```

I'm not a fan of using ``inspect`` too much myself, but in this case it's quite robust. I can add this to the PR if you want.

erikbern:
I remember running into similar issues before

What's the reason we pickle the Tasks in the first place?

riga:
I think parallel scheduling requires that tasks can be pickled. At least it's mentioned in the [docs](http://luigi.readthedocs.org/en/stable/configuration.html#core) at "parallel-scheduling".

Many contrib tasks also make using of it, e.g. in [contrib/spark.py](https://github.com/spotify/luigi/blob/d7464bd5e6bc24c0f2881109e0a8722a0e42951c/luigi/contrib/spark.py#L287).

erikbern:
Hm right makes sense. Tasks are in theory interchangeable given the same name and parameters so instead of pickling you could just de/serialize it using that mechanism (that's how the assistant works). It would be quite easy to fix in worker.py

But you are right that some tasks use pickling (eg Hadoop mapreduce) so it would be mess to avoid pickling everywhere.

Wouldn't your suggestion with inspect run into the same issue when contrib tasks use pickle?

riga:
> Wouldn't your suggestion with inspect run into the same issue when contrib tasks use pickle?

The ``run_callbacks`` would be owned by the ``TaskProcess`` instance, the task itself still doesn't know anything about the callbacks. I just ran the tests, looks good.

riga:
Actually the use of inspect has another advantage. Users might use decorators on Task.run, which are called twice in the current implementation if tracking_url_callback is missing in the signature of run.

dlstadther:
Due to the changes in worker.py, do any of its tests need to be added/updated?

riga:
LGTM. [``test_tracking_url``](https://github.com/spotify/luigi/blob/20f7ce0efa02e547ef9767642652c0871c1ec09c/test/worker_test.py#L269) and [``test_type_error_in_tracking_run``](https://github.com/spotify/luigi/blob/20f7ce0efa02e547ef9767642652c0871c1ec09c/test/worker_test.py#L290) seem to be the only tests that make use of ``tracking_url_callback``. The changes to ``TaskProcess`` and ``Worker`` should be pretty much covered by many other test cases.

test/task_status_message_test.py
```diff
@@ -0,0 +1,41 @@
...
+from helpers import LuigiTestCase
+
+import luigi
+import luigi.scheduler
+import luigi.worker
+
+luigi.notifications.DEBUG = True
+
+
+class TaskStatusMessageTest(LuigiTestCase):
```
Tarrasch: Actually we do have a test suite which is run in two modes automatically (in-memory scheduler and rpc to external process scheduler). Can you make sure that the rpc code path is also tested?

We do also have a set of scheduler tests (they are disabled in Travis for flakiness reasons), but can you add a that case there to and make sure it passes locally?

Tarrasch:
We do also have a set of scheduler tests (they are disabled in Travis for flakiness reasons), but can you add a that case there to and make sure it passes locally?

doc/tasks.rst
```diff
@@ -178,6 +178,24 @@ In other words, you should make sure your Task.run_ method is idempotent.
...
+``status_message_callback`` which is passed opportunistically to Task.run_:
+
+.. code:: python
+
+    class MyTask(luigi.Task):
+        def run(self, status_message_callback):
+            for i in range(100):
+                # do some hard work here
+                if i % 10 == 0:
+                    status_message_callback("Progress: %d / 100" % i)
```
Tarrasch: Is this still correct?! It thought you pass a dict now

luigi/worker.py
```diff
@@ -727,11 +730,19 @@ def update_tracking_url(tracking_url):
                 tracking_url=tracking_url,
             )
 
+        def update_status_message(message):
+            self._scheduler.set_task_status_message(task.task_id, message)
+
+        run_callbacks = {
+            "tracking_url_callback": update_tracking_url,
+            "status_message_callback": update_status_message
+        }
```
Tarrasch: I imagined this being a `__call__`able object where it by default does "update_tracking_url" and prints a deprecation warning.

luigi/worker.py
```diff
@@ -94,27 +95,29 @@ class TaskProcess(multiprocessing.Process):
...
-            if 'unexpected keyword argument' not in str(ex):
-                raise
-            run_again = True
-        if run_again:
-            task_gen = self.task.run()
+        run_spec = inspect.getargspec(self.task.run)
+        if run_spec.keywords:
+            run_kwargs = self.run_callbacks.copy()
+        else:
+            run_kwargs = {key: cb for key, cb in self.run_callbacks.items() if key in run_spec.args}
```
Tarrasch: This is like ... too much magic.

Tarrasch:
I appreciate that you introduce inspect to make `@decorators` not be called twice. But may feel free to skip that change for a different PR if that simplifies things. Such a change for example requires tests that indeed show that decorators are only called once.

what about making the dict callable as I suggested in an inline comment? We would keep it to only one positional parameter for run then. Less magic, no inspection of argument-name.

doc/tasks.rst
riga: Keyword argument unpacking also works for arguments, so

```python
def func(a, b):
    print(a, b)
func(**{"b": "bVal", "a": "aVal"})
# -> ("aVal", "bVal")
```

riga:
Ok. How about this solution:

```python
class TaskProcess(multiprocessing.Process):

    def __init__(self, ..., tracking_url_callback=None, status_message_callback=None):
        ...
        self.status_message_callback = status_message_callback
        ...

    def _run_get_new_deps(self):
        self.task.set_tracking_url = self.tracking_url_callback
        self.task.set_status_message = self.status_message_callback

        def deprecated_tracking_url_callback(*args, **kwargs):
            warnings.warn("tracking_url_callback in run() args is deprecated, use "
                          "set_tracking_url instead.", DeprecationWarning)
            self.tracking_url_callback(*args, **kwargs)

        run_again = False
        try:
            task_gen = self.task.run(tracking_url_callback=deprecated_tracking_url_callback)
        except TypeError as ex:
            if 'unexpected keyword argument' not in str(ex):
                raise
            run_again = True
        if run_again:
            task_gen = self.task.run()

        self.task.set_tracking_url = None
        self.task.set_status_message = None
        ...
```

The callbacks are set right before run() is called, and then reset. This way pickling is not affected, it's backward compatible, and at some point the try-except block can be refactored.

Tarrasch:
Sure. But maybe lets hear from @daveFNbuck first, perhaps there were other reasons he opted for the extra-arg in `run()`.

Though I'm very positive to this. It feels a bit hasted now to have introduced that extra arg in `run()`.

riga:
> Also, of course you'll have to change the places in the code-base that uses run(tracking_url_callback) already, but it seems like your new suggested interface is pretty easy to use. :)

Yep, already done that, mainly in hadoop, hadoop_jar, hive and scaling) =) I'm preparing the commits right now.

riga:
I changed [hadoop.JobTask.dump](https://github.com/spotify/luigi/blob/559df529b455834fc79467aaebeb5176c8b43748/luigi/contrib/hadoop.py#L914) to disregard callbacks, as they're not need in the deserialized job anyway. E.g. the tracking url is not set actively but parsed from the hadoop job stderr.

All tests pass now.

daveFNbuck:
This looks like a good alternative to my solution, and more scalable. I see that the example attached here is of a progress percentage. If this is the main use case, I've been thinking it would be nice to add an optional progress bar that can be automatically displayed in the visualizer. That would be a bit better than having a popup to show a single number.

riga:
> it would be nice to add an optional progress bar that can be automatically displayed in the visualizer

Good idea. Maybe I can add an additional PR that implements a progress bar. However, I think there are more use cases than just sending the progress. We use it for important intermediate output and to some extent for a summary of final results. Maybe one can parse the status message with a regexp (e.g. ``new RegExp('^progress\:\s(\d+)\%.*$', 'i')``) to decide whether to show a progress bar or a pop up.

test/worker_test.py
```diff
@@ -305,6 +305,27 @@ def run(self, tracking_url_callback=None):
...
+        tracking_url = 'http://test_url.com/'
+
+        class A(Task):
+            has_run = False
+
+            def complete(self):
+                return self.has_run
+
+            def run(self):
+                if self.set_tracking_url is not None:
```
Tarrasch: Wait. Do the user really have to do this check before using the tracking_url? Can't we somehow guarantee that it's always present?

Tarrasch:
This looks very good. Thank you so much for doing this! Just see my inline comment/worry.

Other than that. This is good to merge right?

riga:
> Can't we somehow guarantee that it's always present?

Yep, [Worker._create_task_process](https://github.com/spotify/luigi/blob/559df529b455834fc79467aaebeb5176c8b43748/luigi/worker.py#L721) is the only place where ``TaskProcess``es are created, and the callbacks are always present/created here. I made some changes to reflect that. Of course, outside the ``run()`` scope, users will get an exception when a callback is used, but I think that's fine/wanted.

Tarrasch:
@riga, ok, is this "fine to merge" now you think?

riga:
Yes =)

Tarrasch:
Ok. Let's wait for test results and then I'll merge. :)

riga:
@Tarrasch Looks good, glad I could help.

luigi/scheduler.py
```diff
@@ -870,6 +871,7 @@ def _serialize_task(self, task_id, include_deps=True, deps=None):
             'priority': task.priority,
             'resources': task.resources,
             'tracking_url': getattr(task, "tracking_url", None),
+            'status_message': task.status_message
```
Tarrasch: This line is causing a stack-trace like this:

```
  File "/zserver/Python-3.4.3/lib/python3.4/site-packages/tornado/web.py", line 1443, in _execute
  File "/zserver/apps/luigi/code-for-luigid/luigi/server.py", line 101, in get
  File "/zserver/apps/luigi/code-for-luigid/luigi/scheduler.py", line 986, in task_list
    serialized = self._serialize_task(task.id, False)
  File "/zserver/apps/luigi/code-for-luigid/luigi/scheduler.py", line 868, in _serialize_task
    'status_message': task.status_message
```

For anyone using an old pickled state-file.

riga: Do we need to do anything to fix this?

Tarrasch: I think you can copy the line above. Make it like 

`getattr(task, "status_message", None)` instead of `task.status_message`. Then a couple of months later we're changing it back to `task.status_message`. Do you see how it'll work? Do you mind submitting a PR (you can use the online editor)

riga: I created PR #1645 for this issue.

michcio1234:
@riga @Tarrasch 
Hey, sorry for notifying you in something that old. I don't want to open a new issue since I'm not sure if anything is wrong but...
where can I see a status message that I set using `self.set_status_message('msg')`? Can't find it in the scheduler GUI...

riga:
Hi @michcio1234 ,
the status messages are visible in the scheduler GUI. You just have to click on the chat/message icon in the "Actions" column in the task table. If a task has no status message, that icon won't be visible.

PR #1628, opened 2016-04-02
Added module level documentation for a couple use cases of luigi.util

drj42:
Use case documentation for inherits and requires in luigi.util


erikbern:
wow i had forgotten about those decorators! i implemented them like 5 years ago :)

erikbern:
lgtm

PR #1630, opened 2016-04-03
Clone from non-parameters

ChrisBeaumont:
Currently `Task.clone` only propagates values from source to target when both attributes are Parameters.    This PR propagates data from source to target parameter regardless of whether the source attribute is a parameter or something else (property, class or instance-level attribute, etc). 

The use case for this has come up in cases where we interface between generic and specific tasks. A simplified example, consider some generic classes to POST several files referenced in a manifest to an external url:

``` python
class Upload(luigi.Task):
    url = luigi.Parameter()
    file = luigi.Parameter()

class UploadFromManifest(luigi.Task):
    url = luigi.Parameter()
    manifest = luigi.Parameter()

    def run(self):
        files = open(manifest).readlines()
        yield [self.clone(cls=Upload, file=file) for file in files]   
```

as well as a more specific subtask where url is derived from a more fundamental parameter:

``` python
class MonthlyReport(UploadFromManifest):
    month = luigi.Parameter()
    manifest = luigi.Parameter()

    @property
    def url(self):
        return 'http://test.com/reports/%s' % self.month
```

This fails at the clone step because, although url is defined for `MonthlyReport`, it's not a parameter and thus not propagated to `Upload`

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -334,16 +334,15 @@ class Task(object):
         :param kwargs:
         :return:
         """
-        k = self.param_kwargs.copy()
-        k.update(six.iteritems(kwargs))
-
         if cls is None:
             cls = self.__class__
 
         new_k = {}
         for param_name, param_class in cls.get_params():
-            if param_name in k:
-                new_k[param_name] = k[param_name]
+            if param_name in kwargs:
+                new_k[param_name] = kwargs[param_name]
+            elif hasattr(self, param_name):
+                new_k[param_name] = getattr(self, param_name)
 
         return cls(**new_k)
 
--- a/test/clone_test.py
+++ b/test/clone_test.py
@@ -68,3 +68,20 @@ class CloneTest(unittest.TestCase):
         t = PowerSum(lo=42, hi=45, p=2)
         luigi.build([t], local_scheduler=True)
         self.assertEqual(t.s, 42 ** 2 + 43 ** 2 + 44 ** 2)
+
+    def test_inheritance_from_non_parameter(self):
+        """
+        Cloning can pull non-source-parameters from source to target parameter.
+        """
+
+        class SubTask(luigi.Task):
+            lo = 1
+
+            @property
+            def hi(self):
+                return 2
+
+        t1 = SubTask()
+        t2 = t1.clone(cls=LinearSum)
+        self.assertEqual(t2.lo, 1)
+        self.assertEqual(t2.hi, 2)
--- a/test/decorator_test.py
+++ b/test/decorator_test.py
@@ -242,14 +242,6 @@ class Y2(luigi.Task):
     pass
 
 
-@inherits(X)
-class Z(luigi.Task):
-    n = None
-
-    def requires(self):
-        return self.clone_parent()
-
-
 @requires(X)
 class Y3(luigi.Task):
     n = luigi.IntParameter(default=43)
@@ -262,9 +254,6 @@ class CloneParentTest(unittest.TestCase):
         x = X()
         self.assertEqual(y.requires(), x)
         self.assertEqual(y.n, 42)
-
-        z = Z()
-        self.assertEqual(z.requires(), x)
 
     def test_requires(self):
         y2 = Y2()


erikbern:
not sure about this. it seems too magic and i think it could lead to unexpected behavior later.

what's the issue you're trying to solve here? your example isn't very useful because `url` is directly derived from `month`

ChrisBeaumont:
> it seems too magic and i think it could lead to unexpected behavior later.

Yeah maybe, the failing tests (which I haven't digested yet) might suggest this as well. Here's a slightly more fleshed out use case to describe where I'm coming from:

In practice the Upload and UploadFromManifest tasks (which I actually do use), have several more parameters:

```
class RequestMixin(object):
    url = luigi.Parameter()
    headers = DictParameter(default={})
    params = DictParameter(default={})
    username = luigi.Parameter(default=None, significant=False)
    password = luigi.Parameter(default=None, significant=False)

class Upload(luigi.Task, RequestMixin):
    payload_file = luigi.Parameter()

    def run(self):
        # builds and sends request, saves response content as task output

class UploadFromManifest(luigi.Task, RequestMixin):
    manifest = luigi.Parameter()
    def run(self):
         yield [self.clone(cls=Upload, payload_file=file) for file in open(self.manifest).readlines()]
```

we use these the generic UploadFromManifest in several different pipelines to transfer data between services. In most concrete cases all of the parameters in in RequestMixin are either known at task-definition time, or derived from a more context-specific parameter. So they might be specifiable as something like:

```
class UploadToServiceFoo(UploadFromManifest):
    release_number = luigi.IntParameter()

    # static for this task
    username = settings.FooServiceUser
    password = settings.FooServicePassword
    headers = {'Content-Type': 'text/csv'}
    params = {'uploader: 'luigi'}
  
    #url derives from release_number
    @property
    def url(self):
        return "http://foo.com/%s" % self.release_number
```

This PR makes the above code functional, and I rather like its flat+declarative structure. Without it you'd need to write:

```
class UploadToService(luigi.WrapperTask):
    <same declarations as above>

    def requires(self):
         yield UploadManifest(username=self.username, password=self.password, headers=self.headers, params=self.params, url=self.url)
```

ie you lose the intended benefits of clone (not having to re-enumerate every parameter when describing parameters). Furthermore the first example throws an exception like `missing required parameter url`, which surprised me since url is clearly defined by the task.

Another small consideration:  the following code works currently (maybe you'd consider it an antipattern, I'm not sure):

```
class SpecificUpload(Upload):
    username = settings.username
    password = settings.password
    headers = {'Content-Type':'text/csv'}    
    url = 'http://foo.com'
```

In other words, overriding parameters with class-level attributes or properties currently works, but propagating these overrides through clone does not. That feels a little asymmetrical to me. 

However, maybe you disagree, or there's another mechanism for addressing this use case that I'm missing?

Tarrasch:
I'm quite positive to this change and I find this to make sense. We want the clone-behavior for Parameter-attributes to feel as similar as possible for other attributes.

Let me know what insights you can get from the failed test case.

ChrisBeaumont:
@Tarrasch @erikbern ok the [problematic test](https://github.com/spotify/luigi/blob/20f7ce0efa02e547ef9767642652c0871c1ec09c/test/decorator_test.py#L266-L267) case boils down to this:

```python
class X(luigi.Task):
    n = luigi.IntParameter(default=42)

@inherits(X)
class Z(luigi.Task):
    n = None

    def requires(self):
        return self.clone_parent()

assert Z().requires().n == 42
```

or more simply
```python
class X(luigi.Task):
    n = luigi.IntParameter(default=42)

class Z(luigi.Task):
    n = None

assert Z().clone(cls=X).n == 42
```

It was added in [this commit](https://github.com/spotify/luigi/commit/be88a318dce06254a66303a1c398ec28f89c60d7) by @erikbern 

The intention of this PR is to change that behavior, so that `n=None` is treated as an override. Personally I find the behavior in this PR to be easier to reason about. However I don't know what the original intention of the failing test is, or whether it implies that other user code makes the same assumption as this test does. It's easy to rewrite the test failure so that it passes, but will wait to hear from you whether you think this behavior change is appropriate.

erikbern:
So the real issue you are trying to solve here is you want a subclass to be able to override base class parameter and fix them to certain value?

I haven't thought of this use case

ChrisBeaumont:
> So the real issue you are trying to solve here is you want a subclass to be able to override base class parameter and fix them to certain value?

Well this is already possible -- a subclass can override base class parameters with constants (class attributes or properties). I'm proposing the override behavior for `clone` should behave the same way as it does for subclassing.

ChrisBeaumont:
> I haven't thought of this use case

IMO the utility of this comes into play when you extract out common generic tasks. Because they are general-purpose, they often need several parameters to be fully specified. But often times those settings should not be considered "parametrize-able" in the downstream, special-purpose task. For example the `Upload` task takes `url` as a parameter, but in a downstream task `UploadToFoo` I don't want the user to be able to change the value of `url` from the command line -- maybe it always needs to be `foo.com`, or else the implementation of `UploadToFoo` doesn't make sense. Setting `url=foo.com` at the class level is a nice way to do this, and it would be nice if `self.clone(Upload)` passed this value along for me.

Tarrasch:
@ChrisBeaumont any updates on this?

ChrisBeaumont:
@Tarrasch in my opinion this PR is "final", and the question is whether you and @erikbern think it's appropriate to merge. If it's relevant we added this feature to Counsyl's Luigi installation several months ago, and have found it useful.

Tarrasch:
Thanks @ChrisBeaumont. After reading this once and twice. It's clear to me that this new behavior is better and makes more sense. Usually the luigi dictators are pretty hesitant to changes to luigi core hence the prolonged discussion. I hope it wasn't discouraging and sorry for the very long delay in merging!

PR #1631, opened 2016-04-05
Add explicit whitelist of RPC commands for luigid

bwall:
Limit calls on Scheduler from API to only those required for functionality.

Basic functionality testing was conducted, but not extensive testing.

### DIFF
--- a/luigi/server.py
+++ b/luigi/server.py
@@ -68,6 +68,26 @@ class RPCHandler(tornado.web.RequestHandler):
         self._scheduler = scheduler
 
     def get(self, method):
+        if method not in [
+            'add_task',
+            'add_worker',
+            'dep_graph',
+            'disable_worker',
+            'fetch_error',
+            'get_work',
+            'graph',
+            'inverse_dep_graph',
+            'ping',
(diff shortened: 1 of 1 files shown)


erikbern:
nice!

luigi/server.py
```diff
@@ -68,6 +68,26 @@ def initialize(self, scheduler):
...
+            'add_task',
+            'add_worker',
+            'dep_graph',
+            'disable_worker',
+            'fetch_error',
+            'get_work',
+            'graph',
+            'inverse_dep_graph',
+            'ping',
+            'prune',
```
Tarrasch: Not sure if this actually should be public. But whatever.

Tarrasch:
Nice! Thanks :)

luigi/server.py
Tarrasch: To clarify. I'm not sure it makes sense for clients to be able to call `prune`. Many other of the API calls does call `prune` internally themselves, like `graph`, `inverse_dep_graph` etc.

This was just a minor comment so I just merged right away.

PR #1632, opened 2016-04-05
Specify extra spark-submit arguments

smartkiwi:
Currently luigi.contrib.spark has limited support for spark-submit arguments, doesn't support arbitrary spark-submit arguments.

This patch adds the ability for luigi spark task runner to specify custom spark config options (the ones that are not defined in luigiu.contrib.spark). Doing so enables me to schedule spark apps to mesos cluster and be able to specify Docker image to be used to distribute application code and dependencies.

### DIFF
--- a/luigi/contrib/spark.py
+++ b/luigi/contrib/spark.py
@@ -161,8 +161,20 @@ class SparkSubmitTask(ExternalProgramTask):
     def program_args(self):
         return self.spark_command() + self.app_command()
 
+    def extra_spark_submit_args(self):
+        """
+        Override this function to specify extra spark-submit command line arguments.
+        Any arguments that are not supported by luigi.contrib.spark
+
+        (i.e. spark.mesos.executor.docker.image)
+        Returns: list of strings
+
+        """
+        return []
+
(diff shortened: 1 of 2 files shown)


deuxpi:
The only options I can see on Spark 1.6.1 that are missing from `spark_command` are

- `exclude_packages`
- `repositories`
- `proxy-user`
- `verbose`
- `kill` (not sure it makes sense)
- `status` (same thing as `kill`)
- `principal`
- `keytab`

All of those are pretty much uncommon use cases. By the way, you can add as many configuration options you need with the `conf` property (we use it in production to pass spark.yarn.executor.memoryOverhead).

smartkiwi:
Maybe I'm missed some existing functionality:

I need to be able specify extra configuration parameters i.e.
spark-submit --conf spark.mesos.executor.docker.image=<> --conf spark.mesos.coarse=true ...

smartkiwi:
Yep I missed the existing functionality indeed.
There is already way to specify --conf parameters with .conf
https://github.com/spotify/luigi/blob/master/luigi/contrib/spark.py#L175

PR #1633, opened 2016-04-06
Allow central scheduler logging to be configured based on the logging…

brianestlin:
….conf file specified in luigi.cfg.

Although the luigi configuration file has an option to specify a logging conf file, this option currently only applies to the workers, not the central scheduler, which AFAICT ignores any logging configuration and starts up with a basic configuration. I poked around and didn't see any way to configure the central scheduler's logging other than hacking on its entry point; hence this PR.

### DIFF
--- a/luigi/cmdline.py
+++ b/luigi/cmdline.py
@@ -48,7 +48,9 @@ def luigid(argv=sys.argv[1:]):
                 if logging_conf is not None and not os.path.exists(logging_conf):
                     raise Exception("Error: Unable to locate specified logging configuration file!")
             if logging_conf is not None:
+                print("Configuring logging from file: {}".format(logging_conf))
                 logging.config.fileConfig(logging_conf)
             else:
+                print("Defaulting to basic logging; consider specifying logging_conf_file in luigi.cfg.")
                 logging.basicConfig(level=logging.INFO, format=luigi.process.get_log_format())
         luigi.server.run(api_port=opts.port, address=opts.address, unix_socket=opts.unix_socket)
--- a/luigi/cmdline.py
+++ b/luigi/cmdline.py
@@ -1,6 +1,7 @@
 import os
 import argparse
 import logging
+import logging.config
 import sys
 
 from luigi.retcodes import run_with_retcodes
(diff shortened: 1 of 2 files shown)


dlstadther:
+ tests? https://github.com/spotify/luigi/blob/master/test/cmdline_test.py

Tarrasch:
Haha, I just needed this last Monday! I got surprised nobody had sent a PR yet. You're two days late! :D

luigi/cmdline.py
```diff
@@ -40,5 +41,14 @@ def luigid(argv=sys.argv[1:]):
...
                                 filename=os.path.join(opts.logdir, "luigi-server.log"))
         else:
-            logging.basicConfig(level=logging.INFO, format=luigi.process.get_log_format())
+            config = luigi.configuration.get_config()
+            logging_conf = None
+            if not config.getboolean('core', 'no_configure_logging', False):
+                logging_conf = config.get('core', 'logging_conf_file', None)
+                if logging_conf is not None and not os.path.exists(logging_conf):
+                    logging_conf = None
+            if logging_conf is not None:
```
Tarrasch: This code block can be simplified I think. Why first `logging_conf = None` and then `if logging_conf is not None`. Just introduce the variable in the code path that you use it.

Also, can you please also add `print()` statements saying `"Loading logging conf from file"` or `"Defaulting to only logging to stdout. You're highly advised to setup logging"`? Something like that.

Tarrasch:
Also, can you please also add `print()` statements saying `"Loading logging conf from file"` or `"Defaulting to only logging to stdout. You're highly advised to setup logging"`? Something like that.

luigi/cmdline.py
dlstadther: Since L47-L49 may not execute, the variable `logging_conf` needs to be defined as something in order to check if it is not `None` (L50), right? (thus why the variable declaration at L45). At least, that's how i'm thinking through it

Tarrasch: Eh.. Yea.

Actually I think the server should just plain out die and not start if the condition `logging_conf is not None and not os.path.exists(logging_conf)` holds true.

brianestlin:
@dlstadther "What is it that the central scheduler could log that you want to see?"
This came up for me recently when I wanted to turn on SQLAlchemy logging of the database statements made to the task history database.

luigi/cmdline.py
brianestlin: Changed to raise exception if the logging conf file is missing.

brianestlin:
Added unit tests and print statements. PTAL. Also let me know if you want me to squash the commits.

Tarrasch:
Looks good to me. I will squash-accept this PR so no need to merge commits.

PR #1634, opened 2016-04-06
Add scheduled tasks to error emails

neilisaac:
Request for comments

Sometimes my team shares base tasks.  When we get an error email, it's sometimes unclear which upstream task requires() the one that failed. In order to address this, I'm proposing adding the worker's scheduled tasks to the error email.

Some things to consider:

- Is one worker's view on scheduled tasks wide enough to capture all the relevant incomplete tasks?
- Is sorting by task_id meaningful?
- Is there a more general way to show the task tree (ex. execution_summary.summary?)

For my proof-of-concept:

```
import luigi

class SubTask(luigi.Task):
    p = luigi.Parameter()
    def run(self):
        raise Exception('I meant to do that...')

class TestTask(luigi.Task):
    def requires(self):
        yield SubTask('some option')

if __name__ == '__main__':
    luigi.run(main_task_cls=TestTask, local_scheduler=True)
```

produces a plain-text email body:

```
Name: SubTask

Parameters:
  p: some option

Scheduled tasks:
- SubTask(p=some option)
- TestTask()

Runtime error:
  File "/home/nisaac/code/luigi/luigi/worker.py", line 171, in run
  File "/home/nisaac/code/luigi/luigi/worker.py", line 113, in _run_get_new_deps
    task_gen = self.task.run()
  File "testtask.py", line 9, in run
    raise Exception('I meant to do that...')
Exception: I meant to do that...
```

### DIFF
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -315,13 +315,14 @@ def _prefix(subject):
     return subject
 
 
-def format_task_error(headline, task, formatted_exception=None):
+def format_task_error(headline, task, formatted_exception=None, scheduled_tasks=[]):
     """
     Format a message body for an error email related to a luigi.task.Task
 
     :param headline: Summary line for the message
     :param task: `luigi.task.Task` instance where this error occurred
     :param formatted_exception: optional string showing traceback
+    :param scheduled_tasks: list of scheduled tasks
 
     :return: message body
     """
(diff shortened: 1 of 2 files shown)


Tarrasch:
I think the number of tasks can be quite big, so some pretty printing has to be done. And I think the execution_summary will be confusing as you're displaying it midway.

Have you considered to just display the arguments that luigi/python-interpreter was run with, alternatively only the root task? Won't that cover what you want?

neilisaac:
You're right; I think it would make more sense to just show it's parent tasks.

In the mean time, command line arguments would help too: https://github.com/spotify/luigi/pull/1637

Tarrasch:
Well. Yea, root task rather. Let's close this in favor of the simpler and probably better #1637. Thanks for taking time to iterate on a solution. :)

PR #1636, opened 2016-04-07
luigid logging: Log less and change logger name

Tarrasch:
I think it makes sense to log less, because currently I think INFO level
should not be verbose. In particular "No workers connected" message is
verbose and going to be said many more times than it will be actually
removed (when used with assistants).

Further, I think now is a sane time to change the name of the logger to
the "recommended" `getLogger(__name__)` style. Because PR
spotify/luigi#1633 was just merged and it's less likely that people have
relied on the logger name prior to that.


erikbern:
lgtm

PR #1637, opened 2016-04-07
Include sys.argv in error emails

neilisaac:
Sometimes my team shares base tasks. When we get an error email, it's sometimes unclear which upstream task requires() the one that failed.

Following @Tarrasch's suggestion of adding command line arguments to address this.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -34,6 +34,7 @@ import logging
 import multiprocessing  # Note: this seems to have some stability issues: https://github.com/spotify/luigi/pull/438
 import os
 import signal
+import subprocess
 import sys
 
 try:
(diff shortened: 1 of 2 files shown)


neilisaac:
Example email body:

```
A task failed when running. Most likely run() raised an exception.

Name: SubTask

Parameters:
  p: some option

Command line:
  'testtask.py' '--local-scheduler' 'TestTask' '--p' 'asdf'

Runtime error:
  File "/home/nisaac/code/luigi/luigi/worker.py", line 172, in run
  File "/home/nisaac/code/luigi/luigi/worker.py", line 114, in _run_get_new_deps
    task_gen = self.task.run()
  File "testtask.py", line 9, in run
    raise Exception('Intentional failure')
Exception: Intentional failure
```

luigi/worker.py
```diff
@@ -509,7 +510,8 @@ def _email_task_failure(self, task, formatted_traceback):
 
     def _email_error(self, task, formatted_traceback, subject, headline):
         formatted_subject = subject.format(task=task, host=self.host)
-        message = notifications.format_task_error(headline, task, formatted_traceback)
+        command = ' '.join("'{}'".format(arg) for arg in sys.argv)
```
Tarrasch: This makes the command look weird and hard to read I think. What about simply

```
command = subprocess.list2cmdline(sys.argv)
```

Tarrasch:
LGTM. But the formatting looks strange to me. See my inline comment suggestion. :)

Tarrasch:
LGTM! Will merge on green :)

PR #1639, opened 2016-04-08
Bugfix: Scheduler crashes for some configs

Tarrasch:
When the 'disable_hard_timeout' is set but 'disable_failures' is not.

The fix is simply to follow the "good" coding practice to not allow
IntParameters to have a None-value as default.


erikbern:
lgtm

PR #1640, opened 2016-04-08
scheduler: Remove unused prune_done_tasks option

Tarrasch:


erikbern:
lgtm

erikbern:
why "in progress"?

Tarrasch:
@econshick, the bot you've added that add that the label "in progress" seems to only add negative value for now. Can you either help us understand it better or remove it?

PR #1643, opened 2016-04-09
Fix bug in LocalFileSystem.move

ChrisBeaumont:
0bdb009aedf6307fcd066c73f19e8bf96876a183 refactored the implementation of `move` from `LocalTarget` to `LocalFileSystem` but didn't repoint `self.fs.mkdir` to `self.mkdir`. This causes code like ```LocalTarget(src).move(dest)``` to fail if `dest` is in a nonexistent directory.

This adds a fix + regression test


Tarrasch:
Thanks!

PR #1644, opened 2016-04-10
Raise Queue.Empty for get() on empty queue

ej81:
Worker._handle_next_task() expects get() on an empty queue to raise Queue.Empty, but the dummy implementation DequeQueue raises IndexError instead. It probaly only shows up if you create custom workers that derive from Worker, but it is still inconsistent with multiprocessing.Queue.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -271,7 +271,10 @@ class DequeQueue(collections.deque):
         return self.append(obj)
 
     def get(self, block=None, timeout=None):
-        return self.pop()
+        try:
+            return self.pop()
+        except IndexError:
+            raise Queue.Empty
 
 
 class AsyncCompletionException(Exception):


erikbern:
out of curiosity what's the use case for a custom worker in your case?

ej81:
The reason is that we need Luigi to play nice with a cluster that is managed by LSF. We have many jobs with complex dependencies and we want to use Luigi to keep track of them. The only think the worker needs to do is move all tasks without missing dependencies into the LSF queue.

(I didn't want to use contrib.sge, because it creates too many processes on the login node.)

erikbern:
Interesting – and what were the issues with the standard worker that made you implement a custom worker? Asking because maybe we should fix those :)

Tarrasch:
Cool thanks! Did you try to add a test-case? Otherwise these fixes might get re-factored away ^^

ej81:
I want to run jobs on the cluster asynchronously without having an active child process on the login node for every job that is in the queue. The custom worker I use is basically the standard one, but I override _create_task_process() so that it submits the job without forking. In _handle_next_task() I synchronise the job status with the LSF queue before calling the superclass method.

It would probably be better to implement this in Task itself, but I don't think it can be done at the moment without modifying the Luigi code. It needs some refactoring of TaskProcess and Task. In that case the result_queue should be handled by Task and Task.run() should be allowed to return without considering the Task done or failed. If I have time I might try something and send another pull request to discuss.

@Tarrasch No, I didn't add a test case, but it could be useful indeed.

erikbern:
Yes – probably makes more sense to implement this on the Task class instead. Let me know if you need help!

PR #1645, opened 2016-04-11
Make status_message optional in task serialization.

riga:
This PR fixes an issue mentioned [here](https://github.com/spotify/luigi/pull/1625#discussion-diff-59153908).

When scheduler tasks are serialized, their ``status_message`` might not be present when using old state files. Just like ``tracking_url``, the presence of this member is now optional.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -865,7 +865,7 @@ class CentralPlannerScheduler(Scheduler):
             'priority': task.priority,
             'resources': task.resources,
             'tracking_url': getattr(task, "tracking_url", None),
-            'status_message': task.status_message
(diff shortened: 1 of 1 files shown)


Tarrasch:
Thanks for fixing this. It's better to just do this quick-fix than deals with a lot of bug reports about this later. :)

erikbern:
if this is breaking upgrading we should push a new version to pypi quickly

any volunteers? @ulzha @Tarrasch

Tarrasch:
@erikbern, good thinking. Luckily there is no breakage for anybody who only jumps between our versions. The bug and this fix are both committed after 2.1.1. :)

Issue #1647, opened 2016-04-11
Cannot write to a binary file (LocalStorage)


wires:
Trying to implement a `msgpack` target (maybe not right approach...)

``` py
class MessagepackTarget(luigi.file.LocalTarget):
    def __init__(self, filename):
        super(MessagepackTarget, self).__init__(path=filename)

    def dump(self, d):
        with self.open('wb') as f:
            msgpack.pack(d, f)
```

This fails wether using mode `w` or `wb`:

```
  File "... python3.5/site-packages/msgpack/__init__.py", line 38, in pack
    stream.write(packer.pack(o))
TypeError: write() argument must be str, not bytes
```

I also tried with different format `super(MessagepackTarget, self).__init__(path=filename, format=GzipFormat)` but that fails as well:

```
  File "... python3.5/site-packages/luigi/file.py", line 126, in open:
    return self.format.pipe_writer(atomic_file(self.path))
TypeError: pipe_writer() missing 1 required positional argument: 'output_pipe'
```

What am I doing wrong?

erikbern:
sounds like a python 3 thing.

can you figure out what's the underlying issue and send a pull request? much appreciated

wires:
I tried, including various work arounds, but I couldn't get it to work... I think maybe it's due to somehow opening the file in 'w' mode, here https://github.com/spotify/luigi/blob/master/luigi/target.py#L266

I will isolate an example demonstrating the issue!

erikbern:
I think Python 3 has both TextIO and BytesIO. We should probably pick one of them based on whether "b" is in the mode. Not sure

wires:
O_O ... I am not sure either, other than that I'm not enjoying this Python 3 transition ⛅

https://docs.python.org/3/library/io.html#binary-i-o <= on first sight it looks like you are right, picking `BytesIO` on `b` might do the trick.

wires:
OTOH, `FileIO` should be able to `write(bytes)` as well... (as it implements `RawIOBase`) 😕 https://docs.python.org/3/library/io.html#io.RawIOBase.write

wires:
```py
# test.py
import luigi

class Test(luigi.Task):
    def run(self):
        self.output().open('wb').write(b'1234')

    def output(self):
        return luigi.file.LocalTarget(path='/tmp/zzzz')

# python -m luigi --local-scheduler --module test Test
# => TypeError: write() argument must be str, not bytes
```

^ minimal example demonstrating the problem

```py
import io
[...]

PR #1648, opened 2016-04-11
[WIP] fix for  #1647 binary mode for LocalFile 

wires:
Work in progress

### DIFF
--- a/luigi/file.py
+++ b/luigi/file.py
@@ -35,6 +35,9 @@ class atomic_file(AtomicLocalFile):
     Also cleans up the temp file if close is not invoked
     """
 
+    def __init__(self, path, mode):
+        super(AtomicLocalFile, self).__init__(path, mode)
+
     def move_to_final_destination(self):
         os.rename(self.tmp_path, self.path)
 
@@ -123,7 +126,7 @@ class LocalTarget(FileSystemTarget):
         rwmode = mode.replace('b', '').replace('t', '')
         if rwmode == 'w':
             self.makedirs()
-            return self.format.pipe_writer(atomic_file(self.path))
+            return self.format.pipe_writer(atomic_file(self.path, mode))
 
         elif rwmode == 'r':
             fileobj = FileWrapper(io.BufferedReader(io.FileIO(self.path, mode)))
--- a/luigi/target.py
+++ b/luigi/target.py
@@ -260,10 +260,11 @@ class AtomicLocalFile(io.BufferedWriter):
     :class:`luigi.file.LocalTarget` for example
     """
 
-    def __init__(self, path):
+    def __init__(self, path, mode='w'):
         self.__tmp_path = self.generate_tmp_path(path)
         self.path = path
-        super(AtomicLocalFile, self).__init__(io.FileIO(self.__tmp_path, 'w'))
+        self.mode = mode
+        super(AtomicLocalFile, self).__init__(io.FileIO(self.__tmp_path, mode))
 
     def close(self):
         super(AtomicLocalFile, self).close()


luigi/file.py
```diff
@@ -35,6 +35,9 @@ class atomic_file(AtomicLocalFile):
     Also cleans up the temp file if close is not invoked
     """
 
+    def __init__(self, path, mode):
+        super(AtomicLocalFile, self).__init__(path, mode)
+
```
dlstadther: What happens if you remove this? Can't you still initialize this atomic_file class, since it's just AtomicLocalFile with an additional method (move_to_final_destination)?

luigi/target.py
```diff
@@ -260,10 +260,11 @@ class AtomicLocalFile(io.BufferedWriter):
     :class:`luigi.file.LocalTarget` for example
     """
 
-    def __init__(self, path):
+    def __init__(self, path, mode='w'):
         self.__tmp_path = self.generate_tmp_path(path)
         self.path = path
-        super(AtomicLocalFile, self).__init__(io.FileIO(self.__tmp_path, 'w'))
+        self.mode = mode
+        super(AtomicLocalFile, self).__init__(io.FileIO(self.__tmp_path, mode))
```
dlstadther: This should be `super(AtomicLocalFile, self).__init__(io.FileIO(self.__tmp_path, self.mode))` since you define `self.mode = mode`.

luigi/file.py
```diff
@@ -123,7 +126,7 @@ def open(self, mode='r'):
         rwmode = mode.replace('b', '').replace('t', '')
         if rwmode == 'w':
             self.makedirs()
-            return self.format.pipe_writer(atomic_file(self.path))
+            return self.format.pipe_writer(atomic_file(self.path, mode))
```
dlstadther: Shouldn't this reference `rwmode`, not `mode`?

luigi/file.py
wires: @dlstadther probably; I will try. Could you recomment a place to add a unit test for this (roughly the code I pasted in description on top)? I'm not familiar with the Luigi codebase

gpoulin:
Passing a `NopFormat()` should do the trick. The luigi target where only supporting text which was an issue when porting to python3. That was solved by making targets accept only binary but having format that convert text to binary. To preserve compatibility and behaviour between python 2 and python 3, the a default format that transform text to binary is added to a target. Passing explicitly [this](https://github.com/spotify/luigi/blob/master/luigi/format.py#L508) would prevent this behaviour and allow binary.

gpoulin:
FYI there even a [unit test](https://github.com/spotify/luigi/blob/master/test/target_test.py#L187) for that.

wires:
@gpoulin This is what I thought initially, but I couldn't get away with just passing `NopFormat()`, that raises `TypeError: pipe_writer() missing 1 required positional argument: 'output_pipe'`

```py
import luigi

class Nope(luigi.Task):
    def run(self):
        with self.output().open('wb') as f:
            f.write(b'123')

    def output(self):
        return luigi.file.LocalTarget(path='/tmp/zzzz', format=luigi.format.NopFormat)

if luigi.build([Nope()], workers=1, local_scheduler=True):
    print("success")
```

gpoulin:
The issue is that you are passing the class `luigi.format.NopFormat` and not an instance of this class. The following should work:

```python
import luigi

class Nope(luigi.Task):
    def run(self):
        with self.output().open('wb') as f:
            f.write(b'123')

    def output(self):
        return luigi.file.LocalTarget(path='/tmp/zzzz', format=luigi.format.Nop)

if luigi.build([Nope()], workers=1, local_scheduler=True):
    print("success")
```

wires:
@gpoulin !!! You are right! I'll check some more details, but that would void this PR (although that might want to make docs a bit clearer on this). It already seemed strange to me, for reasons explained in #1647

gpoulin:
I refactor `Format` before knowing about twitter's [Bijection](https://github.com/twitter/bijection) but the idea would be similar. I haven't had time to complete it, but the idea would be that every target would write/read a certain "type" and the `Format`s would allow to convert from one "type" to another. So for `LocalTarget` the accepted "type" would be `BinaryStream` and for a SQL DB it would be `Rows`. So dumbing a SQL DB table would simply require to connect the right format together to convert `Rows` to `BinaryStream` and restoring a table would simply require to invert this format.

dlstadther:
@wires If you're finished with your commits, please clean up your commit history. This will reduce the work required for the maintainers (although they can now squash) to accept/merge your PR.

Tarrasch:
@dlstadther, actually I think it's okay. I like to being forced to reword. It makes me think twice what every PR is actually doing, because I rephrase it in my own words.

dlstadther:
I stand corrected. Listen to @Tarrasch , in this setting, his word is law (or at least the first draft of it).

wires:
:ok_hand: Thanks for the comments, will fix up this PR asap.

But plz let me know if I'm correct on this: none of this code is needed.

Because, for `LocalTarget` to support `write(x : bytes)` all that is needed is to set `format=luigi.format.Nop`.

I find this slightly unexpected, because from my (l)user perspective I would expect `self.output().open('wb')` with `LocalTarget` to achieve the same...

So, some things I can do with this PR:

- do nothing on PR and close it, go back to work 😉 
- revert current code changes and add some documentation on using `luigi.format.Nop`
- throw helpful message when `format` is set to one that doesn't support `write(bytes)` (not sure if I can detect that always, maybe on write(bytes) exception and the default TextIO format, throw a more helpful error message)
- or maybe change default format with python3: I think `Nop` also works for `Text`, plus it's maybe good to be explicit about setting Text and the encoding.

ps. While the bijection idea is nice, just beware that it's not really a bijection, there could be subtle bugs arising from this, see for details twitter/bijection#41

pss. why I prefer `luigi` over, say, `airflow` is the simplicity. I would urge you to keep things simple in this way, so in terms of semantics and as little magic as possible...

dlstadther:
I'm personally a fan of making docs/error reporting more useful.

Tarrasch:
@wires, I don't really understand the Format stuff in luigi. I would just appreciate you use your better judgement. Adding docs with examples of best practices is highly appreciated. So is also runtime-checks or renaming variables into something making it less likely to get confused.

PR #1649, opened 2016-04-12
Add the archive parameters to HadoopJobRunner

mfcabrera:
This PR add an additional parameter of archives. This is helpful in particular when running python task and you want them to be run inside a virtual environment. See this gist for an example of how it could be used. 

https://gist.github.com/mfcabrera/63136e233e8eb36392b1ce33cf4e5a06

### DIFF
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -400,8 +400,8 @@ class HadoopJobRunner(JobRunner):
 
     def __init__(self, streaming_jar, modules=None, streaming_args=None,
                  libjars=None, libjars_in_hdfs=None, jobconfs=None,
-                 archives=None, input_format=None, output_format=None,
-                 end_job_with_atomic_move_dir=True):
+                 input_format=None, output_format=None,
+                 end_job_with_atomic_move_dir=True, archives=None):
         def get(x, default):
             return x is not None and x or default
         self.streaming_jar = streaming_jar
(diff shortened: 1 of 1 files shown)


luigi/contrib/hadoop.py
```diff
@@ -478,6 +479,12 @@ def run_job(self, job, tracking_url_callback=None):
         if libjars:
             arglist += ['-libjars', ','.join(libjars)]
 
+        # 'archives' is also a generic option
+        archives = [archive for archive in self.archives]
```
Tarrasch: This looks strange. What does this do?

luigi/contrib/hadoop.py
```diff
@@ -400,7 +400,7 @@ class HadoopJobRunner(JobRunner):
 
     def __init__(self, streaming_jar, modules=None, streaming_args=None,
                  libjars=None, libjars_in_hdfs=None, jobconfs=None,
-                 input_format=None, output_format=None,
+                 archives=None, input_format=None, output_format=None,
```
Tarrasch: Why not add as last argument?

mfcabrera: Well, as it is a  a generic option I thought it would be better next to another one like, like libjars. But I guess I should have added it on the top. Then, I will update the PR to put it at the end.

luigi/contrib/hadoop.py
mfcabrera: I was wondering the same to be honest. I mimicked the `libjars` variable behavior.

Tarrasch:
That was fast. Cool! :)

Tarrasch:
@erikbern, hey, do you mind start using the squashing accept-strategy for PRs instead of  merging?. I think the commit logs looks cleaner (and more chronological). Take a look https://github.com/spotify/luigi/commits/master.

Furthermore I think this PR only should have had 1 commit.

PR #1650, opened 2016-04-13
Add contrib package for MSSQL

traviscook21:
This is primarily based off of the mysqldb.py module, but uses the _mssql library

### DIFF
--- a/luigi/contrib/mssqldb.py
+++ b/luigi/contrib/mssqldb.py
@@ -18,7 +18,6 @@
 import logging
 
 import luigi
-import abc
 
 logger = logging.getLogger('luigi-interface')
 
@@ -34,7 +33,9 @@ class MSSqlTarget(luigi.Target):
     Target for a resource in Microsoft SQL Server.
     """
 
-    marker_table = luigi.configuration.get_config().get('mssql', 'marker-table', 'table_updates')
+    marker_table = luigi.configuration.get_config().get('mssql',
+                                                        'marker-table',
+                                                        'table_updates')
 
     def __init__(self, host, database, user, password, table, update_id):
         """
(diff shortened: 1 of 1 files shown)


erikbern:
looks good!

Tarrasch:
Cool. Some thoughts:

---
 
Can you add some more docs?

 * Is it read/write only or both?
 * Perhaps a minimal example as module level documentation.

---

Can tests be added?

---

If you mostly copy-pasted some other module, did you consider generalizing/inheriiting and stuff??

---

Note, all these things are optional. This can be merged as-is too since it's under `contrib/`.

traviscook21:
Hm, I think generalizing/inhereiting stuff is a good suggestion, but would probably require larger changes.

Generally, the code used for most of the methods is mostly identical to mysqldbpy and postgres.py. 

One thought would be to add a abstract rdbms target class to rdbms.py.  Then refactor the postgres, mysql, and mssql libraries to leverage that abstract class.  I'm not sure if that's a change that y'all have discussed before/would be open to making.

I'm still thinking about how to make generalizable Task classes for the mssql library.  The postgres library has CopyToTable and Query, which aren't use cases that I'm working with, but I'm thinking about a RunProcedure task class.  I'm fairly new to luigi, so I'm still working on having the whole library cement itself in my mind.

Do you have any thoughts on the above, considering your much larger familiarity (duh) with the project?

Tarrasch:
Well. One decent thing to do in these situations when one doesn't have time to dive into the code base and refactor and stuff is to at least add some docs saying `"""This module is mostly a copy-paste of other_module.py. In the future one common interface could be derived."""`. In addition to that I still think you at least should have an usage example in the module-level docs.

traviscook21:
@Tarrasch - Added the comments suggested above.  This can be merged.

PR #1651, opened 2016-04-14
Addresses Unintended Error caused by PR #1624 when user does not specify a logging_conf_file

dlstadther:
Revert default config variables to None, rather than empty strings. Don't warn user if Parameter has value of None.


dlstadther:
See @alexdebrie 's comment on Pr #1624 post-merge.

dlstadther:
Going to stick with warning when `Parameter` value is `None`. Closing....

PR #1653, opened 2016-04-14
Fixes logging error caused by PR #1624

dlstadther:
Set default conf_file value to empty string within function setup_interface_logging such that logging will not try to using a logging.conf file which doesn't exist.

link to pr #1624

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -40,12 +40,12 @@ from luigi import execution_summary
 from luigi.cmdline_parser import CmdlineParser
 
 
-def setup_interface_logging(conf_file=None):
+def setup_interface_logging(conf_file=''):
     # use a variable in the function object to determine if it has run before
     if getattr(setup_interface_logging, "has_run", False):
         return
 
-    if conf_file is None:
+    if conf_file == '':
         logger = logging.getLogger('luigi-interface')
         logger.setLevel(logging.DEBUG)
 


Tarrasch:
@dlstadther, I'm merging as this was pretty urgent. Can you try to also make a test case so this isn't reintroduced? :)

Again thanks for your report @alexdebrie

dlstadther:
@Tarrasch Sure, i can try to put something together.

Tarrasch:
Thanks @dlstadther, I very much appreciate your contributions. :)

dlstadther:
Looking for some unittest help here...

I'm trying to test the `setup_interface_logging(...)` method to determine if it throws an error - specifically, the `ConfigParser.NoSectionError: no section ...`. I have attempted the follow (which is incorrect).

```python 
import luigi
from helpers import unittest, LuigiTestCase

class InterfaceTest(LuigiTestCase):

    def test_setup_interface_logging(self):
        import ConfigParser
        self.assertRaises(ConfigParser.NoSectionError, luigi.interface.setup_interface_logging, '/not/a/path')

if __name__ == '__main__':
    unittest.main()
```

However, this returns `AssertionError: NoSectionError not raised`. I don't understand why this Error is not raised (or caught when raised) when method is provided a non-`''` or path which doesn't exist.

Does anyone have a great python resource that helped unittesting click for them? Thanks,,

alexdebrie:
@dlstadther I'm running a modified version of that (using just `unittest.TestCase` rather than `LuigiTestCase`) and it works for me. One thing you could try is to remove the `self.assertRaises(...)` portion and just call 

```python
def test_setup_interface_logging(self):
    import ConfigParser
    luigi.interface.setup_interface_logging('/not/a/path')
```

and see what error it *does* throw.

dlstadther:
@alexdebrie So you're running the below code and being told that the test passes?

``` python 
class InterfaceTest1(unittest.TestCase):

    def test_setup_interface_logging(self):
        import ConfigParser
        self.assertRaises(ConfigParser.NoSectionError, luigi.interface.setup_interface_logging, '/not/a/path')
```

When I do, I continue to be told that `NoSectionError not raised`. As I am still new to unit tests, it is possible that I'm just doing something wrong, but I'm not sure what it might be.

I have added the above code to `test/interface_test.py` and executed `tox -e py27-nonhdfs test/interface_test.py`.

alexdebrie:
@dlstadther Hmm, this is weird. I played around with it for a bit, and I'm seeing the same behavior you're seeing. I can get it to throw errors if I insert it in different parts of other tests in the module (e.g. [here](https://github.com/spotify/luigi/blob/master/test/interface_test.py#L51)). I can't figure out why it won't throw errors when it's on its own.

PR #1655, opened 2016-04-14
Issue/PR Templates

dlstadther:
Added Issues and PR template files.

Slightly altered [opensource version of template files](https://github.com/TalAter/open-source-templates) to be used as a first draft in this discussion/collaboration. The goal here is to derive at the [minimum] necessary information required by the Luigi community to asses/address Issues and Pull Requests.

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,14 @@
+<!--- This template is optional. Please use it as a starting point to help guide PRs -->
+
+<!--- Provide a general summary of your changes in the Title above -->
+
+## Description
+<!--- Describe your changes -->
+
+## Motivation and Context
+<!--- Why is this change required? What problem does it solve? -->
+<!--- If it fixes an open issue, please link to the issue here. -->
+
+## Have you tested this? If so, how?
+<!--- Valid responses are "I have included unit tests." or --> 
+<!--- "I ran my jobs with this code and it works for me." -->


Tarrasch:
Hmm. I'm afraid that we'll demand from our users to fill in nice bug reports. But then just say "oh sorry we don't have time to fix this".

Can we make bug filing feel less like a choir/filling a form, but still provide some guidance? Like make these shorter, and perhaps hint that this template is just a recommendation and they can describe the bug in whatever way they find most suitable.

Right now I don't feel that the there's too many roundtrips for each PRs/issue because they lack information. Do you think so?

dlstadther:
This PR was prompted by my realization that I'm inconsistent with describing Issues and PRs and I don't have a clear structure when submitting them. Sometimes I don't catch all the edge cases when testing (without writing unit tests), and I'm likely not the only one. By adding structure to Issue and PR submissions, we can more explicitly define expectations. I wouldn't think that we'd lose quality submissions due to a little extra work, but of course I'm still relatively new to the game and need structure.

I think what I originally submitted is a bit too much to ask from Luigi-users, but I just wanted to throw something out to get opinions and narrow it down to the necessary.

Regarding **PRs**, when I read them I'm really looking for two things.  
1. What they think they've changed/add and   
2. Why they are doing so / what purpose does this serve.  
Asking them to explain how they tested it is more so just to remind them to write unit tests (if possible) or at least cause them to make sure they aren't going to induce errors (like my last PR). Instead of asking them to write about how they tested their PR, we might just include some text in there to remind them to make sure they've gone through that step on their own.

**Issue** submissions, I feel, just need to inform the reader/reviewer what behavior is expected, what is actually happening, and (ideally) a hypothesis why.

Tarrasch:
> This PR was prompted by my realization that I'm inconsistent with describing Issues and PRs and I don't have a clear structure when submitting them. Sometimes I don't catch all the edge cases when testing (without writing unit tests), and I'm likely not the only one. By adding structure to Issue and PR submissions, we can more explicitly define expectations. I wouldn't think that we'd lose quality submissions due to a little extra work, but of course I'm still relatively new to the game and need structure.

OK. I read this as you want to guide users submitting PRs. I think that's good, come to think about it many PR submitters for luigi are doing one of their first PRs on github.

Do you think this mindset makes sense?:

 * We *do* put up this template to guide users, so they don't need to think so much when writing PR
 * We *don't* put up this template to constrict contributors or make them do more work

>  Asking them to explain how they tested it is more so just to remind them to write unit tests (if possible) or at least cause them to make sure they aren't going to induce errors (like my last PR). Instead of asking them to write about how they tested their PR, we might just include some text in there to remind them to make sure they've gone through that step on their own.

Yea. I like this thinking.

> Issue submissions, I feel, just need to inform the reader/reviewer what behavior is expected, what is actually happening, and (ideally) a hypothesis why.

Hey. Here's an orthogonal idea. Why don't we start with only having templates for PRs, gather feedback and understanding and the continue with Issues. Sounds good?

dlstadther:
> Do you think this mindset makes sense?:
> 
> We do put up this template to guide users, so they don't need to think so much when writing PR
> We don't put up this template to constrict contributors or make them do more work

Yes, this is exactly what I was trying to get at. Some people will ignore it anyway, so there's no need to make it mandatory.

> Why don't we start with only having templates for PRs, gather feedback and understanding and the continue with Issues.

Sounds good to me! I'll make some edits here and resubmit for iteration.

Issue #1321, opened 2015-10-19
No module named top_artists


autodidacticon:
Executing the following command:
`luigi --module top_artists AggregateArtists --local-scheduler --date-interval 2012-06`

produces the following error:

```
Traceback (most recent call last):
  File "/Users/richard.moorhead/anaconda/bin/luigi", line 11, in <module>
    sys.exit(luigi_run())
  File "/Users/richard.moorhead/anaconda/lib/python2.7/site-packages/luigi/cmdline.py", line 13, in luigi_run
    luigi.interface.run(argv, use_dynamic_argparse=True)
  File "/Users/richard.moorhead/anaconda/lib/python2.7/site-packages/luigi/interface.py", line 421, in run
    tasks = interface.parse(cmdline_args, main_task_cls=main_task_cls)
  File "/Users/richard.moorhead/anaconda/lib/python2.7/site-packages/luigi/interface.py", line 329, in parse
    __import__(module)
ImportError: No module named top_artists
```

Tarrasch:
What do you expect to happen @autodidacticon? The error message "No moudle named top_artists" is easily google-able, I mean,  `__import__` is standard python.

Maybe it would be more user friendly if it *only* said `ImportError: No module named top_artists` without the stack trace?


Oh, and for this particular case I guess you want to pass `--module examples.top_artists`, or cd into the examples folder or change your `PYTHONPATH`.

autodidacticon:
That command came from your docs:

http://luigi.readthedocs.org/en/latest/example_top_artists.html#running-this-locally

Obviously, it was run from within the examples folder, furthermore, running it from the parent folder (examples.top_artists) as you described does not work either.

Tarrasch:
Hmm, maybe

```
PYTHONPATH='' luigi --module top_artists AggregateArtists --local-scheduler --date-interval 2012-06
```

works?

autodidacticon:
Yes thats it; crucial bit here:
```
# my_module.py, available in your sys.path
```

Sorry for the hassle

mbelmadani:
I also ran into this issue. Maybe the PYTHONPATH='' bit should be added in the documentation?

Tarrasch:
Yes. Please send a PR. :)

philly-vanilly:
PYTHONPATH='' luigi --module top_artists AggregateArtists --local-scheduler --date-interval 2012-06

didn't work for me (MacOS) but this did; note the dot:

[...]

PR #1662, opened 2016-04-20
Update example_top_artists.rst

mbelmadani:
## Description

Improved documentation by suggesting to set the python path if an import error occurs.

## Motivation and Context

Since this documentation is part of introductory material, it is probably a good idea to suggest this common mistake. The error this addresses has also been mentioned in a previous issue (#1321). 

Added example command to work around the class missing from PYTHONPATH issue which causes an import error when calling the luigi command line tool.

### DIFF
--- a/doc/example_top_artists.rst
+++ b/doc/example_top_artists.rst
@@ -77,19 +77,11 @@ Try running this using eg.
     $ cd examples
     $ luigi --module top_artists AggregateArtists --local-scheduler --date-interval 2012-06
 
-Note that  *AggregateArtists* needs to be in your PYTHONPATH, or else this can produce an error (*ImportError: No module named AggregateArtists*). A workaround for development is to prepend the path before the command:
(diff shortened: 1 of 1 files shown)


Tarrasch:
This looks a bit confusing. Are you sure you understand PYTHONPATH? I think the reason it works with `PYTHONPATH=''` is that when it's empty it defaults to including to check the current folder. Hence your second example is a bit confusing. Why both `cd` there *and* set the `PYTHONPATH` to that?

alfonsomhc:
It would be nice if this could be added to the documentation. I run into this problem when following the example in http://luigi.readthedocs.io/en/stable/command_line.html
This extra information would have saved me some time...

Tarrasch:
Ah, I missed that this PR was updated since my last comment. Thanks for bumping this thread @alfonsomhc. I'm merging, hoping that the next user will have a better experience. :)

PR #1669, opened 2016-04-27
Adds host resources in addition to global resources

daveFNbuck:
## Description
Host resources are resources used on a specific host. Tasks with host_resource
set will be limited by how many can run on the same machine, but there is no
limit to how many can run globally. For example, if you have a process that
requires 2gb of memory, you might set its host_resources to {'memory': 2} and
then set the host resource for memory on each machine according to how much it
has available. If you have a machine with 8gb of memory, you can run 4 such
tasks on it. If you have another machine with 16gb of memory, you can run
another 8 tasks on it.

## Motivation and Context
I've noticed recently that several people in the mailing list have been asking how to limit resource usage on a machine level. Existing resources are somewhat capable of this by adding a resource_name__machine_name resource, but this only works if tasks are tied to a particular machine. If you have a task that can be run on multiple machines but shouldn't have too many per machine, this solution won't do. I recently created a kind of task that can only be run one at a time per machine, but can be run in parallel across arbitrary machines, so I needed a way to enable this pattern.

I have included unit tests. I have not yet run jobs with it, as I don't have the machines set up to run anything nontrivial that actually requires it yet.


BinrOp:
This is so useful, any plan to merge it soon?

I have tested it on a simple workflow for the current version and it worked perfectly fine.

Tarrasch:
Oh my, It's been over a year now.

> I have included unit tests. I have not yet run jobs with it, as I don't have the machines set up to run anything nontrivial that actually requires it yet.

You have run this in production for a year now, right? Has it been useful? This might be worth merging then. :)

daveFNbuck:
I have not used this in production. We ended up just using a different scheduler per machine and I don't have a different use case for this code yet.

BinrOp:
These guys seem to be using a version of this feature: https://github.com/groupon/luigi-warehouse/blob/d4c9fb47ec3a05dbf91d1cf2b2bd265cd78a3ad0/luigi_warehouse/teradata_to_redshift.py#L60
Might be worth asking them what they think about it (I have opened an issue on their repo).

Btw, if we can manage resources on a host level, we can safely start parallel clients on each host: useful if we want to add tasks to a running workflow and if we do not want to restart the client (and kill long running jobs).

In our case we would use it to share GPUs.

BinrOp:
I got an answer from [luigi-warehouse
](https://github.com/groupon/luigi-warehouse/issues/7), they do use the feature but not that intensively, and there does not seem to be any problem to report.

This feature is a really great asset for **scaling**, at the risk of repeating myself. I am considering rebasing it on my branch and testing it with some bigger workflows, I might tell you how it went afterwards.

Tarrasch:
@BinrOp we would be happy to merge your changes if everything looks good.

srikiraju:
Any updates on this PR? This would be really useful for us.

BinrOp:
Humm we haven't been working on it lately and we're actually moving out from using luigi to handle resources/distribution, and are only keeping luigi to handle parameters and define our dependencies.

So if anyone wants to try rebasing it, feel free to!

PR #1670, opened 2016-04-28
Updated luigi.util.previous to support year and month parameters

drj42:
## Description

The function 'luigi.util.previous' tries to create a "previous" task class based on the existence of a DateParameter (or one of the many variants of DateParameter).  YearParameter and MonthParameter options were not supported.  This patch adds support for both year and month parameters in the 'previous' function.
## Motivation and Context

Updating 'luigi.util.previous' in this manner allows 'luigi.util.get_previous_completed' to work with tasks that have year or month parameters.

Task classes make great "dataset addresses", but when loading them in notebook-style interactive sessions (iPython, Jupyter, Zeppelin, etc) it can be burdensome to remember and specify the exact date for each particular dataset, as a datetime parameter.  Often we just want to grab the latest dataset and get on with it.  That was not easily possible for tasks with date or year parameters without this patch.

I have included unit tests and I have also used these updates in interactive notebook sessions.

### DIFF
--- a/test/util_previous_test.py
+++ b/test/util_previous_test.py
@@ -28,55 +28,55 @@ class DateYearTaskOk(luigi.Task):
 
     def complete(self):
         return self.year in [
-            1996,
-            1998,
-            2000,
+            datetime.date(1996, 1, 1),
+            datetime.date(1998, 1, 1),
+            datetime.date(2000, 1, 1),
         ]
 
 
 class DateYearTaskOkTest(unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


Tarrasch:
Eh, I don't know what @codecov-io is trying to say.

Anyway. this patch looks good to me.

drj42:
I noticed a couple of copy/paste errors in the unit tests, whoops.  Will fix in the morning :)

erikbern:
lgtm

PR #1672, opened 2016-04-29
Fix formatting of example configuration snippet

gszjulcsi:
Code formatting keyword had no effect:

http://luigi.readthedocs.io/en/stable/central_scheduler.html#enabling-task-history


joemeszaros:
Thanks for fixing the documentation, nice catch! cc-ing @Tarrasch

Tarrasch:
Awesome :tada:

Issue #1657, opened 2016-04-18
Task finished yet the luigi failed.


junchenfeng:
I have py script contains the following task

class UpdateUserParameter(luigi.Task):
    udt = luigi.DateParameter()
    subject=luigi.Parameter()

    def output(self):
        return luigi.LocalTarget(local_proj_dir+'/data/crontab/{sub}/{dts}/is_user_parameter_updated.txt'.format(sub=self.subject, dts=self.udt.strftime('%Y-%m-%d')))
    def requires(self):
        return [users.UpdateUserParameterChain(subject=self.subject, udt=self.udt),
                users.UpdateUserKnowledgePointMasteryBuffer(subject=self.subject, udt=self.udt)
            ]

When the task runs, all task output is produced (including the master task) but the master task returns the following error:

Traceback (most recent call last):
  File "run_script.py", line 91, in <module>
    luigi.run()
  File "/usr/local/lib/python2.7/site-packages/luigi/interface.py", line 210, in run
    return _run(*args, **kwargs)['success']
  File "/usr/local/lib/python2.7/site-packages/luigi/interface.py", line 238, in _run
    return _schedule_and_run([cp.get_task_obj()], worker_scheduler_factory)
  File "/usr/local/lib/python2.7/site-packages/luigi/interface.py", line 198, in _schedule_and_run
    logger.info(execution_summary.summary(worker))
  File "/usr/local/lib/python2.7/site-packages/luigi/execution_summary.py", line 384, in summary
    return _summary_wrap(_summary_format(_summary_dict(worker), worker))
  File "/usr/local/lib/python2.7/site-packages/luigi/execution_summary.py", line 333, in _summary_format
    str_output += '{0}\n'.format(_get_str(group_tasks[status], status in _PENDING_SUB_STATUSES))
  File "/usr/local/lib/python2.7/site-packages/luigi/execution_summary.py", line 132, in _get_str
    params = _get_set_of_params(tasks)
  File "/usr/local/lib/python2.7/site-packages/luigi/execution_summary.py", line 173, in _get_set_of_params
    params[param] = {getattr(task, param[0]) for task in tasks}
  File "/usr/local/lib/python2.7/site-packages/luigi/execution_summary.py", line 173, in <setcomp>
    params[param] = {getattr(task, param[0]) for task in tasks}
TypeError: unhashable type: 'dict'

Because this error, other task depends on this cluster cannot be started. I cannot reliably reproduce this error and wonder how to fix it.

[...]

Issue #1627, opened 2016-04-01
Visualizer fails with similar range targets.


nugend:
I'm using Range Targets to guide the loading of some information that appears on a daily basis that we want to load ASAP. (We have the configuration set to retry on external task failures and to keep the workers alive)

When you run two Range Targets with the same parameters at the same time, but the target task has different parameters, the visualizer will only display one of the range targets in any given screen. The sub-tasks are of course still present and Luigi doesn't appear to have any issue with the Range targets being identical, so I am presuming this is an okay way to use Range. I don't know enough about JS to dig in and figure out what's going wrong, but I imagine it has something to do with the task parameters being identical.

My command lines look something like this:

```
# Note that I use 2016.04.01 to mean, start loading today's data. This is set in our cron job to change to whatever the current date is whenever it re-executes (which would be as soon as the job exits)
# 2 Days forward takes care of any issues with the international date line
luigi --module mymodule RangeDaily --start 2016.04.01 --days-forward 2 --of MyTask --MyTask-myparam 1
luigi --module mymodule RangeDaily --start 2016.04.01 --days-forward 2 --of MyTask --MyTask-myparam 2
```

Am I just going about this backassward? I feel like this would be an issue if you had to run a range on a parameterized task for any reason. Just to reiterate, there's *no* problem with the jobs running or anything, this only shows up as an issue in the visualizer.

Tarrasch:
I wish this day would never come ... I was aware of this design issue when redesigning the parameter system. Though I opted for it to make the common case easy.

Yea the problem is indeed that the parameter set to the RangeDaily tasks is identical. If you can't live with it, just hack around and set different `--start` parameters. If you can't live with that bad hack, we should think about a proper solution ...

nugend:
For me, it's a bit of an issue since I actually wrote a subclass of RangeDaily to let me *just* specify `--days-forward` and `--days-back` (so I could have jobs that are always looking forward and filling in anything missing from the past few days).

I mean, it's not a high priority for me since everything seems to work and the Scheduler doesn't really seem to have any problem with it, and we're not heavily relying on the visualizer, but it'd be *nice* if it worked right.

nugend:
Now that each job is in part associated with a hash identifier, maybe that could be used to differentiate between instances in the visualizer?

Tarrasch:
Actually. I totally forgot about this simple solution:

----

Use a new `DictParameter` called "pass-along" or something. This line:

[...]

PR #1675, opened 2016-05-02
Pass through parameters on Range tasks to of tasks

nugend:
## Description
I've added "of_params" as an additional parameter on the RangeBase class. The approach is based on discussion with @Tarrasch 

## Motivation and Context
Please see issue from #1627 

I have included unit tests.

I have not yet tested this in our system, but I will do so if the maintainers think the approach used is fine and report back on the behavior.

### DIFF
--- a/doc/luigi_patterns.rst
+++ b/doc/luigi_patterns.rst
@@ -102,10 +102,17 @@ and stop (exclusive) parameters specified:
 Propagating parameters with Range
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-When your recurring task has a parameter, you'll at first notice that the Range
-tasks do not recognize or propagate parameters passed to them. The easiest
-solution is to set the parameter at the task family level as described
-:ref:`here <Parameter-class-level-parameters>`.
+Some tasks you want to recur may include additional parameters which need to be configured.
+The Range classes provide a parameter which accepts a :class:`~luigi.params.DictParameter`
+and passes any parameters onwards for this purpose.
+
+.. code-block:: console
+
+	luigi RangeDaily --of MyTask --start 2014-10-31 --of-param '{"my-param": 123}'
+
+Alternatively, you can specify parameters at the task family level (as described :ref:`here <Parameter-class-level-parameters>`),
+however these will not appear in the task name for the upstream Range task which
+can have implications in how the scheduler and visualizer handle task instances.
 
 .. code-block:: console
 
--- a/doc/luigi_patterns.rst
+++ b/doc/luigi_patterns.rst
@@ -103,7 +103,7 @@ Propagating parameters with Range
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Some tasks you want to recur may include additional parameters which need to be configured.
-The Range classes provide a parameter which accepts a :class:`~luigi.params.DictParameter`
+The Range classes provide a parameter which accepts a :class:`~luigi.parameter.DictParameter`
 and passes any parameters onwards for this purpose.
 
 .. code-block:: console
--- a/luigi/tools/range.py
+++ b/luigi/tools/range.py
@@ -186,7 +186,10 @@ class RangeBase(luigi.WrapperTask):
 
     @property
     def _param_name(self):
-        return self.of.get_params()[0][0] if self.param_name is None else self.param_name
+        if self.param_name is None:
+            return next(x[0] for x in self.of.get_params() if x[1].positional)
+        else:
+            return self.param_name
 
     def _task_parameters(self, param):
         kwargs = dict(**self.of_params)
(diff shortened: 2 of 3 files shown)


nugend:
Actually, I don't think bulk_complete can work correctly with this change. You'd have to pass in the arguments both explicitly and using the of_params field. I'm working on a fix, but I don't know if there's a way to preserve the existing interface of bulk_complete. The problem being that RangeDaily and RangeHourly were both providing a list of parameters to those functions rather than a complete argument tuple.

The way I'm approaching a change to this is to pass a list of dictionaries. If there's a way to make python use a single value as both the positional and keyword arguments for a function, that'd be nice, but I'm not aware of it.

nugend:
Actually, I see that MixinBulkComplete is set to handle the case that the parameter tuples can be any of a single value, a positional argument tuple, or a keyword argument tuple.

However, the BigQuery code from contribs doesn't expect this behavior.

nugend:
Found one more small issue. Seems to be working correctly in our system now though.

Tarrasch:
Docs are missing my friend. ^^

They can either be added in the code module or in the docs which mention range (you can `git grep Range` for it).

nugend:
Sorry. Hope that addresses the docs issue or is there a specific area that needs to be addressed?

doc/luigi_patterns.rst
```diff
@@ -102,10 +102,17 @@ and stop (exclusive) parameters specified:
...
+The Range classes provide a parameter which accepts a :class:`~luigi.params.DictParameter`
+and passes any parameters onwards for this purpose.
+
+.. code-block:: console
+
+	luigi RangeDaily --of MyTask --start 2014-10-31 --of-param '{"my-param": 123}'
+
+Alternatively, you can specify parameters at the task family level (as described :ref:`here <Parameter-class-level-parameters>`),
+however these will not appear in the task name for the upstream Range task which
+can have implications in how the scheduler and visualizer handle task instances.
```
Tarrasch: Very well written. Thanks! :)

Tarrasch:
A bit unfortunate that we have to resort to code with `__func__` and stuff. And actually `bulk_complete` won't be a `@classmethod` any more.

Other than that I'm fine with this. @ulzha, do you want to take a quick look? Other than that I'll merge day this in a day or two.

Tarrasch:
Actually let's not merge this until we've resolved https://github.com/spotify/luigi/issues/1657#issuecomment-216918636. In particular it seems like execution summary is broken for anything using DictParameter.

nugend:
Ah, I noticed that earlier but thought it must have been some other bug on my end. Since all the tests passed, I figured it was fine. Makes sense to wait for a fix on that before putting in DictParameters retroactively. Please let me know when a solution is available so I can perform refactoring as necessary. I'll be using this solution for the time being since I actually don't have an issue if these error out after completing. Will update if I encounter any weirdness.

I considered changing the interface of bulk_complete, but figured it'd be better to make it a little weird for now and not significantly break the expected API. Then later at a point where you are comfortable making a backwards incompatible change, move it to a staticmethod.

Tarrasch:
There's no need for you to change your implementation once that bug-fix is done. You can just sit back and relax for now. ^^

Yea. I thing you did the right thing regarding the `bulk_complete`. But perhaps you should add an implementation note that one "must" use the `cls` variable to instantiate the class and not use `YourTaskName` as a constructor?

nugend:
Should that be in the docs for Range or for Task?

Tarrasch:
@nugend I guess for Task right? Because when somebody looks at how to implement `bulk_complete`, they're looking at the documentation for Tasks in general.

nugend:
The reason I asked was because nothing actually changed in Task and any other usage of bulk_complete is likely custom created (and likely does not have any sort of similar issue).

Would it make sense to add bulk_complete to RangeBase so that it could be documented inside the Range classes themselves?

Tarrasch:
I don't think it makes sense to add `bulk_complete` to RangeBase. Because it's supposed to be overridden by classes that aren't subclasses of RangeBase. Right?

nugend:
Yes, sorry. I didn't yet have my coffee when I replied earlier. What I mean is just that this isn't *really* a change in how you use bulk_complete. It's a change in how bulk_complete is invoked when used with a RangeBase derived class. Strictly speaking, there could be some other user implemented WrapperTask which uses bulk_complete to calculate the required child tasks and *that* usage of bulk_complete wouldn't change and the ability to use MixinNaiveBulkComplete would support that just as happily.

To the larger point, depending on how it is decided that bulk_complete should change, maybe just more thorough documentation of what the possible types parameter_tuples can inhabit would be sufficient for ensuring correct usage. Or possibly, a helper to generate the tasks associated with each parameter_tuple

Tarrasch:
Ok. Let's wait with the docs change. This PR is still blocked on #1657 though. :)

PR #1676, opened 2016-05-04
raise an error when no columns are found

maxcountryman:


dlstadther:
By default `rdbms.py` sets `columns = []` so your conditional addition should catch when the user does not override this variable.

lgtm

dlstadther:
@Tarrasch Any reason not to merge this?

Tarrasch:
Isn't all the tests failing due to indentation error?

dlstadther:
Ah, yes. The `else` block appears to line up with the method declaration. @maxcountryman , you'll need to indent your additions to line up with rest of the conditional.

Tarrasch:
That also raises the question of how much this was tested (if at all). 

@maxcountryman, can you also add a motivation to this patch? What real world use case scenario does it solve for you?

maxcountryman:
@Tarrasch did you read the commit message...?

Tarrasch:
@maxcountryman I did not, sorry! Thanks, it explains it well. I'm too used to GitHub copying it into the PR description. :)

dlstadther:
Ready to merge?

PR #1677, opened 2016-05-04
Rename move copy functions added to FileSystem and FileSystemTarget child classes

tusharbhat:
## Description
Added/Edited copy, rename and move functions in LocalTarget, LocalFileSystem, MockTarget and MockFileSystem where I felt it would be suitable.

## Motivation and Context
While writing test cases for a rename function which would rename files in S3, I found that MockFileSystem did not have rename or move functions making writing appropriate tests a convoluted and error-prone process, especially when using the OpenerTarget class from luigi.contrib.opener. So, I chose to implement the above functions in file.py and mock.py (s3.py already had support for most functions) in order to present a uniform interface across classes which inherit from FileSystem and FileSystemTarget.

I have included 3 unit tests:
file_test.py:
test_copy()
mock_test.py:
test_copy()
test_rename()
All above test cases pass.

### DIFF
--- a/luigi/file.py
+++ b/luigi/file.py
@@ -156,12 +156,7 @@ class LocalTarget(FileSystemTarget):
         self.fs.remove(self.path)
 
     def copy(self, new_path, raise_if_exists=False):
-        if raise_if_exists and os.path.exists(new_path):
-            raise RuntimeError('Destination exists: %s' % new_path)
-        tmp = LocalTarget(new_path + '-luigi-tmp-%09d' % random.randrange(0, 1e10), is_tmp=True)
-        tmp.makedirs()
-        shutil.copy(self.path, tmp.fn)
-        tmp.move(new_path)
+        self.fs.copy(self.path, new_path, raise_if_exists)
 
     @property
     def fn(self):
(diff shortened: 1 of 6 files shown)


tusharbhat:
@Tarrasch any chance you could take a look at this and see if there's anything that needs to be changed?

dlstadther:
It seems like you should also add `copy()` and `rename()` to `target.FileSystem()` if everything that implements a FileSystem should have these methods.

luigi/file.py
```diff
@@ -83,6 +91,12 @@ def remove(self, path, recursive=True):
         else:
             os.remove(path)
 
+    def rename(self, *args, **kwargs):
+        """
+        Calls ``move()``
+        """
+        self.move(*args, **kwargs)
+
```
dlstadther: Since this seems to be the same implementation for all `rename()` methods here. Why don't you make this the default behavior for `target.FileSystem.rename()` and not worry about rewriting the same thing for each class?

Tarrasch:
@tusharbhat, Overall this looks good to. Just check out @dlstadther's comments and then I think this is pretty much good to go.

tusharbhat:
@dlstadther  , I made the changes, thanks for the feedback. @Tarrasch , code's ready to go

dlstadther:
Assuming tests pass and @Tarrasch approves, this lgtm. Thanks!

Tarrasch:
Thanks @tusharbhat and @dlstadther ! :)

PR #1679, opened 2016-05-06
Fix inherits decorator docs.

pkaleta:
## Description
Just a simple change to the description of how to use the `inherits` decorator. I think there was a bug in the class that's supplied as the argument to `clone` in the example used as the motivation for the `inherits`.

No tests required, since it's just the docs fix.

### DIFF
--- a/luigi/util.py
+++ b/luigi/util.py
@@ -158,7 +158,7 @@ inheritance.
         param_b = luigi.Parameter()
 
         def requires(self):
-            t = self.clone(TaskB)
+            t = self.clone(TaskA)  # or t = self.clone_parent()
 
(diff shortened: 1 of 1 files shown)


Tarrasch:
Cool! LGTM

dlstadther:
@Tarrasch Is this good to be merged? Also, do you normally merge or squash?

Tarrasch:
Usually squash, and also reword it. I find the practice of rewording somebody else's commit message to be a very healthy check for myself that I actually understand what I'm merging. :)

However, if there are many commits, all clean and happening in a logical order that simplifies reviewing and perhaps reverting. I do a merge-commit as the value of the individual commits being merged are high.

PR #1681, opened 2016-05-06
minor improvements

jkukul:
## Motivation and Context
1f1632e is just a tiny optimisation, but since it's in "luigi core", I guess it's worth it

PR doesn't include any changes in the logic


Tarrasch:
Thanks!

PR #1686, opened 2016-05-11
#1685 Add support for multiple results in salesforce batch jobs

### LINKED ISSUE #1685: [WIP] Salesforce does not download all results for large result set
For particularly large result sets, Salesforce will return multiple result IDs for a single batch. 

However, in [`get_batch_results`](https://github.com/spotify/luigi/blob/master/luigi/contrib/salesforce.py#L489), only the first result ID is being returned. The code should utilize `ET.findall` to return a list of result IDs and download each of those to append to the output file.

I'm planning on submitting a PR for this in the next couple of days.

dacort:
Addresses #1685 by adding support to fetch multiple results from a single Salesforce batch job.

## Open Items

- [x] Merge downloaded files

## Description
- New `get_batch_result_ids` method that is more accurately documented
- Backwards compatible `get_batch_results` method, including warning message
- Small set of tests for the changes

## Motivation and Context
Fixes #1685 

Included unit tests as well as verified locally.

### DIFF
--- a/luigi/mock.py
+++ b/luigi/mock.py
@@ -135,10 +135,6 @@ class MockTarget(target.FileSystemTarget):
     def path(self):
         return self._fn
 
-    @property
-    def fn(self):
-        return self._fn
-
     def open(self, mode):
         fn = self._fn
         mock_target = self
--- a/luigi/mock.py
+++ b/luigi/mock.py
@@ -135,6 +135,10 @@ class MockTarget(target.FileSystemTarget):
     def path(self):
         return self._fn
 
+    @property
+    def fn(self):
+        return self._fn
+
     def open(self, mode):
         fn = self._fn
         mock_target = self
--- a/test/contrib/salesforce_test.py
+++ b/test/contrib/salesforce_test.py
@@ -114,4 +114,4 @@ class TestSalesforceQuery(unittest.TestCase):
         qsf = TestQuerySalesforce()
 
         qsf.merge_batch_results(self.result_ids)
-        self.assertEqual(MockTarget(qsf.output().fn).open('r').read(), self.all_lines)
+        self.assertEqual(MockTarget(qsf.output().path).open('r').read(), self.all_lines)
--- a/test/contrib/salesforce_test.py
+++ b/test/contrib/salesforce_test.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2016 Simply Measured
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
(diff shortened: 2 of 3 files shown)


dlstadther:
Based on TravisCI results, looks like you just need to fix the few flake8 errors and then run the tests again. All the others look to be errors when creating the environment (stuff you didn't cause).

I haven't run this myself, but the logic looks good. Thanks!

dacort:
D'oh. Of course I forget to run the flake8 test after I add the unit test. ;) 

Addressing shortly.

dlstadther:
Looks like you need to get @Tarrasch 's updates to Tox (that'll fix these tests).

```
git fetch upstream
git checkout master
git merge upstream/master
git checkout <this branch>
git rebase master

git push origin <this branch> --force
```

dacort:
@dlstadther Thanks for the note re: Tox updates. Just pulled that in and pushed up an update that merges batch results (from CSV).

dacort:
Looks like I have to some work to do re: python 3 compat.

test/contrib/salesforce_test.py
```diff
@@ -88,7 +89,11 @@ def soql(self):
 
 
 class TestSalesforceQuery(unittest.TestCase):
-    @mock.patch('__builtin__.open', side_effect=mocked_open)
+    patch_name = '__builtin__.open'
+    if sys.version_info.major > 2:
+        patch_name = 'builtins.open'
+
+    @mock.patch(patch_name, side_effect=mocked_open)
```
dlstadther: Is it possible to use six.py for the Python 2.x vs 3.x compatibility?

dacort: Will take a look!

dacort: Not sure if I can use it to import a generic builtins, but I can at least use it to check for the Python version. 

The challenge here is the `patch_name` string gets interpreted further down the stack and I'm not actually importing builtins, so flake8 throws an error.

dlstadther: Well, tests passed. So if @Tarrasch approves, then i think this pr is good to go.

luigi/mock.py
```diff
@@ -135,6 +135,10 @@ def rename(self, *args, **kwargs):
     def path(self):
         return self._fn
 
+    @property
+    def fn(self):
```
Tarrasch: Instead of using `.fn` can't you just use `.path` and there's less code in total?

dacort: Looks like I can, yep! Had just been following the convention that was already in the module.

Switched over to using `.path` - that works fine. Also removed extra code that had been added in `mock.py`.

dlstadther: My bad for setting a poor example within `salesforce.py`!

Tarrasch:
Good job. Thanks guys :)

Issue #1688, opened 2016-05-13
SFTP target doesn't allow custom port


jeffcjohnson:
The SFTP target ignores the port parameter and changes it to 22.

```
class RemoteFileSystem(luigi.target.FileSystem):

    def __init__(self, host, username=None, password=None, port=21, tls=False, timeout=60, sftp=False):
        self.host = host
        self.username = username
        self.password = password
        self.port = port
        self.tls = tls
        self.timeout = timeout
        self.sftp = sftp

        if self.sftp:
            self.port = 22
```

The last two lines in that change the port to 22 if SFTP is needed.  In my case, I need to use 522.

I'm not creating a pull request because this would be a breaking change to anyone that is using the default port as removing the problematic code would change the default SFTP port from 22 to 21.

Any suggestions?

dlstadther:
I'm responsible for defining these ports in this code. I set sftp port to be 22 because that is the (port that traditionally corresponds to SFTP)[https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports].

As this ftp.py module is super rough, as well as in the contrib package, breakage may be necessary.

I was given very little time to implement sftp support and since others had been asking about it, I submitted the PR which you can see. We noted that it needed more love, but at that time I did not have the time to give to it.

I'd suggest submitting a PR that you believe would be more flexible for all use cases of FTP and SFTP support regardless of port requirements. Breakage may be necessary.

dlstadther:
This issue can be closed

Tarrasch:
Thanks.

Side note for @dlstadther, I just promoted you to have write access to spotify/luigi, so you should now be able to close issues on your own. Thanks for all your help on this project and I hope these new access rights makes you even more productive. :)

dlstadther:
Thanks @Tarrasch ! For the near future, i'll still leave merges to you and @erikbern , but I will continuing contributing as I can and gaining experience and knowledge. So my first action will be to close this issue! haha. Thanks again!

Tarrasch:
@dlstadther That sounds sensible. Do whatever you feel comfortable :)

PR #1689, opened 2016-05-16
allow sftp default port 22 to be overridden

jeffcjohnson:
## Description
The ftp remote target defaults to port 21 and allows it to be changed.
If sftp is needed, it changes the port to 22 and does not allow it to be changed.

See https://github.com/spotify/luigi/issues/1688

This pull request keeps the intended behavior by defaulting to 21 or 22 but also allows both to be overridden in such a way that it is not a breaking change.

## Motivation
I have to pull files from a non-default sftp port of 522 but cannot.


dlstadther:
lgtm

PR #1456, opened 2015-12-10
fix for 'allow binary opens'

bmaggard:
The previous commit's intent was to enable 'binary' mode. This commit fixes it, preventing 'append' and 'update' modes.

PR #1692, opened 2016-05-19
Clarify file permission error message

xrmx:
Make it clear that permission should either be 'r' or 'w'.

The original "r/w" suggestion was not clear to me, was passing 'w+' originally, then interpreted the error as a request to use 'rw' until i looked at the actual source code :)

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -394,7 +394,7 @@ class RemoteTarget(luigi.target.FileSystemTarget):
                 FileWrapper(io.BufferedReader(io.FileIO(self.__tmp_path, 'r')))
             )
         else:
-            raise Exception('mode must be r/w')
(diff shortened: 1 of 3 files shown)


bmaggard:
I like it, but while making the change, why not also them consistent?  I like the one I switched to back on PR #1456 (b/c: "(got: %s)"), but that can be reverted if others don't:

`raise Exception("mode must be 'r' or 'w' (got: %s)" % mode)`

xrmx:
@bmaggard updated pull request to make them consistent.

bmaggard:
This is good-to-go as far as I'm concerned.

dlstadther:
No real critiques here. Improved error reporting 👍

Tarrasch:
This looks very good, thanks!

One thing we should fix is that it shouldn't be such a general exception as `Exception()` right?

PR #1694, opened 2016-05-22
Add Hdf5 target based on pandas and pytables

kayibal:
## Description
I added a Hdf5 target allowing saving and reading from hdf5 storages using pandas. I use this a lot at work as it allows to integrate luigi nicely with many pandas task we have to run every day, without worrying to much about saving and reading from hdf5 storages. The target also handles synchronous I/O operations as the default pytables storage does not allow parallel I/O. It also comes with chunked reading method to save memory if necessary. 
I figured other people could use this too. It is quite handy if you still can do all your operations on a single machine.

## Motivation and Context
I created this so I can work with pandas in my tasks and nicely pass in and output from one task to another while still maintaining efficiency and low disk usage. 

- I included unittests, which should cover most use cases
- I have been using this at work for some weeks and it seems to work nicely


tox.ini
```diff
@@ -12,6 +12,8 @@ deps=
   nose<2.0
   unittest2<2.0
   boto<3.0
+  pandas==0.18.0
```
Tarrasch: Any reason why it's so strict?

tox.ini
```diff
@@ -12,6 +12,8 @@ deps=
...
+  tables
```
Tarrasch: Any reason why it's so slack?

Perhaps use `<5.0` if the current latest version is `4.x` (for example)

test/contrib/hdf5_test.py
```diff
@@ -0,0 +1,148 @@
...
+        self.df.loc[0, "id"] = True
+        self.df.loc[1, "id"] = "somestring"
+        with self.assertRaises(TypeError):
+            target.write(self.df.iloc[:100])
+
+    def tearDown(self):
+        shutil.rmtree(self.dir)
+
+
+if __name__ == "__main__":
```
Tarrasch: Remove this part please. We run tests with tox+nosetests so this thing is not necessary.

luigi/contrib/hdf5.py
```diff
@@ -0,0 +1,308 @@
...
+    import pandas as pd
+except ImportError:
+    logger.warning("Loading hdf5 module without the python packages pandas. \
+        This will crash at runtime if pandas functionality is used.")
+
+    class Object(): # ensures docs are built will lead to fail at runtime
+        HDFStore = object
+    pd = Object()
+
+if sys.version_info < (3, 3):
```
Tarrasch: If somebody on StackOverflow or something told you to do this. It's always nice to have a url linking back to that SO-answer as a comment in the source code. :)

Tarrasch:
Thanks for including both tests and docs. This looks really well worked through. Thanks for your efforts! :)

If you fix the in-line comments and add some module level documentation. You already wrote a good "Description" in this PR. Perhaps parts of it can be copied as module level documentation?

kayibal:
Thank you @Tarrasch for reviewing. I included the changes you proposed and added some module level documentation.

Tarrasch:
@kayibal. Ok Great. Only two things remain. 1) Fix the broken tests on [travis](https://travis-ci.org/spotify/luigi/builds/132262365) and 2) explain why you seem to post and commit with different github accounts.

Looking forward to merge this once it's all resolved :)

kayibal:
Alright:) 
1) The tests fail because pytables can't install because of the hdf5 C library missing. Not sure how to use apt-get in travis will try to figure this out.
2) Looks like I had mixed up some credentials on my local account. I just borrowed this notebook. Sry for that.

Tarrasch:
@kayibal, I think it's just to do `apt-get install blah`, in particular *don't* use `sudo`.

sixy6e:
Hi,

I'd like to use these features, anything I can do to help things along?

I've taken a look at the build log, and using the "addons" key might do the trick, rather than using sudo.
https://docs.travis-ci.com/user/installing-dependencies/#Adding-APT-Packages

eg:
```yaml
addons:
    apt:
    packages:
    - libhdf5-serial-dev
```

kayibal:
@sixy6e
We have been using the target since May and it evolved a lot providing atomic operations as well. I just didn't have the time to fix the travis.yml yet but will try with addons later today. If it works I will also add the newest version of the target :)

kwilcox:
Also very interested in this. We are currently using a home grown target that isn't nearly as nice as this!

kayibal:
@sixy6e , @kwilcox I updated the pull request to include the new updated target. All tests are passing now.
@Tarrasch I had to add verbosity to the tox commands because the py33 env took too long to build (it seems to compile pandas from source) and as travis received no further output it marked the test as failed after 10min. I tried using `travis_wait` but then all tox commands would wait longer instead of failing after 10min no output.

sixy6e:
Awesome!

Looking forward to trying out the new stuff.

Cheers
Josh

.travis.yml
```diff
@@ -72,4 +82,4 @@ script:
 
 branches:
   only:
-    - master
+    - master
```
Tarrasch: Please add back the trailing newline :)

tox.ini
```diff
@@ -44,7 +46,10 @@ setenv =
   COVERAGE_PROCESS_START={toxinidir}/.coveragerc
   FULL_COVERAGE=true
   nonhdfs: NOSE_WITH_DOCTEST=1
+  hdf5: NOSE_ATTR=hdf5
 commands =
+  # ensure pandas is installed after cython
```
Tarrasch: why? can you link to some known ticket or something?

.travis.yml
kayibal: Will do :)

tox.ini
kayibal: Pandas needs cython installed to build from source because there is no py33 wheel on PyPi. See https://pypi.python.org/pypi/pandas

Tarrasch: Is this only py33? If so. I suggest we simply remove that build.

luigi/contrib/hdf5.py
```diff
@@ -0,0 +1,592 @@
...
+        if need_close_store:
+            store.close()
+        return children
+
+    def __contains__(self, item):
+        with pd.HDFStore(self.store, mode="r") as store:
+            return item in store
+
+
+class Hdf5Table(object):
```
Tarrasch: I'm not sure, but can you use `_Hdf5Table` to make this not appear in the docs?

luigi/contrib/hdf5.py
```diff
@@ -0,0 +1,592 @@
...
+            if "index_table" not in split.keys():
+                raise ValueError("Split dictionary has no main key which specifies the main indexing table")
+
+    @property
+    def fs(self):
+        return self.filesystem
+
+    @property
+    def fn(self):
+        return self.store
```
Tarrasch: I think this just make things less readable. Can we remove it?

luigi/contrib/hdf5.py
```diff
@@ -0,0 +1,592 @@
...
+        if mode not in "war":
+            raise ValueError("Unsupported file mode")
+
+        if isinstance(split, dict):
+            if "index_table" not in split.keys():
+                raise ValueError("Split dictionary has no main key which specifies the main indexing table")
+
+    @property
+    def fs(self):
+        return self.filesystem
```
Tarrasch: I think this just make things less readable. Can we remove it?

luigi/contrib/hdf5.py
```diff
@@ -0,0 +1,592 @@
...
+for more detailed documentation regarding the hdf5 file format and it's parameters be sure to visit
+`pandas doc <http://pandas.pydata.org/pandas-docs/stable/>`_ and/or
+`PyTables doc <http://www.pytables.org/>`_
+
+Below a simple example on how to read and write data from a storage:
+
+.. code-block:: python
+
+    import luigi
+    from luigi.contrib.hdf5 import Hdf5TableTarget
```
Tarrasch: Is this the only exported class from this module?

luigi/contrib/hdf5.py
```diff
@@ -0,0 +1,592 @@
...
+"""
+import io
+import os
+import sys
+import logging
+import tempfile
+
+import luigi
+
+LOGGER = logging.getLogger("luigi-interface")
```
Tarrasch: Isn't  it called `logger` (lowercase) in rest of codebase?

luigi/contrib/hdf5.py
```diff
@@ -0,0 +1,592 @@
...
+
+    pd = Object()
+    LOGGER.warning("Loading hdf5 module without the python packages pandas. \
+        This will crash at runtime if pandas functionality is used.")
+try:
+    import numpy as np
+    from tables import NoSuchNodeError
+except ImportError:
+    LOGGER.warning("Loading hdf5 module without the python packages tables. \
+            This will crash at runtime if tables functionality is used.")
```
Tarrasch: Can the two `try-except ImportError`s be merged into one?

luigi/contrib/hdf5.py
```diff
@@ -0,0 +1,592 @@
...
+import luigi
+
+LOGGER = logging.getLogger("luigi-interface")
+
+try:
+    import pandas as pd
+except ImportError:
+
+    class Object:
+        HDFStore = object
```
Tarrasch: I don't understand this (hence my previous comments, perhaps?)

luigi/contrib/hdf5.py
```diff
@@ -0,0 +1,592 @@
...
+            split=split,
+            complib=complib,
+            selector=selector
+        )
+        # check if parameters are ok
+        AtomicHdf5Table.check_integrity(**self.table_params)
+
+    @property
+    def fs(self):
+        return self.filesystem
```
Tarrasch: Can this be removed? It's both unreadable (`fs`) and undocumented.

luigi/contrib/hdf5.py
```diff
@@ -0,0 +1,592 @@
...
+
+    @property
+    def fn(self):
+        return self.table_params["store"]
```
Tarrasch: This one might be worth keeping, but I would rename it to `filename` and document it.

kmader:
What is the status of this PR? It would be a useful change to have merged.

dlstadther:
Looks like @kayibal never responded to @Tarrasch 's change requests. Any update here @kayibal ?

kayibal:
Unfortunately I forgot about this aswell. I will try to give it a look again asap.

PR #1695, opened 2016-05-25
[WIP] Add docker image build task

tjamet:
Add docker support to luigi tasks
## Motivation and Context

As far as I have seen, currently, the only way to run tasks in docker container or build images is to implement it your way.
I think it would be nice to add a native support of such a feature.
Currently, a very first idea is to add a task to build a docker image.
Then, inspired on ssh module, remote file system and task run will be added if the suggestion is encouraged

There is currently an example running it.
I will be happy to add unittests as soon as the feature suggestion is encouraged

### DIFF
--- a/examples/docker_build.py
+++ b/examples/docker_build.py
@@ -1,0 +1,28 @@
+from __future__ import absolute_import
+
+import luigi
+import luigi.contrib
+import luigi.contrib.docker
+import os
+
+
+class Dockerfile(luigi.Task):
+    context_path = luigi.Parameter(default='.')
+
+    def run(self):
+        if not os.path.exists(self.context_path):
+            os.makedirs(self.context_path)
+        with self.output().open('w') as f:
+            f.write('FROM alpine\n')
+            f.write('RUN apk update && apk add git\n')
+
+    def output(self):
+        return luigi.LocalTarget('%s/Dockerfile' % self.context_path)
+
+
+class BuildImage(luigi.contrib.docker.DockerImageBuildTask):
+    def requires(self):
+        return [Dockerfile(context_path=self.path)]
+
+if __name__ == '__main__':
+    luigi.build([BuildImage(name='luigi-built-image', path='build/context')], local_scheduler=True)
--- a/examples/docker_run.py
+++ b/examples/docker_run.py
@@ -1,0 +1,17 @@
+from __future__ import absolute_import
+
+import luigi
+import luigi.contrib
+import luigi.contrib.docker
+
+
+class DockerHelloWorld(luigi.contrib.docker.DockerTask):
+
+    @property
+    def command(self):
+        return [{
+            'command': ['echo', 'hello world'],
+        }]
+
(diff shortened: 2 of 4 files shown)


Tarrasch:
This sure looks exciting. As anything with docker these days. ^^

I haven't done much docker myself, but there's been any questions about it on the mailing list. As you said the only way to do docker in luigi is to roll your own and it's tempting to at least offer some ready-made implementation, like the one in this PR. May I just suggest that you also send a mail to the mailing list and asking for other docker-enthusiasts to express their thoughts and opinions in this PR?

tym-oao:
Since I don't see a thread on the mailing list, I'll weigh in briefly with my own docker-user thoughts here. I agree this looks like an exciting start. I can see a build image Task being useful. I would also like to see a way to invoke `docker run`, ideally with parameters to pass a command and mount volumes to the resulting container. This would allow for creation execution of task processes that are themselves containerized.
A standard, official Dockerfile for Luigi would probably be a good idea as well, and an aid to accomplishing my other suggestions. I see there a couple of GitHub projects for Luigi on Docker, but the ones I've found are based on `phusion/baseimage` which seems heavier than necessary to me. I've been running Luigi in a small `alpine`-based container for over a year now (the Docker Hub library of official images are all moving to `alpine`), and could contribute something along these lines if there is interest. It is possible to include an s6 supervisor in alpine, if something like the baseimage approach seems necessary. I generally do without it myself, and just use `luigid` as my container command directive.

tjamet:
@tym-oao yes, the idea is to provide a full support of docker containers inside luigi.
Including tasks being ran in `docker run`
I'm very pleased to see the interest this PR creates and will go on working on it

ulzha:
To run an arbitrary task containerized, one can already build an image that packages Luigi together with the task code and dependencies, and run that image.

Building images is usually part of deployment process, for which dedicated build tools are used; I suspect that this would be Luigi reinventing the wheel in that area...

tym-oao:
@ulzha 

> -Building images is usually part of deployment process

I think the key word here is "usually". What I am hoping to achieve for batch jobs in general is that each would come up in a container (based on some standard image or set of images) with the job invocation as the CMD, so that the container would stop itself once the job ended. I can see doing this now for a Luigi workflow (in fact, I am very close to doing this now already, and I know many already do). What I think a Luigi docker task would offer is the ability to do the same sort of thing at the individual Task level, and have all of Luigi's strengths in managing that containerized dependencies, as we enjoy having it do for any other sort of task. Each containerized Task might itself invoke a `luigi` command line job, or (of course) any other sort of desired processing from outside Luigi, including programs in other languages. I suppose even the sort of thing I'm describing now could already be done, using `docker-py` and some logic in a run() call, but like @tjamet said at the beginning, it seems like it would be nice to have a `contrib` module to handle that in a standard, idiomatically Luigi-ish way, so we don't all each have to roll our own.

tjamet:
Thanks all for your feedback, will update the PR with more docker support soon

PR #1696, opened 2016-05-25
Uses more thorough pid check for locking

daveFNbuck:
## Description
Adds `-x` to the `ps` command used in lock.py for checking the command used to run the current process.

## Motivation and Context
I was recently having issues where my jobs started failing once I scheduled them in cron jobs on an osx machine. After investigating, I eventually found that the PID wasn't being found using the standard `ps` command. Adding `-x` includes processes that don't have a controlling terminal, fixing my issue.

I ran this on the machine that was having trouble and it fixed my problems. I also verified that the `-x` option is available on my Ubuntu machines. Not sure how to check its universality though.

### DIFF
--- a/luigi/lock.py
+++ b/luigi/lock.py
@@ -42,7 +42,7 @@ def getpcmd(pid):
                 _, val = lines
                 return val
     else:
-        cmd = 'ps -o pid,args'
+        cmd = 'ps -xo pid,args'
         with os.popen(cmd, 'r') as p:
             # Skip the column titles
             p.readline()


Tarrasch:
LGTM

nugend:
Is this a safe change to make? On Centos 6.5, I'm seeing the following warning: 
```
ps -xo pid,args
Warning: bad syntax, perhaps a bogus '-'? See /usr/share/doc/procps-3.2.8/FAQ
```

Tarrasch:
oh, I actually don't know. Unfortunately we only test on Ubuntu machines with Travis.

daveFNbuck:
I actually started getting the same warning on Ubuntu, just haven't had
time to look into it yet.

nugend:
I'm pretty sure this is behind a bug a colleague is seeing. Do you have a recommended fix?

I was looking at the documentation for our version of `ps` and it seems like it suggests `ps x` (no dash) as a way to remove the tty restriction.

Is it possible that there was a typo made in translating from a shell command that you tested to the patch?

The final command in that case would be:

`ps x -o pid,args`

PR #1700, opened 2016-05-31
Fixes Sphinx version for docs to 1.3b1

daveFNbuck:
## Description
Changes Sphinx requirement for docs from `Sphinx>=1.3b1` to `Sphinx==1.3b1` and updates the sphinx warn_node override to better match the function signature in newer versions.

## Motivation and Context
The docs test is currently broken because it's incompatible with the latest version of Sphinx. By fixing Sphinx to an exact version number, we avoid picking up a newer version that breaks everything. By also fixing the bug that broke the newer version, we can use a newer version if we choose.

I ran the doc test and it succeeded with the lower version with and without the bugfix and with the latest version with the bug fix.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -53,7 +53,7 @@ except ImportError:
     pass
 
 
-def _warn_node(self, msg, node):
+def _warn_node(self, msg, node, *args, **kwargs):
     """
     Mute warnings that are like ``WARNING: nonlocal image URI found: https://img. ...``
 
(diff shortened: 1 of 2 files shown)


Tarrasch:
Thanks. Though I don't understand why you didn't bump the version to today's most recent.

daveFNbuck:
I haven't worked on the documentation much, so I didn't want to be responsible for changing how it's generated. If you want that version, go for it.

Tarrasch:
Sure. Your defensiveness here makes sense. Again thanks for the PR. :)

PR #1701, opened 2016-06-01
[docs] Bump sphinx version for tox builds

Tarrasch:
## Description
As a follow-up of #1700.  I just thought it's better to use a as recent version as possible.

No. I assume nothing breaks if Travis is green.


Tarrasch:
Interesting. Codecov doesn't seem to have gone berserk this time.

daveFNbuck:
Maybe it synced properly with master after the last merge.

PR #1702, opened 2016-06-02
Aws session token

fordb:
## Description
I added the ability to optionally set a [session token parameter](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html) when loading an S3 file into a Redshift table, similar to how one can set the `aws_access_key_id` and the `aws_secret_access_key`.

## Motivation and Context
I was not able to use the `S3CopyToTable` or `S3CopyJSONToTable` because my credentials require an AWS session token.

In both the above classes I added a function `aws_session_token` and changed the credentials string. If `aws_session_token` is set, then `;token=<aws session token>` is appended to the credentials string. If not, then the string will stay the same as it was before.

I ran this locally and it worked for me.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -142,7 +142,6 @@ or held presentations about Luigi:
 * `enbrite.ly <http://enbrite.ly/>`_
 * `Dow Jones / The Wall Street Journal <http://wsj.com>`_
 * `Hotels.com <https://hotels.com>`_
-* `Simulmedia <http://www.simulmedia.com/>`_
 
 We're more than happy to have your company added here. Just send a PR on GitHub.
 
--- a/README.rst
+++ b/README.rst
@@ -142,6 +142,7 @@ or held presentations about Luigi:
 * `enbrite.ly <http://enbrite.ly/>`_
 * `Dow Jones / The Wall Street Journal <http://wsj.com>`_
 * `Hotels.com <https://hotels.com>`_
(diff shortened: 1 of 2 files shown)


dlstadther:
Can you move your readme.rst change to a separate PR?

fordb:
@dlstadther sure, will do that now

dlstadther:
Your changes to redshift.py do not _appear_ to break non-token usage.

At first glance, lgtm.

fordb:
I had print statements to check. if the session token is not set then the credentials string is the same as what is currently in master. Admittedly, I can't _actually_ test it locally because my connection requires the token

dlstadther:
Right, if the optional `aws_session_token

Tarrasch:
@dlstadther, do you want to merge this? I really don't know anything about this module :)

dlstadther:
@Tarrasch I'd be happy to! I'm confident this is non-breaking, and if it works for @fordb then it accomplishes its purpose.

Tarrasch:
Thanks @dlstadther, I'll let you take over this PR then  :)

dlstadther:
Thanks @Tarrasch ! *Fingers crossed* i didn't screw up. I squashed commits (since only one was meaningful) and modified the commit message.

Tarrasch:
:+1:

PR #1704, opened 2016-06-04
Add Task.getinstance(), Task.instantiate(), and getinstances() methods.

keisuke-nakata:
## Description
Automatic instantiation for dependent tasks via `self.getinstance()` instead of `self.input()`.

## Motivation and Context

When tasks are reused many times, I need to write same instantiation codes in `Task.run()`.
For example, if a task `MyTask` whose output is a csv is reused many times, I need to write same instantiation codes in its child tasks as:
```python
class ChildTask1(luigi.Task):
    def requires(self):
        return MyTask()
    def output(self):
        pass
    def run(self):
        ### this instantiation is redundant
        with self.input().open('r') as infile:
            reader = csv.reader(infile)
            indata = [row for row in reader]
        # do stuff

class ChildTask2(luigi.Task):
    def requires(self):
        return MyTask()
    def output(self):
        pass
    def run(self):
        ### this instantiation is redundant
        with self.input().open('r') as infile:
            reader = csv.reader(infile)
            indata = [row for row in reader]
        # do another stuff
```
Instead above, I want to write it like:
```python
class MyTask(luigi.Task):
    def requires(self):
        ...
    def instantiate(self):
        with self.output().open() as outfile:
            reader = csv.reader(outfile)
            outdata = [row for row in reader]
        return outdata

class ChildTask3(luigi.Task):
    def requires(self):
        return MyTask()
    def output(self):
        pass
    def run(self):
        ### automatic instantiation
        indata = self.getinstance()  # get an instance of self.input instead of file pointer
        # do stuff
```

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -464,6 +464,30 @@ class Task(object):
         """
         return getpaths(self.requires())
 
+    def getinstance(self):
+        """
+        Returns the outputs of the Tasks as an object returned by :py:meth:`requires`
+
+        :return: a list of objects of :py:class:`Target` objects which are
+                 specified as outputs of all required Tasks.
+        """
+        return getinstances(self.requires())
+
+    def instantiate(self):
+        """
+        The instantiation method used for :py:meth:`output`.
+        If this method is implemented by a user, `Task` can automatically
+        convert its target to a Python object via `self.getinstance()`.
+
+        .. code:: python
+
+            def instantiate(self):
+                with self.output().open() as outfile:
+                    outobject = [row for row in outfile]
+                return outobject
+        """
+        pass
+
     def deps(self):
         """
         Internal method used by the scheduler.
@@ -590,6 +614,27 @@ def getpaths(struct):
             raise Exception('Cannot map %s to Task/dict/list' % str(struct))
 
         return [getpaths(r) for r in s]
+
+
+def getinstances(struct):
+    """
+    Maps all Tasks in a structured data object to their .instantiate().
+    """
+    if isinstance(struct, Task):
+        return struct.instantiate()
+    elif isinstance(struct, dict):
+        r = {}
+        for k, v in six.iteritems(struct):
+            r[k] = getinstances(v)
+        return r
+    else:
+        # Remaining case: assume r is iterable...
+        try:
+            s = list(struct)
+        except TypeError:
+            raise Exception('Cannot map %s to Task/dict/list' % str(struct))
+
+        return [getinstances(r) for r in s]
 
 
 def flatten(struct):


keisuke-nakata:
Because my coding skill (and English) is poor, this pull request should be rejected.
I just want to show my idea and hear some opinions about it.

erikbern:
why can't this be solved using normal inheritance?

keisuke-nakata:
@erikbern OK here is another implementation for my idea. How about this?
```python
def getinstances(struct):
    if isinstance(struct, luigi.Task):
        return struct.instantiate()
    elif isinstance(struct, dict):
        return {k: getinstances(v) for k, v in six.iteritems(struct)}
    else:
        # Remaining case: assume r is iterable...
        try:
            s = list(struct)
        except TypeError:
            raise Exception('Cannot map %s to Task/dict/list' % str(struct))
    return [getinstances(r) for r in s]

class MyParentTask(luigi.Task):
    def requires(self):...
    def output(self):...
    def run(self):...
    def instantiate(self):
        with self.output().open() as outfile:
            reader = csv.reader(outfile)
            outdata = [row for row in reader]
        return outdata

class MyChildTask(luigi.Task):
    def requires(self):
        return MyParentTask()
    def output(self):...
    def run(self):
        indata = getinstances(self.requires())
        ...
```
Does the method `getinstances()` have any chance to be merged into the source?

erikbern:
i don't get why MyChildTask can't just inherit from MyParentTask?

keisuke-nakata:
@erikbern Sorry I could not understand why MyChildTask can inherit from MyParentTask...
In my understanding, the class inheritance is different from luigi's Task dependency system.

My motivation is: In every `MyChildTask` which requires `MyParentTask`, I need to write the same code to map `MyParentTask`'s output (like csv file) to Python's object (like `list` of `list`s), because `MyChildTask.input()` just returns `target`s. If `MyParentTask` has a method for instantiating its own output for its children tasks, children tasks need not to consider how to map input file to object, because its parent task do it.

erikbern:
yeah but why can't you put that stuff on a parent class and use inheritance to duplicate the functionality?

keisuke-nakata:
I got it. I can implement my motivation as below. I'm sorry to trouble you.
```python
class DependentTask(luigi.Task):
     def requires(self):...
     def output(self):...
     def run(self):...
     def instantiate(self):
          with self.output().open() as outfile:
              reader = csv.reader(outfile)
              outdata = [row for row in reader]
          return outdata

class MyTask(luigi.Task):
    def requires(self):
        return DependentTask()
    def output(self):...
    def run(self):
        indata = self.requires().instantiate()  # instead of self.input().open()
        ...
```

Tarrasch:
Thanks for your pull request @keisuke-nakata. :)

PR #1708, opened 2016-06-06
Ensure DictParameter's are normalized to FrozenOrderedDict

### LINKED ISSUE #1657: Task finished yet the luigi failed.

nugend:
## Description
I normalized dict parameters to instances of Frozen Ordered Dict's

## Motivation and Context
There was an outstanding bug

I have included unit tests

Closes #1657

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -786,6 +786,12 @@ class DictParameter(Parameter):
                 return obj.get_wrapped()
             return json.JSONEncoder.default(self, obj)
 
+    def normalize(self,value):
+        """
+        Ensure that dictionary parameter is converted to a FrozenOrderedDict so it can be JSON serialized.
+        """
+        return FrozenOrderedDict(value)
+
     def parse(self, s):
         """
         Parses an immutable and ordered ``dict`` from a JSON string using standard JSON library.
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -786,9 +786,9 @@ class DictParameter(Parameter):
                 return obj.get_wrapped()
             return json.JSONEncoder.default(self, obj)
 
-    def normalize(self,value):
+    def normalize(self, value):
         """
-        Ensure that dictionary parameter is converted to a FrozenOrderedDict so it can be JSON serialized.
+        Ensure that dictionary parameter is converted to a FrozenOrderedDict so it can be hashed.
         """
         return FrozenOrderedDict(value)
 
--- a/test/execution_summary_test.py
+++ b/test/execution_summary_test.py
@@ -932,3 +932,56 @@ class ExecutionSummaryTest(LuigiTestCase):
         self.run_task(Foo())
         s = self.summary()
         self.assertIn('yellow', s)
+
+    def test_with_dict_dependency(self):
+        """ Just test that it doesn't crash with dict params in dependencies """
+
+        args = dict(start=datetime.date(1998, 3, 23), num=3)
+
+        class Bar(RunOnceTask):
+            args = luigi.DictParameter()
+
+        class Foo(luigi.Task):
+            def requires(self):
+                for i in range(10):
+                    new_dict = args.copy()
+                    new_dict['start'] = str(new_dict['start'] + datetime.timedelta(days=i))
+                    yield Bar(args=new_dict)
+
+        self.run_task(Foo())
+        d = self.summary_dict()
+        exp_set = set()
+        for i in range(10):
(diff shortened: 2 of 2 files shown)


Tarrasch:
Can you fix the flake 8 issue? Otherwise this looks good :)

Tarrasch:
Actually. Can you also add a test case to `parameter_test.py` as well? One day maybe we remove or move out the execution summary functionality, and it would be good if this property is still checked.

nugend:
I've fixed the flake8 error.

After looking into the issue in more detail, this is not something that belongs in the parameter test cases. The only two places I can currently identify that this is an issue is in the execution summary and in the worker (where the worker checks if all the dependent tasks are complete and the inability to hash parameters causes, I *believe*, the lookup of the dependent task instances in the task registry to fail).

If you could help me figure out the best way to test this issue given that, I will add an additional test case in worker_test.py or another test module that you feel is appropriate.

Alternatively, this may be considered an issue with either worker or in how task_register handles its lookup. Honestly, I would consider the latter to be the case now that the task_id property is ostensibly a canonical string.

Regarding the flake8 error: Whenever I run `tox -e flake8` on my dev workstation, I see errors for contrib/ecs and contrib/salesforce which mean I don't get a green result. Is there a particular way I should be invoking flake8 to not trigger those? Or is it something about flake8 + Python 3.5?

Tarrasch:
As for flake8, I have no idea. probably something with python 3.5. I hoped that tox would run exactly the same independent of platform, but it seems not. Oh well.

----

As for the additional test. I think you overcomplicate this. As I see it, hashability is a property for all parameters, so this situations simply warrants another simple test in `parameter_test.py`. That said, any added test case to any module would be greatly appreciated.

---

Lastly, finally we can push for merging in your range-related patch. :)

nugend:
Ahhh, I did not realize that was intended to be a property of all Parameters.

I will add a corresponding simple test immediately. Should we also open an issue to add similar tests for other parameter types so as to ensure that this isn't an issue anywhere else? 

Also, is TaskParameter currently hashable?

Tarrasch:
> Also, is TaskParameter currently hashable?

I think so, the reason I think so is that Range uses it, and there's not issue with combining Range with execution_summary (which apparently crashes when something isn't hashable)

Issue #1706, opened 2016-06-06
ImportError when using pinject on Linux machine


hudswideworld:
Hi, I've been using Luigi for a bit and just recently decided to try using pinject to do some DI for unit testing. However whenever I run my code (which used to work), but with pinject injecting the dependency instead of the direct call, I get this error:

Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/luigi/worker.py", line 175, in run
    new_deps = self._run_get_new_deps()
  File "/usr/lib/python2.7/site-packages/luigi/worker.py", line 117, in _run_get_new_deps
    task_gen = self.task.run()
  File "dispatcher/tasks.py", line 30, in run
    cas_client = client.CasClient()
  File "../../lib/casclient/client.py", line 31, in __init__
    cas_req_obj_graph = pinject.new_object_graph(binding_specs=[CasRequestsBindingSpecs()])
  File "/usr/lib/python2.7/site-packages/pinject/object_graph.py", line 105, in new_object_graph
    found_classes = finding.find_classes(modules, classes)
  File "/usr/lib/python2.7/site-packages/pinject/finding.py", line 32, in find_classes
    all_classes |= _find_classes_in_module(module)
  File "/usr/lib/python2.7/site-packages/pinject/finding.py", line 47, in _find_classes_in_module
    for member_name, member in inspect.getmembers(module):
  File "/usr/lib64/python2.7/inspect.py", line 253, in getmembers
    value = getattr(object, key)
  File "/usr/lib/python2.7/site-packages/luigi/six.py", line 95, in __get__
    result = self._resolve()
  File "/usr/lib/python2.7/site-packages/luigi/six.py", line 118, in _resolve
    return _import_module(self.mod)
  File "/usr/lib/python2.7/site-packages/luigi/six.py", line 85, in _import_module
    __import__(name)
ImportError: No module named _winreg

Doing some research, I found that this is a bug in six 1.9.0, which they fix in 1.10.0 with this 
[bit of code](https://bitbucket.org/gutworth/six/pull-requests/55/_winreg-is-added-to-the-moves-module-under/diff). I was able to work around it by just moving the updated six.py file into the luigi folder. Just thought I'd give a heads up

dlstadther:
If this is an update to `six.py`, would you mind submitting a PR with the changes?

hudswideworld:
Done

Tarrasch:
Fixed in #1709

PR #1709, opened 2016-06-06
Updated six to version 1.10.0

hudswideworld:
## Description
Updated six.py from v1.9.0 to v.1.10.0

## Motivation and Context

Six v.1.9.0 imports the _winreg module, which only works on Windows, and results in an ImportError when running on Linux machines. Six v.1.10.0 fixes that.

I have run my code and it works. I was only running into problems when I tried to use the pinject library, and this fixed my ImportError

### DIFF
--- a/luigi/six.py
+++ b/luigi/six.py
@@ -1,9 +1,4 @@
-"""Utilities for writing code that runs on Python 2 and 3
-
-In luigi, we hard-copy this file into the project itself, to ensure that all
-luigi users use the same version of six.
-
-"""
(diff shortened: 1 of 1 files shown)


dlstadther:
resolves #1706

dlstadther:
There are quite a number of newline additions. Did you modify the `six.py` module from Pypi, or is this the original?

hudswideworld:
I did not modify anything. This should be the original

Tarrasch:
Thanks!

PR #1710, opened 2016-06-07
Fix RemoteFileSystem._ftp_exists once more

andreabedini:
## Motivation and Context

Commit bdfd225 from PR  #1602 breaks _ftp_exists. The problem is that nlist(path)
returns a list of paths starting with path.

E.g.

```python
In [40]: conn.nlst()
Out[40]: ['public', 'pub', 'ietf', 'apnic', 'uploads', 'welcome.msg']

In [41]: conn.nlst('pub/stats')
Out[41]:
['pub/stats/lacnic',
 'pub/stats/ripe-ncc',
 'pub/stats/iana',
 'pub/stats/afrinic',
 'pub/stats/apnic',
 'pub/stats/arin']
```

The test `fn in files` will always be false since fn is only the basename of
the path.

Also, conn.sendcmd('MDTM ' + path) will fail if path is a not a regular file.
This is never triggered because the above test always fails.

Works for me.

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -131,14 +131,11 @@ class RemoteFileSystem(luigi.target.FileSystem):
         return exists
 
     def _ftp_exists(self, path, mtime):
-        path_parts = path.split('/')
-        path = '/'.join(path_parts[:-1])
-        fn = path_parts[-1]
-
-        files = self.conn.nlst(path)
+        dirname = os.path.dirname(path)
+        files = self.conn.nlst(dirname)
 
         exists = False
-        if fn in files:
+        if path in files:
             if mtime:
                 mdtm = self.conn.sendcmd('MDTM ' + path)
                 modified = datetime.datetime.strptime(mdtm[4:], "%Y%m%d%H%M%S")


andreabedini:
I am still not entirely happy about the way it's done. We only want to check if a single file exist and whether it is has been updated since a given timestamp but `nlist` fetches the entire directory content, which potentially includes a large number of files.

dlstadther:
I realize that I did not write tests for this ftp module (and they do seem more and more necessary). However, I do not get the same result as you for `conn.nlst(...)`.

I edited the beginning of the ftp exists method as follows:

```python
def _ftp_exists(self, path, mtime):
        path_parts = path.split('/')
        path = '/'.join(path_parts[:-1])
        fn = path_parts[-1]

        print(path)  # view the ftp path (without the filename)
        files = self.conn.nlst(path)
        pprint(files)  # see the list of files within that path
        ...
```

Suppose the ftp I connect to has the following directory structure:

```
/
/root.zip
/import/
/import/test.txt
/import/test.csv
/import/ds_test/
/import/ds_test/sample.txt
/export/
```

`exists('/import/test.txt')` prints the path `/import` and the listed files/dirs in that path `['ds_test','test.txt','test.csv']`, as well as returns `true` because `test.txt` exists on the ftp server on the path `/import/test.txt`.

This is the same form of output for `exists('/root.zip')`, `exists('/import/test.csv')`, and `exists('/import/ds_test/sample.txt')`.

Lastly, to prove to myself that `exists(...)` will return `false` for a file which doesn't exist, I mistyped the extension types for the 3 level paths mentioned in the previous paragraph. All cases returned `false` for existence.

All this to say, I'm not sure how your `conn.nlst('pub/stats')` returns a list of files with fully qualified path names.

Since we get different outputs for this method, it may be worth you or I (or both) writing the mock tests for this module (or at least for the ftp side) and submitting a PR.

------------------

Regarding your second concern about `mdtm = self.conn.sendcmd('MDTM ' + path)` failing for non-regular file paths (which I'm not sure what you mean), I never use this - it existed in `ftp.py` prior to any of my modulations. I kept it and assumed it functioned.

andreabedini:
thanks for the feedback @dlstadther, I appreciate it 💯. Might it be a up to the ftp server? Here is my reproducible test (Python 3.5.1):

```python
#!/usr/bin/env python3

import ftplib
from pprint import pprint

conn = ftplib.FTP('ftp.apnic.net')
conn.set_debuglevel(1)
conn.login()

pprint(conn.nlst())
pprint(conn.nlst('pub/stats'))
```

my output is

```bash
❯ python3 test.py
*cmd* 'USER anonymous'
*resp* '331 Anonymous login ok, send your complete email address as your password'
*cmd* 'PASS **********'
*resp* '230-Welcome to the master FTP site for APNIC Pty. Ltd.\n There are currently 4 users out of 30 possible.\n \n \n     **********************************************************************\n     ** This Archive is also accessible via HTTP at http://ftp.apnic.net **\n     **********************************************************************\n \n The local time is Tue Jun 07 23:01:07 2016.\n \n If you have problems, please try using a dash (-) as the first character\n of your password -- this will turn off the continuation messages that may\n be confusing your ftp client.\n \n NOTE: All transactions with this server are logged. \n       If you do not like this, disconnect now!\n \n Please send mail to technical@apnic.net if you experience any problems.\n \n230 Anonymous access granted, restrictions apply'
*cmd* 'TYPE A'
*resp* '200 Type set to A'
*cmd* 'EPSV'
*resp* '229 Entering Extended Passive Mode (|||58394|)'
*cmd* 'NLST'
*resp* '150 Opening ASCII mode data connection for file list'
*resp* '226 Transfer complete'
['public', 'pub', 'ietf', 'apnic', 'uploads', 'welcome.msg']
*cmd* 'TYPE A'
*resp* '200 Type set to A'
*cmd* 'EPSV'
*resp* '229 Entering Extended Passive Mode (|||58113|)'
*cmd* 'NLST pub/stats'
*resp* '150 Opening ASCII mode data connection for file list'
*resp* '226 Transfer complete'
['pub/stats/lacnic',
 'pub/stats/ripe-ncc',
 'pub/stats/iana',
 'pub/stats/afrinic',
 'pub/stats/apnic',
 'pub/stats/arin']
```

Night time here, I''ll have a better look tomorrow.

dlstadther:
All ftp servers I work with are not under my control. So i cannot speak to the possibility that this is a server setting.

I use Python 2.7.10, but I just ran my same code with Python 3.5.1 and received results without fully qualified path names.

I then noticed that you leave out the root directory slash `/`. I also removed that initial `/`, but continue to get the same result. 

Running your code on my machine returns the same result you provided...

I'll git blame some people and see if someone knows anything about this.

dlstadther:
@steenzout @ajornetic3 @interskh @themalkolm @a86c6f7964 @gpoulin @jeffcjohnson @daveFNbuck @xrmx 

You all were the results of git blame on `ftp.py`. Can anyone weigh-in on why @andreabedini and I receive different format results for `conn.nlst(...)` for our different ftp servers?

a86c6f7964:
just an FYI. 
```
python --version
Python 2.7.11
```
when I use the raw ftp lib am seeing the same behavior as @andreabedini

```
conn.cwd('/')
print conn.nlst('pub/tmp')
```
 
```
*cmd* 'CWD /'
*resp* '250 CWD command successful.'
*cmd* 'TYPE A'
*resp* '200 Type set to A.'
*cmd* 'PASV'
*resp* '227 Entering Passive Mode (204,64,182,14,104,103)'
*cmd* 'NLST pub/tmp'
*resp* '150 Opening ASCII mode data connection for file list.'
*resp* '226 Transfer complete.'
['pub/tmp/20160512.dmp']
```

a86c6f7964:
it does not look like the RFC (https://www.w3.org/Protocols/rfc959/4_FileTransfer.html) specifies what the server should return. So if you are seeing different behavior (what is the ftp server you are running?) and I guess the code will have to be defensive about getting both types of data back.

dlstadther:
The ftp servers i use are owned and managed by companies whose services my company uses for our operations. I'm only given a host name, u/n, and p/w. That's all i know.

andreabedini:
Reading the RFC it looks like it's common for the FTP server to return error code 550, if the file is not available. Perhaps we could just try to fetch it can catch the exception with that error code? Happy to go down this route if people agree ...

daveFNbuck:
Trying to fetch is a bit expensive if the file exists. Couldn't we just do `if fn in files or path in files`?

Side note: the RFC specifies that the file names returned must be valid pathnames. The RFC does not give a specification for how pathnames should look (and explicitly says so), but does specify that they must uniquely identify a file on the server. So giving just the file name without a full path would probably violate the RFC unless file names are globally unique.

luigi/contrib/ftp.py
```diff
@@ -131,14 +131,11 @@ def _sftp_exists(self, path, mtime):
         return exists
 
     def _ftp_exists(self, path, mtime):
-        path_parts = path.split('/')
-        path = '/'.join(path_parts[:-1])
-        fn = path_parts[-1]
-
-        files = self.conn.nlst(path)
+        dirname = os.path.dirname(path)
```
daveFNbuck: The split and join was probably done to ensure that this still works on Windows because os.path.dirname works differently depending on the os you're running under :(. It would probably be a little better to do `dirname, _, fn = path.rpartition('/')` if you want something cleaner than the split/join.

dlstadther: I did the split and join to separate the filename from the file path. `os.path.dirname(path)` is the better way to go about this.

daveFNbuck: I just looked it up and we can use posixpath.basename (or posixpath.split to get the dirname too) to ensure that this works consistently across systems.

luigi/contrib/ftp.py
```diff
@@ -131,14 +131,11 @@ def _sftp_exists(self, path, mtime):
...
+        files = self.conn.nlst(dirname)
 
         exists = False
-        if fn in files:
+        if path in files:
             if mtime:
```
dlstadther: @daveFNbuck , I believe i used split and join over `os.path.dirname()` and `posixpath.basename` because I also needed the filename such that I could check for existence in the returned results from `nlst`.

This is where @andreabedini and I's FTP servers appear to be performing `nlst` differently. The servers I deal with only return filenames, whereas his returns fully qualified path names. I believe this is the main concern we need to code around.

daveFNbuck: Yeah, the simple solution is

```
dirname, fn = posixpath.split(path)
...
if path in files or fn in files:
```

now everyone's happy

dlstadther: @andreabedini Do you want to make this change?

andreabedini: oh gosh, sorry guys got busy and completely forgot about this 😞

andreabedini:
superseded by #1751

PR #1713, opened 2016-06-07
Can yield (not return) a list of tasks.

bmaggard:
Returning tasks from run(), even in a list, does not add dependencies.  yield does.

## Description

Clarifying the documentation to specify the proper mechanism by which to add dynamic dependencies.

## Motivation and Context

The documentation is incorrect.

I tested with the following cases.  TestYieldTask performs as desired.  TestRunTask does not.
```
class TestYieldTask(DoDynDepTestTask):

    def run(_s):
        yield [ DynDepBTestTask(), DynDepCTestTask(), ]
        super().run()
```
```
class TestRunTask(DoDynDepTestTask):

    def run(_s):
        super().run()
        return [ DynDepBTestTask(), DynDepCTestTask(), ]
```

### DIFF
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -156,7 +156,7 @@ Sometimes you might not know exactly what other tasks to depend on until runtime
 In that case, Luigi provides a mechanism to specify dynamic dependencies.
 If you yield another :class:`~luigi.task.Task` in the Task.run_ method,
 the current task will be suspended and the other task will be run.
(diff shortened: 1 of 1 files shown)


Tarrasch:
Thanks. Usually for these small docs fixes, you can get away with a more lightly motivated PR (as the maintainers know already if you must `yield` or `return`). :)

While at it, I guess it should say "yield any iterable", not only "yield a list"

PR #1716, opened 2016-06-09
DateParameter docs: Add example with interpolation

Tarrasch:
There's many many times in my time using luigi where I've wanted to look this up. Hopefully not only I but also others save time by having this reminder.

I got motivated to do this after getting this as a question in the email group.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -317,6 +317,20 @@ class DateParameter(_DateParameterBase):
 
     A DateParameter is a Date string formatted ``YYYY-MM-DD``. For example, ``2013-07-10`` specifies
     July 10, 2013.
+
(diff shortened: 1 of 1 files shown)


dlstadther:
LGTM. This is exactly how i use `DateParameter` - filenames, dir names, etc.

luigi/parameter.py
```diff
@@ -317,6 +317,20 @@ class DateParameter(_DateParameterBase):
 
     A DateParameter is a Date string formatted ``YYYY-MM-DD``. For example, ``2013-07-10`` specifies
     July 10, 2013.
+
+    DataParameters are 90% of the time used to be interpolated into file system paths or the like.
```
joemeszaros: I think there is a typo: `DataParameters` - `DateParameters`

PR #1717, opened 2016-06-09
corrected typo and added parameter for cluster (#1)

Zacmatix:
## Description

I corrected a typo and added a parameter to the task run
## Motivation and Context

It both solves a bug ( the typo was critical) and allows for more configuration. It was blocking for us at some point to not be able to run tasks on clusters other than the default one.

I ran my code (not sharable) and it worked just fine.


Zacmatix:
@jfeala

tox.ini
```diff
@@ -12,6 +12,7 @@ deps=
   nose<2.0
   unittest2<2.0
   boto<3.0
```
Tarrasch: I know nothing about boto. But is it possible to remove `boto` dependency now that you introduce `boto3`?

wblancqu:
We are facing the same issue, any progress on merging this PR?

Kr, Wouter

vanhove:
This PR #2045 fixes the same issue.

PR #1718, opened 2016-06-09
Adding release step for Debian packages.

linsomniac:
Debian packages build with the version number in the debian/changelog
file.  The 2.1.1 release and probably others have a version number of
0.0 in that file, so if you grab the release and do a "debuild" you get
a package of version 0.0.


Tarrasch:
Ah right :)

Thanks!

PR #1719, opened 2016-06-09
Added tests for hashing parameters except ListParameter

nugend:
ListParameter cannot be hashed. This may or may not be a problem.

Meant to add this for the other PR. Lists can't be hashed of course. Not sure what should be done wrt that.

I realize the tests are a bit repetitive, but it seems that many of the parameter tests are similar in this respect.

The reason I went through the trouble of actually instantiating a task is due to the distinction between how the parsing of the string handles the creation of the value for the parameter and that the actual instantiation logic only invokes the normalization code. This is what caused the two separate ways that errors showed up for DictParameters in the execution summary.

@Tarrasch

### DIFF
--- a/test/parameter_test.py
+++ b/test/parameter_test.py
@@ -286,6 +286,85 @@ class ParameterTest(LuigiTestCase):
         self.assertEqual(b_tuple, a.parse(a.serialize(b_tuple)))
 
 
+class TestParametersHashability(LuigiTestCase):
+    def test_date(self):
+        class Foo(luigi.Task):
+            args = luigi.parameter.DateParameter()
+        p = luigi.parameter.DateParameter()
+        self.assertEqual(hash(Foo(args=datetime.date(2000, 1, 1)).args), hash(p.parse('2000-1-1')))
+
+    def test_dateminute(self):
+        class Foo(luigi.Task):
+            args = luigi.parameter.DateMinuteParameter()
+        p = luigi.parameter.DateMinuteParameter()
+        self.assertEqual(hash(Foo(args=datetime.datetime(2000, 1, 1, 12, 0)).args), hash(p.parse('2000-1-1T1200')))
+
+    def test_dateinterval(self):
+        class Foo(luigi.Task):
+            args = luigi.parameter.DateIntervalParameter()
+        p = luigi.parameter.DateIntervalParameter()
+        di = luigi.date_interval.Custom(datetime.date(2000, 1, 1), datetime.date(2000, 2, 12))
+        self.assertEqual(hash(Foo(args=di).args), hash(p.parse('2000-01-01-2000-02-12')))
+
+    def test_timedelta(self):
+        class Foo(luigi.Task):
+            args = luigi.parameter.TimeDeltaParameter()
+        p = luigi.parameter.TimeDeltaParameter()
+        self.assertEqual(hash(Foo(args=datetime.timedelta(days=2, hours=3, minutes=2)).args), hash(p.parse('P2DT3H2M')))
+
+    def test_boolean(self):
+        class Foo(luigi.Task):
+            args = luigi.parameter.BoolParameter()
+
+        p = luigi.parameter.BoolParameter()
+        self.assertEqual(hash(Foo(args=True).args), hash(p.parse('true')))
+
+    def test_int(self):
+        class Foo(luigi.Task):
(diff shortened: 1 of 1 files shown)


test/parameter_test.py
```diff
@@ -286,6 +286,85 @@ def test_tuple_serialize_parse(self):
...
+
+        p = luigi.parameter.EnumParameter(enum=MyEnum)
+        self.assertEqual(hash(Foo(args=MyEnum.A).args), hash(p.parse('A')))
+
+    def test_dict(self):
+        class Foo(luigi.Task):
+            args = luigi.parameter.DictParameter()
+
+        p = luigi.parameter.DictParameter()
+        self.assertEqual(hash(Foo(args=dict(foo=1, bar="hello")).args), hash(p.parse('{"foo":1,"bar":"hello"}')))
```
Tarrasch: I thought the frozen dict sorted the keys. Shouldn't `bar` be before `foo`?

Tarrasch:
Really?! Is all hashing in python done by first turning the data into a string representation and then hashing it??

test/parameter_test.py
Tarrasch: Actually no: https://github.com/spotify/luigi/pull/1574#issuecomment-194234738

Hmm. this is a bit of a bad property in hindsight.

Tarrasch:
> Lists can't be hashed of course. Not sure what should be done wrt that.

Why of course? I think it was just an oversight from us when merging ListParameter. If we make it a frozen structure, it should be hashable, no?

test/parameter_test.py
nugend: Honestly, I was just thinking with the general structure of the test cases that a decent test to ensure hashability would be to run the parameter versions through the same as the value versions. That way, the hashes should be stable.

I can see where this could be an issue with the way that the DictParameter is implemented right now though (the normalization of the plain dict won't necessarily result in an item set ordered in the same way from instance to instance, and then result in equivalent dictionaries not hashing identically).

I could do some fixing up on the implementation to provide the required properties, if you like.

nugend:
I just meant standard Python lists. If a new frozen class is introduced, I'm sure it can be made to work. But given the current state of the code, adding a list test case would fail. My presumption was that you wouldn't want to merge a failing test case. I can add one if you like though.

> Really?! Is all hashing in python done by first turning the data into a string representation and then hashing it??

I have no clue. I figured using the standard `hash` function was right though since my understanding is that that is how dictionaries hash their keys.

dlstadther:
@nugend @Tarrasch What is the status of this? Do I need to update `ListParameter` to be a frozen structure before this PR can be merged?

nugend:
If ListParameters are made frozen, a list test case would need to be added.

Tarrasch:
@dlstadther, oh, if you can change the ListParameter to be a frozen structure that would be great!

dlstadther:
@Tarrasch, Am I correct that when we refer to a frozen structure, we mean that the data structure is immutable? If so, isn't a frozen list just a tuple?

Tarrasch:
@dlstadther that's correct. I just didn't realize that it would be that simple as you just said to normalize everything into the `tuple` type.

dlstadther:
@Tarrasch Does it make more sense to add tuple casting to ListParameter's normalize method? Or just  have ListParameter call TupleParameter (essentially removing any differences between ListParameter and TupleParameter)?

Tarrasch:
Oh holy macaroni we already have `TupleParameter` I just realized.

We might still call it a ListParameter although it fact the internal structure is a Tuple. Immutability should be seen as an implicit property. Super confusingly enough it seems like tuple-parameter allows 2D structures (tuples of tuples)

Also the name List is intuitive. Luigi is a tool also used by non-programmers, and "List" is more familiar than "Tuple".

-----

My summary: Yes, just add the casting in the `normalize()` method. Though be careful with implementation as when `parsing` and `serializing` we still should print the `[1,2,3]` syntax. Does this make sense? :)

dlstadther:
Yeah, `TupleParameter` allows for 2D structures since redshift/postgres read `columns` as a tuple of tuples.

I believe I understand. I'll work on this as time allows this week.

dlstadther:
After messing with this a bit, I have something. However, i'm not overly confident with it. I'll submit a PR and we can move the discussion there.

PR #1724, opened 2016-06-21
Fix default content type (salesforce)

dlstadther:
## Description
By default, `content_type` is now passed through to `create_batch()` and `create_operation_job()`. Resolves error caused by attempted reference to `self.content_type` which does not exist within `SalesforceAPI`.

## Motivation and Context
Fixes errors induced by PR #1686 . Lines 402 and 478 throw errors because `self.content_type` does not exist in `SalesforceAPI`, but rather `QuerySalesforce`. 

Tests ran successfully, locally, in addition to successful execution of my code.

### DIFF
--- a/luigi/contrib/salesforce.py
+++ b/luigi/contrib/salesforce.py
@@ -142,7 +142,9 @@ class QuerySalesforce(Task):
 
     @property
     def content_type(self):
-        """Override to use a different content type. (e.g. XML)"""
+        """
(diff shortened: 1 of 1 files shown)


dacort:
This looks good to me, thanks @dlstadther - I had come across this after the merge, but hadn't had time to address it yet.

dlstadther:
Unless I hear otherwise, I'll merge this Monday (2016-06-27).

Issue #294, opened 2014-03-19
MRJob adaptation


rudaoshi:
MRJob is a good tool to write simple multi-step hadoop jobs. I'd like luigi use MRJob to write a single node in the workflow.

Jyrsa:
Are there many people who would like to use this feature? I'm currently working on triggering S3->S3 mrjob EMR tasks using luigi for the simplicity it would offer to data flow management.

@danielhfrank wrote [Waluigi](https://github.com/danielhfrank/waluigi) a while ago and I believe by adding a few more batteries I could make it work at least for simple use cases in the abovementioned context. I can't make any guarantees about having the time to clean it up to publish, though.

@Tarrasch are there any objections on principle to having a contrib.mrjob.MRJobTask?

Tarrasch:
Nope, no objections. Though some form of tests are expected. Also, don't hesitate to reuse/refactor/generalize other modules relating to map reduce if it makes sense.

Jyrsa:
I wouldn't trust the quality of my own code without at least some tests.

The general idea is to assume that the user knows what they're doing with MRJob and just provide a simple wrapper that instantiates the object with user given parameters and and lets MRJob handle the rest. It comes with plenty of batteries already so I don't expect the bridge to be too complex or smart.

Tarrasch:
Jyrsa. It's sane too assume the user knows what they are doing. But don't forget to add documentation as pydocs, at least describing how to inherit your base task. Looking forward to your PR :)

Jyrsa:
Sorry about the delay. It was surprisingly difficult to find the person at my employer to OK my open sourcing this.

Tarrasch:
Added by @Jyrsa in #1731. Thanks again!

rudaoshi:
I wrote a extension package for luigi to support mrjob protocol and default parameters. 

https://github.com/rudaoshi/super_luigi

bnewbold:
PR #1731 never got merged; i'm not sure why this issue was closed before that happens.

PR #1731, opened 2016-06-23
Add support for running MRJob jobs as luigi Tasks.

Jyrsa:
I added a luigi.contrib.mrjob package to support running MRJob steps (particularly on EMR) as part of Luigi workflows.

It's based on waluigi by @danielhfrank with a couple of bells and whistles added that I thought would make sense when testing workflows with small loads on local machines so that moving to EMR or Hadoop would be as effortless as possible.

## Description

A class MRJobTask that parses it's input into the correct format for a MRJob JobRunner and initializes and runs the runner. Support for adding configuration parameters to the MRJob, either everything or an external config file. 

The user is assumed to have some previous experience with MRJob. 

Also a class LocalMRJobTarget for testing workflows using LocalRunner on local machine and a LocalMRJobTask to boot that I believe follows the naming convention.

## Motivation and Context

I liked the idea of luigi and wanted to use it an envisioned automated data pathway that consisted of both local and EMR steps.

Also, 
https://github.com/spotify/luigi/issues/294

I have included tests that are more integration than unit tests for the basic functionality. 

I'll admit that there are a few cryptic functions that could use more testing.

## Notes

I'm not sure about the copyright notices. I put in my company's name as I did this on their time (and got permission to publish). If you want to change that to Spotify like everywhere else let me know.

Also this is my first meaningful pull request to an open source project. Be gentle :)


luigi/contrib/mrjob.py
```diff
@@ -0,0 +1,288 @@
...
+        destination = luigi.Parameter()
+
+        job_cls = mrjob.examples.mr_word_freq_count.MRWordFreqCount
+
+        def mrjob_opts(self):
+            return {
+                '-r', 'inline'
+                }
+
+        def input(self):
```
Tarrasch: In luigi we do `require()` other tasks. Tasks in luigi has an automatic convenient method `input()` which you usually doesn't override. That `input()` is the same structure as `requires()` but with the corresponding `output()` of those tasks.

Given that you likely want requirements that luigi can understand even for MRJobtasks too. Overriding `input()` and not overriding `requires()` doesn't seem right to me at first glance.

luigi/contrib/mrjob.py
```diff
@@ -0,0 +1,288 @@
...
+            opts_list += ['--no-output']
+            return opts_list + input_files + ['-o', output_dir]
+        elif not output_remote:
+            # if user wants the output to be a local file, enable streaming of output
+            # so we can write the file in runner
+            self.output_is_local_file = True
+            return opts_list + input_files
+
+
+class LocalMRJobTarget(luigi.file.LocalTarget):
```
Tarrasch: Can you inherit a more general target? What does LocalTarget give you over [FileSystemTarget](https://github.com/spotify/luigi/blob/831e390763d9b2c83faeef43985debf54d92403a/luigi/target.py#L203)?

Tarrasch: Oh wait, I see what you do here. Isn't this some sort of a directory target???

luigi/contrib/mrjob.py
```diff
@@ -0,0 +1,288 @@
...
+    def __init__(self, *args, **kwargs):
+        super(MRJobTask, self).__init__(*args, **kwargs)
+        self.opts = self.make_mrjob_opts(*args, **kwargs)
+
+    def run(self):
+        # if the last step fails it's possible for mrjob to
+        # have created the target dir at least on s3
+        # without the actually being successfull. this causes an error and
+        # requires manual intervention. should we just delete the dir on s3 if
+        # it doesn't contain a _SUCCESS ?
```
Tarrasch: The problem you mention in the inline comments here is the one of atomic movedir. Can you make MRJob always write to a temp-name and in the end rename the file to the original filename if MRJob gives 0 exit code?

Tarrasch:
This LGTM par some comments and thoughts I wrote as inline comments. Let me know if they make sense to you :)

I really appreciate you taking the time to document everything and writing useful comments and having test coverage right from the start. Good job!

luigi/contrib/mrjob.py
Jyrsa: Ahh, OK I may have misunderstood something about the structure of requiring things. I'll have to look into it.

Jyrsa:
OK, so I had misunderstood the relationship between requires() and input() and believe I fixed that. 

Unfortunately there is no atomic movedir in S3, but rather renaming a lot of files which is a bit problematic as well. This is why the S3FlagTarget exists in general.

For testing data workflows in the local system the batteries included with MRJob already do an atomic movedir or some command that is very similar.

Jyrsa:
@Tarrasch does this look better?

I had to fix a couple of things in #1769 to get the tests to pass and they're also in this branch.

bnewbold:
Did this PR stall for a specific reason, or did folks just get busy?

erikbern:
probably just forgot about it.

seems fine to rebase and reopen

Issue #1726, opened 2016-06-22
Luigi ignores luigi.cfg


dmyersturnbull:
`luigi.cfg`:
```
[X]
y=100
```

`test.py`:
```python
import luigi

class X(luigi.Task):
        y = luigi.Parameter(default='10')]
```

```
python3 -m luigi --module test X --local-scheduler
```

Output:
```
Scheduled 1 tasks of which:
* 1 ran successfully:
    - 1 X(y=10)
```

Side note: Incidentally, I'm unable to run luigi this way:
```
luigi --module test X --local-scheduler # ImportError: No module named 'test'
```

Tarrasch:
Make sure you're running everything from the correct folders and stuff. Also try to explicitly set the [LUIGI_CONFIG_PATH](https://github.com/spotify/luigi/blob/master/luigi/configuration.py) env var.

dmyersturnbull:
I'm running from the same directory that contains `luigi.cfg` and `test.py`. I also tried setting `LUIGI_CONFIG_PATH`. I'm running on Python 3.5.

dlstadther:
Your simple program appears to function correctly. (At least, according to my usage).

If you're wanting `y` to evaluate to 100, shouldn't you use the `Config` class?

```
# luigi.cfg
[X]
y=100
```

```python
# test.py
import luigi
class X(luigi.Config):
    y = luigi.IntParameter(default=10)

class Hello(luigi.Task):
    def output(self):
        return luigi.LocalTarget('/path/to/somewhere.txt')
    def run(self):
        with open(self.output().fn, 'w') as outfile:
            outfile.write(X().y)
```

File would then contain `100` if `luigi.cfg` was found. Otherwise, file would contain `10`.

Hope this helps!

dmyersturnbull:
Thanks for the reply, @dlstadther. What version of Python are you using? I'm using 3.5.1.

First, http://luigi.readthedocs.io/en/stable/configuration.html#paramconfigingestion seems to say that `luigi.Config` is not needed.

Unfortunately, your example doesn't work for me. I had to modify `outfile.write(X().y)` to read `outfile.write(str(X().y))`, but then running `python3 -m luigi --module new_test Hello --local-scheduler` produces a file at `/path/to/somewhere.txt` with contents `10`. Similarly, running `python3 -m luigi --module new_test X --local-scheduler` produces this output:
```
Scheduled 1 tasks of which:
* 1 ran successfully:
    - 1 X(y=10)
```
[...]

PR #1733, opened 2016-06-24
Added a note about the problem in issue #1726.

dmyersturnbull:

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -75,6 +75,31 @@ We can create a :py:class:`~luigi.Config` class:
 
     mysection().option
     mysection().intoption
+    
+Note that when using ``inherits`` or ``requires``
+(http://luigi.readthedocs.io/en/stable/api/luigi.util.html), the parameter needs to be set
+in the section corresponding to the calling class rather than the inner class that
+declares the parameter. For example:
+
+.. code:: python
+
+    class Inner(luigi.Task):
+        x = luigi.Parameter(default='ten')
+        def run(self):
+            with open(self.x + '.tmp', 'w') as f:
+                f.write(self.x)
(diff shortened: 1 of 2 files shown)


Tarrasch:
LGTM

daveFNbuck:
I'm not familiar with this issue, and I still don't understand it after reading and re-reading the new documentation several times. Could you please be more clear about what will happen if you do
```
[Outer]
x=eleven
```
versus
```
[Inner]
x=eleven
```
and create instances of both Outer and Inner?

dmyersturnbull:
@Tarrasch Do you have an idea how to make this clearer?

erikbern:
would prefer if you put this in the documentation for `@requires` instead. configuration is a lot more common and i don't want to add unnecessary complexity to the documentation there

Tarrasch:
@dmyersturnbull,

Here's what I think you could do to make it clearer:

 * Move the docs as @erikbern said
 * In the docs for configuration. Perhaps only add a sentence saying something like "Any configuration done to a luigi task will never affect anything but instances of that class,  not superclasses nor subclasses.". Luigi never cares in what way a task parameter came to be (weather it's through inheritance, `@requires` or explicit retyping). Parameter objects are also immutable and more functional than object oriented.
 * Come up with a more creative name than Inner/Outer.

Tarrasch:
@dmyersturnbull, there's a flake8 issue. Can you resolve that?

dmyersturnbull:
@Tarrasch It looks fine to me. Do you see where the issue is?

Tarrasch:
Not sure. I would look for other existing examples in the docs and see how your code differs. As noted in CONTRIBUTING.md, you can debug this locally too using `tox`

PR #1734, opened 2016-06-25
Removes redundant function definitions from rpc and server

daveFNbuck:
## Description
When defining a new API call in the scheduler, there are three steps that
must be taken. First, you create the function in CentralPlannerScheduler.
Next, you define the same function with the same signature in
RemoteScheduler. This second function just aggregates its arguments in a
dictionary and passes them to _request. Finally, the function must be
registered in server.py.

These last two steps are extremely automatable, so I moved them into a
decorator that gets applied to all rpc-accessible functions in
CentralPlannerScheduler.

## Motivation and Context
We've wanted to simplify the way the rpc works in the past. The addition of registration in server makes this even more important. I'm tired of bugs during development caused by inconsistencies between the three locations a function must be mentioned.

The current rpc tests should be enough. I think they miss the graph-related functions and maybe a few others, but there's nothing different about those and an error would probably get caught by some other integration test I'm hoping.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -22,6 +22,8 @@ See :doc:`/central_scheduler` for more info.
 """
 
 import collections
+import inspect
+
 try:
     import cPickle as pickle
 except ImportError:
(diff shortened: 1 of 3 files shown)


luigi/scheduler.py
```diff
@@ -77,6 +79,38 @@ class Scheduler(object):
...
 
+RPC_METHODS = {}
+
+
+def rpc_method(fn=None, **request_args):
+
+    # If request args are passed, return this function again for use as
+    # the decorator function with the request args attached.
+    if fn is None:
+        return functools.partial(rpc_method, **request_args)
```
Tarrasch: Can we avoid this magic if we just force everyone to do `@rpc_method()` instead of the slightly more convenient `rpc_method`?

Tarrasch:
Seem to be issues with Python 3. Other than that this looks fantastic. :)

luigi/scheduler.py
daveFNbuck: Sure, I'll make that update.

Tarrasch:
LGTM. Unless anyone objects I'll merge this. :)

erikbern:
LGTM

PR #1743, opened 2016-07-01
Fix bug with GCSFlagTarget __init__

marcpare:
A single line fix to GCSFlagTarget __init__ that passes along kwargs properly.

## Description

Change __init__ to pass along kwargs (`format` and `client`)

## Motivation and Context

Was initializing a `GCSFlagTarget` with a GCSClient that was created with a json key file, rather than the default auth via environment variable. There wasn't a way to get the GCSFlagTarget to pick this up since it ignored the `client` kwarg.


Tarrasch:
LGTM. I suppose it was an oversight :)

marcpare:
Great, thanks a lot y'all.

PR #1744, opened 2016-07-01
Caches get_autoconfig_client on a per-thread basis

### LINKED ISSUE #1740: HdfsTarget commands fail when many targets are instantiated
I've recently added an existence check to a large mapreduce task for which some input files may be missing. With a large enough set of inputs, it will fail every time. I've simplified it to the following code:

```python
from luigi.contrib.hdfs import HdfsTarget

many_targets = [HdfsTarget('/') for _ in range(2000)]
all(target.exists() for target in many_targets)
```

This will break if I use any past 1000 or so. Here the client uses snakebite. For a more direct triggering, we can also do

```python
from snakebite.client import AutoConfigClient

clients = [AutoConfigClient() for _ in range(10000)]
all(client.test('/', exists=True) for client in clients)
```

In either case, the bug goes away if I use a generator expression rather than a list comprehension. The problem is that when I'm dealing with objects coming out of luigi calls like input_hadoop, it's too late for me to decide between lists and iterators. I can code around this by instantiating all of my HdfsTargets with the same client, but I'm not sure this is safe. It could also be fixed in luigi if we had get_autoconfig_client return the same object each time. Is there any reason this wouldn't work?

daveFNbuck:
## Motivation and Context
If too many different autoconfig clients are open at the same time, it can
create too many open files for the operating system to handle. To avoid
this issue, we can simply cache the autoconfig client to ensure that the
same one is being used every time. However, the autoconfig client is not
thread safe so we must have a different one for each thread.

This resolves #1740

I added unit tests and ran the minimal HdfsTarget example in #1740. Both pass.

### DIFF
--- a/luigi/contrib/hdfs/clients.py
+++ b/luigi/contrib/hdfs/clients.py
@@ -19,33 +19,42 @@
 The implementations of the hdfs clients. The hadoop cli client and the
 snakebite client.
 """
-
+import logging
+import threading
 
 from luigi.contrib.hdfs import config as hdfs_config
 from luigi.contrib.hdfs import snakebite_client as hdfs_snakebite_client
 from luigi.contrib.hdfs import webhdfs_client as hdfs_webhdfs_client
 from luigi.contrib.hdfs import hadoopcli_clients as hdfs_hadoopcli_clients
 import luigi.contrib.target
-import logging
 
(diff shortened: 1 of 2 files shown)


Tarrasch:
LGTM

Issue #1741, opened 2016-06-30
S3 Feature Addition - will break current usage of list and listdir


dlstadther:
### This feature addition will break current use of `S3Client.list()` and `S3Client.listdir()`.

My question is... **_Since this is a niche-case feature that will break people's existing code who use one of these functions, would it still be valuable to submit a PR for this feature addition?_**

Due to a specific use-case where I need to know the total number and size of files copied between s3 buckets, I have modified `S3Client.listdir()` and `S3Client.list()` such that they return a generator consisting of `(path/fn, s3_item)`. The inclusion of this modification allows one to get [metadata about each file](http://boto.cloudhackers.com/en/latest/ref/s3.html#module-boto.s3.key) listed.

Current use of `list` or `listdir` looks like the following:

```python
from luigi import s3
s3_client = s3.S3Client()
for item in s3_client.list('s3://path/to/dir'):
    print(item)  #prints filename
for item in s3_client.listdir('s3://path/to/dir'):
    print(item)  #prints filename with path
```

If a PR for this feature is submitted and approved, usage would be the following:

```python
from luigi import s3
s3_client = s3.S3Client()
for path, item in s3_client.list('s3://path/to/dir'):
    print(path)  #prints filename
    print(item.last_modified)
    print(item.size)
for path, item in s3_client.listdir('s3://path/to/dir'):
    print(path)  #prints filename with path
    print(item.last_modified)
    print(item.size)
```

dlstadther:
Code can be found [here](https://github.com/dlstadther/luigi/tree/s3_copy_size).

Changed files:  
* luigi/s3.py
* test/s3_test.py

Tarrasch:
Why not extend the function (like pass a kwarg saying you also want `item) or create a new function?

dlstadther:
Good idea. Thanks @Tarrasch

PR #1745, opened 2016-07-01
S3 list with key object

dlstadther:
## Description
Updated `S3Client.copy()` such that when copying a directory of files, the number and total size of files will be returned. To accomplish this, the `list()` and `listdir()` now have an optional parameter `return_key` which when set `True` will return the `boto.s3.key.Key` objects rather than the string filenames.

Module should remain backwards compatible.

## Motivation and Context
I need to perform a few dynamic calculations based on the number and size of copied files. To accomplish this, `S3Client.copy()` needs to return the number and size of files copied.

Tests are included.

### DIFF
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -449,13 +449,14 @@ class S3Client(FileSystem):
         self.copy(source_path, destination_path, **kwargs)
         self.remove(source_path)
 
-    def listdir(self, path, start_time=None, end_time=None):
+    def listdir(self, path, start_time=None, end_time=None, return_key=False):
         """
         Get an iterable with S3 folder contents.
         Iterable contains paths relative to queried path.
 
         :param start_time: Optional argument to copy files with modified dates after start_time
         :param end_time: Optional argument to copy files with modified dates before end_time
+        :param return_key: Optional argument, when set to True will return a boto.s3.key.Key (instead of the filename)
         """
         (bucket, key) = self._path_to_bucket_and_key(path)
 
@@ -472,12 +473,18 @@ class S3Client(FileSystem):
                     (not start_time and end_time and last_modified_date < end_time) or  # end defined, prior to end
                     (start_time and end_time and start_time < last_modified_date < end_time)  # both defined, between
                ):
-                yield self._add_path_delimiter(path) + item.key[key_path_len:]
+                if return_key:
+                    yield item
+                else:
+                    yield self._add_path_delimiter(path) + item.key[key_path_len:]
 
-    def list(self, path, start_time=None, end_time=None):  # backwards compat
+    def list(self, path, start_time=None, end_time=None, return_key=False):  # backwards compat
         key_path_len = len(self._add_path_delimiter(path))
-        for item in self.listdir(path, start_time=start_time, end_time=end_time):
-            yield item[key_path_len:]
+        for item in self.listdir(path, start_time=start_time, end_time=end_time, return_key=return_key):
+            if return_key:
+                yield item
+            else:
+                yield item[key_path_len:]
 
     def isdir(self, path):
         """
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -307,6 +307,8 @@ class S3Client(FileSystem):
 
         When files are larger than `part_size`, multipart uploading will be used.
 
+        When copying a directory, method will return a tuple (number_of_files_copied, total_size_copied_in_bytes).
+
         :param source_path: The `s3://` path of the directory or key to copy from
         :param destination_path: The `s3://` path of the directory or key to copy to
         :param threads: Optional argument to define the number of threads to use when copying (min: 3 threads)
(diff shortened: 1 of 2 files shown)


dlstadther:
@Tarrasch Does this satisfy your suggestion within #1741? If so, I'll merge Monday (EST).

luigi/s3.py
```diff
@@ -449,7 +456,7 @@ def move(self, source_path, destination_path, **kwargs):
         self.copy(source_path, destination_path, **kwargs)
         self.remove(source_path)
 
-    def listdir(self, path, start_time=None, end_time=None):
+    def listdir(self, path, start_time=None, end_time=None, with_key=False):
         """
         Get an iterable with S3 folder contents.
         Iterable contains paths relative to queried path.
```
Tarrasch: Maybe also add docs like `:param with_key: if True this function will instead return blah blah`?

Tarrasch:
LGTM

dlstadther:
@jpiper Since you've made the most recent s3 changes, could you take a look at this PR? Thanks

luigi/s3.py
```diff
@@ -449,13 +458,14 @@ def move(self, source_path, destination_path, **kwargs):
...
 
         :param start_time: Optional argument to copy files with modified dates after start_time
         :param end_time: Optional argument to copy files with modified dates before end_time
+        :param with_key: Optional argument, when set to True will return a tuple (filename, boto_key_object)
```
jpiper: Sorry to nitpick, but can you change this to something along the lines of

> Optional argument, when set to True causes function to yield (filename, boto.s3.key.Key) tuples

So we're a bit more explicit in the documentation - thanks!

luigi/s3.py
```diff
@@ -472,12 +482,20 @@ def listdir(self, path, start_time=None, end_time=None):
...
         key_path_len = len(self._add_path_delimiter(path))
-        for item in self.listdir(path, start_time=start_time, end_time=end_time):
-            yield item[key_path_len:]
+        for path in self.listdir(path, start_time=start_time, end_time=end_time):
+            yield path[key_path_len:]
+
+    def list_with_key(self, path, start_time=None, end_time=None):
+        key_path_len = len(self._add_path_delimiter(path))
+        for path, item in self.listdir(path, start_time=start_time, end_time=end_time, with_key=True):
+            yield path[key_path_len:], item
```
jpiper: My spidey senses tell me this could be condensed to a single function with an extra argument, but I'm not too concerned by it

jpiper:
Apart from two minor nitpicks, this LGTM!

luigi/s3.py
dlstadther: You're right. I'll update this docstring

luigi/s3.py
dlstadther: Yeah... I couldn't think of a more eloquent way to combine them than the following:

```python
def list(self, path, start_time=None, end_time=None, with_key=False):
    key_path_len = len(self._add_path_delimiter(path))
    if with_key:
        for path, item in self.listdir(path, start_time=start_time, end_time=end_time, with_key=with_key):
            yield path[key_path_len:], item
    else:
        for path in self.listdir(path, start_time=start_time, end_time=end_time):
            yield path[key_path_len:]
```

The issue is that `listdir` return a generator of values when `with_key=False` and a generator of tuples when `with_key=True`. So I'm not sure there's a best way to handle these two generator formats with a single loop.

Perhaps, there's a better way to handle `listdir` when `with_key=True`. Possibly only return `boto.s3.key.Key` (such that listdir only yields a single value)? Thoughts?

dlstadther:
@jpiper Thanks for reviewing so quickly!

dlstadther:
@jpiper Could i get you to take one more look at this? I've changed it such that `list()` and `listdir()` will only ever return a single value from their generators - either a `str` or a `boto.s3.key.Key` object. I then updated `copy()` accordingly.

dlstadther:
@jpiper @f355 @mikekap You all have had your hands in the `copy`, `list`, and `listdir` methods for which i have modified. Thoughts?

luigi/s3.py
```diff
@@ -338,16 +340,23 @@ def copy(self, source_path, destination_path, threads=100, start_time=None, end_
...
+            total_size_bytes = 0
             src_prefix = self._add_path_delimiter(src_key)
             dst_prefix = self._add_path_delimiter(dst_key)
-            for key in self.list(source_path, start_time=start_time, end_time=end_time):
+            for item in self.list(source_path, start_time=start_time, end_time=end_time, return_key=True):
+                (bucket, key) = self._path_to_bucket_and_key(source_path)
+                key_path = self._add_path_delimiter(key)
+                key_path_len = len(key_path)
+                path = item.key[key_path_len:]
+                print(path)
```
f355: Debugging leftover?

luigi/s3.py
```diff
@@ -338,16 +340,23 @@ def copy(self, source_path, destination_path, threads=100, start_time=None, end_
...
             management_pool = ThreadPool(processes=threads)
 
+            total_size_bytes = 0
             src_prefix = self._add_path_delimiter(src_key)
             dst_prefix = self._add_path_delimiter(dst_key)
-            for key in self.list(source_path, start_time=start_time, end_time=end_time):
+            for item in self.list(source_path, start_time=start_time, end_time=end_time, return_key=True):
+                (bucket, key) = self._path_to_bucket_and_key(source_path)
+                key_path = self._add_path_delimiter(key)
+                key_path_len = len(key_path)
```
f355: Three lines above can be moved outside of the `for` loop I think.

f355:
Apart from that looks good!

luigi/s3.py
dlstadther: Ah, yes they can

luigi/s3.py
dlstadther: Oops, oversight...

dlstadther:
@f355 thanks for taking a look! Will fix those two small things

jpiper:
LGTM :+1:

dlstadther:
I will merge this Monday (07/18/2016) morning (US Eastern time) unless I hear otherwise.

PR #1749, opened 2016-07-06
Update call to iteritems in luigi/tools/deps: deprecated in Python 3

sguermond:
## Description

dict.iteritems() is deprecated in Python 3 (PEP 3106), was replaced with dict.items()

## Motivation and Context

Running luigi-deps to get a list of all upstream outputs returned the following error: `AttributeError: 'dict' object has no attribute 'iteritems'`

The change worked for me (Python 3.5.0)

### DIFF
--- a/luigi/tools/deps.py
+++ b/luigi/tools/deps.py
@@ -117,7 +117,7 @@ def main():
         task_output = task.output()
 
         if isinstance(task_output, dict):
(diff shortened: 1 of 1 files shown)


Tarrasch:
Thanks! For whatever reasons that tool is only tested with Python 2.

PR #1751, opened 2016-07-06
Fix exist method for ftp server

dlstadther:
## Description
Ftp exists method can now handle FTP servers which return fully qualified name or strictly the filename in question.

## Motivation and Context
Resolves PR #1710 (orphaned PR)

Works for me.

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -131,14 +131,12 @@ class RemoteFileSystem(luigi.target.FileSystem):
         return exists
 
     def _ftp_exists(self, path, mtime):
-        path_parts = path.split('/')
-        path = '/'.join(path_parts[:-1])
-        fn = path_parts[-1]
+        dirname, fn = os.path.split(path)
 
-        files = self.conn.nlst(path)
+        files = self.conn.nlst(dirname)
 
         exists = False
-        if fn in files:
+        if path in files or fn in files:
             if mtime:
                 mdtm = self.conn.sendcmd('MDTM ' + path)
                 modified = datetime.datetime.strptime(mdtm[4:], "%Y%m%d%H%M%S")


Tarrasch:
Is it possible to have a test case? Other than that LGTM

dlstadther:
Does anyone have experience unittesting ftp? I'm really unsure where to start. Is it possible that an [external project](https://github.com/tarttelin/Python-Stub-Server) may be needed to mock a FTP server?

Tarrasch:
If it's so much hassle, perhaps just ask somebody to review. :)

dlstadther:
@andreabedini @daveFNbuck Could y'all take a look at this and verify that it'll work for all cases? I can confirm that it works for my use (where the FTP does not return fully qualified file names).

daveFNbuck:
LGTM, matches how I've modified my clone since the discussion on #1710

daveFNbuck:
If you want to unit test, you don't need to mock a whole ftp server. You just need to mock the functions you use in the connection object. You'd just need one test with nlst mocked to return the file name and one with it mocked to return the full path.

luigi/contrib/ftp.py
```diff
@@ -131,14 +131,12 @@ def _sftp_exists(self, path, mtime):
...
-        path = '/'.join(path_parts[:-1])
-        fn = path_parts[-1]
+        dirname, fn = os.path.split(path)
 
-        files = self.conn.nlst(path)
+        files = self.conn.nlst(dirname)
 
         exists = False
-        if fn in files:
+        if path in files or fn in files:
```
andreabedini: this should work but requires iterating over the file list twice, which is expensive if the folder contains many files. No reason to block on this, we can change it when we have a better idea.

interskh:
👍

luigi/contrib/ftp.py
andreabedini: how about we issue `MDTM path` and `MDTM fn` (we already use MDTM below) and if both return status code 550 we decide the file doesn't exist? we might need to wrap the command in a `try` block because Python ftp module will raise if it receives a 5xx result code (see [here](https://hg.python.org/cpython/file/3.5/Lib/ftplib.py#l234)). Sorry I had no time to work on the PR.

daveFNbuck: I wonder how universally MDTM is implemented on ftp servers. Our current usage is controlled by a conditional, so users can decide whether to use it.

dlstadther: I prefer the idea of just checking the first index of `files`, seeing if it contains `'/'` and then checking for file existence accordingly. Thus, we'd have theta(n)+1 for this part.

dlstadther:
@daveFNbuck @andreabedini Do y'all approve of this new version?

daveFNbuck:
I think it's more complicated than necessary and solves a non-problem, so I'd vote for sticking with the simpler version.

dlstadther:
@daveFNbuck Where the simpler version potentially loops through the directory twice?

daveFNbuck:
I don't think that looping through the directory twice is a problem. The directory listing is only retrieved once, and that probably takes at least an order of magnitude longer than looping through it. I did a quick test via
```timeit("path in files or fn in files", "files = ['/aaaa/bbbbb/ccccc/ddddd/%i' % i for i in range(50000)];path='/aaaa/bbbbb/ccccc/ddddd/100000';fn='100000'", number=1000)```
and the thousand runs took about 1.2 seconds on my laptop. So we're talking about turning a 1.2-millisecond operation into a 0.6-millisecond operation if we run into some crazy directory with 50k items. This is also a worst-case where the file doesn't exist and the listing is all full paths.

Meanwhile, you've now hard-coded the assumption that if the first item in the directory listing contains a `/`, the listings will all be full paths rather than just filenames. I don't know how safe this assumption is, and the correctness is less obvious than the simpler code. Maybe some ftp servers show subdirectory names with a trailing slash, or recursively list subdirectory contents with relative paths.

Premature optimization being the root of all evil, I'd prefer not to do something a bit weird to save that half-millisecond in the exaggerated case.

**Comment by dlstadther** [comment]
Wow, thanks for the very thorough explanation @daveFNbuck ! I'll go back to the previous version.

Thanks so much for your input!

**Comment by daveFNbuck** [comment]
No problem, I've been having a lot of trouble with ftp upload pipelines lately so it's important to me that this code is solid :)

**Comment by dlstadther** [comment]
I completely understand. My ftp upload job is held together with rubber bands

File: luigi/contrib/ftp.py (hunk 1 above)
Comment (a1k0n) [inline luigi/contrib/ftp.py]: Or you could just change files into a set and then you have a more efficient test regardless of the circumstances
```
    files = set(self.conn.nlst(dirname))
```

daveFNbuck: I tested that too. It takes 3x the time of just comparing twice against
each element. So it wouldn't make sense as an optimization unless we were
doing 7 or more comparisons.

dlstadther:
So, are we good to merge? I'm not going to have the time to mock FTP functions for a full test suite for awhile.

daveFNbuck:
LGTM

PR #1693, opened 2016-05-20
tox: Remove all code coverage tools

Tarrasch:
## Description
Remove all code coverage tools.

## Motivation and Context
I removed "coverage" and "codecov" (I don't even know their difference).
At the moment I want to minimize the cruft. Also the output is smaller
and less verbose when our first-time-users test their code with tox.

I wasn't sure what "sitecustomize.py" is, but the less magic the better
for now. Given that nobody currently have strong ownership over luigi,
it's better to strip the project to a manageable size.

PR #1755, opened 2016-07-08
Remove sitecustomize file

Tarrasch:
## Description
Less files, less confusion. This came to be from discussion in spotify/luigi#1693

Not tested. I entirely rely on Travis for this. In particular there could be flags misspelled and other unknown incompatibilities as of the new version of coverage.

Luigi developers might need to run `rm .coverage` after applying this change as a new version of coverage will be in use after this patch.


Tarrasch:
@gpoulin, is this kinda what you said in https://github.com/spotify/luigi/pull/1693#issuecomment-220613066?

Tarrasch:
Die you incomprehensible root folder file! >_<

PR #1756, opened 2016-07-11
Remove the confusing "dummy_test_module" directory

Tarrasch:
## Description
Remove the confusing "dummy_test_module" directory

## Motivation and Context
The purpose is simply to remove a confusing module from the luigi root
directory. It was only used from tests but it couldn't be put in the
test module for whatever technical reasons.

This is only affecting tests. If tests pass it should be OK.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,7 @@
 [report]
 omit =
     luigi/mrrunner.py
+    test/_test_time_generated_module*.py
     */python?.?/*
     */site-packages/nose/*
     *__init__*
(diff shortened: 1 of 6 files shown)


Tarrasch:
@daveFNbuck, does this look sane to you?

daveFNbuck:
I'm not sure it makes sense to import the unloaded module for the dynamic
loading test. I think importing it invalidates the test.

Tarrasch:
> I'm not sure it makes sense to import the unloaded module for the dynamic loading test. I think importing it invalidates the test.

Not sure we're on the same page.

The helper creates a module (a file within our PYTHONPATH) *without* loading it. Later, we run `luigi --module that_yet_unloaded_module ...`, and see if luigi is actually loading it or not.

daveFNbuck:
Sounds good.

PR #1760, opened 2016-07-13
docs: Set minimum versions for sphinx

Tarrasch:
I saw that the readthedocs build was failing, hence this is my attempt to fix it,
altough I don't know how to test this before merging it.
See https://readthedocs.org/projects/luigi/builds/4189801/

Unfortunately I couldn't find a way to connect a webhook between GitHub
PRs and readthedocs: https://github.com/rtfd/readthedocs.org/issues/1340

The reason I believe a version bump would fix it is because I see
exactly the same error as in sphinx-doc/sphinx#2465 and readthedocs
seem to use 1.3.5, causing the pdf docs to be outdated.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -82,7 +82,7 @@ autoclass_content = 'both'
 # -- General configuration ------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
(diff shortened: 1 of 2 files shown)


Tarrasch:
This patch wasn't enough, but it got fixed in #1761!

PR #1766, opened 2016-07-14
Added pass_params method to luigi.Task

Talmaj:
## Description

This method takes another luigi subtask and returns the parameters that can/should be passed on (have the same naming).
## Motivation and Context

It's an enhancement, often used when needing to pass on the parameters from the top task to the bottom. 

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -350,6 +350,19 @@ class Task(object):
 
         return cls(**new_k)
 
+    def pass_params(self, task_class, include_significant=True):
+        """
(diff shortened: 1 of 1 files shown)


Tarrasch:
How does this relate to the help functions in the util module? http://luigi.readthedocs.io/en/stable/api/luigi.util.html


For this PR and also the related #1767, I would like to see how you use this function in the real world and what problem it solves. You've already  motivated it, but an example would make it easier for me to understand. :)

PR #1767, opened 2016-07-14
Added assert_passed_params classmethod for luigi.Task 

Talmaj:
## Description

Added a simple classmethod that raises assert error when not all expected kwargs passed on to a luigi.Task.
## Motivation and Context

I find it useful when we want to be sure that the upper tasks is passing on all expected/needed parameters to the required subtasks.


erikbern:
this should just be a static method in luigi.util or something like that  – no reason to make it a method

Issue #1768, opened 2016-07-15
Dataproc module assumes Google application credentials exist


Jyrsa:
The dataproc module assumes that if it is being imported, then the Google application credentials must be available.  The related tests raise skiptest if this is the case but also attempting to import the module will result in an exception. 

https://github.com/spotify/luigi/blob/master/luigi/contrib/dataproc.py#L18

PR #1769, opened 2016-07-15
dataproc import just warns and doesn't fails if google credentials do…

Jyrsa:
This pull-request fixes the Issue #1768 . 

I haven't set up any Google cloud default credentials so I got odd errors in import tests and other places that don't really test the dataproc functionality but just import the module for one reason or another.


Jyrsa:
This ended up being 3 commits as the related tests and gcs tests had a similar issue. 

I fixed them so I could get #1731 tests to pass so they're mingled with other commits in that branch so I'm not squashing this. Sorry about that.

PR #1770, opened 2016-07-17
Compression atomic local file

walterhan:
## Description

This Pull Request is meant to add a compressed boolean option to AtomicLocalFile to compress data when writing to the file. This particular pull request isn't necessarily the final code, but meant to start a conversation on whether or not this is the right direction to go. One alternative is to explicitly create an AtomicCompressedLocalFile (and associated inherited classes)
## Motivation and Context

In my particular use case, I'd like to compress CSV data (using the csv module) when writing to S3 using the AtomicS3LocalFile. The method I used beforehand was to use zlib to compress data before writing the content to a file - however, it ends up being a little bulky/cumbersome if you want to use the python csv.writer. In particular, with the csv module, I would write to a stringio w/ csv.writer to format the data before then compressing the contents of the stringio when writing to the final atomic file. The usage of AtomicLocalFile with compression=True would be then be similar to using GzipFile.

Ran this code with my jobs - specifically, I have only run this with AtomicS3File

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -324,7 +324,7 @@ class AtomicFtpFile(luigi.target.AtomicLocalFile):
     Also cleans up the temp file if close is not invoked.
     """
 
-    def __init__(self, fs, path):
+    def __init__(self, fs, path, compressed=False):
         """
         Initializes an AtomicFtpfile instance.
         :param fs:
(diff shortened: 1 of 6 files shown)


erikbern:
This doesn't seem super clean. Boolean arguments are generally an antipattern. I also don't like how it's hardcoded to gzip using a boolean flag.

You should extend it to support the same `format` as the `File` class uses – should be doable right?

Issue #1660, opened 2016-04-20
When worker-task-limit is exceeded, return code is 0 even though tasks are left pending


ulzha:
Should probably return scheduling_error code in the end.

(In https://github.com/spotify/luigi/pull/1612 we addressed a few other spurious return code 0 cases, but descoped this one.)

ulzha:
I found another case where return code is 0 despite tasks left pending - that also happens when a worker is forgotten by scheduler. Then we get "left pending because of unknown reason" tasks.

If we want return code 0 to reliably correspond to completeness of the "root task", then perhaps the best idea is to actually explicitly check the completeness as the last thing before exiting with 0.

PR #1771, opened 2016-07-18
Update retcodes to handle new cases

fabriziodemaria:
## Description
This pull request is a continuation of #1612. The purpose of this pull request is to cover more cases for which tasks does not run successfully and Luigi exit code is 0.
More specifically, the following two cases are covered:
- Tasks are not run because task-limit is reached counts as _scheduling_error_ 
- Return code can be set for cases in which tasks fail or are left pending for _unknown_reason_ 

## Motivation and Context
This relates to #1660.

I have included unit tests.


Tarrasch:
Cool! I did a quick glance. One question quickly rose for me. Should we really convey that "unknown-->bad"? I think of unknown like literally unknown. Sometimes it's out of resources (not bad) and sometimes that task is disabled or still have a timeout for it's FAILED state (maybe bad, but the worker that previously failed has already sent error emails and stuff).

What I'm saying is that this is a really good patch, because it allows for luigi users to react exactly the way the want based on return codes. But I propose that (1) we don't use `:(` as I don't consider it bad. Maybe we make things confusing, but perhaps we could introduce `:D` for indicating that the root task is done, and unknown stays as `:)`.

doc/configuration.rst
```diff
@@ -481,18 +481,21 @@ We recommend that you copy this set of exit codes to your ``luigi.cfg`` file:
   missing_data=20
   task_failed=30
   scheduling_error=35
+  unknown_reason=38
```
Tarrasch: ... and (2) I suggest we say `38 or 0`. And then explain more detailed when you might want what number.

doc/configuration.rst
```diff
@@ -481,18 +481,21 @@ We recommend that you copy this set of exit codes to your ``luigi.cfg`` file:
...
   unhandled_exception=40
 
 unhandled_exception
   For internal Luigi errors.  Defaults to 4, since this type of error
   probably will not recover over time.
+unknown_reason
+  For when a task fails or is left pending for unknown reasons.
```
Tarrasch: Don't we already have a error code for a task have failed?

fabriziodemaria: Is it correct to say that a task in the _unknown_reason_ set is always in state "left pending"?

Tarrasch: Actually, I think it could be any of these:

 * PENDING - Out of resources
 * FAILED - There's a timeout for retries
 * DISABLED - Similar to FAILED

fabriziodemaria: > PENDING - Out of resources

Another reason for this could be workers forgotten by the scheduler? (from #1660)

fabriziodemaria: Perhaps a more correct (but generic) description could be: _For when a task does not run successfully for unknown reasons_

Tarrasch: > Another reason for this could be workers forgotten by the scheduler? (from #1660)

Hmm, I'm not sure what @ulzha meant by "forgotten", I think the scheduler will not prune the task if it's something that was uploaded by the that's still alive worker.

> Perhaps a more correct (but generic) description could be: For when a task does not run successfully for unknown reasons

I like it! Let's not be specific or detailed, as that risks to quickly get out of date or be slightly incorrect to begin with.

fabriziodemaria:
> But I propose that (1) we don't use :( as I don't consider it bad. 

I changed `:(` with `:|` for the _unknown_reason_ case, would that be good enough in your opinion?

Xeago:
What about `:¿`? I make that to indicate my confusion.

ulzha:
Name it "pending_for_unknown_reason" maybe? That's what it is. And that hopefully also shows how it is worse than "pending". After all we're using ":|" for `still_pending_ext` ("pending_for_known_reason"). A situation where uncertainty is added is not better.

doc/configuration.rst
```diff
@@ -481,18 +481,24 @@ We recommend that you copy this set of exit codes to your ``luigi.cfg`` file:
...
   unhandled_exception=40
 
 unhandled_exception
   For internal Luigi errors.  Defaults to 4, since this type of error
   probably will not recover over time.
+unknown_reason
+  For when a task does not run successfully because of an unknown reason. Despite
+  this case can be expected in an error free execution, it does not guarantee
+  completeness of the root task. Return code 38 is advised only for cases in which
+  Luigi return code 0 is used by the user to guarantee root tasks' completeness.
```
ulzha: Missed the [comment suggesting "38 or 0"](https://github.com/spotify/luigi/pull/1771#discussion_r71119648). The comment doesn't describe the benefit. I would -1 this convoluted wording as overcomplication.

0 is already the default value. The existing users who perform retcode configuration won't be taken by surprise.

luigi/execution_summary.py
```diff
@@ -270,15 +277,15 @@ def _get_comments(group_tasks):
...
     ("still_pending", 'were left pending, among these'),
     ("still_pending_ext", 'were missing external dependencies'),
     ("run_by_other_worker", 'were being run by another worker'),
     ("upstream_failure", 'had failed dependencies'),
     ("upstream_missing_dependency", 'had missing external dependencies'),
     ("upstream_run_by_other_worker", 'had dependencies that were being run by other worker'),
-    ("upstream_scheduling_error", 'had dependencies whose complete() or requires() failed'),
-    ("unknown_reason", 'were left pending because of unknown reason'),
+    ("upstream_scheduling_error", 'had dependencies whose scheduling failed'),
+    ("unknown_reason", 'did not run successfully because of unknown reason'),
```
ulzha: When the task fails `run()` we do get the reason back, as the exception, right? Was there a reason to alter wording "were left pending"?

doc/configuration.rst
ulzha: By "forgotten" I meant "pruned", yeah. If a worker loses connection intermittently for a certain time (and there are no redundant workers triggering the same workflow), then it ends up with parts of its task graph pruned and never given to it in `get_work()`, translated to "pending because of unknown reason" in summary.

luigi/execution_summary.py
ulzha: Mhm, it stems from [this part of discussion](https://github.com/spotify/luigi/pull/1771#discussion_r71123666). Thinking about the cases of disabling and resources missing, it's still true that the instance remained "pending". (In the common sense common denominator of those more detailed statuses. If we invested more in status propagation we would come up with summaries like "pending because missing resources" etc.)

I'm saying that in my view the higher level conceptual task's flow hasn't changed much and doesn't risk changing much. It's first pending, then becomes running (the moment when `run()` is invoked). The reason for failing running is always "known" via the exception; the reason for not running at all can vary, but "left pending" remains valid wording.

ulzha: Alternatively, a valid wording would be "did not run", because that's equivalent to the common sense "left pending" (didn't enter `run()`). "Did not run successfully" is what I object against, because that can also be interpreted to mean that it did enter `run()`.

Tarrasch:
Sure, let's call it pending for unknown reason. Actually, I oversaw perhaps the most common "reason", that is when you let two separate workers (typically on 2 separate computers) run the same task, but one isn't allowed to run it because the other one has already finished running it.

Actually, even better name would probably be "not_run_for_unknown_reason". If we call it "pending_for_unknown_reason", users might think that it's state according to the central scheduler also is pending, that's usually not the case.

doc/configuration.rst
Tarrasch: You're right, lets not complicate things. The benefit I was looking for was to say that this return code is "no problem, don't worry".

What about using a number but a much lower number? I noticed we already say they are in increasing order of severity (I had forgotten this). Maybe use the number `15`?

doc/configuration.rst
Tarrasch: Oh I see. Makes sense. There are truly a lot of cases.

luigi/execution_summary.py
Tarrasch: I think you're spot on Uldis. Let's say "did not run" or "wasnt permitted to run" (to remind users that who decides this is actually the central scheduler).

fabriziodemaria:
> Actually, I oversaw perhaps the most common "reason", that is when you let two separate workers (typically on 2 separate computers) run the same task, but one isn't allowed to run it because the other one has already finished running it.

Shouldn't this case be covered in the `run_by_other_worker` set?

doc/configuration.rst
ulzha: Indeed, I was going to suggest 25. (As said, for a human operator, "not run for unknown reason" IMO evaluates worse than "not run for \<insert known reason, like missing data>".)

Tarrasch: Ok. I imagined `10 < x < 30`. So 25 sounds good because of the reason you said.

Tarrasch:
> Shouldn't this case be covered in the run_by_other_worker set?

Sometimes but not always. The summary-printing worker will only know that somebody else ran it if it was RUNNING and that worker asked for work at that time. It could be the case that the other worker was fast and the tasks where DONE already when the summary-printing worker asks for work.

fabriziodemaria:
I have one more proposal for this.
Tasks in this set did not enter the `run()` phase and there is a certain number of reasons causing this; such reasons have been listed in the various comments of this pull request. In such a context, I think `unknown_reason` might be misleading since we are not dealing with an unexpected/unknown error. Nevertheless, the set is too broad to better define the specific motivation for which the task didn't run.
What if we simply call the set `not_run`, and we list the possible causes in the docs. I believe this name fits the other sets' names in terms of descriptive granularity (i.e. `completed`, `already_run`, `not_run`, `failed`,... ). I would agree on using `25`  as error code.
Regarding the info logs and execution summary, `did not run` might be not enough. We might just mention the most probable causes for it.

Tarrasch:
@fabriziodemaria Your proposal sounds good to me in it's entirety. I like it.

Did you consider to have the info logs say `wasn't permitted to run` as I suggested before? I think it conveys that the worker did actually ask for work but was denied.

ulzha:
Denial can also occur passively/accidentally, as in the case when connection failure causes task purge, or in the case of an eventual scheduler bug. The "wasn't permitted" presents some controlledness flair when it may not be the case.

But I agree that the pattern how worker gets work out is something users often are confused about. The description of the probable cases can include the phrasing "wasn't permitted to run [by scheduler [because X or Y or Z]]".

doc/configuration.rst
```diff
@@ -479,28 +479,37 @@ We recommend that you copy this set of exit codes to your ``luigi.cfg`` file:
...
   the [worker] keep_alive option.
-scheduling_error
-  For when a task's ``complete()`` or ``requires()`` method fails with an
-  exception.
+not_run
+  For when a task is not granted run permission by the scheduler. Typically
+  because of lack of resources, because the task has been already run by
+  another worker or because the attempted task is in DISABLED state.
+  Connectivity issues with the central scheduler might also cause this.
+  This does not include the cases for which a run is not allowed due to missing
```
ulzha: Very nice.

Come to think about it, I find it illogical that already_running has a lower severity in this scheme than the not_run case where a task has already been run and succeeded. But I don't consider that a blocker now.

ulzha:
\o/ Looks good.

Tarrasch:
I squashed this as there were to many implementation-detail commits. Hopefully the reworded commit is clearer and more direct to the point of what changed. Good job! :)

PR #1772, opened 2016-07-18
Assistants: Don't affect longevity of tasks

Tarrasch:
## Motivation and Context

When assistants where introduced, we imagined that one worker would
run tasks (the assistant) and one would upload them.  In order to ensure
that the tasks don't get pruned before tasks get completed, we added
a what we now consider an "antifeature" that made tasks not getting
pruned while assistants where alive.

Real world use have showed that it wasn't a good feature. It have
caused bugs like spotify/luigi@736c0f1, but worst, it's a feature that
makes luigid change behavior for all tasks if you just submit 1
assistant. I didn't expect that the first time I tested assistants.

There's an email thread where it was discussed to remove this feature.
In there one can also read about how to get something similar to the old
behavior.

https://groups.google.com/forum/#!topic/luigi-user/b7Acym0n-g4

In addition to the modified tests, I'll try to test this in production for a few days. I'll report back.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -444,11 +444,12 @@ class SimpleTaskState(object):
 
     def update_status(self, task, config):
         # Mark tasks with no remaining active stakeholders for deletion
-        if not task.stakeholders:
-            if task.remove is None:
-                logger.debug("Task %r has stakeholders %r but none remain connected -> might remove "
-                             "task in %s seconds", task.id, task.stakeholders, config.remove_delay)
-                task.remove = time.time() + config.remove_delay
+        if (not task.stakeholders) and (task.remove is None) and (task.status != RUNNING):
+            # We don't check for the RUNNING case, because that is already handled
+            # by the fail_dead_worker_task function.
+            logger.debug("Task %r has no stakeholders anymore -> might remove "
+                         "task in %s seconds", task.id, config.remove_delay)
+            task.remove = time.time() + config.remove_delay
 
         # Re-enable task after the disable time expires
         if task.status == DISABLED and task.scheduler_disable_time is not None:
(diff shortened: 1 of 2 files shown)


test/central_planner_test.py
```diff
@@ -343,19 +343,27 @@ def test_re_enable_failed_task_assistant(self):
...
-    def test_fail_job_from_dead_worker_with_live_assistant(self):
+    def test_assistant_doesnt_keep_alive_task(self):
         self.setTime(0)
         self.sch.add_task(worker='X', task_id='A')
         self.assertEqual('A', self.sch.get_work(worker='X')['task_id'])
-        self.sch.add_worker('Y', [('assistant', True)])
+        self.sch.add_worker('Y', {'assistant': True})
 
-        self.setTime(600)
+        self.setTime(self.sch._config.remove_delay + 1)
```
daveFNbuck: You shouldn't access private variables like _config. This breaks on the RPC tests.

Tarrasch:
I found 1 bug already which I've amended a fix for now. It was that RUNNING tasks got pruned while the assistant was running them.

Looking through the code for the `scheduler.py`, I feel it can be improved in so many ways. Maybe I'll take a stab at it unless somebody else wants to. :)

luigi/scheduler.py
```diff
@@ -444,11 +444,12 @@ def fail_dead_worker_task(self, task, config, assistants):
...
         # Mark tasks with no remaining active stakeholders for deletion
-        if not task.stakeholders:
-            if task.remove is None:
-                logger.debug("Task %r has stakeholders %r but none remain connected -> might remove "
-                             "task in %s seconds", task.id, task.stakeholders, config.remove_delay)
-                task.remove = time.time() + config.remove_delay
+        if (not task.stakeholders) and (task.remove is None) and (task.status != RUNNING):
+            # We don't check for the RUNNING case, because that is already handled
+            # by the fail_dead_worker_task function.
+            logger.debug("Task %r has no stakeholders anymore -> might remove "
```
Tarrasch: Sorry. The changed log message is not related to this patch. Just see it as a bonus update of legacy code.

Tarrasch:
Merging. I want to keep refactor other parts of the code without conflicts. :)

PR #1777, opened 2016-07-20
tests: Skip a inttermittently failing s3 test

Tarrasch:
I've seen it fail many times now.


dlstadther:
Do we know why this test fails intermittently on Travis? I believe the copy dir test is important because it covers the bulk of the code within the `copy` method.

PR #1778, opened 2016-07-21
Remove abstract Scheduler class

Tarrasch:
## Description

Remove abstract Scheduler class.

This is an internal refactoring. The purpose is just to remove code and
complexity.

No, I rely on existing unittests run by Travis.


Tarrasch:
Self-approving. Another motivation is that I want the luigi code source to feel hackable. Unnecessary abstractions greatly hinders that.

PR #1779, opened 2016-07-21
Adding DateSecondParameter to parameter.py (close #1775)

### LINKED ISSUE #1775: Adding DateSecondParameter to parameter.py
Hi,

Would you be open to such addition ?

I can send a PR replicating the existing logic for hours & minutes:
https://github.com/spotify/luigi/blob/master/luigi/parameter.py#L477

Lucas-C:
## Description
Addition of `parameter.DateSecondParameter` and modification of all related unit tests.
I copied what was already done for `DateHourParameter` & `DateMinuteParameter`.

## Motivation and Context
I have a need for a `DatetimeParameter` specified to the second.

TravisCI checks.

### DIFF
--- a/test/date_parameter_test.py
+++ b/test/date_parameter_test.py
@@ -34,6 +34,10 @@ class DateMinuteTask(luigi.Task):
     dm = luigi.DateMinuteParameter()
 
 
+class DateSecondTask(luigi.Task):
+    ds = luigi.DateSecondParameter()
+
+
 class MonthTask(luigi.Task):
     month = luigi.MonthParameter()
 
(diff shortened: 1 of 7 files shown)


Tarrasch:
This makes sense and is good to merge, but if you don't mind, could you share why you need this? Do you plan to use "every second" of it?

Lucas-C:
What I need is a single timestamp shared by all my tasks when I trigger a `run`, so that I can date the starting point of their chained execution.
Granularity to the minute is not enough, as several pipeline "builds" could happen over a single minute. But granularity to the second would suit me perfectly.

Lucas-C:
I left a mistake in the tests:
https://github.com/Lucas-C/luigi/blob/master/test/parameter_test.py#L546
I'm gonna fix it tomorrow.

Lucas-C:
Only the `flake` check fail now, and I don't understand why.
Is this ok for you to merge ?

test/date_parameter_test.py
```diff
@@ -108,6 +112,24 @@ def test_serialize_task(self):
         self.assertEqual(str(t), 'DateMinuteTask(dm=2013-02-01T1842)')
 
 
+class DateSecondParameterTest(unittest.TestCase):
+    def test_parse(self):
+        dh = luigi.DateSecondParameter().parse('2013-02-01T184227')
```
Tarrasch: should be called `ds` right?

Tarrasch:
I'll ask you to rebase your change once we've fixed the actual problem, that's supposedly when #1786 is merged.

test/date_parameter_test.py
Lucas-C: Yes, well spotted.
I fixed that

PR #1781, opened 2016-07-22
Rename CentralPlannerScheduler to Scheduler

Tarrasch:
## Description

Rename CentralPlannerScheduler to Scheduler

## Motivation and Context

I find very little value in having that complicated name. In particular
it should be clear that the main class of scheduler.py is in fact
the Scheduler() class.

I've also renamed the tests testing the api to
test/scheduler_api_test.py. This should feel more intuitive to have
"scheduler" in the name of the file

I rely on Travis for this. I could try this in production before merging to, just in case. :)

### DIFF
--- a/luigi/server.py
+++ b/luigi/server.py
@@ -16,7 +16,7 @@
 #
 """
 Simple REST server that takes commands in a JSON payload
-Interface to the :py:class:`~luigi.scheduler.CentralPlannerScheduler` class.
+Interface to the :py:class:`~luigi.scheduler.Scheduler` class.
 See :doc:`/central_scheduler` for more info.
 """
 #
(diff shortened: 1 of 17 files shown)


luigi/server.py
```diff
@@ -247,7 +247,7 @@ def run(api_port=8082, address=None, unix_socket=None, scheduler=None, responder
     Runs one instance of the API server.
     """
     if scheduler is None:
-        scheduler = CentralPlannerScheduler()
+        scheduler = Scheduler()
```
Tarrasch: This kind of code is more readable IMO, requires no brain cycles to process anymore.

erikbern:
lgtm

Issue #1073, opened 2015-07-14
Retry settings/logic at task level instead of at "global" scheduler level


soxofaan:
Hi,

At the moment we have several luigi runs, for which we'd like to have different retry behaviour (some tasks should be retried every hour, some every ten minutes, for some tasks it's fine to be disabled after first failure, etc)

As far as I understand with current luigi we would have to use different configs for each run. And even then it is not possible to have different retry settings within the same run.

Wouldn't it make more sense to be able do define retry settings (retry delay, disable-num-failures, window, ...) or even retry related logic (e.g. custom back-off delay function, send warning mail, switching to a fallback task implementation, ...) at the task level instead of having this as global/scheduler properties?

I don't have a concrete proposal at the moment, but I just wanted to check with the maintainers if there are already ideas around this

Tarrasch:
I haven't had the intent to implement this. The reason is that we don't have one big execution of luigi that depends on all other tasks. We've broken it up into many smaller executions with `ExternalTask`s and whatnot. In that case we are fine with per-execution settings. :)

soxofaan:
We also don't have the intent to have one big luigi execution to cover it all.
The issue is that splitting your execution run based on desired retry behaviour is a bit weird and cumbersome.

mfcabrera:
Closing this as it was fixed by #1791

PR #1782, opened 2016-07-22
Task Level Configuration and `upstream-status-when-all` scheduler config

javrasya:
## Description
Defining some scheduling configuration at task level as it is defined in #1073 . Also have new configuration named as `upstream-status-when-all` in `[scheduler]` section.

### Task Level Configuration
It may be required to have different scheduling configuration for different tasks like retry-count(`disable-num-failures`). Assume that, a task is retrieving data from Hive which responses in long time and other one is retrieving data from RDBM which responses in short time. I may want Hive task not to retry or has small retry-count not to block all my system again an again. But, retrying RDBM task many times may not be problem for me. Assume that, I have some network issue for a while and I should be able to give its retry-count more than Hive task. So I would need to define those configuration at task level.

This fixes the issue defined in #1073 

### `upstream-status-when-all` config

Normally luigi scheduler sets a wrapper task status according to its upstream task severity statuses. For example; if **any** of its upstream task **severity** is `DISABLED` or `FAILED`, scheduler sets the task status as FAILED. If this configuration is set as `True`; scheduler won't make the task as `UPSTREAM_FAILED` or `UPSTREAM_DISABLED` unless **all** of its tasks excluding `SUCCESS` 
ones are `FAILED` or `DISABLED`. 

This is not a problem when there was `task level configuration` feature. But After this feature, some upstream task may retry even one of other upstream tasks is `DISABLED` or `FAILED`.
Please check the example which is added in this PR.

There is also a unit test for those features. Debugging with PyCharm, running available task with tox (Some of them are not able to run in my local machine. I don't know, maybe tests include hdfs or other outer integration systems) and finally with travis(with my account) on my forked repo. I met some optimization problem. Adding task by scheduler was so slow and I realised it because of a test. Thanks to `test_get_work_speed` in `central_planner_test.py`. 

## What is added extra?

* Documentation
* Unit-test
* An example


examples/task_spesific_configuration.py
```diff
@@ -0,0 +1,107 @@
+# -*- coding: utf-8 -*-
+
+#
+# To make this run, you probably want to edit /etc/luigi/client.cfg and add something like:
```
Tarrasch: Something like what? :)

examples/task_spesific_configuration.py
```diff
@@ -0,0 +1,107 @@
...
+        pending tasks or failed tasks which will be retried. While keep-alive is active, workers
+        are not shutdown while there is/are some pending task(s). This task also has a task level
+        config which is ``upstream-status-when-all``;This configuration is about how workers behave
+        to set ``MyTaskHasConfig`` status according to it's upstream tasks which are ``ErrorTask1``,
+        ``ErrorTask2`` and ``SuccessTask``. If this config is set as ``False``, Any of upstream task
+        ``FAILED`` or ``DISABLED`` will make ``MyTaskHasConfig`` ``FAILED`` or ``DISABLED`` which also means,
+        shut the worker down immediately. While this is ``True``, all upstream tasks should be ``FAILED``
+        or ``DISABLED``, worker to set ``MyTaskHasConfig`` status as ``FAILED`` or ``DISABLED``. ``MyTaskHasConfig`` will
+        remain as PENDING unless all of its upstream task severities (so does not contains SUCCESS tasks)
+        are ``FAILED`` or ``DISABLED``
```
Tarrasch: Can all this be promoted to module-level config?

examples/task_spesific_configuration.py
```diff
@@ -0,0 +1,107 @@
+# -*- coding: utf-8 -*-
+
+#
+# To make this run, you probably want to edit /etc/luigi/client.cfg and add something like:
+# DO NOT FORGET to set keep-alive as True.
```
Tarrasch: May I suggest that you include a command line on how to run this? Here's some inspiration: https://github.com/spotify/luigi/blob/5bd9b436990d48acc8def675382c29f8fb0e729e/examples/execution_summary_example.py#L21

examples/task_spesific_configuration.py
```diff
@@ -0,0 +1,107 @@
...
+        sleep(30)
+
+        with self.output().open('w') as output:
+            output.write('SUCCESS Test Task 4.1\n')
+
+    def output(self):
+        return luigi.LocalTarget(path='/tmp/_docs-%s.ldj' % self.task_id)
+
+
+if __name__ == '__main__':
```
Tarrasch: Can you remove this? We try to promote the cli for running luigi over `luigi.run()` :)

luigi/scheduler.py
```diff
@@ -135,6 +135,10 @@ class scheduler(Config):
 
     prune_on_get_work = parameter.BoolParameter(default=False)
 
+    upstream_status_when_all = parameter.BoolParameter(default=False,
```
Tarrasch: Remove the `config_path=dict(...)`, that is automatic!

luigi/scheduler.py
```diff
@@ -691,6 +717,18 @@ def update_resources(self, **resources):
             self._resources = {}
         self._resources.update(resources)
 
+    def _get_task_level_config(self, config):
```
Tarrasch: I don't like this function too much. Is `config` a `dict()` type? If so call it `config_dict`, as I thought initially it was a `Configuration` variable (from the standard libraries).

luigi/scheduler.py
```diff
@@ -691,6 +717,18 @@ def update_resources(self, **resources):
...
+        if config:
+            return {
+                'disable_failures': config.get('disable-num-failures', self._config.disable_failures),
```
Tarrasch: I would guess this could all be simplified and this conversion-code made go away. Also consider packing things in a named-tuple. Just an idea :)

luigi/scheduler.py
```diff
@@ -655,9 +672,15 @@ def add_task(self, task_id=None, status=PENDING, runnable=True,
...
             task.deps = set(deps)
 
+        if deps_configs is not None:
+            task.deps_configs = deps_configs
+
         if new_deps is not None:
             task.deps.update(new_deps)
 
+        if new_deps_configs is not None:
+            task.deps_configs.update(new_deps_configs)
```
Tarrasch: This looks plain wrong. Actually, I think this feature could be implemented totally agnostic of dynamic dependencies. Have you considered completely removing `new_deps_configs` and not thread it around? If you think it's necessary, can you point out a test case that would othewise fail without it?

luigi/scheduler.py
```diff
@@ -536,7 +549,9 @@ def __init__(self, config=None, resources=None, task_history_impl=None, **kwargs
         self._make_task = functools.partial(
             Task, disable_failures=self._config.disable_failures,
             disable_hard_timeout=self._config.disable_hard_timeout,
-            disable_window=self._config.disable_window)
+            disable_window=self._config.disable_window,
+            upstream_status_when_all=self._config.upstream_status_when_all
```
Tarrasch: Please add a trailing comma to make code style cleaner :)

luigi/scheduler.py
```diff
@@ -667,7 +690,10 @@ def add_task(self, task_id=None, status=PENDING, runnable=True,
             # Task dependencies might not exist yet. Let's create dummy tasks for them for now.
             # Otherwise the task dependencies might end up being pruned if scheduling takes a long time
             for dep in task.deps or []:
-                t = self._state.get_task(dep, setdefault=self._make_task(task_id=dep, status=UNKNOWN, deps=None, priority=priority))
+                t = self._state.get_task(dep, setdefault=self._make_task(task_id=dep, status=UNKNOWN, deps=None,
+                                                                         priority=priority,
+                                                                         **self._get_task_level_config(
```
Tarrasch: ah, this is where the magic happens

luigi/task.py
```diff
@@ -146,6 +146,8 @@ class MyTask(luigi.Task):
     #: Only works when using multiple workers.
     worker_timeout = None
 
+    config = {}
```
Tarrasch: Please comment this. Also the name `config` is both short and unclear. I imagine something very explicit like `per_task_retry_policy`, or perhaps the shorter `retry_policy` or `retry_policy_dict`. A non-luigi user reading luigi code would then have a chance to understand this.

There probably would be some user who already use `config`

Tarrasch:
Summary of my comments:

 * Excellent documentation
 * Despite it's pretty short
 * I think there's 1 bug related to run()time dependencies feature (`new_deps_config`)

test/scheduler_test.py
```diff
@@ -112,5 +112,129 @@ def __init__(self):
 
         worker.prune(TmpCfg())
 
+    @with_config({'scheduler': {'disable-num-failures': '44'}})
+    def test_scheduler_with_task_level_config(self):
+        cps = luigi.scheduler.CentralPlannerScheduler()
```
Tarrasch: Sorry, as I merged this you have to update this line :)

https://github.com/spotify/luigi/pull/1781

examples/task_spesific_configuration.py
javrasya: You mean renaming the task?

luigi/scheduler.py
javrasya: I haven't try it actually. I haven't master all luigi code yet. I should ask that, may dynamic dependencies also have such task level config?

luigi/task.py
javrasya: You are right, more spesific and readable naming should be used. But this contains some configuration from luigi config. It may seems only for `retry-policy` at the moment. But I thought that this feature may includes another configuration besides `retry-policy` by the end of the day. I don't know, `email-sender`, maybe `email-prefix` or another suitable config. So maybe naming should not be related with retry-policy. What would you think about it, am I missing it?

examples/task_spesific_configuration.py
Tarrasch: Sorry, I was a bit unclear. I imagined that you make the comment like in: https://github.com/spotify/luigi/blob/5bd9b436990d48acc8def675382c29f8fb0e729e/examples/execution_summary_example.py#L16-L46.

However, your text would be reworded to be written to describe the whole example, instead of talking from the perspective of the "MyTaskHasConfig" class.

luigi/scheduler.py
Tarrasch: I suppose that they do have task-level config too. Makes sense right? However, from this code (when you read this in combination with the `worker.py` code related to the `new_deps_configs` variable) it seems like you "inherit" the task-level config which just is nonsensical, right? I might have misunderstood your code though. :)

luigi/task.py
Tarrasch: Wait a second ..., this is not the first time we do task-level config. Perhaps you could just follow the way previous contributors have done?

https://github.com/spotify/luigi/blob/e02ba97c8d29e61cdd7d4c7a8845dcd05ed9a35b/luigi/task.py#L133-L156

I believe it as you should do like it's done for `owner_email`, that is (1.) use `@property` and (2.) add docs under it.

This method seems much better than the dict approach. For instance you can hyperlink to config options [like this](http://luigi.readthedocs.io/en/stable/api/luigi.task.html#luigi.task.Task.owner_email)

luigi/scheduler.py
javrasya: On scheduler side, whenever a task is tried to be added, its retry-policy should also be sent from wherever it is invoked. If it is not sent, the config in luigi.cfg will be active for retry-policy. To provide that, I looked all places which invoke scheduler `_add_task` method and also made them send task-level config. I saw two kind of places which  are runing task normally and handling dynamic dependencies. I though that, even a task is added as dynamically, its task-level config must be passed to add_task method to apply task-level config of those kind of tasks. Am I thinking wrongly?

luigi/task.py
javrasya: Oh, I will check that, this may be more proper way of doing it.

luigi/scheduler.py
Tarrasch: Maybe think of it from the other side. Say you run TaskX(), which uses dynamic dependencies. You need `new_deps` because after each `yield`-from-a-`run()`, you will add a few more dependencies. However, for each `yield`-from-a-`run()` you don't update the retry-policy of the already added task TaskX(). Sure, the new deps laso have their own retry-policies but they are added in *their* `_add_task` call. Makes sense?

javrasya: Oh, yes. I see it now.

examples/task_spesific_configuration.py
javrasya: Removing means, removing `sleep` right?

javrasya:
@Tarrasch I don't want to bother more by pushing more commit in this PR without being sure. Can you review [changes according to your comments](https://github.com/javrasya/luigi/commit/ca8aef4bb35848c0afefb9f738d9e3f7863a6d0e) . I made sort of major changes. 

* I changed the way of using dict to @property
* named-tuples are used to make it more readable and maintainable.
* config -> retry-policy as term. (task_cofigs -> retry_policy, deps_task_configs -> deps_retry_policies)
* new_deps_configs is removed.
* It is added that how to run the added example on command line.

Please keep commenting there and I will correct it and merge it into my master branch to update PR after all.

Tarrasch:
Actually it's totally fine to push to this PR, anyone not interested may unsubscribe. No hurt feelings. :)

Anyway. I'll commenting on the link you sent :)

Tarrasch:
So it looks better now. Although you didn't remove `new_deps_configs` concept right?

-----

Lastly, I'm pretty unconvinced that `upstream-status-when-all` is something that we want. I would like it removed. At least from this patch. Main concerns of mine:

 * You wrote a motivation accompanied with this PR, but I don't understand it
 * Currently I can say to my boss that "Here is the luigi visualiser, an upstream disabled task is a task who yet can't run because some dependency is disabled". Short and concise. After this patch, we make the system complicated.
 * You add it to the retry-policy NamedTuple. But it has nothing to do with retry-policy.

But naturally, I'm open to be persuaded to think otherwise. :)

javrasya:
Actually I removed `new_deps_config` which is related with dynamic dependencies. 

`upstream-status-when-all` is a bit critical after retry-policy per task feature. Assume that, I have multiple etl report to be done at night and some of them have different retry-policy. For example one of them has retry-count as 5, another has it as 2. Luigi is marking parent task as `DISABLED` and shutting down the worker and does not let the task retry itself 5 times because of the other task is disabled after 2 times retry. In the example which I added, this can be seen. I added this config because, I may want to wait for all tasks to retry itself even one is `DISABLED` and settings the task as `UPSTREAM_DISABLED` when all its upstream tasks are `DISABLED` (not includes SUCCESS). This config   is more critical after the per task retry-policy config as I said. But it does not break anything, because its default value is `False` and it works how luigi works currently.

Tarrasch:
Ok. I think we aren't on the same page when it comes to `upstream-status-when-all`. Maybe I've misunderstood parts of the PR. Also comment https://github.com/javrasya/luigi/commit/44415be794239c2899728d37182f92e54a965689#commitcomment-18391330 comes to mind.

Let's render an example. 

-----

Task X depends on Y2 and Y5. Y2 has retry-count 2 and Y5 has retry-count 5. Let's say that In our example X will never run because *both* Y2 and Y5 will not both get DONE.

Now let's say Y5 failed and is now marked as red while Y2 have not run yet and is pending.

`upstream-status-when-all`: If this config is set to false (default) for task X, then X will be in the set of UPSTREAM_FAILED because Y5 is. But if this variable is set to true, it will not be in UPSTREAM_FAILED, becuase Y2 is still pending.

We note that UPSTREAM_FAILED is not something a task ever is marked as, it's just a *view* that's only exists in the luigi scheduler. It's possible that a task is UPSTREAM_FAILED, UPSTREAM_DISABLED, both, or none.

Are we on the same page so far?

-----

Actually. I downloaded your branch and tested to run your example. And I do see what you mean. now. It only works when that variable is true. What about this proposal. Because I really don't want to introduce a configuration option that takes a luigi maintainer hours to understand, can you try to change the default behaviour to do as if your new variable was true?

Yes, it's a bit provocative to change this over 3 years old concept in luigi. But I don't really seeing it being used anywhere except for visualization. There's only one place (as of https://github.com/spotify/luigi/pull/871) and in that case I believe the author mistakenly assumed the behaviour works as I propose we change it too.

I checked that no scheduler-api tests fail after the change, and I can help you to provide a test case that will fail until we change the behaviour. How does that sound? Less code for both of us and I think we fix a design-flaw in luigi. And yes, you will run into this bug in production as of today, it's just that in your case both child tasks crash after 0 seconds, if one task has a delay you will get into the same problem even if both have the same retry-count.

Tarrasch:
Also linking in https://github.com/spotify/luigi/pull/686, even there the author said "if *all* dependencies", it seems that `upstream-status-when-all` should always be `True` and it's `False` for some ancient mistake years ago.

javrasya:
I would go for not including this configuration if we have a chance and changing the default behavior for all instead of any.

Tarrasch:
Yes. I think the default behaviour should change, today I'll work on creating a test case which only passes with the new behaviour.

javrasya:
A few tests fail after I change the current behavior of luigi on that. I will fix them and update the PR soon.

Tarrasch:
@javrasya, it would be so ideal if that could be made into a separate PR that we merge first. Could you arrange for that? Then we have one diffview where it's clear what that change affects. I remember scheduler_api tests did not start failing, but maybe I ran it incorrectly.

javrasya:
The one which fails is `test_task_list_upstream_status`. I fixed it and pushed it but I can rollback it if you want.

doc/configuration.rst
```diff
@@ -749,52 +736,42 @@ user
...
-| disable-hard-timeout     | scheduler |
-+--------------------------+-----------+
-| disable-window-seconds   | scheduler |
-+--------------------------+-----------+
-| upstream-status-when-all | scheduler |
-+--------------------------+-----------+
++------------------------+-----------+
+| Config                 | Section   |
++========================+===========+
+| retry-count            | scheduler |
```
Tarrasch: `retry_count`. Use underscores.

doc/configuration.rst
```diff
@@ -749,52 +736,42 @@ user
...
++------------------------+-----------+
+| disable-hard-timeout   | scheduler |
```
Tarrasch: The correct in config is to always use underscores. There are places where that's not the case, those cases are just yet not updated. :)

doc/configuration.rst
javrasya: True it is from first version of this feature.

luigi/task.py
```diff
@@ -149,6 +149,30 @@ class MyTask(luigi.Task):
     config = {}
 
     @property
+    def disable_num_failures(self):
```
Tarrasch: Should now say `retry_count`

test/target_test.py
```diff
@@ -255,7 +255,7 @@ def test_move_on_fs(self):
         fs.move(t.path, t.path+"-yay")
         self.assertFalse(t.exists())
 
-    def test_rename_dont_move_on_fs(self):
+    def test_renamegs_dont_move_on_fs(self):
```
Tarrasch: ??

luigi/scheduler.py
```diff
@@ -108,12 +108,12 @@ class scheduler(Config):
...
 
-    # Jobs are disabled if we see more than disable_failures failures in disable_window seconds.
+    # Jobs are disabled if we see more than retry_count failures in disable_window seconds.
     # These disables last for disable_persist seconds.
     disable_window = parameter.IntParameter(default=3600,
                                             config_path=dict(section='scheduler', name='disable-window-seconds'))
-    disable_failures = parameter.IntParameter(default=999999999,
-                                              config_path=dict(section='scheduler', name='disable-num-failures'))
+    retry_count = parameter.IntParameter(default=999999999,
+                                         config_path=dict(section='scheduler', name='retry-count'))
```
Tarrasch: noooo... `s/retry-count/disable_failures`.

Sorry for the confusion. But `config_path` should be called `deprecated_config_path`. The confusion is as of our bad naming.

test/target_test.py
javrasya: This was mistakenly pushed sorry. I fixed it.

luigi/scheduler.py
javrasya: Ok probably, I got it wrongly. I first didn't change it but, after a comment of yours, I changed it. I can rollback

javrasya: OK I am bit of confused about your comment `Oh, even more important. Change this line to say name='disable_failures'! Also update the docs to tell people to use retry_count.` . Should I rollback it?

Tarrasch: Code should be `name='disable_failures`` so we keep backward compatiblity. Docs should encourage usage of the newest name `retry_count`, so people dont get deprecation warnings.

Just add upon commits. There's not point in "rolling back". In the end I'll squash everything anyway as the history is a bit messy already. :)

javrasya: Oh, I guess I got it now. You mean, If I describe it as `retry_count` in `.cfg` file, even the name is `disable-num-failures`, both will be able to be used ? So all I need to commit is to change the part as `name='disable-num-failures'`, not the doc right?

luigi/scheduler.py
```diff
@@ -862,9 +885,8 @@ def _upstream_status(self, task_id, upstream_status_table):
                     elif upstream_status_table[dep_id] == '' and dep.deps:
                         # This is the postorder update step when we set the
                         # status based on the previously calculated child elements
-                        status = max((upstream_status_table.get(a_task_id, '')
-                                      for a_task_id in dep.deps),
-                                     key=UPSTREAM_SEVERITY_KEY)
+                        upstream_severities = list(upstream_status_table.get(a_task_id) for a_task_id in dep.deps if a_task_id in upstream_status_table) or ['']
+                        status = min(upstream_severities, key=UPSTREAM_SEVERITY_KEY)
```
javrasya: Here is the change of luigi default behavior to detect upstream severity. Parent task marked as `UPSTREAM_DISABLED` when any of its upstream task is `DISABLED` and the worker was shutdown because of it. This behavior is changed to wait for all upstream tasks to be `DISABLED` or `FAILED` to mark parent task as `UPSTREAM_DISABLED` or `UPSTREAM_FAILED`.

Tarrasch:
Ok so your first next step here is to create a new pr with only the "`upstream-status-when-all`" change. I'll try to quickly merge it and then you can create a follow up PR with the rest of the changes.

This PR is really messy now. It's better to have many small PRs that get merged so we have the feeling that we're progressing. :)

PR #1783, opened 2016-07-26
README: Remove monthly downloads badge

Tarrasch:
## Description

README: Remove monthly downloads badge

## Motivation and Context

The fact that there's only 4 downloads per month is not something we can really brag about. Let's just remove this bag. A historic note is that at some point that number fell from the tens of thousands downloads per month to just a handful. I suppose pypi fixed a bug on their side.


erikbern:
i think pypi stopped publishing those numbers a while ago, not sure why it's showing 4 though

PR #1784, opened 2016-07-27
Enables running of multiple tasks in batches

daveFNbuck:
## Description
Implements batching with worker-side batch methods.

## Motivation and Context
Sometimes it's more efficient to run a group of tasks all at once rather than
one at a time. With luigi, it's difficult to take advantage of this because your
batch size will also be the minimum granularity you're able to compute. So if
you have a job that runs hourly, you can't combine their computation when many
of them get backlogged. When you have a task that runs daily, you can't get
hourly runs.

For the user, batching is accomplished by setting batch_method in the parameters
that you wish to batch. This takes a list of parameter values (already parsed)
and combines them into a single parameter. You can limit the number of tasks
allowed to run simultaneously in a single batch by setting the class variable
batch_size. If you want to exempt a specific task from being part of a batch,
simply set its is_batchable property to False.

Unlike #1538, this does batch combining on the worker side, allowing arbitrary
batch methods and keeping better track of task ids for things like execution
summary.

Added many unit tests. Will test this in production if I get an okay on the idea.


luigi/scheduler.py
```diff
@@ -746,13 +820,18 @@ def get_work(self, host=None, assistant=False, current_tasks=None, worker=None,
         if assistant:
             self.add_worker(worker_id, [('assistant', assistant)])
 
+        batched_params, unbatched_params, batched_tasks, max_batch_size = None, None, [], 1
```
Tarrasch: Would using `{}` make more sense than `None`?

Tarrasch: Oh it's a `dict()` not a `set()`, you see how easy it is to get confused? ^^

luigi/scheduler.py
```diff
@@ -804,6 +889,20 @@ def get_work(self, host=None, assistant=False, current_tasks=None, worker=None,
...
+                            batched_params = {
+                                name: [task.params[name]] for name in batch_param_names
+                            }
+                            unbatched_params = {
+                                name: value for name, value in task.params.items()
+                                if name not in batched_params
+                            }
+                            batched_tasks.append(task)
+                        except KeyError:
+                            batched_params, unbatched_params = None, None
```
Tarrasch: Same here of course about `None`/`dict()`

luigi/scheduler.py
```diff
@@ -804,6 +889,20 @@ def get_work(self, host=None, assistant=False, current_tasks=None, worker=None,
             if self._schedulable(task) and self._has_resources(task.resources, greedy_resources):
                 if in_workers and self._has_resources(task.resources, used_resources):
                     best_task = task
+                    batch_param_names, max_batch_size = self._state.get_batcher(
+                        worker_id, task.family)
+                    if batch_param_names and task.is_batchable():
+                        try:
+                            batched_params = {
+                                name: [task.params[name]] for name in batch_param_names
```
Tarrasch: You create a generator of singleton lists here???

luigi/scheduler.py
```diff
@@ -822,7 +921,22 @@ def get_work(self, host=None, assistant=False, current_tasks=None, worker=None,
...
                  'n_unique_pending': n_unique_pending}
 
-        if best_task:
+        if len(batched_tasks) > 1:
+            batch_id = hashlib.md5('|'.join(task.id for task in batched_tasks)).hexdigest()
+            for task in batched_tasks:
+                self._state.set_batch_running(task, batch_id, worker_id)
+
+            combined_params = best_task.params.copy()
+            combined_params.update(batched_params)
```
Tarrasch: This is the line that made me shiver seeing `batched_params` is ever set to `None`

Tarrasch:
Ok. This looks more promising for every iteration. Thought caught my attention the most is that it seems like `batch_mode=my__pure_python_func` is strictly stronger than the decorator method that you also have created. Why wouldn't one just use something like `batch_method=(lambda x: x)` or `batch_mode=list`?

luigi/scheduler.py
daveFNbuck: `{}` is a `dict`, not a `set`.

I think it's less clear to set it to an empty dict than to set it to None. An empty dict looks like something you're going to fill rather than replace. If you think it's more clear, I'm happy to change it though.

luigi/scheduler.py
daveFNbuck: It's a dictionary comprehension of singleton lists. It has the list of batch parameters for each parameter name. When we first create it, each batch parameter only has one value.

luigi/scheduler.py
daveFNbuck: batched_params couldn't be None here because batched_tasks gets filled in only when batched_params is filled in. If we managed to get here with an empty batched_params, that's a huge logic failure in the scheduler too.

luigi/scheduler.py
Tarrasch: Ah. I see :)

daveFNbuck:
Instead of `batch_mode=list`, you'd need something like `batch_mode=flatten` as otherwise you'd have a list of lists. You'd then also need to implement parsing and serialization to handle lists and worry about escaping your separators, etc. I agree that just setting `batch_mode=function` would be better but unfortunately list parameters are tricky. Maybe we should just leave this feature out?

daveFNbuck:
It's basically the same problem we had previously with list parameters, it's difficult to do with just a single argument.

luigi/parameter.py
```diff
@@ -222,6 +257,23 @@ def parse(self, x):
...
+
+        :param xs: list of values to parse and combine
+        :return: the combined parsed values
+        """
+        if self._batch_method is None:
+            raise NotImplementedError('No batch method found')
+        elif not xs:
+            raise ValueError('Empty parameter list passed to parse_list')
+        else:
+            return self._batch_method(map(self.parse, xs))
```
Tarrasch: About if `batch_method=list` will work or not:

@daveFNbuck, you've really got me confused. Here it really looks like that `xs` is one iterable of values. Ignoring the `parse`-step. If we have `xs=iterable(1,2,3)`, then as I see it:

 * `batch_method=max` ==> `3`
 * `batch_method=min` ==> `1`
 * `batch_method=','.join` ==> `'1,2,3'` (string)
 * `batch_method=list` ==> `[1,2,3]` (list of int)
 * `batch_method=(lambda x: x)` ==> `iterable(1,2,3)`

If it was this simple it would be fantastic. Because we wouldn't need `batch_tuple_parameter`. But that's not the case huh?

daveFNbuck: Suppose you have the following code:

```
class ExampleTask(luigi.Task):
    param = IntParameter(batch_method=list)
```

Now `ExampleTask(1).param` is an int, `ExampleTask(2).param` is an int, and so on. But if I send these both to the scheduler and get back a batch of both of them, I get `ExampleTask([1, 2])`. `ExampleTask([1, 2]).param` is a list of ints. If I call serialize on it, it's going to fail. I'll also have to write weird code that handles both ints and lists of ints.

To make this work cleanly, I'll need to change the type of this parameter from int to list of ints. I'll need to change the parse and serialize functions. Now `batch_method=list` won't work because if I batch `ExampleTask([1])` and `ExampleTask([2])`, I get `ExampleTask([[1, 2]])`. Again, the types are wrong because `list` changes the type from `X` to `list<X>`. So I also need to import or write a flatten function to return a list.

So in order to get a `list` batch method, I need to write custom parse and serialize methods and find a flatten function. The parse and serialize methods might also need to handle weird escaping in case the parameter value could include the list separator. The only way I can see to make this all easy and likely to succeed for the average user is to provide a well-tested decorator. Otherwise, they're likely to fall into the same trap you did (and I did in some of my unit tests before getting this far) and just set `batch_method=list`.

Tarrasch: Ah. I see. I understand the issue.

If so. I suggest the following:

 * We start simple, lets not aim for support having lists yet as it's not as easy as one might think
 * Clearly say in the docs that `<func>` in `batch_method=<func>` must be a pure python function of signature `<func> :: [A] -> A`. (or however you would explain this in simple terms)
 * Also add a runtime check that checks that the type returned by `<func>` is correct.

Does that make sense @daveFNbuck?

daveFNbuck: The runtime check could be tricky, as you could have `[A] -> B` where `A` and `B` are both subclasses of the same superclass. Unfortunately, that's true for any pair of new-style classes where object is the superclass. We probably don't want to have people explicitly pass the class. We could do a simple check like `B` can only be iterable if `A` is iterable.

daveFNbuck: Example: `u','.join` can take a list of `str` and return a `unicode`.

daveFNbuck: I guess I can't think of a reason this would actually happen with realistic code though.

Tarrasch: I rely on you to come up with any simple compromise here :)

daveFNbuck: I can move this into a separate PR.

daveFNbuck: Removed from this PR, not sure why the tests aren't re-running.

daveFNbuck: Seems like it doesn't get triggered on pushes anymore.

daveFNbuck:
Are there any more issues with this? I think I've addressed them all.

doc/luigi_patterns.rst
```diff
@@ -118,6 +118,87 @@ can have implications in how the scheduler and visualizer handle task instances.
...
+setting the batch_method to max, like so:
+
+.. code-block:: python
+
+    class ExampleTask(luigi.Task):
+        date = luigi.DateParameter(batch_method=max)
+
+What's exciting about this is that if you send multiple ExampleTasks to
+the scheduler, it can combine them and return one. So if
+ExampleTask(date=2016-07-28), ExampleTask(date=2016-07-29) and
```
Tarrasch: Perhaps you could format this into `code text`?

I'm not sure, but perhaps just calling the task `A` will make it more readable as `ExampleTask` is quite big and draws user attention from what's important.

Tarrasch: By the way this is how it looks rendered (almost, it's not identical to readthedocs) https://github.com/Houzz/luigi/blob/985522850ecfc8d84b9e56cbeb67bde95042dfcc/doc/luigi_patterns.rst

doc/luigi_patterns.rst
```diff
@@ -118,6 +118,87 @@ can have implications in how the scheduler and visualizer handle task instances.
...
+ExampleTask(date=2016-07-30) are all ready to run, you will start running
+ExampleTask(date=2016-07-30). While this is running, the scheduler will
+show ExampleTask(date=2016-07-28), ExampleTask(date=2016-07-29) as batch
+running while ExampleTask(date=2016-07-30) is running. When
+ExampleTask(date=2016-07-30) is done running and becomes FAILED or DONE,
+the other two tasks will be updated to the same status.
+
+If you want to limit how big a batch can get, simply set batch_size.
```
Tarrasch: Should be "simply set `max_batch_size`" right?

doc/luigi_patterns.rst
```diff
@@ -118,6 +118,87 @@ can have implications in how the scheduler and visualizer handle task instances.
...
+
+Now if you have multiple tasks such as
+DailyOverwriteTask(date=2016-06-01),
+DailyOverwriteTask(date=2016-06-02),
+DailyOverwriteTask(date=2016-06-03), the scheduler will just tell you to
+run the highest available one and mark the lower ones as batch_running.
+Using a unique resource will prevent multiple tasks from writing to the
+same location at the same time if a new one becomes available while
+others are running.
+
```
Tarrasch: I just finished reading the docs. It looks pretty good! :)

luigi/parameter.py
```diff
@@ -117,7 +117,7 @@ def run(self):
     _counter = 0  # non-atomically increasing counter used for ordering parameters.
 
     def __init__(self, default=_no_value, is_global=False, significant=True, description=None,
-                 config_path=None, positional=True, always_in_help=False):
```
Tarrasch: You really should add docs for this. :)

luigi/parameter.py
```diff
@@ -117,7 +117,7 @@ def run(self):
...
+                 config_path=None, positional=True, always_in_help=False, batch_method=None):
```
Tarrasch: You really should add docs for this. Perhaps just linking to the `luigi_patterns.rst` file

luigi/scheduler.py
```diff
@@ -224,10 +226,18 @@ def __init__(self, task_id, status, deps, resources=None, priority=0, family='',
         self.status_message = status_message
         self.scheduler_disable_time = None
         self.runnable = False
+        self.batchable = False
+        self.batch_id = None
 
     def __repr__(self):
         return "Task(%r)" % vars(self)
 
+    def is_batchable(self):
```
Tarrasch: The long term plan is just to replace this with `self.batchable` right? Can you write a comment that this is only for backward compatability (and perhaps include a date)

luigi/scheduler.py
```diff
@@ -366,13 +379,31 @@ def get_active_tasks(self, status=None):
...
     def get_running_tasks(self):
         return six.itervalues(self._status_tasks[RUNNING])
 
     def get_pending_tasks(self):
         return itertools.chain.from_iterable(six.itervalues(self._status_tasks[status])
                                              for status in [PENDING, RUNNING])
 
+    def set_batcher(self, worker_id, family, batcher_args, max_batch_size):
+        if max_batch_size is None:
+            max_batch_size = float('inf')
```
Tarrasch: Can we avoid in `None` in the code by much earlier setting this default? Perhaps as the default value for the `rpc_method`

luigi/scheduler.py
```diff
@@ -634,25 +687,35 @@ def add_task(self, task_id=None, status=PENDING, runnable=True,
...
+            task.batch_id = batch_id
+        if status == RUNNING and not task.worker_running:
+            task.worker_running = worker_id
+
         if tracking_url is not None or task.status != RUNNING:
             task.tracking_url = tracking_url
+            for batch_task in self._state.get_batch_running_tasks(task.batch_id):
+                batch_task.tracking_url = tracking_url
+
+        if batchable is not None:
```
Tarrasch: Can we avoid tribool logic here? Like now we have 3 possible values (`None`, `True`, `False`). Could the default value for `batchable` be set to `False`?

luigi/worker.py
```diff
@@ -678,6 +707,26 @@ def _log_remote_tasks(self, running_tasks, n_pending_tasks, n_unique_pending):
...
+            )
+            self._scheduler.add_task(
+                worker=self._id,
+                task_id=task.task_id,
+                module=get_work_response.get('task_module'),
+                family=get_work_response['task_family'],
+                params=task.to_str_params(),
+                status=RUNNING,
+                batch_id=get_work_response['batch_id'],
+            )
```
Tarrasch: Wait, this `add_task` worries me a bit. Previously a worker have not needed to say "I'm RUNNING TaskABC()", because the scheduler have already told the worker to run a task. Why do things need to be different for the BATCH case?

I'm not sure how useful the invariant of `get_work` is never called with `status=RUNNING` actually is. But I'm interested to hear why it can't be kept. :)

test/execution_summary_test.py
```diff
@@ -95,6 +95,59 @@ def requires(self):
...
+        self.run_task(MaxBatches())
+        d = self.summary_dict()
+        expected_failed = {
+            MaxBatchFailTask(0),
+            MaxBatchFailTask(1),
+            MaxBatchFailTask(2),
+            MaxBatchFailTask(3),
+            MaxBatchFailTask(4),
+        }
+        self.assertEqual(expected_failed, d['failed'])
```
Tarrasch: Thanks for also testing the execution summary. :)

luigi/parameter.py
```diff
@@ -223,6 +227,23 @@ def parse(self, x):
         """
         return x  # default impl
 
+    def parse_list(self, xs):
+        """
+        Parse a list of values from the scheduler.
+
+        Only possible if this parameter has a batch method. This will combine the list into a single
```
Tarrasch: Actually shouldn't it be "Only possible if `is_batchable()` is true".

luigi/parameter.py
```diff
@@ -223,6 +227,23 @@ def parse(self, x):
         """
         return x  # default impl
 
+    def parse_list(self, xs):
```
Tarrasch: May I suggest we start with defining `is_batchable` and `parse_list` as `_private`? Or do you already have use cases in mind where you want to override them?

luigi/parameter.py
```diff
@@ -223,6 +227,23 @@ def parse(self, x):
...
+        """
+        Parse a list of values from the scheduler.
+
+        Only possible if this parameter has a batch method. This will combine the list into a single
+        parameter value using batch method.
+
+        :param xs: list of values to parse and combine
+        :return: the combined parsed values
+        """
```
Tarrasch: Can we also add in the docs that in most cases, this does not need to be overriden? (Or am I wrong about this?)

luigi/worker.py
```diff
@@ -389,11 +389,14 @@ def __init__(self, scheduler=None, worker_id=None, worker_processes=1, assistant
         self.host = socket.gethostname()
         self._scheduled_tasks = {}
         self._suspended_tasks = {}
+        self._batch_running_tasks = {}
+        self._batch_classes_sent = set()
```
Tarrasch: Maybe rename to `_batch_families_sent`? And/or add comment `  # set of task families`

luigi/scheduler.py
```diff
@@ -366,13 +379,31 @@ def get_active_tasks(self, status=None):
             for task in six.itervalues(self._tasks):
                 yield task
 
+    def get_batch_running_tasks(self, batch_id):
+        if batch_id is None:
```
Tarrasch: If possible I would prefer that you would not be allowed to pass in `None`. Don't you think it would increase readability?

Tarrasch:
Starting to look good. I would like to get this in and then release 2.4.0. There's a lot of new cool features added recently. :)

doc/luigi_patterns.rst
daveFNbuck: Everything else in the docs uses code-block. I'm having trouble finding code-text in the documentation for reStructuredText (it's not very searchable). Do I just replace `code-block` with `code-text` or did you mean to leave a space there? What's the difference between this and `code-block`? I think it looks fine in the link...

I changed ExampleTask to A.

doc/luigi_patterns.rst
daveFNbuck: Right, good catch!

doc/luigi_patterns.rst
daveFNbuck: Thanks!

luigi/parameter.py
daveFNbuck: How do I link to that? Also, how should I mark the type? I've made an attempt but it's probably wrong.

luigi/scheduler.py
daveFNbuck: Right, added a TODO comment for a year from now since it'll probably be a while before this makes it to most end users.

luigi/scheduler.py
daveFNbuck: I had to set it as the default value in Task, as rpc_method always receives an override to the default.

luigi/scheduler.py
daveFNbuck: batchable is False by default when a new task is created. It'll never be None in the internal representation in the scheduler.

This just says not to override batchable when an add_task doesn't specify it. add_tasks for changing job statuses don't always contain every argument, so we don't want to change task.batchable if it isn't supplied. It's the same pattern we use for other parameters nearby in the same function.

luigi/worker.py
daveFNbuck: We need this because we moved the batching logic into the worker. When the scheduler sends the worker a bunch of batched tasks, it doesn't know what task the worker will end up running. The worker might start running one of the batched tasks or it could create a new task that the scheduler doesn't know about yet. So it has to inform the scheduler how it's running the batch.

We could add a separate API call for this response, but it would likely have to call `add_task` at least as a first step. I guess it could do that and then mark it running after to keep the `status != RUNNING` invariant if that's important.

test/execution_summary_test.py
daveFNbuck: I've been using a version that doesn't handle execution summary for a very long time, so I'm excited that this one will be able to do it properly.

luigi/parameter.py
daveFNbuck: Sounds like a better condition. Changed.

luigi/parameter.py
daveFNbuck: They're public because they're used by Task. is_batchable gets called in Task.batch_param_names and parse_list gets called in Task.from_str_params

luigi/parameter.py
daveFNbuck: You're right about that. Added to the docs.

luigi/worker.py
daveFNbuck: renamed

luigi/scheduler.py
daveFNbuck: I'll just have to add the same conditional around each call to this function. I'll do that if you prefer.

daveFNbuck: Turns out 2 out of 4 calls already had this condition anyway.

daveFNbuck:
Thanks so much for all the comments! I'm very excited that we're finally going to get this feature into Luigi. It'll be very nice to end the ~2 years of having a significantly different fork.

doc/luigi_patterns.rst
Tarrasch: Sorry. My previous comment was confusing. I just mean to turn "A(date=2016-07-28)" into "`A(date=2016-07-28)`". That is done using double-backticks (\`\`) in RST.

luigi/worker.py
Tarrasch: What you say makes sense. I just found the code a bit hard to read, in particular the `params_str=get_work_response['task_params']` still looks like we're talking about a single task, but `get_work_response['task_params']` is in fact a dict of lists now. 

-------

It's fine to keep this I think. I don't know any reason for us to keep the RUNNING invariant,

doc/luigi_patterns.rst
daveFNbuck: Got it, thanks. Updated.

luigi/parameter.py
```diff
@@ -140,6 +140,10 @@ def __init__(self, default=_no_value, is_global=False, significant=True, descrip
                                 ``positional=False`` for abstract base classes and similar cases.
         :param bool always_in_help: For the --help option in the command line
                                     parsing. Set true to always show in --help.
+        :param function(iterable[A]) -> A batch_method: Method to combine an iterable of parsed
+                                                        parameter values into a single value. Used
+                                                        when receiving batched parameter lists from
+                                                        the scheduler.
```
Tarrasch: I just created #1814 which is going to be able to help you.

As for referencing to a section in the documentation. See this example:

 * Place where you refer: https://github.com/spotify/luigi/blob/298a0d13e812f5613e39d86a24f51ac9348b66be/luigi/task.py#L591
 * Place where you refer to: https://raw.githubusercontent.com/spotify/luigi/298a0d13e812f5613e39d86a24f51ac9348b66be/doc/configuration.rst (the line containing `.. _ConfigClasses:`)

luigi/scheduler.py
```diff
@@ -380,13 +382,11 @@ def get_active_tasks(self, status=None):
...
 
     def get_batch_running_tasks(self, batch_id):
-        if batch_id is None:
-            return []
-        else:
-            return [
-                task for task in self.get_active_tasks(BATCH_RUNNING)
-                if task.batch_id == batch_id
-            ]
+        assert batch_id is not None
```
Tarrasch: I really like this style. :)

luigi/parameter.py
Tarrasch: Actually. This is one of the things I like about Java vs Python. You seem to consider "private means only the class may use it". I consider it as "private means that it's Luigi-private". In particular Luigi-contributors might at any time remove it. The reason I think the way I do is that everything that's not prepended with `_` is going to end up on the [docs page](http://luigi.readthedocs.io/en/stable/api/luigi.parameter.html#luigi.parameter.Parameter).

doc/luigi_patterns.rst
```diff
@@ -183,19 +183,17 @@ resource:
...
-run the highest available one and mark the lower ones as batch_running.
-Using a unique resource will prevent multiple tasks from writing to the
-same location at the same time if a new one becomes available while
-others are running.
+Now if you have multiple tasks such as ``A(date=2016-06-01)``,
+``A(date=2016-06-02)``, ``A(date=2016-06-03)``, the scheduler will just
+tell you to run the highest available one and mark the lower ones as
+batch_running. Using a unique resource will prevent multiple tasks from
+writing to the same location at the same time if a new one becomes
+available while others are running.
```
Tarrasch: While it's implicit from this text if you think hard about it. We perhaps should mention that "if task X, Y, Z gets batch-run as task Z, those 3 tasks will only take up as many resources as Z alone would have had". (and perhaps the word resources is a hyperlink to the docs about luigi resources)

luigi/parameter.py
daveFNbuck: I guess that's a fair way to look at it. I'll rename them both.

doc/luigi_patterns.rst
daveFNbuck: Added to the documentation of the previous section, where it's much less obvious that it would work that way.

daveFNbuck: Turns out there's no documentation specifically about resources. I linked to the [resources] section in configuration, which comes closest to documenting it.

luigi/parameter.py
Tarrasch: If you happen to know any reading reference with better conventions. I'll be super interested to know. :)

daveFNbuck: [PEP 8](https://www.python.org/dev/peps/pep-0008/) seems to agree with you:

> Public attributes are those that you expect unrelated clients of your class to use, with your commitment to avoid backward incompatible changes. Non-public attributes are those that are not intended to be used by third parties; you make no guarantees that non-public attributes won't change or even be removed.

Tarrasch: Ah. Cool. :)

Tarrasch:
Hi @daveFNbuck, I was hopeful to start using this. But in my case I would actually need something like the `batch_tuple_parameter` you had described here. I have an idea for a slightly different and hopefully even better API I'm willing to implement, but I couldn't find your original implementation. I see the docs in a commit [here](https://github.com/spotify/luigi/pull/1784/commits/b31018c2be7620be436c1994aa6df4268603b659#diff-2cde74a15862b5bef34fd70738b3540eR140), but I don't see the code being in that commit. Do you still have the code lying around somewhere? Please push a branch if you have anything. Otherwise I can re-implement it.

My thought for api could be something like:

```
class MyTask(luigi.Task)
     city, cities = batch_tuple_parameter(ChoiceParameter)(choices=('stockholm', 'hcmc'))
```

Where you still do `luigi MyTask --city stockholm` and `MyTask(city='hcmc')`, but within `run` you aren't supposed to touch `self.city` but only `self.cities`.

Would this make sense? Perhaps we should even recommend `_cities` as to clarify that you can optimize your task by batching really shouldn't affect the API.

daveFNbuck:
@Tarrasch sorry I didn't notice your comment until now. I don't think I have the implementation anymore. I like your idea, as it would simplify an issue I have where I can't batch tasks that have standardized parameters.

PR #1785, opened 2016-07-27
Excludes .tox from flake8 to prevent checking third-party libraries

daveFNbuck:
## Description
Ads .tox to the exclude list in flake8 commands

## Motivation and Context
flake8 is failing in the CI due to errors found in the .tox directory. Since these have nothing to do with our code, they can be safely ignored.

We'll see if it passes CI. Since other stuff is currently broken, we'll have to check the flake8 test in particular.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -71,7 +71,7 @@ commands =
 
 [testenv:flake8]
 deps = flake8
-commands = flake8 --max-line-length=160 --exclude=doc,luigi/six.py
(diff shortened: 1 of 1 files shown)


Tarrasch:
While it's still red I believe it's less red :)

daveFNbuck:
I can't replicate the other failures locally, I think the server might have a weird environment.

PR #1786, opened 2016-07-27
flake8: Unbreak travis build

Tarrasch:
## Description

flake8: Unbreak travis build

## Motivation and Context

I belive this might have broken as of a flake8 being published to pypi.

Furthermore, I removed the deprecate_kwarg module. Anyone should feel
free to put it back again when they want.

This relates to https://github.com/spotify/luigi/pull/1785


tox.ini
```diff
@@ -70,7 +70,7 @@ commands =
   nosetests -v --tests=test/visualiser
 
 [testenv:flake8]
-deps = flake8
+deps = flake8>=3.0.2
```
daveFNbuck: This doesn't fix the issue going forward, as newer versions of flake8 can still break the build without code changes.

Tarrasch: I'm aware of it. While I never like to wake up to a red build status. It's not a bad thing to force ourselves to keep the code up to date to the latest regulations. :)

PR #1788, opened 2016-07-27
Removes comment from NOQA in util.py

daveFNbuck:
Corrects the flake8 failures on CI and sets a specific flake8 version in tox.ini so that it won't break again without at least a code change.


daveFNbuck:
Closed because #1786 does roughly the same thing.

PR #1791, opened 2016-07-28
Per Task Retry-Policy

javrasya:
## Description
Defining retry-policy per task as it is defined in #1073 

## Motivation and Context

It may be required to have different retry-policy for different tasks according to their priority and resource usage. Assume that, a task is retrieving data from Hive which responses in long time and other one is retrieving data from RDBM which responses in short time. I may want Hive task not to retry many times and not to drain all my resource again an again. But, retrying RDBM task many times may not be problem for me. Assume that, I have some network issue for a while and I should be able to give its retry-count more than Hive task. So I would need to define retry-policy at task level.

This fixes the issue defined in #1073

* Some unit tests are added about the feature.
* Run available tests(Not all of them, I don't know, some of them maybe includes hadoop ones, I couldn't make passed. ) via tox and travis.
* PyCharm Debugging.

## What is added?
* The feature
* Unit tests
* Documentation
* An example

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -209,6 +209,7 @@ smtp_timeout
   Optionally sets the number of seconds after which smtp attempts should
   time out.
 
+.. _worker-config:
 
 [worker]
 --------
@@ -742,33 +743,35 @@ Luigi also supports defining retry-policy per task.
 
     class GenerateWordsFromHdfs(luigi.Task):
 
-       retry_count = 5
+       retry_count = 2
 
         ...
 
     class GenerateWordsFromRDBM(luigi.Task):
 
-       retry_count = 3
+       retry_count = 5
 
         ...
 
     class CountLetters(luigi.Task):
 
         def requires(self):
-            return [GenerateWordsFromHdfs(),GenerateWordsFromRDBM()]
+            return [GenerateWordsFromHdfs()]
+
+        def run():
+            yield GenerateWordsFromRDBM()
 
         ...
 
-Supported Configurations
-************************
-The configurations below are also definable in luigi config file. Check :ref:`scheduler-config`
+If none of retry-policy fields is defined per task, the field value will be **default** value which is defined in luigi config file. 
 
-+------------------------+-----------+
-| Config                 | Section   |
-+========================+===========+
-| retry_count            | scheduler |
-+------------------------+-----------+
-| disable_hard_timeout   | scheduler |
-+------------------------+-----------+
-| disable_window_seconds | scheduler |
-+------------------------+-----------+
+To make luigi sticks to the given retry-policy, be sure you run luigi worker with `keep_alive` config. Please check ``keep_alive`` config in :ref:`worker-config` section.
+
+Retry-Policy Fields
+-------------------
+
+The fields below are in retry-policy and they can be defined per task.
+
+* retry_count
+* disable_hard_timeout
+* disable_window_seconds
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -269,7 +269,7 @@ retry_external_tasks
   This means that if external dependencies are satisfied after a workflow has
   started, any tasks dependent on that resource will be eligible for running.
   Note: Every time the task remains incomplete, it will count as FAILED, so
-  normal retry logic applies (see: `disable-num-failures` and `retry-delay`).
+  normal retry logic applies (see: `retry_count` and `retry-delay`).
   This setting works best with `worker-keep-alive: true`.
   If false, external tasks will only be evaluated when Luigi is first invoked.
   In this case, Luigi will not check whether external dependencies are
@@ -550,10 +550,10 @@ disable-hard-timeout
   **again** after this amount of time. E.g. if this was set to 600
   (i.e. 10 minutes), and the task first failed at 10:00am, the task would
   be disabled if it failed again any time after 10:10am. Note: This setting
-  does not consider the values of the `disable-num-failures` or
+  does not consider the values of the `retry_count` or
   `disable-window-seconds` settings.
 
-disable-num-failures
+retry_count
   Number of times a task can fail within disable-window-seconds before
   the scheduler will automatically disable it. If not set, the scheduler
   will not automatically disable jobs.
@@ -563,7 +563,7 @@ disable-persist-seconds
   Defaults to 86400 (1 day).
 
 disable-window-seconds
-  Number of seconds during which disable-num-failures failures must
+  Number of seconds during which retry_count failures must
   occur in order for an automatic disable by the scheduler. The
   scheduler forgets about disables that have occurred longer ago than
   this amount of time. Defaults to 3600 (1 hour).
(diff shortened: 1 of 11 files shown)


doc/configuration.rst
```diff
@@ -731,3 +731,44 @@ user
...
+
++------------------------+-----------+
+| Config                 | Section   |
++========================+===========+
+| retry_count            | scheduler |
++------------------------+-----------+
+| disable_hard_timeout   | scheduler |
++------------------------+-----------+
+| disable_window_seconds | scheduler |
++------------------------+-----------+
```
Tarrasch: Actually we shouldn't have this information here as it's already somewhere else in the documentation.

luigi/scheduler.py
```diff
@@ -656,8 +668,11 @@ def add_task(self, task_id=None, status=PENDING, runnable=True,
 
             # Task dependencies might not exist yet. Let's create dummy tasks for them for now.
             # Otherwise the task dependencies might end up being pruned if scheduling takes a long time
+            deps_retry_policy_dicts = _get_default(deps_retry_policy_dicts, {})
```
Tarrasch: What's `_get_default`??? Is it from global scope?

luigi/worker.py
```diff
@@ -643,7 +649,9 @@ def _add(self, task, is_complete):
                        resources=task.process_resources(),
                        params=task.to_str_params(),
                        family=task.task_family,
-                       module=task.task_module)
+                       module=task.task_module,
+                       retry_policy_dict=_get_retry_policy_dict(task),
+                       deps_retry_policy_dicts=deps_retry_policy_dicts)
```
Tarrasch: As I've asked you in every iteration before. Why do we need the `deps_retry_policy_dicts`?

luigi/scheduler.py
```diff
@@ -108,12 +119,12 @@ class scheduler(Config):
...
 
-    # Jobs are disabled if we see more than disable_failures failures in disable_window seconds.
+    # Jobs are disabled if we see more than retry_count failures in disable_window seconds.
     # These disables last for disable_persist seconds.
     disable_window = parameter.IntParameter(default=3600,
                                             config_path=dict(section='scheduler', name='disable-window-seconds'))
-    disable_failures = parameter.IntParameter(default=999999999,
-                                              config_path=dict(section='scheduler', name='disable-num-failures'))
+    retry_count = parameter.IntParameter(default=999999999,
+                                         config_path=dict(section='scheduler', name='disable-num-failures'))
```
Tarrasch: this is still wrong....

test/scheduler_test.py
```diff
@@ -110,3 +110,122 @@ def __init__(self):
...
+
+        # Task 9 which is disable-failures 1 and its has_excessive_failures method returns False under 44
+        for i in range(10):
+            task_9.add_failure()
+        self.assertFalse(task_9.has_excessive_failures())
+        task_9.add_failure()
+        self.assertTrue(task_9.has_excessive_failures())
+
+
+if __name__ == '__main__':
```
Tarrasch: remove this

Tarrasch:
Can you also add a test case with with dynamic dependencies? It should be super easy, and it immediately shows if it's necessary to thread along the `deps_retry_policy_dicts` var or not.

luigi/worker.py
javrasya: I saw that dependency tasks are added for the first time with their parent task and it is cashed at this point. Even you send its retry policy to its own task adding method，it won't be updated because it is cached.

doc/configuration.rst
javrasya: I think we should include a part about what is in retry policy and what they can use per task，what do you think about it?

luigi/scheduler.py
javrasya: I have never touched the name in config path. But if you say，I will change it.

luigi/scheduler.py
javrasya: Yes, it is global and luigi scheduler module has that method already and I just used it to get default value of a variable if it is `None`

javrasya:
@Tarrasch, I added multiple test cases in `PerTaskRetryPolicyBehaviorTest` in `worker_test`. You can see `deps_retry_policy` is important by removing the part below and run `PerTaskRetryPolicyBehaviorTest` again. 

`scheduler.py -> add_task()`
```
retry_policy=self._generate_retry_policy(deps_retry_policy_dicts.get(dep))
```


This is  because of luigi task adding order. Luigi adds a dependency task indirectly with its parent task first. So we need to send `deps_retry_policy` of its parent task unless luigi adds dependency task directly instead of indirectly with its parent task.

test/worker_test.py
```diff
@@ -1296,3 +1298,358 @@ def store_task(t, error_msg):
         w._handle_next_task()
 
         self.assertEqual(result, [task])
+
+
+class PerTaskRetryPolicyBehaviorTest(unittest.TestCase):
+    def setUp(self):
+        self.per_task_retry_count = 2
+        self.default_retry_count = 1
+        self.sch = Scheduler(retry_delay=0.1, retry_count=self.default_retry_count, prune_on_get_work=True, record_task_history=False)
```
Tarrasch: You don't need to be so explicit and say `record_task_history=False`. But it's ok. :)

test/worker_test.py
```diff
@@ -1296,3 +1298,358 @@ def store_task(t, error_msg):
         w._handle_next_task()
 
         self.assertEqual(result, [task])
+
+
+class PerTaskRetryPolicyBehaviorTest(unittest.TestCase):
```
Tarrasch: Inherit form `LuigiTestCase` instead of  `unittest.TestCase`, it will handle the cleanup of the classes that you've introduced in every test case.

test/worker_test.py
```diff
@@ -1296,3 +1298,358 @@ def store_task(t, error_msg):
...
+    def setUp(self):
+        self.per_task_retry_count = 2
+        self.default_retry_count = 1
+        self.sch = Scheduler(retry_delay=0.1, retry_count=self.default_retry_count, prune_on_get_work=True, record_task_history=False)
+
+        self.test_instance_id = str(uuid4())
```
Tarrasch: This mechanism can be removed when you use `LuigiTestCase`

test/worker_test.py
```diff
@@ -1296,3 +1298,358 @@ def store_task(t, error_msg):
...
+        a = A()
+
+        with Worker(scheduler=self.sch, worker_id='X', keep_alive=True, wait_interval=0.1) as w1:
+            self.assertTrue(w1.add(a))
+
+            self.assertFalse(w1.run())
+
+            self.assertEqual([a.task_id], list(self.sch.task_list('PENDING', 'UPSTREAM_DISABLED').keys()))
+
+            self.assertEqual(sorted([b.task_id, c.task_id]), sorted(self.sch.task_list('DISABLED', '').keys()))
```
Tarrasch: Nitpick: Using `set` vs `sorted` is more closely representing our intent. :)

test/worker_test.py
```diff
@@ -1296,3 +1298,358 @@ def store_task(t, error_msg):
...
+            Task C has default retry_count which is 1, but Task B has retry_count at task level as 2.
+
+            This test is running on single worker
+        """
+        class C(luigi.Task):
+            task_namespace = self.test_instance_id
+
+            retry_index = 0
+
+            def run(self):
```
Tarrasch: Can you make your test cases shorter by having a task you inherit from. I've seen the `self.retry_index += 1` repeated many many times in this patch :)

luigi/scheduler.py
```diff
@@ -108,12 +119,12 @@ class scheduler(Config):
...
 
-    # Jobs are disabled if we see more than disable_failures failures in disable_window seconds.
+    # Jobs are disabled if we see more than retry_count failures in disable_window seconds.
     # These disables last for disable_persist seconds.
     disable_window = parameter.IntParameter(default=3600,
                                             config_path=dict(section='scheduler', name='disable-window-seconds'))
-    disable_failures = parameter.IntParameter(default=999999999,
-                                              config_path=dict(section='scheduler', name='disable-num-failures'))
+    retry_count = parameter.IntParameter(default=999999999,
+                                         config_path=dict(section='scheduler', name='disable-failures'))
```
Tarrasch: almost, `name='disable_failures'` :)

Tarrasch:
@javrasya, I still think `deps_retry_policy` can be removed. To make it clearer how, I created javrasya/luigi#1 :)

test/worker_test.py
javrasya: Even I use `LuigiTestCase`, `TaskClassAmbigiousException: Task 'D' is ambiguous` error is thrown in `test_with_dynamic_dependencies_with_single_worker` with tox. Can you help me on that?

test/worker_test.py
javrasya: Sometimes, ordering may change. For example, I run tests in PyCharm and also with tox. Ordering the elements in PyCharm may be different the ordering in tox. This is why I use `sorted`.

Tarrasch:
And this is now finally in!! :)

PR #1793, opened 2016-08-01
Use deps instead of requires.

jslovan:
## Description

Replace calls to _self.requires()_ with _self.deps()_ in the task module.
Handling of dependencies gets inconsistent when one redefines method __requires_.
## Motivation and Context

We have workflow of a daily task with a whole tree of dependencies.  Now, we have just discovered an issue with the task _A_ in the middle of the tree of deps.  We need to recompute it and all of its successors but we want not to touch other fine tasks.

We solved it by the daily job having a taskParameter "startAt" to limit the dependency tree only up to task _A_.  The implementation is in the redefined method __requires_.

We run our jobs with overwritten methods _complete_.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -440,10 +440,12 @@ class Task(object):
         provide consistent end-user experience), yet need to introduce
         (non-input) dependencies.
 
-        Must return an iterable which among others contains the _requires() of
-        the superclass.
+        Must return a structure of Tasks which among others contains
+        the _requires() of the superclass. The return value can be a single
+        Task, a list of Task instances, or a dict whose values are
+        Task instances.
         """
-        return flatten(self.requires())  # base impl
+        return self.requires()  # base impl
 
     def process_resources(self):
         """
@@ -459,10 +461,11 @@ class Task(object):
 
         See :ref:`Task.input`
 
-        :return: a list of :py:class:`Target` objects which are specified as
-                 outputs of all required Tasks.
+        :return: a structure of :py:class:`Target` objects which are specified
+                 as outputs of all required Tasks. Normally, the structure is
+                 the same as that returned by :py:meth:`requires`.
         """
-        return getpaths(self.requires())
+        return getpaths(self._requires())
 
     def deps(self):
         """
@@ -559,7 +562,7 @@ class WrapperTask(Task):
     """
 
     def complete(self):
-        return all(r.complete() for r in flatten(self.requires()))
+        return all(r.complete() for r in self.deps())
 
 
 class Config(Task):
@@ -636,6 +639,6 @@ def flatten_output(task):
     """
     r = flatten(task.output())
     if not r:
-        for dep in flatten(task.requires()):
+        for dep in flatten(task.deps()):
             r += flatten_output(dep)
     return r
--- a/luigi/execution_summary.py
+++ b/luigi/execution_summary.py
@@ -77,7 +77,7 @@ def _depth_first_search(set_tasks, current_task, visited):
         upstream_missing_dependency = False
         upstream_run_by_other_worker = False
         upstream_scheduling_error = False
-        for task in current_task._requires():
+        for task in current_task.deps():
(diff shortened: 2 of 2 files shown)


Tarrasch:
I really don't understand this patch. Is the issue that you overrided `_requires()` in this task? If so, you cold solve this by a simple renaming. Thankfully python supports `__double_underscore()` method names are not overrideable. So you could just rename `_requires()` to `__requires()`.

jslovan:
No, my point is somewhere else.

Task has 3 methods for getting dependencies:
1. _requires_, which is inteded to be overriden,
2. _\_requires_ (internal), intended to manipulate the deps defined in _requires_ dynamically,
3. _deps_ (internal), which are actual dependencies in the defined form.

It seems to me that the Task should use the internal methods in internal affairs - _\_requires_ or _deps_.

Of course, when the tests are passing.

jslovan:
It seems I am not able to present the main idea, let me sum it up again.  Just close this PR immediately if you will not find any new information in it.

1.  There is a valid "business use case" for overriding `_requires()` in order to restrict the dependency tree.
2. When one does that, he must also override `complete()` because current implementation checks all dependencies (got by `requires()`), though one need to check only the actual dependencies (got by `_requires()` or `deps()`.

This PR deals with the item "2" -- no need to override `complete()` or other methods when one overrides `_requires()`.

### Why should I override `_requires()` at all?

Imagine a whole tree of jobs for daily batch data processing.  The root is a wrapper job representing the daily pipeline and is run every day and it imports data and performs analytics processing.

Today, we have discovered a bug in the middle of analytics processing.  We need to recompute the analytics tasks but we don't want to reimport all the source data again.  

This need is IMHO the valid case to override `_requires()`.  We declared a Task parameter `startAt` and overrid `_requires()` to exclude all tasks not affected by the given Task in the dependency tree.
Thus, we don't have to code one-shot tasks to fix data errors every time we discover a bug, effectively duplicating the part of existing task tree.  Is there a better, more proper way we should follow?

luigi/task.py
```diff
@@ -459,10 +461,11 @@ def input(self):
...
         See :ref:`Task.input`
 
-        :return: a list of :py:class:`Target` objects which are specified as
-                 outputs of all required Tasks.
+        :return: a structure of :py:class:`Target` objects which are specified
+                 as outputs of all required Tasks. Normally, the structure is
+                 the same as that returned by :py:meth:`requires`.
         """
-        return getpaths(self.requires())
+        return getpaths(self._requires())
```
dlstadther: If `_requires()` is left as the default, this behavior is unchanged.

luigi/task.py
```diff
@@ -440,10 +440,12 @@ def _requires(self):
...
 
-        Must return an iterable which among others contains the _requires() of
-        the superclass.
+        Must return a structure of Tasks which among others contains
+        the _requires() of the superclass. The return value can be a single
+        Task, a list of Task instances, or a dict whose values are
+        Task instances.
         """
-        return flatten(self.requires())  # base impl
+        return self.requires()  # base impl
```
dlstadther: If someone wants the `flatten`ed version of requirements, they should use `deps()`.

luigi/task.py
```diff
@@ -636,6 +639,6 @@ def flatten_output(task):
     """
     r = flatten(task.output())
     if not r:
-        for dep in flatten(task.requires()):
+        for dep in flatten(task.deps()):
```
dlstadther: Why would you `flatten` an already `flatten`ed struct? Bc this would be something like `flatten(tasks.(flatten(self._requires()))` (incorrect syntax for the purpose of getting my question across).

Tarrasch:
> I don't believe @Tarrasch 's last review negated the validity of your patch. But rather, asked that you PR a couple Task refactors to make this patch more clear (while also cleaning up old code).

Exactly this. @jslovan, is my suggested list of re-factorings something you would be willing to take on?

PR #1795, opened 2016-08-01
Parameterized poll_time for SGEJobTask.

brcopeland:
## Description
I added the --poll-time parameter to SGEJobTask.

## Motivation and Context
This change is helpful in my experience because I run into scaling issues with the default of waiting 5 seconds for every SGEJobTask running.  I frequently have hundreds of these tasks running on a cluster, and so raising the wait time for long-running jobs reduces the overhead.

I have been using this change for several months on my cluster successfully.  We experience better responsiveness on our head node when adjusting the poll time up when we're running my tasks simultaneously.


erikbern:
oops looks like you need to rebase

PR #1798, opened 2016-08-01
Added --dont-remove-tmp-dir parameter to SGEJobTask.

brcopeland:
## Description
Added parameter to retain the temporary directory created to run the Task on the cluster.

## Motivation and Context
This change is helpful for debugging.  Without the change, the temporary directory is always removed, even when, say, it fails it can be helpful to examine the directory to see what was copied, the log files, etc.

I've been using this change without issue for several months.


erikbern:
will merge if you rebase :)

brcopeland:
Rebase done.

PR #1799, opened 2016-08-01
Added NumericalParameter Parameter class that does automatic error ch…

brcopeland:
## Description
Added a new Parameter class, NumericalParameter, that validates that the specified input parameter, an int or float, is in the specified range.  The comparison functions, i.e. <= or < may be overridden, and the min and max values are adjustable.

## Motivation and Context
This Parameter type is helpful to reduce repetitive code, such as repeatedly checking that various input numeric parameters are in the permissible ranges.  I argue it is also cleaner to specify the constraints on the parameter when it is defined in a Task class.

I've been using this class for several weeks successfully.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -29,6 +29,7 @@ from collections import OrderedDict, Mapping
 import operator
 import functools
 from ast import literal_eval
+from sys import maxsize
 
 try:
     from ConfigParser import NoOptionError, NoSectionError
@@ -959,3 +960,30 @@ class TupleParameter(Parameter):
         :param x: the value to serialize.
         """
         return json.dumps(x)
+
+
+class NumericalParameter(Parameter):
+    """Parameter that restricts values to the given numerical type, i.e. int or
+    float, and to the specified range
+    """
+    def __init__(self, var_type=int, min_value=0, max_value=maxint,
+                 left_op=operator.lt, right_op=operator.le, *args, **kwargs):
+        super(NumericalParameter, self).__init__(*args, **kwargs)
+        self.var_type = var_type
+        self.min_value = min_value
+        self.max_value = max_value
+        self.left_op = left_op
+        self.right_op = right_op
+
+    def parse(self, s):
+        value = self.var_type(s)
+        if (self.left_op(self.min_value, value) and
+            self.right_op(value, self.max_value)):
+            return value
+        else:
+            raise ValueError(
+                "{s} is not in the permitted range: {left_endpoint}{min_value} "
+                ", {max_value}{right_endpoint}".format(
+                    s=s, min_value=self.min_value, max_value=self.max_value,
+                    left_endpoint="(" if self.left_op == operator.lt else "[",
+                    right_endpoint="]" if self.right_op == operator.le else ")"))
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -1015,8 +1015,8 @@ class NumericalParameter(Parameter):
             "{var_type} in {left_endpoint}{min_value}, {max_value}{right_endpoint}".format(
                 var_type=var_type.__name__,
                 min_value=min_value, max_value=max_value,
-                left_endpoint="(" if self.left_op == operator.lt else "[",
-                right_endpoint=")" if self.right_op == operator.lt else "]"))
+                left_endpoint="(" if left_op == operator.lt else "[",
+                right_endpoint=")" if right_op == operator.lt else "]"))
         if self.description:
             self.description += " "
         else:
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -963,11 +963,48 @@ class TupleParameter(Parameter):
 
 
 class NumericalParameter(Parameter):
-    """Parameter that restricts values to the given numerical type, i.e. int or
-    float, and to the specified range
+    """
+    Parameter whose value is a number of the specified type, e.g. ``int`` or
+    ``float`` and in the range specified.
+
+    In the task definition, use
+
+    .. code-block:: python
+
+        class MyTask(luigi.Task):
+            my_param = luigi.NumericalParameter(
+                var_type=int, min_value=-3, max_value=7, left_op=operator.le, right_op=operator.lt)
+
+    At the command line, use
+
+    .. code-block:: console
+
+        $ luigi --module my_tasks MyTask --my-param -3
     """
     def __init__(self, var_type=int, min_value=0, max_value=maxint,
                  left_op=operator.lt, right_op=operator.le, *args, **kwargs):
+        """
+        :param function var_type: The type of the input variable, e.g. int or float.
+                                  Default: ``int``.
+        :param min_value: The minimum value permissible in the accepted values
+                          range.  May be inclusive or exclusive based on left_op parameter.
+                          This should be the same type as var_type.
+                          Default: ``0``.
+        :param max_value: The maximum value permissible in the accepted values
+                          range.  May be inclusive or exclusive based on right_op parameter.
+                          This should be the same type as var_type.
+                          Default: ``sys.maxsize``.
+        :param function left_op: The comparison operator for the left-most comparison in
+                                 the expression ``min_value left_op value right_op value``.
+                                 This operator should generally be either
+                                 ``operator.lt`` or ``operator.le``.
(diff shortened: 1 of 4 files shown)


erikbern:
lgtm although would be cool if this works with `--help`

brcopeland:
I fixed the python3 compatibility issue and added the permitted range to --help.  Should be good to go.

erikbern:
awesome – seems like unfortunately the latest merge broke something. can you rebase?

luigi/parameter.py
```diff
@@ -959,3 +960,37 @@ def serialize(self, x):
         :param x: the value to serialize.
         """
         return json.dumps(x)
+
+class NumericalParameter(Parameter):
+    """Parameter that restricts values to the given numerical type, i.e. int or
+    float, and to the specified range
+    """
+    def __init__(self, var_type=int, min_value=0, max_value=maxsize,
+                 left_op=operator.lt, right_op=operator.le, *args, **kwargs):
```
Tarrasch: What is a `left_op`/`right_op`, can you add docs? What values can a user replace them with?

luigi/parameter.py
```diff
@@ -959,3 +960,37 @@ def serialize(self, x):
...
+            "{var_type} in {left_endpoint}{min_value}, {max_value}{right_endpoint}".format(
+                var_type=var_type.__name__,
+                min_value=min_value, max_value=max_value,
+                left_endpoint="(" if self.left_op == operator.lt else "[",
+                right_endpoint=")" if self.right_op == operator.lt else "]"))
+        if self.description:
+            self.description += " "
+        else:
+            self.description = ""
+        self.description += "permitted values: " + self.permitted_range
```
Tarrasch: Can you make all variables private? I guess the exception is `.description` as the parent class will use it, but there's not reason to expose these variables to users of this class.

Tarrasch:
This must have tests before we can merge it.

brcopeland:
Thanks for the feedback guys.  I will look this over and incorporate your suggestions.

brcopeland:
I updated the docstring for the class and \__init__ and added a test suite for the class.  Let me know if the changes are sufficient or if more is needed.

Tarrasch:
This looks *much better*. I still have some minor suggestions relating to comments and tests. After that this can be merged pretty quick I think.

luigi/parameter.py
```diff
@@ -995,3 +996,74 @@ def parse(self, s):
...
+    Parameter whose value is a number of the specified type, e.g. ``int`` or
+    ``float`` and in the range specified.
+
+    In the task definition, use
+
+    .. code-block:: python
+
+        class MyTask(luigi.Task):
+            my_param = luigi.NumericalParameter(
+                var_type=int, min_value=-3, max_value=7, left_op=operator.le, right_op=operator.lt)
```
Tarrasch: What about having two lines here instead, in particular I want to clarify what the default value of `left_op` and `right_op` is:

```python

my_param_1 = luigi.NumericalParameter(
    var_type=int, min_value=-3, max_value=7)  # -3 <= my_param_1 < 7
my_param_2 = luigi.NumericalParameter(
    var_type=int, min_value=-3, max_value=7, left_op=operator.lt, right_op=operator.le) # -3 < my_param_1 <= 7
```

------

Actually, isn't the default values for the operators quite unintuitive? I just would assume the default would be `[, )` or `lb <= x < ub` (I imagine `0 <= x < n` being the common usecase). But it looks like the default is the opposite right???

test/numerical_parameter_test.py
```diff
@@ -0,0 +1,64 @@
...
+# limitations under the License.
+#
+
+from helpers import unittest
+
+import luigi
+from operator import le, lt
+
+
+class NumericalParameterTest(unittest.TestCase):
```
Tarrasch: Thanks for adding the tests. but please also quickly throw together some tests that checks the edge-cases (for both "edges") for when you have not specified given the `left_op` and `right_op` kwargs.

Tarrasch:
Again this looks good. Please just finish the last iteration and we can merge this. :)

luigi/parameter.py
brcopeland: Yeah, I think you're right - this sort of just grew organically from a particular use case I had where the way I implemented were sensible defaults (autoincrement database IDs).  I have no attachment to the defaults as they are so will go with yours, as they're more natural assumptions given the language.

test/numerical_parameter_test.py
```diff
@@ -0,0 +1,72 @@
...
+                                left_op=le, right_op=lt)
+        self.assertRaises(ValueError, lambda: d.parse(7))
+
+    def test_float_max_value_exclusive(self):
+        d = luigi.NumericalParameter(var_type=float, min_value=-3, max_value=7,
+                                left_op=le, right_op=lt)
+        self.assertRaises(ValueError, lambda: d.parse(7))
+
+    def test_defaults_start_range(self):
+        d = luigi.NumericalParameter()
```
Tarrasch: Seeing this, I really feel like we have to make `min_value` and `max_value` not optional. Imagine you read somebody else's code and you see `d = luigi.NumericalParameter()`. My first reaction would maybe be "Why didn't the author just use `IntParameter` or `FloatParameter`?". Perhaps `var_type` should be explicit too, as a reader of the code would probably want to know if the "numerical parameter" is actually an `int` or a `float`.

Tarrasch:
What do you think? Should we make `var_type`, `min_value` and `max_value` not optional? There are no defaults that is given, hence a reader of luigi code will be confused.

test/numerical_parameter_test.py
brcopeland: Well, `min_value` having a default of 0 means there is the advantage of excluding negative integers (if of course the user is relying on that).  But yes, point noted that something implicit like this is fine for me, since I wrote the function to test it, but this is definitely unclear.  I'll make these changes.

brcopeland:
I agree with your logic, so I've made var_type, min_value, and max_value all required parameters and added tests to confirm that.

Tarrasch:
Thanks! As soon as the build is green we could merge this I believe. By the way, you know that you can run `flake8` [locally](https://github.com/spotify/luigi/blob/master/CONTRIBUTING.rst) right?

brcopeland:
I've never used tox before (installed it today) and had some issues getting it working (I don't have sudo on the machine I've been working on).

Tarrasch:
Ah that's too bad. Perhaps installing tox inside a virtualenv would be possible. I have no idea if it works though :)

brcopeland:
Fixed a minor indenting issue that caused a flake8 failure; got tox working and reran the flake8 test successfully, so should be good to go when the tests finish here.

dlstadther:
Wouldn't it be worth adding tests for `NumericalParameter` to mirror tests for existing `Parameter`s? (i.e. [hashability](https://github.com/spotify/luigi/blob/master/test/parameter_test.py#L289), [serialize/parse](https://github.com/spotify/luigi/blob/master/test/parameter_test.py#L278-L286), command line usage, etc.)

Question: how does one specify a `NumericalParameter` in their `luigi.cfg`?

brcopeland:
I added hashability and serialize/parse tests that work properly.  I haven't personally used a `luigi.cfg` before, but I assume it would be like any other parameter, e.g. `MyNumericalParameter=5`.  Any reason to think otherwise?

Tarrasch:
> Any reason to think otherwise?

I don't think so. But you should add a test to make sure. :)

brcopeland:
Added a test for that.

dlstadther:
Awesome!

Why aren't the contents of `numerical_parameter_test.py` just in `parameter_test.py`?

brcopeland:
There is a separate script there for DictParameter tests. It wasn't clear to me where to put the tests.

Tarrasch:
Yea I think doing whichever is fine. I have a slight preference for `numerical_parameter_test.py` as it reduces chances of conflicts and I easier can find and run tests related to a particular parameter type.

Tarrasch:
Thanks! No more do I need to write code like this:

```
class AppIdParameter(luigi.IntParameter):
    def parse(self, x):
        assert 10000 <= int(x) < 20000
        return super().parse(x)
```

brcopeland:
Sure, glad it helped. It'll also help me not have to maintain my own changes upon updating.

PR #1800, opened 2016-08-01
Add ChoiceParameter to restrict parameter options to those specified.

brcopeland:
## Description
As in description.

## Motivation and Context
This allows the user to define the permissible values for a parameter and the value type, and reject any parameter that does not meet these criteria.

I've been using the class for several weeks without issue.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -1046,9 +1046,12 @@ class ChoiceParameter(Parameter):
         super(ChoiceParameter, self).__init__(*args, **kwargs)
         self.choices = set(var_type(choice) for choice in choices)
         self.var_type = var_type
-        self.description += ("(" + ",".join(str(choice) for choice in choices)
-                             + ")")
-        print(self.description)
+        if self.description:
+            self.description += " "
+        else:
+            self.description = ""
+        self.description += (
+            "Choices: {" + ",".join(str(choice) for choice in choices) + "}")
 
     def parse(self, s):
         var = self.var_type(s)
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -1061,6 +1061,13 @@ class ChoiceParameter(Parameter):
     desired.
     """
     def __init__(self, var_type=str, *args, **kwargs):
+        """
+        :param function var_type: The type of the input variable, e.g. str, int,
+                                  float, etc.
+                                  Default: str
+        :param choices: An iterable, all of whose elements are of `var_type` to
+                        restrict parameter choices to.
+        """
         if "choices" not in kwargs:
             raise ParameterException("A choices iterable must be specified")
         self._choices = set(kwargs.pop("choices"))
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -1034,3 +1034,23 @@ class NumericalParameter(Parameter):
             raise ValueError(
                 "{s} is not in the set of {permitted_range}".format(
                     s=s, permitted_range=self._permitted_range))
+
+
+class ChoiceParameter(Parameter):
+    """Parameter that restricts options to the specified set
+    All choices must be of the same type that so that parse can
+    correctly infer type
+    Variable type is specified with var_type=...
+    """
+    def __init__(self, choices, var_type=str, *args, **kwargs):
+        super(ChoiceParameter, self).__init__(*args, **kwargs)
+        self.choices = set(var_type(choice) for choice in choices)
+        self.var_type = var_type
+
+    def parse(self, s):
+        var = self.var_type(s)
+        if var in self.choices:
+            return var
+        else:
+            raise ValueError("{s} is not a valid choice from {choices}".format(
+                s=s, choices=self.choices))
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -1063,8 +1063,9 @@ class ChoiceParameter(Parameter):
     def __init__(self, var_type=str, *args, **kwargs):
         if "choices" not in kwargs:
             raise ParameterException("A choices iterable must be specified")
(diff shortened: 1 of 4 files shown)


erikbern:
lgtm although would be cool if this works with `--help`

brcopeland:
@erikbern, can you give me an example of what you mean?  I'd be happy to make the change(s).  When I create a Task using a ChoiceParameter, the parameter and its description show up when I run `luigi --module <module> <Task> --help`.

erikbern:
would be cool if the `--help` also shows what options are available. but no biggie :)

luigi/parameter.py
```diff
@@ -959,3 +959,21 @@ def serialize(self, x):
...
+        super(ChoiceParameter, self).__init__(*args, **kwargs)
+        self.choices = set(var_type(choice) for choice in choices)
+        self.var_type = var_type
+
+    def parse(self, s):
+        var = self.var_type(s)
+        if var in self.choices:
+            return var
+        else:
+            raise ValueError("{s} is not a valid choice from {choices}".format(
```
erikbern: this line is broken – can you fix

brcopeland:
Ah, yeah.  I'm trying to figure out a way to do that - updating self.description in __init__ does not appear to be sufficient.

luigi/parameter.py
brcopeland: Yes, fixed.  Don't know how I didn't copy that last line.

brcopeland:
Ok, got it working.  My system has a luigi install I was importing instead of the one I was editing.

erikbern:
awesome! flake8 is complaining: https://travis-ci.org/spotify/luigi/jobs/149050059

Tarrasch:
Can you also add test cases please?

luigi/parameter.py
```diff
@@ -959,3 +959,28 @@ def serialize(self, x):
...
+
+class ChoiceParameter(Parameter):
+    """Parameter that restricts options to the specified set
+    All choices must be of the same type that so that parse can
+    correctly infer type
+    Variable type is specified with var_type=...
+    """
+    def __init__(self, choices, var_type=str, *args, **kwargs):
+        super(ChoiceParameter, self).__init__(*args, **kwargs)
+        self.choices = set(var_type(choice) for choice in choices)
```
Tarrasch: Can you make this into an assertion? I imagine people creating new `Parameter`s know enough luigi+python that they would rather be explicit in types over than luigi magically converting stuff behind their back.

luigi/parameter.py
```diff
@@ -959,3 +959,28 @@ def serialize(self, x):
         :param x: the value to serialize.
         """
         return json.dumps(x)
+
+class ChoiceParameter(Parameter):
+    """Parameter that restricts options to the specified set
+    All choices must be of the same type that so that parse can
+    correctly infer type
+    Variable type is specified with var_type=...
```
Tarrasch: Please format all docs and make sure it renders nice. See CONTRIBUTING.md and other examples in this file.

Tarrasch:
This class is somewhat similar to `EnumParameter`. Would you mind interlinking those two in the docs? Maybe add something like "If you want a more typed and structured version of a ChoiceParameter, you should consider using an EnumParameter instead." could guide our users to pick the right choice?

brcopeland:
You could be entirely right here.  I am a pretty new user of luigi and may have overlooked this class before coming up with my own version.  I will clean it up if requested, but maybe at this point it just makes more sense to not include this.

Tarrasch:
I still think this makes sense. Not everyone likes to create a new Enum type for every input-restriction and add all the conversion overhead if they want to eventually go back to a string type. I would be happy to include this, as long as the docs comes with recommended good practices. Like we've previously  done [here](https://github.com/ehdr/luigi/blob/f771622fe2b8cca39b1165c13a47f444f7c46b03/luigi/contrib/external_program.py#L24-L29) for example. :)

brcopeland:
I changed the docstring to I believe the proper format and added a test suite to the test/ directory.  Please let me know if any further changes would be welcome.

luigi/parameter.py
```diff
@@ -995,3 +995,49 @@ def parse(self, s):
...
+
+    Consider using :class:`~luigi.EnumParameter` for a typed, structured
+    alternative.  This class can perform the same role when all choices are the
+    same type and transparency of parameter value on the command line is
+    desired.
+    """
+    def __init__(self, var_type=str, *args, **kwargs):
+        if "choices" not in kwargs:
+            raise ParameterException("A choices iterable must be specified")
+        self._choices = set(var_type(choice) for choice in kwargs.pop("choices"))
```
Tarrasch: As I pointed out [earlier](https://github.com/spotify/luigi/pull/1800#discussion_r73090575), I probably would prefer if we could somehow would assert instead of convert. But I'm not sure really what's best :)

brcopeland: Oh, ok sure I think that's perfectly reasonable.

Tarrasch:
Apart from the small comment about assert/convert. This looks *very good*. I must say I'm very impressed how you managed to fix all the shortcomings in a single review turnaround! :)

brcopeland:
Made a couple changes to assert the choices are the proper type as you suggest.  And sure, glad it's helpful.

brcopeland:
I rebased, updated docstrings, and added unit tests for the class.

Tarrasch:
Awesomeness! :)

Tarrasch:
A post-merge note. Some users might want to use the ["Functional API"](https://docs.python.org/3/library/enum.html#functional-api) of enums. Instead of 

    ChoiceParameter(choices=['a', 'b', 'c'])

Some users might prefer

    EnumParameter(enum=Enum('Character', 'a b c'))

PR #1801, opened 2016-08-01
Added InputFileParameter and OutputFileParameter classes.

brcopeland:
## Description
1. Added class, InputFileParameter, that parses to the real path of the parameter if it's a file, and raises an OSError otherwise.
2. Added class, OutputFileParameter, that parses to the real path of the parameter if it's a valid file path, creates the required output directories if necessary, and raises an OSError if it can't do this successfully.

## Motivation and Context
I commonly create luigi Tasks that take input and output file parameters.  It is helpful to me, and I expect to others, to restrict a parameter to existing files/valid paths (for output files).  This also has the benefit of converting a given relative path to the absolute path, which is useful in my experience when the working directory changes, as it does with, for example, SGEJobTasks.

I have been using these two classes successfully for several weeks.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -961,6 +961,7 @@ class TupleParameter(Parameter):
         """
         return json.dumps(x)
 
+
 class InputFileParameter(Parameter):
     """
     Paramater whose value is an existing file.
@@ -974,6 +975,7 @@ class InputFileParameter(Parameter):
             return os.path.realpath(s)
         else:
             raise OSError("{s} does not exist".format(s=s))
+
 
 class OutputFileParameter(Parameter):
     """
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -29,6 +29,7 @@ from collections import OrderedDict, Mapping
 import operator
 import functools
 from ast import literal_eval
+import os
 
 try:
     from ConfigParser import NoOptionError, NoSectionError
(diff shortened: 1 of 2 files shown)


Tarrasch:
Hey @erikbern I think you should not have merged this. For changes to luigi-core changes, could you at least let the PR be open for 24 hours (and also during a workday)?

@erikbern, Besides also please start using the merge+squash feature. I recall you agreed to do it before, so you just need to click one button and get started.

luigi/parameter.py
```diff
@@ -959,3 +960,38 @@ def serialize(self, x):
...
+        """
+        Returns the real path to the file and attempts to create any directories
+        needed.
+        """
+        s = os.path.realpath(s)
+        if not os.path.isdir(os.path.dirname(s)):
+            try:
+                os.makedirs(os.path.dirname(s))
+            except OSError:
+                raise OSError("could not make parent directories for {s}".format(
```
Tarrasch: I don't think this reraising adds any value. It's just makes the stack trace longer and nested.

Tarrasch:
@brcopeland, can you follow up this PR with test cases and fixing all the inline comments I will add?

erikbern:
sorry. i thought it was fine since the PR was pretty tiny and didn't touch any existing code. but you are right

luigi/parameter.py
```diff
@@ -959,3 +960,38 @@ def serialize(self, x):
         :param x: the value to serialize.
         """
         return json.dumps(x)
+
+
+class InputFileParameter(Parameter):
+    """
+    Paramater whose value is an existing file.
```
Tarrasch: Please clarify to "existing file on the local file system"

luigi/parameter.py
```diff
@@ -959,3 +960,38 @@ def serialize(self, x):
...
+
+    def parse(self, s):
+        """
+        Returns the real path to the file and attempts to create any directories
+        needed.
+        """
+        s = os.path.realpath(s)
+        if not os.path.isdir(os.path.dirname(s)):
+            try:
+                os.makedirs(os.path.dirname(s))
```
Tarrasch: I'm pretty sure this will cause way more damage than the little convenience that it adds. Can you instead just check that the directory exists and throw otherwise?

Tarrasch:
Actually. I think we can not keep these two parameters as they are now. Since the `parse` function performs IO, and we will have weird failures where the file exists on the node submitting the task but some other task getting the task from the server will crash upon receiving the task.

-----

Furthermore it feels like these checks makes no sense given that luigi is all about having dependencies resolved. I mean consider this example:

```
MyTask(Task):
    input_file = InputFileParameter('some_path')

    requires(self):
         return TaskThatWillCreateTheFile(output_file=self.input_file)

    run(self):
         # code using self.input_file

dlstadther:
I support the revert in order to spend more time cleaning this and making it luigi-friendly (plus including tests where applicable - I remember getting bitten by my `Parameter` PRs).

Tarrasch:
Okay. I'll revert it tomorrow then unless I hear compelling arguments against it. :)

PR #1803, opened 2016-08-02
Revert "Added InputFileParameter and OutputFileParameter classes."

Tarrasch:
Reverts spotify/luigi#1801

### DIFF
--- a/luigi/__init__.py
+++ b/luigi/__init__.py
@@ -35,8 +35,7 @@ from luigi.parameter import (
     DateParameter, MonthParameter, YearParameter, DateHourParameter, DateMinuteParameter, DateSecondParameter,
     DateIntervalParameter, TimeDeltaParameter,
     IntParameter, FloatParameter, BooleanParameter, BoolParameter,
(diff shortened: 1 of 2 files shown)


brcopeland:
I suppose these could be problematic for systems that don't have access to the same files.  Personally though I have found them helpful to reduce frequent boilerplate code.  Anything occur to you to make it safer, or is something like this just not something you think should be a standard feature?

Tarrasch:
I revert for now. But I'm still as open as over to merging an improved version of this. In addition to my comments in #1801, I think we need to introduce some mechanism in addition to `Parameter.parse()`, something like `Parameter.parse_from_command_line`. The idea is that it has a default implementation to just call out to `parse`. Meanwhile `parse` may only perform pure logic `prase_from_command_line` could do all kinds of conveniences for the user, in this case, it could convert a relative path to an absolute path.

Technically, this [call to parse](https://github.com/spotify/luigi/blob/master/luigi/cmdline_parser.py#L133) should be call `parse_from_command_line` instead.

As a practical advice to I would first focus on getting your other two less controversial Parameter-related patches merged first and then continue on this. :)

Issue #1615, opened 2016-03-30
Do not show non-significant parameters in history


fjavieralba:
I've configured luigi to store history in a SQLite.

When I access my task history at `luigi-scheduler:8082/history`, the interface is showing non-significant parameters such as passwords etc.

It would be great to not store that info in the DB, or at least avoid showing them in the GUI.

I'm using the latest stable release (Version 2.0.1)

erikbern:
sounds like a good idea – can you send a pull request? thanks!

fjavieralba:
Sure, will look into it :)

mfcabrera:
Closing this as there is already many PRs dealing with this and there is a decision to be made regarding how to handle this properly. See https://github.com/spotify/luigi/pull/1892/files#diff-f5f076fb9da1107cb1221ae176bbc955R91

PR #1816, opened 2016-08-12
Worker now passes only significant parameters to the scheduler.

impguard:
The scheduler now only receives the significant parameters when a task is sent to it by the worker.

This design currently assumes that [http://luigi.readthedocs.io/en/stable/api/luigi.parameter.html?highlight=significant#luigi.parameter.Parameter](the parameter documentation) still holds true, and that insignificant parameters still indicate sensitive information that should not be exposed in addition to parameters that should not make up the task_id.
## Description

Simply not passing significant parameters to the scheduler from the worker.
## Motivation and Context

These changes are based off an internal use case of ours to protect passwords/access keys from being exposed via the web GUI and history API. These keys are parameters to the task but should not be exposed because the web GUI and API isn't locked behind authentication.

It resolves issue #1615.

Ran jobs with code and the history API and the Web GUI does not expose the information anymore. Most of the tests seemed to pass bar the server_test which I presumed needed a server up and running to ping and I presumed the failures was an environmental issue.


Tarrasch:
Can you also add a test case?

impguard:
Yea sure, I'll try spinning something up tomorrow. Mind shooting me in the right direction of where you think it should belong to save me some time? It's an integration point between the scheduler/luigi client.

Tarrasch:
I would run commands like `ag significant test` and `ag Worker test` (if you don't use `ag` use `git grep`) to find various existing examples and copy the parts you find necessary. I really don't know an exact location.

-----

Also. If the docs doesn't already mention it, you should mention clearly that insignificant parameters can be used for secrets only to be known by the worker(s).

ifiddes:
This would be a great PR to have integrated, as I also dislike the disconnect between the logs hiding the insignificant parameters and the web GUI displaying them.

It would be even nicer if overriding `__repr__` worked for the web GUI as it does for the logs. Would that work here? Is there anything I can do to help?

impguard:
Haven't looked at __repr__. Our team has a task to write up some tests and push it here, but it most likely won't be handled until our next couple of sprints. 

If you would like to help out it would be greatly appreciated!

fjavieralba:
It would be great to have this PR merged!

I've made some local tests and it solves #1615   💯 

thanks @impguard

impguard:
Yea, still haven't gotten more time to create a simple test, might be able to get to it this or next weekend.

If you have some local tests, go ahead and push them up!

fjavieralba:
Hi!
I don't have any specific tests apart from testing my own luigi workflows with this codebase and running tox tests.

Is anything else we should be testing here?
do you think is possible to merge this already? 
It would be great, since I think it's an important security issue to expose non-significant parameters such as passwords etc on the luigi GUI

thanks!!

PR #1817, opened 2016-08-16
Adds BATCH_RUNNING to the visualiser

daveFNbuck:
## Description
In the earlier batching PR, I left out any visualiser elements. This
adds BATCH_RUNNING to the Task List selectors and Dependency Graph
legend. The legend needed to be widened a bit to accomodate the longer
name, and I replace commas with spaces while capitalizing the first
letter of each word in the legend.

## Motivation and Context
I realized while working to merge the latest trunk with my fork that my visualizer code was left out.

I've been using this for a few months now with the batch running backend from the dropped PR. This part should be the same, as it's just about the statuses.

### DIFF
--- a/luigi/static/visualiser/index.html
+++ b/luigi/static/visualiser/index.html
@@ -490,6 +490,16 @@
                   </div>
 
                   <div class="col-md-3 col-sm-6 col-xs-12">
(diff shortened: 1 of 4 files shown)


Tarrasch:
LGTM

Issue #1765, opened 2016-07-14
TypeError: can't pickle function objects with v2.2.0 PySparkTask


alextmeyer:
EDIT: (Similar issue to: https://github.com/spotify/luigi/issues/1714 but simplified and directly related to PySparkTask on 2.2.0)

Running simple module as a test (after running into this issue with a separate PySparkTask on a recently spun up EMR cluster with the newest version of Luigi installed - 2.2.0). Only running into this issue on Luigi==2.2.0, tested and works fine every version up until 2.2.0: 

Simple module:
```
import luigi
from luigi.contrib.spark import PySparkTask

class MyTask(PySparkTask):
    x = luigi.IntParameter()
    y = luigi.IntParameter(default=45)

    def main(self, sc, *args):
        sc.parallelize([self.x, self.y])
```

Continually running into the following error and the task fails out: 
```
ERROR: [pid 25382] Worker Worker(salt=572741934, workers=1, host=ip-<<>>, username=hadoop, pid=25382) failed    MyTask(x=123, y=456)
Traceback (most recent call last):
  File "/home/hadoop/venv/local/lib/python2.7/site-packages/luigi/worker.py", line 177, in run
    new_deps = self._run_get_new_deps()
  File "/home/hadoop/venv/local/lib/python2.7/site-packages/luigi/worker.py", line 115, in _run_get_new_deps
    task_gen = self.task.run()
  File "/home/hadoop/venv/local/lib/python2.7/site-packages/luigi/contrib/spark.py", line 281, in run
    self._dump(fd)
  File "/home/hadoop/venv/local/lib/python2.7/site-packages/luigi/contrib/spark.py", line 294, in _dump
    pickle.dump(self, fd)
  File "/home/hadoop/venv/lib64/python2.7/copy_reg.py", line 70, in _reduce_ex
    raise TypeError, "can't pickle %s objects" % base.__name__
TypeError: can't pickle function objects
```

(Not sure if there were any significant changes in PysparkTask or potentially change in pickle version being used under the hood)

**Steps to reproduce (on EMR cluster):**
```
virtualenv -p /usr/bin/python2.7 venv
source venv/bin/activate
pip install --upgrade pip
pip install luigi # (this works when specify any version 2.0.1 to 2.1.0)
# TO FIX: pip install --upgrade 'luigi==2.1.1'
luigi --module my_module MyTask --x 123 --y 456 --local-scheduler
```

Tarrasch:
[...]

Issue #1714, opened 2016-06-08
can't pickle function objects with SGEJobTask


mmendez12:
I've been using luigi.Task to run jobs on a SGE queuing system without any problem but when I try the example from the SGEJobTask documentation I get the following error:

```
Traceback (most recent call last):
  File "/home/toto/anaconda2/lib/python2.7/site-packages/luigi-2.1.1-py2.7.egg/luigi/worker.py", l
ine 191, in run
    new_deps = self._run_get_new_deps()
  File "/home/toto/anaconda2/lib/python2.7/site-packages/luigi-2.1.1-py2.7.egg/luigi/worker.py", l
ine 129, in _run_get_new_deps
    task_gen = self.task.run()
  File "/home/toto/anaconda2/lib/python2.7/site-packages/luigi-2.1.1-py2.7.egg/luigi/contrib/sge.p
y", line 214, in run
    self._init_local()
  File "/home/toto/anaconda2/lib/python2.7/site-packages/luigi-2.1.1-py2.7.egg/luigi/contrib/sge.p
y", line 205, in _init_local
    self._dump(self.tmp_dir)
  File "/home/toto/anaconda2/lib/python2.7/site-packages/luigi-2.1.1-py2.7.egg/luigi/contrib/sge.p
y", line 232, in _dump
    d = pickle.dumps(self)
  File "/home/toto/anaconda2/lib/python2.7/copy_reg.py", line 70, in _reduce_ex
    raise TypeError, "can't pickle %s objects" % base.__name__
TypeError: can't pickle function objects
```

Here is the code I tried to run on a cluster, it is not exactly what's on the documentation but both code returned the same error above:

``` python
import luigi
from luigi.contrib.sge import SGEJobTask

class TestJobTask(SGEJobTask):

    i = luigi.Parameter()

    def work(self):

        with open(self.output().path, 'w') as f:
            f.write('this is a test')

    def output(self):
        return luigi.LocalTarget('./testfile_' + str(self.i))

if __name__ == '__main__':
    tasks = [TestJobTask(i=str(i), shared_tmp_dir='./') for i in range(3)]
    luigi.build(tasks, local_scheduler=True, workers=3)
```

I checked the source code but I am not familiar with pickle so I am not sure what is going wrong here.

Does anyone know how to fix this? maybe @jfeala ?

Tarrasch:
Hmm, maybe you can @-ping the authors of the SGE module?

mmendez12:
[...]

PR #1819, opened 2016-08-17
Bug fix for spark task can not be pickled

ivannotes:
This PR is addressing https://github.com/spotify/luigi/issues/1765 

## Description
The change of this PR is trying to set problematic methods to None before dump and resume them
after dump which is consistent to the behavior of hadoop job which you can find [here](https://github.com/spotify/luigi/blob/master/luigi/contrib/hadoop.py#L947)

## Motivation and Context
https://github.com/spotify/luigi/issues/1765 

I ran this code with my own job and it works, also travis build is passed you can find it here https://travis-ci.org/ivannotes/luigi/builds/152888222

let me know it you need more information.

### DIFF
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -944,7 +944,7 @@ class JobTask(BaseHadoopJobTask):
         """
         Dump instance to file.
         """
-        with self._without_unpicklable_properties():
+        with self.no_unpicklable_properties():
             file_name = os.path.join(directory, 'job-instance.pickle')
             if self.__module__ == '__main__':
                 d = pickle.dumps(self)
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -944,23 +944,16 @@ class JobTask(BaseHadoopJobTask):
         """
         Dump instance to file.
         """
-        _set_tracking_url = self.set_tracking_url
-        self.set_tracking_url = None
-        _set_status_message = self.set_status_message
-        self.set_status_message = None
+        with self._without_unpicklable_properties():
+            file_name = os.path.join(directory, 'job-instance.pickle')
+            if self.__module__ == '__main__':
+                d = pickle.dumps(self)
+                module_name = os.path.basename(sys.argv[0]).rsplit('.', 1)[0]
+                d = d.replace(b'(c__main__', "(c" + module_name)
+                open(file_name, "wb").write(d)
 
-        file_name = os.path.join(directory, 'job-instance.pickle')
-        if self.__module__ == '__main__':
-            d = pickle.dumps(self)
-            module_name = os.path.basename(sys.argv[0]).rsplit('.', 1)[0]
-            d = d.replace(b'(c__main__', "(c" + module_name)
-            open(file_name, "wb").write(d)
-
-        else:
-            pickle.dump(self, open(file_name, "wb"))
-
-        self.set_tracking_url = _set_tracking_url
-        self.set_status_message = _set_status_message
+            else:
+                pickle.dump(self, open(file_name, "wb"))
 
     def _map_input(self, input_stream):
         """
--- a/luigi/contrib/sge.py
+++ b/luigi/contrib/sge.py
@@ -203,6 +203,7 @@ class SGEJobTask(luigi.Task):
         description="don't tarball (and extract) the luigi project files")
 
     def __init__(self, *args, **kwargs):
+        super(SGEJobTask, self).__init__(*args, **kwargs)
         if self.job_name:
             # use explicitly provided job name
             pass
@@ -270,15 +271,15 @@ class SGEJobTask(luigi.Task):
 
     def _dump(self, out_dir=''):
         """Dump instance to file."""
-        self.job_file = os.path.join(out_dir, 'job-instance.pickle')
-        if self.__module__ == '__main__':
-            d = pickle.dumps(self)
-            module_name = os.path.basename(sys.argv[0]).rsplit('.', 1)[0]
-            d = d.replace('(c__main__', "(c" + module_name)
-            open(self.job_file, "w").write(d)
-
-        else:
-            pickle.dump(self, open(self.job_file, "w"))
+        with self._without_unpicklable_properties():
+            self.job_file = os.path.join(out_dir, 'job-instance.pickle')
+            if self.__module__ == '__main__':
+                d = pickle.dumps(self)
+                module_name = os.path.basename(sys.argv[0]).rsplit('.', 1)[0]
+                d = d.replace('(c__main__', "(c" + module_name)
+                open(self.job_file, "w").write(d)
+            else:
+                pickle.dump(self, open(self.job_file, "w"))
 
     def _run_job(self):
 
--- a/luigi/contrib/sge.py
+++ b/luigi/contrib/sge.py
@@ -271,7 +271,7 @@ class SGEJobTask(luigi.Task):
 
     def _dump(self, out_dir=''):
         """Dump instance to file."""
-        with self._without_unpicklable_properties():
+        with self.no_unpicklable_properties():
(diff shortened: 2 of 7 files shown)


Tarrasch:
Normally we expect bugfixes to come with a testcase so that the bug is not reintroduced.

ivannotes:
@Tarrasch will work on it later today :)

Tarrasch:
Great. Thanks! Let's let this live for a day or so so others have time to comment. Then this can be merged.

erikbern:
👍

ehdr:
👍

petergaultney:
@ivannotes, thank you for working on this. 

I would suggest that this functionality be moved into task.py as a pair of methods - `_pre_pickle_dump()` and `_post_pickle_dump()`. It would seem that [SGETask has the same issue](https://github.com/spotify/luigi/issues/1714), and obviously HadoopTask did as well. Then, inheriting tasks that need to pickle themselves could simply call the methods before and after the pickling, and there would be a central place to maintain the portions of the base Task that cannot be pickled.

```
def _pre_pickle_dump(self):
    _set_status_message = self.set_status_message
    _set_tracking_url = self.set_tracking_url
    self.set_status_message = None
    self.set_tracking_url = None
    return (_set_status_message, _set_tracking_url)

def _post_pickle_dump(self, unpickleable_tuple):
    self.set_status_message = unpickleable_tuple[0]
    self.set_tracking_url = unpickleable_tuple[1]
```

example usage:

```
    def _dump(self, fd):
        unpickleable_tuple = self._pre_pickle_dump()

        if self.__module__ == '__main__':
            d = pickle.dumps(self)
            module_name = os.path.basename(sys.argv[0]).rsplit('.', 1)[0]
            d = d.replace(b'(c__main__', "(c" + module_name)
            fd.write(d)
        else:
            pickle.dump(self, fd)

       self._post_pickle_dump(unpickleable_tuple)
```

Tarrasch:
@petergaultney, thanks for thinking long-term like this. I really like your suggestion. I would only suggest that  we only define 1 context-manager method instead. You would use it like this 

```python
def _dump(self, fd):
    with self._no_unpickleable_properties():
        if self.__module__ == '__main__':
            d = pickle.dumps(self)
            module_name = os.path.basename(sys.argv[0]).rsplit('.', 1)[0]
            d = d.replace(b'(c__main__', "(c" + module_name)
            fd.write(d)
        else:
            pickle.dump(self, fd)
```

Would that work @petergaultney?

ivannotes:
@Tarrasch @petergaultney I think context manager is a good idea,  I will change it in this PR and apply it to SparkTask. And another question is do we need to change hadoop task and SGETask in this PR ?

Tarrasch:
@ivannotes I wouldn't say it's necessary but it would be nice. :)

petergaultney:
@Tarrasch, yes, that's definitely an improvement on my suggestion. @ivannotes, thanks again for chasing this one down; it will be a big help to us at Eventbrite.

test/contrib/sge_test.py
```diff
@@ -21,9 +21,10 @@
 from glob import glob
 import unittest
 import logging
+from mock import patch
 
 import luigi
-from luigi.contrib.sge import SGEJobTask, _parse_qstat_state, _clean_task_id
```
ivannotes: @Tarrasch it surprised me that this task can not work on my laptop, I found out the `_clean_task_id` is not there anymore so I just remove it in this test case.

FYI the command I ran the test on my lap top is
```
tox -e py-nonhdfs -- -s --tests=test/contrib/sge_test.py
```
let me know if I'm wrong

luigi/contrib/sge.py
```diff
@@ -203,6 +203,7 @@ class SGEJobTask(luigi.Task):
         description="don't tarball (and extract) the luigi project files")
 
     def __init__(self, *args, **kwargs):
+        super(SGEJobTask, self).__init__(*args, **kwargs)
```
ivannotes: @Tarrasch this is another change you need to pay attention, it seems `SGEJobTask` is not calling `super.__init__` which break the test case some how.

test/contrib/sge_test.py
Tarrasch: Oh, I didn't know it was optional to leave out the python version. Cool.

Anyway the command looks sane. You might want to see the COTNRIBUTING file too though.

luigi/task.py
```diff
@@ -550,6 +551,37 @@ def on_success(self):
...
+                    with self._without_unpicklable_properties():
+                        pickle.dumps(self)
+
+        """
+        unpicklable_properties = ('set_tracking_url', 'set_status_message')
+        reserved_properties = {}
+        for property_name in unpicklable_properties:
+            if hasattr(self, property_name):
+                reserved_properties[property_name] = getattr(self, property_name)
+                setattr(self, property_name, None)
```
Tarrasch: Here is an opportunity to save somebody from great debugging frustration I think. What if instead of `None` we have a human-readable value like `placeholder_during_pickling`. That is so easy to `grep` to.

What do you think?

ivannotes: agree, will do this

Tarrasch:
Really cool to see that you so beautifully documented and tested your method. Although it's not strictly necessary for private methods. Thanks for caring about the code quality!

luigi/task.py
```diff
@@ -550,6 +551,37 @@ def on_success(self):
         Default behavior is to send an None value"""
         pass
 
+    @contextmanager
+    def _without_unpicklable_properties(self):
```
Tarrasch: given that the syntax for python's context manager invocation is `with xxx:`, perhaps change `without` to `no`. Contrast

```
with self._without_unpicklable_properties():
    ...
```

and


```
with self._no_unpicklable_properties():
    ...
```

ivannotes: reasonable, will do

luigi/contrib/sge.py
Tarrasch: How is this possible??? Parameters won't be set if you don't call the Task's constructor (I thought).

Anyway looks sane.

luigi/contrib/hadoop.py
```diff
@@ -944,23 +944,16 @@ def dump(self, directory=''):
...
-        _set_status_message = self.set_status_message
-        self.set_status_message = None
-
-        file_name = os.path.join(directory, 'job-instance.pickle')
-        if self.__module__ == '__main__':
-            d = pickle.dumps(self)
-            module_name = os.path.basename(sys.argv[0]).rsplit('.', 1)[0]
-            d = d.replace(b'(c__main__', "(c" + module_name)
-            open(file_name, "wb").write(d)
+        with self._without_unpicklable_properties():
```
Tarrasch: Excellent! :)

Tarrasch:
This looks good. Do others want to review again?

luigi/task.py
```diff
@@ -566,16 +566,17 @@ def _without_unpicklable_properties(self):
             class DummyTask(luigi):
 
                 def _dump(self):
-                    with self._without_unpicklable_properties():
+                    with self.no_unpicklable_properties():
                         pickle.dumps(self)
 
         """
         unpicklable_properties = ('set_tracking_url', 'set_status_message')
+        placeholder_during_pickling = None
```
Tarrasch: Nonono....

My point was that when somebody is debugging why something is not working. and they suddenly see that  `mytask.set_tracking_url` is `None`, they will be clueless. I basically suggest you edit this to 

```
placeholder_during_pickling = 'placeholder_during_pickling'
```

But of course inline it ^^

ivannotes: aha, I see ;)

ivannotes: fixed in d3c6b51

Tarrasch:
Cool! Thanks. Will merge once other +1's ^^

peterg-eb:
@Tarrasch, any chance we can get this into 2.3.1 as well, since PySparkTask has been unusable since 2.2?

PR #1820, opened 2016-08-18
Added to S3Target the ability to write temp file on S3 instead of to local filesystem

peterg-eb:
## Description

Added to S3Target the ability to write temp file on S3 instead of to local file system. This requires boto3 >= 1.4.0, but is only enabled if requested by the caller, so it should not affect current users at all.
## Motivation and Context

When working with very large files, it is impractical (or impossible) to write an entire file to a local filesystem. S3 is much more scalable, and it makes more sense to write the temporary/in-progress file there instead of locally. This significantly increases the flexibility of working with S3Target for my use cases, and I imagine it will come in handy for others as well. 

I have tested this thoroughly with my own code and it works well. I have not included unit tests, but can write some if someone can point me to an example set of tests for the AtomicS3File. The particular issues that need to be tested in a formal test are small files vs large files. Because boto3 handles these differently (because of S3 rules about multipart upload sizes), it was necessary to write some logic to intelligently buffer to meet boto3's expectations. The file sizes to test would be files < 5MB, files < 8MB, files <16MB, and files >16MB.

### DIFF
--- a/luigi/contrib/ecs.py
+++ b/luigi/contrib/ecs.py
@@ -58,11 +58,13 @@ logger = logging.getLogger('luigi-interface')
 
 try:
     import boto3
-    client = boto3.client('ecs')
+    try:
+        import botocore.exceptions
+        client = boto3.client('ecs')
+    except botocore.exceptions.NoRegionError:
+        logger.warning('Your AWS config is missing Region information, ECSTask requires a working config.')
 except ImportError:
     logger.warning('boto3 is not installed. ECSTasks require boto3')
-except NoRegionError:
-    logger.warning('Your AWS config is missing Region information, ECSTask requires a working config.')
 
 POLL_TIME = 2
 
--- a/luigi/contrib/ecs.py
+++ b/luigi/contrib/ecs.py
@@ -61,6 +61,8 @@ try:
     client = boto3.client('ecs')
 except ImportError:
     logger.warning('boto3 is not installed. ECSTasks require boto3')
+except NoRegionError:
+    logger.warning('Your AWS config is missing Region information, ECSTask requires a working config.')
 
 POLL_TIME = 2
 
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -672,6 +672,9 @@ class AtomicRemoteWritableS3File(object):
         if exc_type:
             print('**** WARNING: Failed to properly close AtomicRemoteWritableS3File because of error.')
             self._internal_queue.error('Pipe not properly closed.')
+            if self._upload_future:
+                self._upload_future.result()
+                self.s3_client.delete_object(Bucket=self.s3_bucket, Key=self.s3_key + '.TMP')
             return
         else:
             self.close()
--- a/luigi/s3.py
+++ b/luigi/s3.py
@@ -30,6 +30,7 @@ import os.path
 
 import time
 from multiprocessing.pool import ThreadPool
+import threading
 
 try:
     from urlparse import urlsplit
@@ -570,7 +571,7 @@ class S3Client(FileSystem):
 
 class AtomicS3File(AtomicLocalFile):
     """
-    An S3 file that writes to a temp file and put to S3 on close.
+    An S3 file that writes to a local temp file and put to S3 on close.
 
     :param kwargs: Keyword arguments are passed to the boto function `initiate_multipart_upload`
     """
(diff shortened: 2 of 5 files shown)


Tarrasch:
If nobody steps up and review his can be just merged...

jpiper:
Let me have a crack at this

luigi/s3.py
```diff
@@ -584,6 +586,180 @@ def move_to_final_destination(self):
...
+    Works around AWS S3's multipart transfer size requirements and boto3's
+    idiosyncratic implementation that requires an initial buffer size
+    larger than the multipart transfer threshold in order to correctly
+    select the 'read-until-empty' behavior needed for a streaming upload.
+    """
+    _boto3_default_multipart_threshold = 8 * 1024 * 1024
+
+    def __init__(self, s3_bucket, s3_key, boto3_s3_client=None):
+        import boto3
+        from s3transfer.manager import TransferManager, TransferConfig
```
jpiper: Is there any reason the imports are here and not at the top of the file? This might mask `importError`s until this task is encountered

jpiper:
A problem I can foresee here is that there appears to be no way of passing in kwargs to the boto functions here - i.e. for encryption, ACLs, etc.

luigi/s3.py
dlstadther: I'd guess it's because these aren't required for other `s3.py` functionality.

dlstadther:
I don't really have anything negative to say about this PR. 

It would be great if you could include some tests, if possible. Perhaps there are luigi tests for other Atomic file writes that you can reference, or maybe some boto tests somewhere on the internet.

luigi/s3.py
peterg-eb: yeah, this is the reason. It's certainly a tradeoff, but I've seen the pattern elsewhere in the luigi codebase. See RemoteFileSystem in https://github.com/spotify/luigi/blob/master/luigi/contrib/ftp.py for an example where pysftp is not a hard requirement, because some users of the RemoteTarget will only care about FTP and won't have pysftp installed.

peterg-eb:
@jpiper, that's an interesting point. It wouldn't be too hard to enable that access; ironically my last pull request was adding a kwargs parameter for PySFTP in RemoteTarget. Let me take a stab at adding this, hopefully without complicating things overly.

peterg-eb:
@dlstadther - It looks to me like the tests in test/s3_test.py actually run against moto mock s3 instead of boto. It seems like moto has at least limited support for boto3, but I don't know whether moto mocks at a deep enough level that it will be able to handle the new boto3 1.4.0 API that I'm using. 

I'll try writing up a basic test and see how far I get. :)

peterg-eb:
i've checked in a simple test, but I am unsure whether it will work properly because I don't really know how to let tox know about the boto3 dependency, which doesn't come into play until the code runs. If the tests fail, that is probably the issue. If that happens, perhaps someone with a pre-existing understanding of tox can come to my rescue and tell me what to do.

peterg-eb:
looks like the test is failing because it didn't know that boto3 was a dependency. I imagine I could fix this easily enough by importing boto3 in s3_test, though I don't know if that's the proper fix. 

Unfortunately I've been unable to run the test suite itself locally - I just get the same exception over and over again: `HostRequiredError: BotoClientError: When using SigV4, you must specify a 'host' parameter.`  Obviously I'm not providing some kind of configuration that boto wants, but I don't really know how it wants me to provide it, and I'm reluctant to monkey with my preexisting config for boto just to get the tests to run on my local machine.  If anyone knows the workaround for this, please let me know.

peterg-eb:
sheesh. that was worse. 

well, i'm going to refrain from further attempts to get the tests working until someone advises me on how to make tox prepare the boto3 dependency without declaring it as a top-level import in s3.py.

luigi/s3.py
jpiper: Ah ok, in that case best leave it that way :)

dlstadther: @peterg-eb from where does s3transfer.manager come? Is it part of boto3? Or is it something separate?

Some of travis fails because of an import error when importing this module.

test/s3_test.py
```diff
@@ -116,6 +116,43 @@ def test_get_path_sse(self):
...
+
+    def _run_remote_temp_upload_test(self, file_size):
+        file_contents = b"a" * file_size
+
+        tmp_file = tempfile.NamedTemporaryFile(mode='wb', delete=True)
+        tmp_file_path = tmp_file.name
+        tmp_file.write(file_contents)
+        tmp_file.flush()
+
+        s3_client = S3Client(AWS_ACCESS_KEY, AWS_SECRET_KEY)
```
dlstadther: Won't this actually try to connect to s3 using those creds (rather than pretending to connect)?

test/s3_test.py
```diff
@@ -116,6 +116,43 @@ def test_get_path_sse(self):
...
+        s3_client.s3.create_bucket('mybucket')
+
+        s3_path = 's3://mybucket/remote_test_file'
+        t = S3Target(s3_path, remote_temp_write=True)
```
dlstadther: line 98 above specifies `client=s3_client` within `S3Target`, that may be needed

peterg-eb: thanks. that should help, but there's a bunch of other stuff going on here too, because no AWS config is present in the testing environment. I can pretty easily work around it in my code, but ECSTask tests are failing for the same reason now that tox is pulling in the boto3 dependency, which activates the ECSTask tests, but still in a config-less environment.

luigi/contrib/ecs.py
```diff
@@ -58,7 +58,11 @@
 
 try:
     import boto3
-    client = boto3.client('ecs')
+    try:
+        import botocore.exceptions
+        client = boto3.client('ecs')
+    except botocore.exceptions.NoRegionError:
+        logger.warning('Your AWS config is missing Region information, ECSTask requires a working config.')
```
dlstadther: Umm... this isn't relevant to this PR

peterg-eb:
@Tarrasch, @dlstadther : Looks like I've gotten as close as I can to getting these tests working. Seems like there's something down inside boto3 that causes a nose.proxy.ImportError -- this is what the Travis CI build is reporting:

```
...
File "/home/travis/build/spotify/luigi/.tox/py33-nonhdfs/lib/python3.3/site-packages/botocore/hooks.py", line 210, in _emit
    response = handler(**kwargs)
  File "/home/travis/build/spotify/luigi/.tox/py33-nonhdfs/lib/python3.3/site-packages/boto3/utils.py", line 61, in _handler
    module = import_module(module)
  File "/home/travis/build/spotify/luigi/.tox/py33-nonhdfs/lib/python3.3/site-packages/boto3/utils.py", line 52, in import_module
    __import__(name)
  File "/home/travis/build/spotify/luigi/.tox/py33-nonhdfs/lib/python3.3/site-packages/boto3/s3/inject.py", line 17, in <module>
    from boto3 import utils
nose.proxy.ImportError: cannot import name utils
```

You'll notice that the stack trace shows that execution is passing through the very module that `nose` is complaining about not existing. Apparently this is something of a known issue with nose, or at least not totally unprecedented.  boto3 certainly works just fine when not run through nose.

I've looked for workarounds for this, but haven't found anything that doesn't require me to either modify the boto3 package or changing the Nose configuration in a way that I don't think I can control directly in the Travis CI build.

For the time being, I'm just going to prevent the tests from running, but I'll leave them in the diff so that, if at a later date someone is able to fix nose so that it doesn't mess up the tests, they will be ready to be activated.

references: 
http://stackoverflow.com/questions/16200333/import-errors-when-running-nosetests-that-i-cant-reproduce-outside-of-nose
http://stackoverflow.com/questions/2799852/importerror-using-nose-no-importerror-using-raw-unittest

luigi/contrib/ecs.py
peterg-eb: it's necessary in order to make these unit tests work, because ECSTask does not use boto3 in an extensible manner.

Basically, the ecs_test module appears to have been written for someone's personal testing instead of for Travis CI. When the boto3 package is available for testing in Travis (since I added it to tox.ini), the ecs_test module is suddenly activated by its presence. It then proceeds to fail because boto3.client will always fail when no local AWS configuration is available. 

Since boto3 is now available, I have contributed a fix for their tests so that their tests won't fail the build when my tests are present.

dlstadther:
I think i'm good with this - so long as the `ecs` module still functions as desired (i never use it). Passing tests (for `ecs`) are also a nice thing to see. It unfortunate that the new s3 tests aren't working. @Tarrasch , any suggestions on overcoming the Travis issue?

PR #1821, opened 2016-08-20
Don't kill the worker when it cannot connect to smtp server

interskh:
## Description
Catch socket.error exception when trying to connect smtp server in error email notification.

## Motivation and Context
Worker dies when it tries to send out error email but cannot connect to smtp server. This is rather unfortunate in production environment. This PR is intended to fix it.

I have included unit test.


Tarrasch:
Nice catch!

erikbern:
LGTM

PR #1824, opened 2016-08-23
Tries using path for non-hdfs targets in hadoop_jar

daveFNbuck:
## Description
Tries using the path attribute of arguments in fix_paths before resorting to str.

## Motivation and Context
When writing a map-reduce job that takes S3 inputs, I found that the arguments were incorrect. As S3Target does not inherit from HdfsTarget, the code goes to the base case of calling str.

I've been running this code a while and have unit tests.

### DIFF
--- a/luigi/contrib/hadoop_jar.py
+++ b/luigi/contrib/hadoop_jar.py
@@ -52,7 +52,12 @@ def fix_paths(job):
                 logger.info('Using temp path: %s for path %s', y.path, x.path)
                 args.append(y.path)
         else:
-            args.append(str(x))
+            try:
+                # hopefully the target has a path to use
+                args.append(x.path)
+            except AttributeError:
+                # if there's no path then hope converting it to a string will work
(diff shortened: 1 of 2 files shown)


luigi/contrib/hadoop_jar.py
```diff
@@ -52,7 +52,10 @@ def fix_paths(job):
                 logger.info('Using temp path: %s for path %s', y.path, x.path)
                 args.append(y.path)
         else:
-            args.append(str(x))
+            try:
+                args.append(x.path)
```
Tarrasch: Can you add a comment to this `try-except` dance? Perhaps just linking back to this exact PR.

daveFNbuck: I'm not sure what the str is supposed to accomplish, so hopefully my comment is ok.

Tarrasch:
Forgot about this PR. I just merged it :)

PR #1826, opened 2016-08-24
Allow port reuse for central scheduler server

darabos:
## Description

Allows port reuse for the central scheduler server.
## Motivation and Context

Otherwise it is impossible to restart the scheduler. For a minute or more startup fails with:

```
  File "/usr/bin/luigid", line 11, in <module>
    sys.exit(luigid())
  File "/usr/lib/python3.5/site-packages/luigi/cmdline.py", line 56, in luigid
    luigi.server.run(api_port=opts.port, address=opts.address, unix_socket=opts.unix_socket)
  File "/usr/lib/python3.5/site-packages/luigi/server.py", line 260, in run
    unix_socket=unix_socket,
  File "/usr/lib/python3.5/site-packages/luigi/server.py", line 237, in _init_api
    api_sockets = tornado.netutil.bind_sockets(api_port, address=address)
  File "/usr/lib64/python3.5/site-packages/tornado/netutil.py", line 194, in bind_sockets
    sock.bind(sockaddr)
OSError: [Errno 98] Address already in use
```

Manually tried restarting it after this change. Works okay now.

### DIFF
--- a/luigi/server.py
+++ b/luigi/server.py
@@ -234,7 +234,7 @@ def _init_api(scheduler, responder=None, api_port=None, address=None, unix_socke
     if unix_socket is not None:
         api_sockets = [tornado.netutil.bind_unix_socket(unix_socket)]
     else:
-        api_sockets = tornado.netutil.bind_sockets(api_port, address=address)
+        api_sockets = tornado.netutil.bind_sockets(api_port, address=address, reuse_port=True)
     server = tornado.httpserver.HTTPServer(api_app)
     server.add_sockets(api_sockets)
 


luigi/server.py
```diff
@@ -234,7 +234,7 @@ def _init_api(scheduler, responder=None, api_port=None, address=None, unix_socke
     if unix_socket is not None:
         api_sockets = [tornado.netutil.bind_unix_socket(unix_socket)]
     else:
-        api_sockets = tornado.netutil.bind_sockets(api_port, address=address)
```
graingert: This can fail:
    if reuse_port and not hasattr(socket, "SO_REUSEPORT"):
        raise ValueError("the platform doesn't support SO_REUSEPORT")

graingert:
have you tried using the unix socket and using an NGINX proxy?

darabos:
> have you tried using a unix socket and exposing TCP via an NGINX reverse proxy?

No. Is that the recommended way to run the Luigi UI? I'm completely lost :).

Tarrasch:
I don't understand this PR. I've never had problems restarting the scheduler unless it's already running, which seem like the exact desired behaviour. What have I missed? :)

graingert:
Here's some docs on what SO_REUSEPORT is for and what it does:
https://lwn.net/Articles/542629/

It's basically so you can have multiple servers share the same port. This
doesn't make sense for a server that you're using as a central lock to
schedule processes.

You should probably look at what process is locking that port:
http://www.cyberciti.biz/faq/what-process-has-open-linux-port/

If it's luigi even after you've done a clean shutdown that's a different
bug.

On 25 Aug 2016 03:12, "Arash Rouhani" <notifications@github.com> wrote:

I don't understand this PR. I've never had problems restarting the
scheduler unless it's already running, which seem like the exact desired
behaviour. What have I missed? :)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/spotify/luigi/pull/1826#issuecomment-242262849>, or mute
the thread

PR #1829, opened 2016-08-26
Fixes backward compatibility with disable-num-failures

daveFNbuck:
## Description
Changes the name `disable_failures` to `disable-num-failures` in the scheduler config backward compatibility check.

## Motivation and Context
I recently noticed that tasks don't get disabled anymore in my scheduler since updating luigi. After digging into it a bit, I found that my `retry_count` was set to the default value of 999999999. For backward compatibility, it checks `disable_failures`, but my config contains `disable-num-failures`. Some people may be using `disable_num_failures` (which I think was also valid before this change) but I don't know how we can support both cases.

I ran a test scheduler with my config and a job set to fail in its run. It disabled appropriately again.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -126,7 +126,7 @@ class scheduler(Config):
     disable_window = parameter.IntParameter(default=3600,
                                             config_path=dict(section='scheduler', name='disable-window-seconds'))
     retry_count = parameter.IntParameter(default=999999999,
-                                         config_path=dict(section='scheduler', name='disable_failures'))
+                                         config_path=dict(section='scheduler', name='disable-num-failures'))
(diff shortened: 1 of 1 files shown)


daveFNbuck:
I'd like to note that retry_count is a bit of a misnomer because it includes the initial attempt (which is not a retry). I'd expect that if I set retry_count to 1, it'll run twice before disabling, but it'll actually just run once as I would expect from the name disable_num_failures.

Tarrasch:
:-1:, luigi previously printed a deprecation warning suggesting to use `disable_failures`. Though I see how this can annoy users. But it's a one-time change only.

daveFNbuck:
Oh, I missed those because I don't normally look at the server logs. It's
very difficult to keep up with repeated name changes of random parameters
throughout the code base.

PR #1830, opened 2016-08-26
Readable partitioned targets

peterg-eb:
## Description

Added to S3FlagTarget the ability to read a set of underlying partition files as though they were a single file. 

The code has been written to be fairly generic, so it has been added as a separate module. It should theoretically have nearly drop-in suitability with other "Flag" Targets such as the GCS Flag Target. However, I am unable to test other systems, and so have not inserted the proper shims into contrib/gcs or any other parts of Luigi.
## Motivation and Context

Though it is generally advisable to refrain from dealing with any Hadoop-style distributed data in a serial manner, it is nevertheless occasionally necessary to combine data at the end of a workflow into a single file suitable for use in serial systems.

In my own case, I find it necessary to combine the results of a Spark CSV write into a single Spark file for upload to an external service. It would sometimes be possible to have Spark coalesce the data into a single partition before writing, but in cases where the data is very large, this might be prohibitive or even impossible depending on available memory.

The more general solution to this problem is the ability to have Luigi read multiple "partition" files in sequence, as though they were a single file target. The data from the partition files can then be streamed through memory in small amounts as they are written elsewhere - in my case, to the external service.

I also recognize that sometimes certain types of files cannot be correctly concatenated so simply. In my example, a CSV file with a header defining the column names, there is a single header line per partition file, but this is not suitable for the external service. Therefore, the PartitionedFilesReader allows callable functions to be submitted at creation time that will allow a savvy user to selectively modify or ignore parts of the files as they are being read, while still providing a clean Luigi Target interface to the pipeline.

I have tested this thoroughly with my own systems, including with the line-based and byte-based filtering function. 
I have also added a basic unit test.

### DIFF
--- a/luigi/partitions.py
+++ b/luigi/partitions.py
@@ -1,0 +1,172 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2016 Eventbrite Inc
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+def splitting_prefix_chooser(fs_url, prefix='part-'):
+    filename = fs_url.split('/')[-1]
+    return filename.startswith(prefix)
+
+
+class PartitionedFilesReader(object):
+    """
+    Pass this a Target representing a directory full of part-* files,
+    such as an S3FlagTarget, and it will allow you to read the part-* files
+    underneath that directory one by one as though they were a single file.
+    A target creator callable, such as S3Target, is also necessary.
+    It will be passed the canonical URL of each part file in turn, and must return a
+    Target-like object that can be opened as readable (e.g., open('r')).
+
+    The user should check that the flag target 'exists()' before using
+    this class, otherwise the user may end up reading partial part-* files.
+
+    You can additionally pass a function to help it make proper decisions
+    about how to handle the start of a file, e.g. if your CSV files have a header
+    that you want to ignore for all files except the first one. This function
+    will differ based on whether the reader will be reading by bytes or by lines.
+
+    Examples:
+
+    # The line-by-line filter function receives each line once. It can choose to
+    # accept (return), reject (return blank), or modify (return bytes)
+    # the line. It will not receive this line again.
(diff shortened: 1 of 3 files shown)


Tarrasch:
Please let us know when this can be reviewed again.

petergaultney:
It's ready to be reviewed. Only delay was trying to get sphinx and flake happy, since I still can't get the tests to run on my dev machines. :/

peterg-eb:
It's ready to be reviewed now. Just took forever to get sphinx and flake happy. I wish I could get the tests to run on my dev machines to make this step easier on all of us, but for various reasons it's not possible at the moment.

Tarrasch:
Can you find people who knows s3 to review this?

peterg-eb:
Only folks I know would be at my company. I can ask someone to swing by. 

But except for the couple of lines added to S3FlagTarget, this isn't actually S3-specific. It's just a generic wrapper for multiple file-like objects. It depends only on listdir() returning files in alphabetical order, which is pretty common for a filesystem interface. I can add some unit tests to prove the general functionality of the PartitionedFilesReader via the local file system, if that helps.

Tarrasch:
A person from your company is totally fine too I think! :)

peterg-eb:
@eb-brandonhamric, @alextmeyer, mind stopping by and giving this a look?

eb-brandonhamric:
i'de probably call it something else, saying "partitioned" about a hadoop target makes me think it's for actual partitions using fields like "table/partition_field=partition_value/part-*" instead of just "table/part-*".  I would also make it support any hadoop compatible file system if it's called something this generic since the purpose is to bring results out of hdfs/s3 into a single file.

the glob pattern should also be configurable since not all hadoop processes use the part-* convention

eb-brandonhamric:
probably just wrap "hadoop fs -getmerge" and just make it a luigi task called HadoopGetMerge taking HdfsTargets for input and output. I would assume getmerge will work over s3. This also implies that the process of pulling the single file out of hdfs should be separate IMO.

peterg-eb:
the glob pattern is configurable, it's just embedded in the listdir_chooser. In the very unlikely scenario that your filesystem uses something other than '/' to split paths, you'd have to write your own. In the much more common scenario that you want to glob differently, you could simply supply a lambda wrapping "splitting_prefix_chooser" with "prefix" pre-defined.

What do you mean by "support any hadoop compatible file system"? I think it already does. Perhaps the confusion here is stemming from the adaptation of S3FlagTarget to use this code? I don't particularly have the expertise to find all of the Luigi targets that are hadoop-related and add this functionality to all of them, but anyone who does have that expertise should find it very easy to shim this into them just like I did for S3FlagTarget. 

In any case, this is meant to be more generic than Hadoop (though, again, I don't have the expertise to suggest what else it would used for). For instance, if you wanted to read local files from a unix filesystem as though they were one file, it would work equally well.

PR #1836, opened 2016-08-31
Properly use Python3's buffer interface.

sceeter89:
## Description
When using `luigi` with Python 3 there is issue running PigTask. Because output in `track_and_progress` is both read to unicode string and copied to temporary file, it failed as conversion to string happens before writing to file. I simply moved decoding after writing to temporary file.

## Motivation and Context
Without this change, running PigTask ended with following error:

>   File "/projects/big-data/luigi-tasks/libs/pig_task.py", line 57, in on_execute
>    PigTask.run(self)
>  File "/home/etl/.projects/big-data/lib/python3.4/site-packages/luigi/contrib/pig.py", line 127, in run
>    self.track_and_progress(cmd)
>  File "/home/etl/.projects/big-data/lib/python3.4/site-packages/luigi/contrib/pig.py", line 149, in >track_and_progress
>    temp_stdout.write(line)
>TypeError: 'str' does not support the buffer interface

And while Pig script finished successfully, `luigi` marked it as failed.

After making this change locally our Pig scripts run without issues using Python3.4

### DIFF
--- a/luigi/contrib/pig.py
+++ b/luigi/contrib/pig.py
@@ -145,8 +145,9 @@ class PigJobTask(luigi.Task):
                         line = proc.stderr.readline().decode('utf8')
                         err_lines.append(line)
                     if fd == proc.stdout.fileno():
-                        line = proc.stdout.readline().decode('utf8')
-                        temp_stdout.write(line)
+                        line_bytes = proc.stdout.readline()
+                        temp_stdout.write(line_bytes)
+                        line = line_bytes.decode('utf8')
 
                 err_line = line.lower()
                 if err_line.find('More information at:') != -1:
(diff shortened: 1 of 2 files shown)


Tarrasch:
Can you add a test case?

sceeter89:
I'll give it a shot afternoon.

sceeter89:
@Tarrasch Instead of adding new test case I changed fake Popen stuff to let PigTask read input (by returning None from poll at least once). It was enough to make all tests fail. And applying fix fixed that.

Tarrasch:
LGTM. But someone knowing pig should review.

sceeter89:
Actually this issue has nothing to do with Pig itself, but if you know such a person it would be great to let her/him know.

Tarrasch:
Okay. If nobody comments for a while I suppose this can be merged.

gpoulin:
LGTM, I would just change line 130 to `temp_stdout = tempfile.TemporaryFile('wb')` to make it explicit that the temporary file is expecting binary and not string.

sceeter89:
@gpoulin Good point, thank you for that, I've just pushed this change.

erikbern:
lgtm

PR #1838, opened 2016-09-01
Add and refactor tests for disabling workers

Tarrasch:
## Description

Increase readability and exhaustiveness for tests relating to disabling workers.


Tarrasch:
Only affects tests and basically only a test added. Hence I self-approve without waiting for reviewers.

PR #1839, opened 2016-09-01
Scheduler: don't give tasks to disabled assistants

Tarrasch:
## Description

Previously assistants would keep getting tasks and never stop. Despite
that a user have clicked "disable worker" in the WebUI.

## Motivation and Context

This is a bugfix.

I have included unit tests. I have yet to test in production environment.

### DIFF
--- a/luigi/static/visualiser/index.html
+++ b/luigi/static/visualiser/index.html
@@ -283,22 +283,22 @@
               </div>
             </div>
             {{#workers}}
-            {{#num_pending}}
-            <div class="box">
-            {{/num_pending}}
-            {{^num_pending}}
+            {{#is_disabled}}
             <div class="box box-solid box-default">
-            {{/num_pending}}
+            {{/is_disabled}}
+            {{^is_disabled}}
+            <div class="box">
+            {{/is_disabled}}
                 <div class="box-header with-border">
                     <h3 class="box-title">{{name}}</h3>
                     <div class="box-tools pull-right">
-                      {{#num_pending}}
+                      {{^is_disabled}}
                       <div class="button-tooltip" data-toggle="tooltip" title="Disable Worker">
                         <button type="button" class="btn btn-danger btn-disable-worker" data-toggle="modal" data-target="#disableWorkerModal" data-worker="{{name}}">
                           <i class="fa fa-fire-extinguisher"></i>
                         </button>
                       </div>
-                      {{/num_pending}}
+                      {{/is_disabled}}
                     </div>
                 </div>
                 <div class="box-body">
--- a/luigi/static/visualiser/index.html
+++ b/luigi/static/visualiser/index.html
@@ -274,6 +274,9 @@
                   </div>
                   <div class="modal-body">
                     Are you sure you want to disable this worker?
+                    <p>
+                        A disabled worker will finish its existing tasks but not start new ones.
+                    </p>
                   </div>
                   <div class="modal-footer">
                     <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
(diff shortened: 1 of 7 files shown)


Tarrasch:
@daveFNbuck, as the founder of assistants feature. Do you mind reviewing this?

daveFNbuck:
Won't we need to add worker logic to kill the assistant after this? It's not ideal if the assistant process stays permanently alive after this.

Tarrasch:
Oh. I assumed assistants would break their get_work loop once the scheduler says there's no more tasks. Let me double check this.

Tarrasch:
This PR is yet not done although still sensible to review. I need to add tests that actually disable a worker. The tests I've touched upon so far are about sending interrupts to the worker (only vaguely related).

Tarrasch:
Feels a bit scary adding another thing to `get_work`. But does it make sense to you @daveFNbuck?

daveFNbuck:
Looks good, but it's missing a way to tell the viz about whether the worker is disabled. You can't disable a worker if it doesn't have any pending jobs without that.

Tarrasch:
Really? I thought the visualizer already had this information in full. Because after clicking on the fire extinguisher button, the worker "bar" is greyed out. Eh,, I'll let pics explain this:

before click:

selection_045

after click:

selection_046

daveFNbuck:
It uses whether the worker has any pending tasks to determine whether to
gray the bar out. If you have an assistant with no pending tasks, the bar
is already grayed out and you won't have the button to kill it.

Tarrasch:
Oh, put your safety belt on everyone, I'm going to mess around in frontend land then. ^^

Thanks for the clarification @daveFNbuck, I'll try to see where I can get with this.

daveFNbuck:
I can do it if this part is too unfamiliar for you.

Tarrasch:
I'll take a stab at it today. If I fail miserably I'll see how we can cooperate :)

Tarrasch:
@daveFNbuck, I've got stuck on the `.html` templating part. I initially thought we use tornado for html templating ([here](http://www.tornadoweb.org/en/stable/template.html) and [here](http://www.tornadoweb.org/en/stable/guide/templates.html). But it seems like no.

I'm editing `luigi/static/visualiser/index.html` and I don't see syntax like `{% if blah%}` but instead it uses `{{#var_name}}` which doesn't seem to be Tornado's thingie. Confusingly, the html file `luigi/templates/layout.html` seem to use the tornado syntax.

Can you say what template engine provides `{{#var_name}}` syntax so I can google for docs?

daveFNbuck:
I also don't know the answer on that one. I've mostly just tweaked around the edges on what was already there.

Tarrasch:
@freider @erikbern @stephenpascoe, quick question. Do you know what HTML templating syntax luigi's visualiser uses? See the comment above.

Tarrasch:
Ah, luigi is using [Mustache](https://github.com/janl/mustache.js/) as of #80. I'm not sure about what version and stuff though.

Tarrasch:
Oh boy am I excited after my first line of written javascript! 

Mostly unrelated: Kudos to @fabriziodemaria and everyone involved in the recent improvements to the execution summary and return codes. The summary prints exactly what one expects after disabling the worker:

selection_048

Tarrasch:
@daveFNbuck, I would appreciate a final round of review. :)

luigi/static/visualiser/index.html
```diff
@@ -308,6 +311,9 @@ <h3 class="box-title">{{name}}</h3>
                     Running: {{num_running}}<br>
                     Pending: {{num_pending}}<br>
                     Unique Pending: {{num_uniques}}<br>
+                    {{#is_disabled}}
+                    This worker is <b>disabled</b>. It will not start on new tasks.<br>
```
daveFNbuck: It will not start new tasks. No "on"

PR #1844, opened 2016-09-08
Disables siginterrupt for SIGUSR1

daveFNbuck:
## Description
Sets siginterrupt to False for SIGUSR1 when installing the shutdown handler.

## Motivation and Context
I've recently started using SIGUSR1 to stop old workers on deploy and each deploy has resulted in all MapReduce tasks dying with `IOError: [Errno 4] Interrupted system call`. To avoid this, we want the SIGUSR1 signal to not interrupt system calls.

I wrote some unit tests and tested it heavily in production both by sending lots of SIGUSR1 signals to running MRs and by seeing the regular stream of errors during deploys stop.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -404,6 +404,7 @@ class Worker(object):
         if not self._config.no_install_shutdown_handler:
             try:
                 signal.signal(signal.SIGUSR1, self.handle_interrupt)
+                signal.siginterrupt(signal.SIGUSR1, False)
             except AttributeError:
                 pass
 
--- a/test/worker_test.py
+++ b/test/worker_test.py
@@ -176,6 +176,26 @@ class WorkerTest(unittest.TestCase):
         self.w.run()
         self.assertFalse(d.complete())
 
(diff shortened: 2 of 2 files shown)


Tarrasch:
Looks good to me. The concept of `siginterrupt` is something new to me (and I feel I still don't understand it after lots of trying). What is the "system call" that is interrupted? Is it a system-call that opens child-processes (like map reduce jobs)

daveFNbuck:
Siginterrupt is new to me as well, but I was able to solve the problem thanks to some Googling. I can't find the Stack Overflow thread that helped me solve it again though :(. The system call that got interrupted most of the time was reading the stderr pipe in the spawned subprocess for the mapreduce on line 295 of luigi/contrib/hadoop.py. https://github.com/spotify/luigi/blob/master/luigi/contrib/hadoop.py#L295

If the signal happened during scheduling, it could kill the multiprocessing queue get during a call to conn.recv() which I'm guessing is also a pipe read.

PR #1845, opened 2016-09-08
Extended inherits/requires decorators support multiple parents

ifiddes:
## Motivation and Context

These decorators simplify creating more complicated dependency graphs where one task either depends on or inherits characteristics from more than one parent task.

Technically, you could replace `@requires` and `@inherits` entirely with these versions, as they support singular input. If you want me to do that, I can change the PR.

 

I have not produced unit tests for this yet, but it works within my own code base. It is a fairly simple logical extension of the base versions of these decorators.

### DIFF
--- a/luigi/util.py
+++ b/luigi/util.py
@@ -298,6 +298,54 @@ class inherits(object):
         return Wrapped
 
 
+class multiple_inherits(object):
+    """
+    Task inheritance.
+
+    Usage:
+
+    .. code-block:: python
+
+        class TaskPathA(luigi.Task):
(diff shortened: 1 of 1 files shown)


Tarrasch:
This seem like sensible addition. I rarely use these util functions. So if somebody cold review that would be sweet.

But as you hinted in your comments, I think one might as well replace `@requires` and `@inherits` to use these versions. Also adding unit tests would be good.

erikbern:
lgtm

dlstadther:
Closing due to merge of #2475

PR #1846, opened 2016-09-08
Prevents batch tasks from running before their dependencies

daveFNbuck:
## Description
Checks whether batch tasks are schedulable before adding them to a batch in get_work.

Note that even though self._schedulable(task) gets called twice in the get_work loop, they fall under different branches of whether best_task is defined, so this shouldn't slow get_work down much.

## Motivation and Context
When adding tasks to an existing batch in get_work, we neglected to check whether the tasks were schedulable. This meant that in production I found my pipelines running batches that included jobs with PENDING dependencies. In order to fix this, we simply add a check that the task is schedulable.

I have included unit tests.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -894,7 +894,8 @@ class Scheduler(object):
 
             if (best_task and batched_params and task.family == best_task.family and
                     len(batched_tasks) < max_batch_size and task.is_batchable() and all(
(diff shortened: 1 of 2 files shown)


Tarrasch:
Thanks. Am I correct in that this is a pretty serious bug as it would run a task before it's dependencies are complete?

daveFNbuck:
Yes, that was what happened in production after I deployed a version of Luigi with this bug.

PR #1847, opened 2016-09-08
Allows clients to specify max_shown_tasks in task_list

daveFNbuck:
## Description
Adds an optional max_shown_tasks argument to task_list to override the server-configured value.

## Motivation and Context
I have some scripts that process task_list for reporting. When the response length goes beyond the maximum shown tasks, the scripts all break. In order to allow the scripts to keep reporting as things get backed up, this adds an optional argument to override the config's max shown tasks in task_list, while keeping the configured default for the visualizer.

This could also potentially be used in the visualizer to make the number of tasks shown configurable by users.

I included a unit test and I've been using this in production for about 3 months.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -1121,7 +1121,8 @@ class Scheduler(object):
             task_id, dep_func=lambda t: inverse_graph[t.id], include_done=include_done)
 
     @rpc_method()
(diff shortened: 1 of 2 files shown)


Tarrasch:
Looks good to me! :)

PR #1848, opened 2016-09-08
Sets upstream status for BATCH_RUNNING jobs to UPSTREAM_RUNNING

daveFNbuck:
This was an oversight and is now fixed.


Tarrasch:
Thanks. Was the consequence a runtime crash in the scheduler?

daveFNbuck:
I'm not sure there was a consequence to this, as we use `STATUS_TO_UPSTREAM_MAP.get(dep.status, '')` when retrieving from the map and UPSTREAM_RUNNING is never used anywhere. That's probably why I overlooked it in the PR even though it was in my fork. It might cause reporting scripts that use the API or a future feature to report incorrect results.

PR #1850, opened 2016-09-08
Fails batch_running tasks if worker dies early

daveFNbuck:
## Description
Compares task.status to both BATCH_RUNNING and RUNNING in fail_dead_worker_task instead of just to RUNNING.

## Motivation and Context
If a worker were to die between receiving batch tasks and responding with the task that will run them, we want the tasks to fail. Our existing methods to fail batch tasks don't work in this case, so I've extended the method for failing RUNNING tasks to also work for BATCH_RUNNING in order to handle this extra case.

I have included tests.


Tarrasch:
Ah, thanks for fixing this. :)

PR #1851, opened 2016-09-08
Adds time_running to batch running tasks

daveFNbuck:
## Description
Sets time_running when marking a task as BATCH_RUNNING in the scheduler.

## Motivation and Context
We neglected to set time_running when setting a task as batch_running. By adding it, we get a better view of these tasks in the visualizer.

I've been using it in production, and I added a test to verify that the time is set.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -431,6 +431,7 @@ class SimpleTaskState(object):
         self.set_status(task, BATCH_RUNNING)
         task.batch_id = batch_id
         task.worker_running = worker_id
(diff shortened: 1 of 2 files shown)


Tarrasch:
LGTM, but it seems that you need to rebase this :)

daveFNbuck:
Done. One of the other PRs added tests starting on the same line.

Tarrasch:
Thanks for the PR! :)

daveFNbuck:
No problem, thanks for all the merges!

PR #1856, opened 2016-09-16
Fix sample code in SGE package description

rolfschr:
## Description
Sample code was missing an import


Tarrasch:
Thanks @rolfschr

PR #1858, opened 2016-09-20
Handles errors when loading batch tasks from multiple ids

daveFNbuck:
## Description
Catches and handles exceptions when loading batch tasks from multiple batch ids.

## Motivation and Context
I ran into an error when I used `operator.add` as a batch_method. This fails because operator.add has the wrong signature. The problem is that this failure kills the worker rather than just failing that particular job. In order to allow the rest of the pipeline to continue, we trap the error just like when we fail loading a single task.

I have included unit tests.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -536,6 +536,21 @@ class Worker(object):
         message = notifications.format_task_error(headline, task, command, formatted_traceback)
         notifications.send_error_email(formatted_subject, message, task.owner_email)
 
+    def _handle_task_load_error(self, exception, task_ids):
+        msg = 'Cannot find task(s) sent by scheduler: {}'.format(','.join(task_ids))
+        logger.exception(msg)
+        subject = 'Luigi: {}'.format(msg)
+        error_message = notifications.wrap_traceback(exception)
+        for task_id in task_ids:
+            self._add_task(
+                worker=self._id,
+                task_id=task_id,
+                status=FAILED,
+                runnable=False,
+                expl=error_message,
+            )
+        notifications.send_error_email(subject, error_message)
+
     def add(self, task, multiprocess=False):
         """
         Add a Task for the worker to check and possibly schedule and run.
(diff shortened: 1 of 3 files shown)


test/worker_test.py
```diff
@@ -131,6 +132,14 @@ def run(self, result=None):
         if time.time != self.time:
             time.time = self.time
 
+    def setUp(self):
+        # ensure that other tests don't leak information into these via the task register
+        luigi.task_register.Register.clear_instance_cache()
+
+    def tearDown(self):
+        # ensure these tasks don't leak into others via the task register
+        luigi.task_register.Register.clear_instance_cache()
```
Tarrasch: Instead of setUp and tearDown, did you consider using the [LuigiTestCase](https://github.com/spotify/luigi/blob/6cf7091326af6eba64965767cef83f62a1bff53b/test/helpers.py#L141) helper?

luigi/worker.py
```diff
@@ -713,11 +713,25 @@ def _get_work_task_id(self, get_work_response):
...
-                task_name=get_work_response['task_family'],
-                params_str=get_work_response['task_params'],
-            )
+            try:
+                task = load_task(
+                    module=get_work_response.get('task_module'),
+                    task_name=get_work_response['task_family'],
+                    params_str=get_work_response['task_params'],
+                )
+            except Exception as ex:
```
Tarrasch: Since you catch such a broad exception, the code isn't self-documenting. I'm sure your use-case is fine, but can you add a comment within the `except` block of what typically goes wrong and makes us end up here?

luigi/worker.py
```diff
@@ -713,11 +713,25 @@ def _get_work_task_id(self, get_work_response):
...
+                batch_task_ids = get_work_response['batch_task_ids']
+                msg = 'Cannot load batch task for %s' % ', '.join(batch_task_ids)
```
Tarrasch: Perhaps use `format()` or just  `+`?

luigi/worker.py
```diff
@@ -713,11 +713,25 @@ def _get_work_task_id(self, get_work_response):
...
+                logger.exception(msg)
+                subject = 'Luigi: %s' % msg
```
Tarrasch: Is this copy-pasted from somewhere?  Is it worth to refactor out a helper? (maybe?)

test/worker_test.py
daveFNbuck: I wasn't aware of that, I'll try to switch to it.

luigi/worker.py
daveFNbuck: I used such a broad catch because there's no limit to what could go wrong here. batch_method can throw arbitrary errors because it can be an arbitrary function supplied by the user.

luigi/worker.py
daveFNbuck: Moved it into a helper, and now this PR subsumes #1859 :)

PR #1859, opened 2016-09-20
Sends an error message to the scheduler upon failing to load tasks

daveFNbuck:
## Description
Adds the traceback to add_task when updating a task status to FAILED after failing to load it from a task id in the worker.

## Motivation and Context
Since moving to an assistant-based workflow, assistants often pick up tasks that users are developing in their own branches. As assistants have no way of running these, we get a fair amount of failures. Unfortunately, the scheduler never receives the reason for these failures, leaving many
people confused as to what exactly happened. Luckily, there's an easy fix for this by simply adding the traceback when informing the scheduler of the failure.

I have included unit tests.

### DIFF
--- a/test/worker_test.py
+++ b/test/worker_test.py
@@ -1322,6 +1322,17 @@ class UnimportedTask(luigi.Task):
             self.assertTrue(self.assistant.run())
             self.assertEqual(list(self.sch.task_list('DONE', '').keys()), [task.task_id])
 
+    def test_unimported_job_sends_failure_message(self):
+        class NotInAssistantTask(luigi.Task):
+            task_family = 'Unknown'
+            task_module = None
+
+        task = NotInAssistantTask()
+        self.w.add(task)
+        self.assertFalse(self.assistant.run())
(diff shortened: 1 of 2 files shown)


test/worker_test.py
```diff
@@ -1322,6 +1322,17 @@ class NotImportedTask(luigi.Task):
             self.assertTrue(self.assistant.run())
             self.assertEqual(list(self.sch.task_list('DONE', '').keys()), [task.task_id])
 
+    def test_unimported_job_sends_failure_message(self):
+        class NotInAssistantTask(luigi.Task):
```
Tarrasch: You can make this more realistic by using the [`temporary_unloaded_module`](https://github.com/spotify/luigi/blob/6cf7091326af6eba64965767cef83f62a1bff53b/test/helpers.py#L193) helper.

daveFNbuck: That's only used when you want the module to exist. I'm only testing what happens when it doesn't exist.

Tarrasch:
Subsumed by #1858

PR #1860, opened 2016-09-20
Enables worker timeouts with only one worker

daveFNbuck:
## Description
Always runs tasks with timeouts in a separate process.

## Motivation and Context
Worker timeouts rely on running the task in a separate process. In order to get them to work with a single worker, we need to run tasks that have timeouts in separate processes. This change does that while still allowing tasks that don't have a timeout set to run in the main process.

I have included unit tests and tried it with some simple test jobs. I've also deployed it to production and I no longer see jobs running forever on single workers.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -101,7 +101,7 @@ class TaskProcess(multiprocessing.Process):
     Mainly for convenience since this is run in a separate process. """
 
     def __init__(self, task, worker_id, result_queue, tracking_url_callback,
-                 status_message_callback, random_seed=False, worker_timeout=0):
+                 status_message_callback, use_multiprocessing=False, worker_timeout=0):
         super(TaskProcess, self).__init__()
         self.task = task
         self.worker_id = worker_id
@@ -111,7 +111,7 @@ class TaskProcess(multiprocessing.Process):
         if task.worker_timeout is not None:
             worker_timeout = task.worker_timeout
         self.timeout_time = time.time() + worker_timeout if worker_timeout else None
-        self.random_seed = random_seed or self.timeout_time is not None
+        self.use_multiprocessing = use_multiprocessing or self.timeout_time is not None
 
     def _run_get_new_deps(self):
         self.task.set_tracking_url = self.tracking_url_callback
@@ -146,7 +146,7 @@ class TaskProcess(multiprocessing.Process):
     def run(self):
         logger.info('[pid %s] Worker %s running   %s', os.getpid(), self.worker_id, self.task)
 
-        if self.random_seed:
+        if self.use_multiprocessing:
             # Need to have different random seeds if running in separate processes
             random.seed((os.getpid(), time.time()))
 
@@ -782,7 +782,7 @@ class Worker(object):
 
         self._running_tasks[task_id] = task_process
 
-        if task_process.random_seed:
+        if task_process.use_multiprocessing:
             with fork_lock:
                 task_process.start()
         else:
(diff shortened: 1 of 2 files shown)


luigi/worker.py
```diff
@@ -786,7 +782,7 @@ def _run_task(self, task_id):
 
         self._running_tasks[task_id] = p
 
-        if self.worker_processes > 1:
+        if p.random_seed:
```
Tarrasch: Can we rename `p` to `task_process`?

luigi/worker.py
```diff
@@ -108,10 +108,10 @@ def __init__(self, task, worker_id, result_queue, tracking_url_callback,
         self.result_queue = result_queue
         self.tracking_url_callback = tracking_url_callback
         self.status_message_callback = status_message_callback
-        self.random_seed = random_seed
         if task.worker_timeout is not None:
             worker_timeout = task.worker_timeout
         self.timeout_time = time.time() + worker_timeout if worker_timeout else None
+        self.random_seed = random_seed or self.timeout_time is not None
```
Tarrasch: * What is `random_seed`?
 * Is there any code-path where it's not set?
 * Is there any code-path but [this one](https://github.com/Houzz/luigi/blob/afe01d32eb8a7fac1249554a8a658393a66a530e/luigi/worker.py#L149-L151) where it's used?

To me it seems like we can remove it entirely. For a moment I thought it could be a `TaskProcess` thingie, but I only found luigi-related results when [googling](https://www.google.com/search?client=ubuntu&channel=fs&q=multiprocess.Process+random_seed&ie=utf-8&oe=utf-8) it.

luigi/worker.py
```diff
@@ -410,11 +410,7 @@ def __init__(self, scheduler=None, worker_id=None, worker_processes=1, assistant
                 pass
 
         # Keep info about what tasks are running (could be in other processes)
-        if worker_processes == 1:
-            self._task_result_queue = DequeQueue()
```
Tarrasch: Interesting. Before this patch, was this just a premature optimization done before?

luigi/worker.py
daveFNbuck: It won't be set if you're running a task that doesn't have a timeout on 1 worker.

luigi/worker.py
daveFNbuck: Yeah we didn't need a multiprocess queue so we used the single-process one here.

luigi/worker.py
daveFNbuck: Yeah, but I like to keep PRs focused. We can do that as a separate change.

luigi/worker.py
Tarrasch: Sorry. By set I don't mean `True`/`False`/`Whatever`. I mean that we shouldn't have any default values as they are not used anywhere. That's my claim.

Furthermore. I still don't see anywhere where it's used. Can you convince me that there's a codepath where it's used? From what I see it's not used anywhere.

---

Second read-through. [Here](https://github.com/Houzz/luigi/blob/afe01d32eb8a7fac1249554a8a658393a66a530e/luigi/worker.py#L149-L151) it's referenced. I guess we should rename this variable to `use_random_seed`. As it's not the random seed itself we're providing.

luigi/worker.py
Tarrasch: I like PR's to be focused as well. But if you decide to only minimally touch hard-to-understand code, it will remain hard-to-understand for the reviewers too. I would appreciate it if you could change the name of this local variable at this one place. Not strictly necessary though (as with all my suggestions :)).

luigi/worker.py
daveFNbuck: I've renamed random_seed to use_multiprocessing so it makes sense in more contexts.

luigi/worker.py
daveFNbuck: Added to this PR as a separate commit.

Tarrasch:
A nice separation of commits worthy of a merge-commit! :)

daveFNbuck:
Thanks!

PR #1861, opened 2016-09-20
Enables running with 0 workers to support assistants

daveFNbuck:
## Description
Changes how worker.run works to make it compatible with running 0 workers. This allows scheduler verification of whether the worker needs to stay alive.

## Motivation and Context
In order to allow assistants to take care of running all tasks, we need a
way to run scheduling with 0 workers. This adds a slightly tweaked worker
code path to allow a process with 0 workers to stay alive while skipping
any get_work calls to the scheduler. Instead, the worker makes a
count_pending call to see if it's still necessary.

count_pending returns the pending count and unique count as well as a new
pending last scheduled count, which counts the number of pending tasks
which haven't been scheduled by another worker since the current worker
first scheduled it. This last count is important because it allows enough
workers to stay alive to keep all tasks scheduled. If we use unique counts
to keep workers alive, all workers could die at the same time leaving
nothing running. If we use pending counts, all workers will stay alive,
preventing new tasks from being scheduled via locks.

In order to minimize code changes, the pending last scheduled count is
stored into the same variable used for unique pending counts in standard
workers. This makes slightly misleading messages but that can be fixed
later if it bothers anyone.

These workers should be used with keep_alive and count_uniques both set
to true. I'll add documentation about this to the luigi patterns doc once
I've solved more of the issues I'm having with assistant-based workloads.

I have included unit tests and I've been using this in production for about a month.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -200,38 +200,72 @@ def _get_default(x, default):
         return default
 
 
-class OrderedSet(object):
-    def __init__(self):
-        self._order = set()
-        self._items = []
+class OrderedSet(collections.MutableSet):
+    """
+    Standard Python OrderedSet recipe found at http://code.activestate.com/recipes/576694/
 
-    def add(self, item):
-        if item not in self._items:
-            self._order.add(item)
-            self._items.append(item)
+    Modified to include a peek function to get the last element
+    """
 
-    def _filter_items(self):
-        self._items = [item for item in self._items if item in self._order]
-
-    def difference_update(self, removed_items):
-        self._order.difference_update(removed_items)
-        self._filter_items()
-
-    def __contains__(self, item):
-        return item in self._order
-
-    def __nonzero__(self):
-        return bool(self._items)
+    def __init__(self, iterable=None):
+        self.end = end = []
+        end += [None, end, end]         # sentinel node for doubly linked list
+        self.map = {}                   # key --> [key, prev, next]
+        if iterable is not None:
+            self |= iterable
 
     def __len__(self):
-        return len(self._items)
+        return len(self.map)
+
+    def __contains__(self, key):
+        return key in self.map
+
+    def add(self, key):
+        if key not in self.map:
+            end = self.end
+            curr = end[1]
+            curr[2] = end[1] = self.map[key] = [key, curr, end]
+
+    def discard(self, key):
+        if key in self.map:
+            key, prev, next = self.map.pop(key)
+            prev[2] = next
+            next[1] = prev
 
     def __iter__(self):
-        return iter(self._items)
+        end = self.end
+        curr = end[2]
+        while curr is not end:
+            yield curr[0]
+            curr = curr[2]
 
-    def __getitem__(self, idx):
-        return self._items[idx]
+    def __reversed__(self):
+        end = self.end
+        curr = end[1]
+        while curr is not end:
+            yield curr[0]
+            curr = curr[1]
 
+    def peek(self, last=True):
+        if not self:
+            raise KeyError('set is empty')
+        key = self.end[1][0] if last else self.end[2][0]
+        return key
+
+    def pop(self, last=True):
+        key = self.peek(last)
+        self.discard(key)
+        return key
+
+    def __repr__(self):
+        if not self:
+            return '%s()' % (self.__class__.__name__,)
+        return '%s(%r)' % (self.__class__.__name__, list(self))
+
+    def __eq__(self, other):
+        if isinstance(other, OrderedSet):
+            return len(self) == len(other) and list(self) == list(other)
+        return set(self) == set(other)
 
 class Task(object):
     def __init__(self, task_id, status, deps, resources=None, priority=0, family='', module=None,
@@ -599,7 +633,7 @@ class SimpleTaskState(object):
         for task in self.get_active_tasks():
             if remove_stakeholders:
                 task.stakeholders.difference_update(workers)
-            task.workers.difference_update(workers)
+            task.workers -= workers
 
     def disable_workers(self, worker_ids):
         self._remove_workers_from_tasks(worker_ids, remove_stakeholders=False)
@@ -885,7 +919,7 @@ class Scheduler(object):
             elif task.status == PENDING:
                 num_pending += 1
                 num_unique_pending += int(len(task.workers) == 1)
-                num_pending_last_scheduled += int(task.workers[-1] == worker_id)
+                num_pending_last_scheduled += int(task.workers.peek(last=True) == worker_id)
         return {
             'n_pending_tasks': num_pending,
             'n_unique_pending': num_unique_pending,
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -867,7 +867,7 @@ class Scheduler(object):
     def count_pending(self, worker):
         worker_id, worker = worker, self._state.get_worker(worker)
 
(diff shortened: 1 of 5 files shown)


luigi/scheduler.py
```diff
@@ -830,6 +836,26 @@ def _reset_orphaned_batch_running_tasks(self, worker_id):
...
+            num_unique_pending += int(len(task.workers) == 1)
+            if task.workers:
+                last_scheduled = max(task.workers, key=worker_starts.get)
+                num_last_scheduled += int(last_scheduled == worker_id)
+        return {
+            'n_pending_tasks': num_pending,
+            'n_unique_pending': num_unique_pending,
+            'n_pending_last_scheduled': num_last_scheduled,
+            'worker_state': worker.state,
+        }
```
Tarrasch: Did you think about if we were able to refactor out some of this logic to share between this method and `get_work`?

luigi/scheduler.py
```diff
@@ -556,6 +556,12 @@ def get_worker_ids(self):
     def get_worker(self, worker_id):
         return self._active_workers.setdefault(worker_id, Worker(worker_id))
 
+    def get_worker_starts(self):
```
Tarrasch: what is this??? Does it mean `get_started_workers`? If you can't rename it to something self-describing. Can you add a comment instead?

luigi/scheduler.py
```diff
@@ -830,6 +836,26 @@ def _reset_orphaned_batch_running_tasks(self, worker_id):
...
+        for task in worker.get_pending_tasks(self._state):
+            num_pending += 1
+            num_unique_pending += int(len(task.workers) == 1)
+            if task.workers:
+                last_scheduled = max(task.workers, key=worker_starts.get)
+                num_last_scheduled += int(last_scheduled == worker_id)
+        return {
+            'n_pending_tasks': num_pending,
+            'n_unique_pending': num_unique_pending,
+            'n_pending_last_scheduled': num_last_scheduled,
```
Tarrasch: I think you should rename the one to match the other. like

`n_pending_last_scheduled': num_pending_last_scheduled`

or

`n_last_scheduled': num_last_scheduled`

luigi/worker.py
```diff
@@ -732,25 +732,37 @@ def _get_work_task_id(self, get_work_response):
...
+                'running_tasks': running_tasks,
+            })
+
+        # Just keeping tasks alive for assistants
+        else:
+            logger.debug("Checking if tasks are still pending")
+            r = self._scheduler.count_pending(worker=self._id)
+            n_pending_tasks = r['n_pending_tasks']
+            running_tasks = 0
+            n_unique_pending = r['n_pending_last_scheduled']
```
Tarrasch: Same here. Rename one to match the other.

luigi/scheduler.py
Tarrasch: aha, is it `get_worker_starttimes`?

luigi/worker.py
Tarrasch: Hmmm... Reading this again, I feel this is not only a name mis-match, but they actually sound like two different things. Even if worker X was last to schedule task T it's not necessarily unique to that worker.

daveFNbuck: Yeah, I did this intentionally to avoid making other annoying changes, as mentioned in the description of this PR. I guess I should just put in the extra work and do it right now.

luigi/scheduler.py
daveFNbuck: yes

luigi/scheduler.py
daveFNbuck: Yeah, I bet we could actually just call count_pending at the end of get_work rather than doing it inline without significantly affecting runtimes. Pulling the logic out and sharing it between both functions would probably be more awkward, confusing and slow. If we add running_tasks to count_pending, we could even exit the get_work loop early when we pick a best_task that isn't batchable.

luigi/scheduler.py
Tarrasch: Would you mind changing the name for clarity? :)

luigi/scheduler.py
Tarrasch: Okay. As long as you considered it seriously I trust you to do a smart thing. :)

luigi/scheduler.py
daveFNbuck: Yeah, as soon as I have some time today I'll update everything.

daveFNbuck: Turns out this function really slows down the quadratic behavior test because it makes count_pending linear in the number of workers. I've instead written a quick SortedSet data structure so we can just check the last worker id in Task.workers.

luigi/scheduler.py
```diff
@@ -200,12 +200,45 @@ def _get_default(x, default):
         return default
 
 
+class OrderedSet(object):
```
Tarrasch: Is this copy-pasted from somewhere? If yes, then link to that place, if not, write it down explicitly as a comment.

luigi/worker.py
```diff
@@ -94,6 +94,16 @@ class TaskException(Exception):
     pass
 
 
+GetWorkResponse = collections.namedtuple('GetWorkResponse', (
+    'task_id',
+    'running_tasks',
+    'n_pending_tasks',
+    'n_unique_pending',
+    'n_pending_last_scheduled',
+    'worker_state'
```
Tarrasch: Add trailing comma

luigi/worker.py
```diff
@@ -789,7 +802,20 @@ def _get_work(self):
...
+        return GetWorkResponse(
+            task_id=task_id,
+            running_tasks=running_tasks,
+            n_pending_tasks=r['n_pending_tasks'],
+            n_unique_pending=r['n_unique_pending'],
+
+            # TODO: For a tiny amount of time (a month?) we'll keep forwards compatibility
+            #  That is you can user a newer client than server (Sep 2016)
+            n_pending_last_scheduled=r.get('n_pending_last_scheduled', 0),
+            worker_state=r.get('worker_state', 0),
```
Tarrasch: Shouldn't `0` be `WORKER_STATE_ACTIVE`? And why is there both `get()` and `setdefault`?

daveFNbuck: Bad refactoring on my part as I rewrote this a few times. Fixed now. The setdefault hid the bad get.

luigi/worker.py
daveFNbuck: done

luigi/scheduler.py
daveFNbuck: I had written it myself but after thinking a bit more I've decided it would be better to use the recipe at http://code.activestate.com/recipes/576694/ which is mentioned at the bottom of the [collection docs](https://docs.python.org/2/library/collections.html).

daveFNbuck:
A squash merge should be fine, unless there's something you wanted in a separate commit.

luigi/scheduler.py
```diff
@@ -963,12 +1041,7 @@ def get_work(self, host=None, assistant=False, current_tasks=None, worker=None,
 
                             break
 
-        reply = {'n_pending_tasks': locally_pending_tasks,
-                 'running_tasks': running_tasks,
-                 'task_id': None,
-                 'n_unique_pending': n_unique_pending,
-                 'worker_state': worker.state,
-                 }
```
Tarrasch: I've hated this code so much ...

daveFNbuck: Yeah, get_work is a little less crazy now :)

PR #1862, opened 2016-09-21
Add u prefix in the format statement to avoid encoding error

rossilxx:
## Description

Fix encoding error issue
## Motivation and Context

The `stdout` and `stderr` is an unicode object after decoded in the _clean_output_file function
And if we have some character not in ascii range, the python `format` function will return an error. 

It works fine in my production environment. python 2.7


Tarrasch:
Can you illustrate the error by adding a test-case?

Tarrasch:
Cool. Once you get the tests to pass we'll merge this! :)

PR #1863, opened 2016-09-23
Add support to RangeByMinutes backfilling tasks

j-santander:
## Description
This change adds RangeByMinutesBase (inheriting from existing RangeBase) and RangeByMinutes (inheriting from RangeByMinutesBase) providing support for back-filling classes operating in units of minutes.

Behavior (and coding) follows closely the existing RangeHourly and RangeDaily classes. Only addition is 
the ability to specify the number of minutes considered as one interval (defaulting to 5 minutes)

## Motivation and Context

In our project we produce reports on 5 minutes intervals and the existing backfilling tasks do not cover
such a short interval.

We've tested the code using on our existing jobs, working succesfully.

In addition, I've added additional test cases to test/range_test.py, following the same structure as existing cases for hours and days intervals.

### DIFF
--- a/luigi/tools/range.py
+++ b/luigi/tools/range.py
@@ -428,7 +428,7 @@ class RangeByMinutesBase(RangeBase):
 
     minutes_interval = luigi.IntParameter(
         default=1,
-        description="separation between events in minutes"
+        description="separation between events in minutes. It must evenly divide 60"
     )
 
     def datetime_to_parameter(self, dt):
@@ -460,6 +460,11 @@ class RangeByMinutesBase(RangeBase):
         """
         Simply returns the points in time that correspond to a whole number of minutes intervals.
         """
+        # Validate that the minutes_interval can divide 60 and it is greater than 0.
+        if self.minutes_interval <= 0:
+            raise ParameterException('minutes-interval must be > 0')
+        if (60 / self.minutes_interval) * self.minutes_interval != 60:
+            raise ParameterException('minutes-interval does not evenly divide 60')
         # start of a complete interval, e.g. 20:13 and the interval is 5 -> 20:10
         start_minute = int(finite_start.minute/self.minutes_interval)*self.minutes_interval
         datehour_start = datetime(
--- a/luigi/tools/range.py
+++ b/luigi/tools/range.py
@@ -402,6 +402,84 @@ class RangeHourlyBase(RangeBase):
         return luigi.DateHourParameter().serialize(dt)
 
 
+class RangeByMinutesBase(RangeBase):
+    """
+    Produces a contiguous completed range of an recurring tasks separated a specified number of minutes.
+    """
+    start = luigi.DateMinuteParameter(
+        default=None,
+        description="beginning date-hour-minute, inclusive. Default: None - work backward forever (requires reverse=True)")
+    stop = luigi.DateMinuteParameter(
+        default=None,
+        description="ending date-hour-minute, exclusive. Default: None - work forward forever")
+    minutes_back = luigi.IntParameter(
+        default=60*24,  # one day
+        description=("extent to which contiguousness is to be assured into "
+                     "past, in minutes from current time. Prevents infinite "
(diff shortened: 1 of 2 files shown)


luigi/tools/range.py
```diff
@@ -402,6 +402,84 @@ def _format_datetime(self, dt):
...
+                     "be set shorter to that, too, to prevent the oldest "
+                     "outputs flapping. Increase freely if you intend to "
+                     "process old dates - worker's memory is the limit"))
+    minutes_forward = luigi.IntParameter(
+        default=0,
+        description="extent to which contiguousness is to be assured into future, "
+                    "in minutes from current time. Prevents infinite loop when stop is none")
+
+    minutes_interval = luigi.IntParameter(
+        default=5,
```
Tarrasch: This seems way to arbitrary. Why not just `default=1`?

j-santander: @Tarrasch, thanks for reviewing it. 
Not sure what you mean with your first comment, is it just the choice of name of RangeByMinutes vs RangeMinutely? English is not my mother tongue, but [minutely ](http://english.stackexchange.com/questions/3091/weekly-daily-hourly-minutely)didn't sound correct to me.

About the default value of 5 minutes interval. Ultimately it responds to the fact it is the interval that was needed in our project. I could justify it as being a reasonable interval in the minutes range, making every minute a too frequent interval... but it is not trouble to change it.

Tarrasch:
@j-santander I think your name of `RangeByMinutes` makes sense. You're right that `RangeMinutely` is weird English. I can't think of a better name than `RangeByMinutes`. Does @ulzha have suggestions?

----

Thanks for changing the default to `=1`. I also realize your point that as low as `1` very rarely is useful as it's so frequent.

----

A lost point which I would like you to fix. Can you throw a `ParameterException` if the value does not evently divide 60? Surely it's still possible to have 7 or 8, but it's certainly a very bad idea. Can you add that to the patch and have a small test case alongside it?

j-santander:
Added the validations, let me know if anything else is needed.
Thanks
Julian

luigi/tools/range.py
```diff
@@ -460,6 +460,11 @@ def finite_datetimes(self, finite_start, finite_stop):
         """
         Simply returns the points in time that correspond to a whole number of minutes intervals.
         """
+        # Validate that the minutes_interval can divide 60 and it is greater than 0.
+        if self.minutes_interval <= 0:
```
Tarrasch: While you're at it, can you change this to

`if not (0 < self.minutes_interval < 60)`

test/range_test.py
```diff
@@ -592,6 +592,49 @@ def missing_datetimes(a, b, c):
...
+        # finite_datetimes constitute many years to consider
+        task = RangeByMinutes(now=datetime_to_epoch(datetime.datetime(2016, 4, 1)),
+                              of=SomeByMinutesTask,
+                              start=datetime.datetime(2014, 3, 20, 17),
+                              minutes_interval=-1)
+        try:
+            task.requires()
+        except luigi.parameter.ParameterException:
+            return
+        self.fail("Expected a parameter exception")
```
Tarrasch: Can you use this helper instead? https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises

test/range_test.py
```diff
@@ -592,6 +592,49 @@ def missing_datetimes(a, b, c):
...
+        class SomeByMinutesTask(luigi.Task):
+            d = luigi.DateMinuteParameter()
+
+            def output(self):
+                return MockTarget(self.d.strftime('/data/2014/p/v/z/%Y_/_%m-_-%doctor/20/%HZ%MOOO'))
+
+        for task_path in task_a_paths:
+            MockTarget(task_path)
+        # this test takes a few seconds. Since stop is not defined,
+        # finite_datetimes constitute many years to consider
```
Tarrasch: Can we remove this comment? (and some for the other test)

Tarrasch:
Thanks @j-santander! :)

PR #1866, opened 2016-09-26
Add connection options for sqlalchemy in db history classes by using connection string query params.

oivoodoo:
## Description
1. `connection_string` is passing by config in luigi.
2. `connection_options` is query params of `connection_string`

Example:
`postgresql://username:password@example.com:5432/luigi?pool_size=25&pool_timeout=50`
## Motivation and Context

On actively using luigi(db history) in our master node I got timeout issues on workers on pinging master node. When I discovered luigi logs, I found that db history is producing timeout issues on connecting to our PostgreSQL instance. I've decided to increase the pool size from default 5 upto 25.

I deployed this change by using git tarball url in our `requirements.txt` and it's working in the prod environment properly.
## Questions:

As part of the change I used `future` library to support Python 2 on using `urllib.parse` module. is it good enough way for this change?  How to add `future` library as dependency properly?

### DIFF
--- a/luigi/db_task_history.py
+++ b/luigi/db_task_history.py
@@ -34,6 +34,8 @@ See :ref:`TaskHistory` for information about how to turn out the task history fe
 # Copyright 2015 Naver Corp.
 # Author Yeseul Park (yeseul.park@navercorp.com)
 #
+from future.standard_library import install_aliases
+install_aliases()
 
 import datetime
 import logging
@@ -44,6 +46,8 @@ from luigi import six
 from luigi import configuration
 from luigi import task_history
 from luigi.task_status import DONE, FAILED, PENDING, RUNNING
+
+from urllib.parse import urlparse, parse_qs
 
 import sqlalchemy
 import sqlalchemy.ext.declarative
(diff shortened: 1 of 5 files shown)


Tarrasch:
The future package is still something that's in beta (still no 1.0). Was there no way to skip including it?

----

I still don't exactly understand what this patch does. Does this patch basically make it so that we don't ignore query-params (things after `?`) in the task_history_db database-URL?

oivoodoo:
@Tarrasch yeah. it's not ignoring query params. it converts them to `sqlalchemy` database connection settings if you want to include `pool_size`, `max_overflow` or other options. I will skip future.

oivoodoo:
@Tarrasch going to apply few more changes for this pull request.

oivoodoo:
@Tarrasch I don't know exactly how to do it better. But in `create_engine` we can pass additional options ([sqlalchemy doc](http://docs.sqlalchemy.org/en/latest/core/engines.html)) . in our luigi master server we are using db history to review the previous success and failed jobs. and sometimes it's causing the issues with db connection too much sql actions and it can't provide connections. as result I increased `max_overflow` and `pool_size`.

At the beginning I was thinking about to use in connection string query params(it's similar approach like when you setup the instance in heroku and add your postgresql addon(`DATABASE_URL` could contain `pool_size` there).  But sqlalchemy could receive `int` or `string` depends on the one possible connection option. As result I deleted query params and move few of the significant options as part of config in `history` section with prefix `db_`.

Tarrasch:
@oivoodoo If you've tested both params-in-connection-url vs config variables and found that the latter was better. I trust you on that.

Please go ahead and improve the code quality. If you can find somebody else to review this that would be good too.

oivoodoo:
Hi @Tarrasch. Is it look better now? I've decided to leave ability to extend by adding more options to `CONNECTION_OPTIONS`.

oivoodoo:
Hi @Tarrasch. Do you have some time to review it again? Thank you!

luigi/db_task_history.py
```diff
@@ -79,13 +78,32 @@ def _session(self, session=None):
...
+            'db_max_overflow': {'type': 'int', 'name': 'max_overflow'},
+            'db_pool_timeout': {'type': 'int', 'name': 'pool_timeout'},
+            'db_pool_recycle': {'type': 'int', 'name': 'pool_recycle'},
+            }
+
+    def _prepare_connection_options(self, config):
+        options = {}
+        for (config_key, column_settings) in six.iteritems(self.CONNECTION_OPTIONS_MAPPING):
+            column_name = column_settings['name']
+            if column_settings['type'] == 'int':
```
Tarrasch: This is a `_private` function, yet you anticipate something else than `'int'`, but `'int'` is hard-coded in the code already. This can be simplified.

PR #1867, opened 2016-09-26
Prevents batched tasks from all getting marked RUNNING

daveFNbuck:
## Description
Adjusts SimpleTaskState.set_status to not set batched tasks to RUNNING when setting an already running batch task to RUNNING again.

## Motivation and Context
I've recently seen some of my workers simultaneously running multiple tasks from the same batchable class, when the should all be running together as a single batch. Furthermore, these tasks exceeded both resource constraints and the number of workers supposed to be running at the same time.

After checking the logs, I saw that the worker would actually run these one at a time upon receiving each one from the scheduler. My best guess after looking into it for a week is that the worker does not receive a response from the scheduler after marking the batch runner as RUNNING, and the scheduler updates the entire batch's status on the retry. I verified that this is indeed a possibility with a unit test and have patched that bug here.

I have included unit tests.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -532,7 +532,7 @@ class SimpleTaskState(object):
             elif task.scheduler_disable_time is not None and new_status != DISABLED:
                 return
 
-        if task.status == RUNNING and task.batch_id is not None:
+        if task.status == RUNNING and task.batch_id is not None and new_status != RUNNING:
             for batch_task in self.get_batch_running_tasks(task.batch_id):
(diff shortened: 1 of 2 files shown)


daveFNbuck:
It's been very helpful for me, I'm sure more people will come to love it
soon too :)

Tarrasch:
@daveFNbuck congrats on luigi's first merge with rebase strategy! https://github.com/spotify/luigi/commit/6a8ce3aab9346100ac9008d2aa9db6c5eacf672e :D

PR #1868, opened 2016-09-26
Uses task name instead of task_id as job identifier for MR

daveFNbuck:
## Description
Uses task name instead of task_id as job identifier for MR

## Motivation and Context
It's much more useful to see MyTask(param=value) in the job tracker rather than MyTask_value_abf1238eb238eb1832, especially when the latter form usually removes most of the useful information for figuring out which task is which. Since changing the task id format, we're no longer intending for task ids to be human-readable, and this is a context where we want something human-readable.

I updated the unit tests and ran MRs with this code to validate that their names are now human-readable.


Tarrasch:
Thanks!

PR #1871, opened 2016-10-01
Moves e-mail config into luigi.Config classes

daveFNbuck:
## Description
Moves all configuration access in notifications.py to use luigi.Config classes.

Also sets a default 10 second timeout for smtp.

## Motivation and Context
Today when I was trying to figure out why my luigi server stopped responding at 5am, my only clue was that the last line in the log had it sending an e-mail. When I went to the notifications code to see if e-mail has a timeout, I found the config fetching code a bit outdated and hard to read. So I decided to move it to a luigi.Config class like we use in worker and scheduler. This ended up cascading until I had moved all the config accesses into Config classes. As part of this, the parameters are no longer under `[core]` but all current settings are still backward-compatible.

Since I had started this due to a nasty result of not having an smtp timeout, I gave timeout a 10-second default while I was at it.

I left the unit tests mostly as-is and they remained passing (except for the ones that don't pass on my machine in master). I only adjusted for changed function signatures in the send_email functions.

I deployed my server with unchanged configuration and triggered a disable e-mail and got it.

I ran the docs test and opened the configuration docs to verify they look right.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -105,34 +105,6 @@ default-scheduler-url
   non-standard URI scheme: ``http+unix``
   example: ``http+unix://%2Fvar%2Frun%2Fluigid%2Fluigid.sock/``
 
-email-prefix
-  Optional prefix to add to the subject line of all e-mails. For
-  example, setting this to "[LUIGI]" would change the subject line of an
-  e-mail from "Luigi: Framework error" to "[LUIGI] Luigi: Framework
-  error"
-
-email-sender
-  User name in from field of error e-mails.
-  Default value: luigi-client@<server_name>
-
-email-type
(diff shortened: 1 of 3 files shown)


Tarrasch:
Yes, yes, yes! :)

There still seems to be some valid travis failures. Can you look into it?

daveFNbuck:
Yeah, it broke my pipeline over the weekend too, I'll be fixing it first
thing tomorrow. I guess I should have tested it with workers too before
deploying and leaving...

daveFNbuck:
Updated. Glad you like the new name :)

Tarrasch:
This is the kind of patch that make a maintainers day! Great job! :)

daveFNbuck:
Glad you like it :)

PR #1876, opened 2016-10-08
busybox compatibility for ps command usage in getpcmd, also use proc …

nmandery:
…filesystem first

This pull request makes the lock.getpcmp function compatible with the busybox implementation of the `ps` command. Also the function now attempts to use the `/proc` filesystem first.

## Description

The incompatibility to busybox `ps` came from the `ww` flags added to the command in PR #1809.

## Motivation and Context

This avoids error messages when using luigi for example on android devices.

Manually tested the function. Also ran the unittests, but these give me currently 8 errors on the master branch without any of my changes.

### DIFF
--- a/luigi/lock.py
+++ b/luigi/lock.py
@@ -23,6 +23,7 @@ from __future__ import print_function
 
 import hashlib
 import os
+import subprocess
 
 from luigi import six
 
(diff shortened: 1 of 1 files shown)


Tarrasch:
I'm okay with luigi not working for everyone on every OS. Android feels like one such operating system. I'm suggesting you don't keep the backwards compatibility so the code can be simplified *a lot*

nmandery:
I see, busybox compatibility may indeed stretch the scope of luigi a bit, although android devices are geting more and more powerful. Are you interested in the part of the pull-request using the /proc-filesystem when I just drop the part regarding the `ps` command?

Tarrasch:
Yes. In fact I'm very interested. I'm sorry if my gratefulness didn't shine through in my previous comment. :)

However, this is what I suggest for now:

 * Close this PR.
 * Open a new one, but only keep the simplyfing `ps`-removing part. And as a comment in the code mentioning the limitation (busybox/android etc), and link to this PR.

Does that make sense?

nmandery:
No worries ;)

I just checked [this article](https://www.kernel.org/doc/Documentation/filesystems/proc.txt) and indeed the `ps` command just reads the /proc filesystem:
 
> In  fact,  ps  uses  the  proc  file  system  to  obtain its information.

I guess there may be some setuid magic which allows `ps` to gather information from /proc without the user being able to read the filesystem directly.

Just to see if I understand you correctly: I am going to remove the parts using `ps` completely and add a few notes regading my experience on android.

Tarrasch:
> Just to see if I understand you correctly: I am going to remove the parts using ps completely and add a few notes regading my experience on android.

Exactly, I would even link back to this PR. The purpose would be that you're first implementation isn't deleted, in case we some day in the future would like to actually support `ps` as fallback.

PR #1877, opened 2016-10-08
Allow a file to be moved cross-device.

miku:
## Description

Implementation of `luigi.File` would now allow movement of a file across filesystems.

## Motivation and Context

If you use `luigi.File(src).move(target)` when source and target are on different filesystems, you get an error:

```
OSError: [Errno 18] Invalid cross-device link
```

Also, this would align implementation of `File.move()` with `File.copy()` which already uses `shutil`.

Semantic stays the same. Existing tests apply.

> If the destination is on the current filesystem, then os.rename() is used. - [shutil.move](https://docs.python.org/3/library/shutil.html#shutil.move)

Important tradeoff: If you move a file across filesystems, you lose atomicity of `os.rename`.

### DIFF
--- a/test/file_test.py
+++ b/test/file_test.py
@@ -32,7 +32,7 @@ from target_test import FileSystemTargetTestMixin
 
 import itertools
 import io
-from errno import EEXIST
+from errno import EEXIST, EXDEV
 
 
 class LocalTargetTest(unittest.TestCase, FileSystemTargetTestMixin):
@@ -129,6 +129,32 @@ class LocalTargetTest(unittest.TestCase, FileSystemTargetTestMixin):
         t.move(self.copy)
         self.assertFalse(os.path.exists(self.path))
         self.assertTrue(os.path.exists(self.copy))
+
+    def test_move_across_filesystems(self):
+        t = LocalTarget(self.path)
+        with t.open('w') as f:
+            f.write('test_data')
+
+        def rename_across_filesystems(src, dst):
+            err = OSError()
+            err.errno = EXDEV
+            raise err
+
+        real_rename = os.rename
+
+        def mockrename(src, dst):
+            if '-across-fs' in src:
+                real_rename(src, dst)
+            else:
+                rename_across_filesystems(src, dst)
+
+        copy = '%s-across-fs' % self.copy
+        with mock.patch('os.rename', mockrename):
+            t.move(copy)
+
+        self.assertFalse(os.path.exists(self.path))
+        self.assertTrue(os.path.exists(copy))
+        self.assertEqual('test_data', LocalTarget(copy).open('r').read())
 
     def test_format_chain(self):
         UTF8WIN = luigi.format.TextFormat(encoding='utf8', newline='\r\n')
--- a/luigi/file.py
+++ b/luigi/file.py
@@ -25,6 +25,7 @@ import shutil
 import tempfile
 import io
 import warnings
+import errno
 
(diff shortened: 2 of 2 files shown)


erikbern:
lgtm

miku:
Thanks for the review, @Tarrasch, I tried to address your points. As for the docs, I didn't find a perfect place yet, so added a short section to the tasks docs.

erikbern:
> Mention the atomicity things in the docs? Atomicity of renames are a very important thing in luigi. If it can't be maintained for some functions, the docs should say under what condition.

pretty sure cross device file moving will never be atomic

miku:
@erikbern, closest thing to atomic would probably be: copy file to a temp location on the target device, rename it to destination, then unlink the source? But I wanted to keep the code short.

Luigi allows non-atomicity, e.g. in tasks with multiple outputs:

> If multiple outputs are returned, atomicity will be lost ... -- [Task.output](https://luigi.readthedocs.io/en/stable/tasks.html#task-output)

So I thought it would be ok to drop atomicity if this special case, if the user chooses to (chooses to work with two file systems).

test/file_test.py
```diff
@@ -130,6 +130,24 @@ def test_move(self):
         self.assertFalse(os.path.exists(self.path))
         self.assertTrue(os.path.exists(self.copy))
 
+    def test_move_across_filesystems(self):
+        t = LocalTarget(self.path)
+        f = t.open('w')
+        f.write('test_data')
+        f.close()
```
Tarrasch: Hmm, can't you do the much more elegant context-manager thingie? `with`-statements.

doc/tasks.rst
```diff
@@ -307,3 +307,19 @@ Luigi also does some metaclass logic so that
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Atomicity is crucial when working with targets. In the following scenarios
+atomicity cannot be guaranteed.
+
+Moving of a :class:`~luigi.target.LocalTarget` across filesytems:
+
+.. code:: python
+
+    luigi.LocalTarget(path=src).move(dst)
```
Tarrasch: Wouldn't it be much more easy to put this in the docs for `LocalTarget.move()` docs?? You don't even need a code example. Just a warning saying to be aware of the risks when doing this across filesystems's.

erikbern:
> @erikbern, closest thing to atomic would probably be: copy file to a temp location on the target device, rename it to destination, then unlink the source? But I wanted to keep the code short.

great idea, let's do this

doc/tasks.rst
miku: Yes, you're right. I addressed this the [next PR](https://github.com/spotify/luigi/pull/1877/commits/93f064d3bc0a7602b65a6af532a7f0f20f2c7dc0).

miku:
> great idea, let's do this

Thanks for the feedback. I implemented an almost-atomic cross-device move operation and updated the docs. (I'll be AFK for ten days, can only look into this when I'm back).

Tarrasch:
Does this work for you in production as well? If so, let's merge. :)

erikbern:
lgtm – nice!

miku:
> Does this work for you in production as well?

@Tarrasch, yes it does.

Background: We have `$TEMPDIR` and data on different partitions plus we sometimes do writes to a tempfile first, then *manually* move the file into place with a 

    luigi.LocalTarget(sometempfile).move(self.output().path)

miku:
Thanks!

PR #1878, opened 2016-10-11
Fixed bug where DataprocPysparkTask overrides the job name parameter …

matt-miller:
## Description
The name of the main pyspark script was hardcoded to be main_job.py, it now uses the provided parameter.

I have made my own version of DataprocPysparkTask with this change to successfully run pyspark jobs on Dataproc, whereas the original version failed due to an inability to find main_job.py. Also note that there is a unit test for this that failed to catch the issue because the unit test happened to use main_job.py as its main job name.

### DIFF
--- a/luigi/contrib/dataproc.py
+++ b/luigi/contrib/dataproc.py
@@ -126,7 +126,7 @@ class DataprocPysparkTask(DataprocBaseTask):
     job_args = luigi.Parameter(default="")
 
     def run(self):
-        self.submit_pyspark_job(job_file="main_job.py",
+        self.submit_pyspark_job(job_file=self.job_file,
                                 extra_files=self.extra_files.split(",") if self.extra_files else [],
                                 job_args=self.job_args.split(",") if self.job_args else [])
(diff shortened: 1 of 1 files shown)


dlstadther:
I don't use dataproc, but based on code lgtm.

Tarrasch:
Can you also fix the unit test?

constantijn:
Change looks good to me ... not sure the unit test was broken, seems I forgot to replace a string (the job name) with the actual parameter passed in, and instead passed in a hard coded string of the job file i normally use. The test is there, it just didn't show up because the value I was using for the parameter was the same as the hard coded string i accidentally left in. 

Speaking of tests ... do you have any idea why travis is reporting an error? It seems the running of the tests on some other part of the code broke

matt-miller:
The unit test is fine, it just happened to not catch the bug.

I have no idea what the Travis errors are all about. I distinctly remember master having a failed build when I forked the repo yesterday but looking back at the Travis builds I can't find any evidence to support this claim.

PR #1881, opened 2016-10-13
Add UDF support for BigQuery

ukarlsson:


luigi/contrib/bigquery.py
```diff
@@ -594,7 +600,8 @@ def run(self):
                     'allowLargeResults': True,
                     'createDisposition': self.create_disposition,
                     'writeDisposition': self.write_disposition,
-                    'flattenResults': self.flatten_results
+                    'flattenResults': self.flatten_results,
+                    'userDefinedFunctionResources': [{"resourceUri": v} for v in self.udf_resource_uris]
```
Tarrasch: Add a trailing comma please

luigi/contrib/bigquery.py
```diff
@@ -567,6 +567,12 @@ def query_mode(self):
         """The query mode. See :py:class:`QueryMode`."""
         return QueryMode.INTERACTIVE
 
+    @property
+    def udf_resource_uris(self):
+        """List of code resource to load from a Google Cloud Storage URI (gs://bucket/path).
```
Tarrasch: s/List/Iterable?

ukarlsson: Fixed.

luigi/contrib/bigquery.py
ukarlsson: Fixed.

ukarlsson:
I have added tests. Yes, we use this in production

ukarlsson:
@Tarrasch OK thanks have fixed the whitespacing and rebased :)

luigi/contrib/bigquery.py
Tarrasch: what happened to the trailing comma you added??

Tarrasch:
I merge as this have had many minor iterations already. But I see that @ukarlsson actually never filled out the template in the PR. While it's indeed optional, I think you could jot down some words about why your company needed this patch. What can be done now that you couldn't do before?

Thanks for the patch! :)

PR #1883, opened 2016-10-14
[bigquery] point out Avro support

ulzha:
Google recently introduced Avro support in the BigQuery load API. It is also nice with automatic table schema inference.

We have tested that `'AVRO'` is a valid `source_format` value and it does what it says.


Tarrasch:
Cool. Can this be tested in any way? Apart from that I'm all +1 and happy they finally added Avro support. :)

ulzha:
This is subsumed in the bigger https://github.com/spotify/luigi/pull/1888.

PR #1885, opened 2016-10-14
Lets use Text type for value column of TaskParameter table

oivoodoo:
## Description

Change `value` column type from `string(256)` to `text` because of the issues on insert to task parameters table.

[error](https://gist.github.com/oivoodoo/0fb78f5a041bb08939f30f4af10c00c6) 
## Motivation and Context

Found this issue from the log file in our production system.

Going to test on 17.10.2016 in the next our deploy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -170,3 +170,4 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+.python-version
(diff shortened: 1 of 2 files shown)


Tarrasch:
LGTM but if somebody else could review that would be good.

.gitignore
```diff
@@ -170,3 +170,4 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+.python-version
```
kalvdans: This change in .gitignore is unrelated and should be in a separate patch.

kwilcox:
Seems like you need to add an upgrade step to `_upgrade_schema` to migrate old schemas on a `luigi` upgrade?

PR #1886, opened 2016-10-15
Use the proc filesystem to gather process information

nmandery:
## Description
This pull request uses the proc filesystem to gather process information instead of calling the `ps` command.

## Motivation and Context
See the discussion at PR #1876

### DIFF
--- a/luigi/lock.py
+++ b/luigi/lock.py
@@ -42,14 +42,17 @@ def getpcmd(pid):
                 _, val = lines
                 return val
     else:
-        cmd = 'ps x -wwo pid,args'
-        with os.popen(cmd, 'r') as p:
-            # Skip the column titles
-            p.readline()
-            for line in p:
-                spid, scmd = line.strip().split(' ', 1)
-                if int(spid) == int(pid):
-                    return scmd
+        # Use the /proc filesystem
+        # At least on android there have been some issues with not all
+        # process infos being readable. In these cases using the `ps` command
+        # worked. See the pull request at
+        # https://github.com/spotify/luigi/pull/1876
+        try:
+            with open('/proc/{0}/cmdline'.format(pid), 'r') as fh:
+                return fh.read().replace('\0', ' ').rstrip()
+        except IOError:
+            pass
+
     # Fallback instead of None, for e.g. Cygwin where -o is an "unknown option" for the ps command:
     return '[PROCESS_WITH_PID={}]'.format(pid)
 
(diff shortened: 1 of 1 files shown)


erikbern:
cool

luigi/lock.py
```diff
@@ -42,14 +42,17 @@ def getpcmd(pid):
...
+        # Use the /proc filesystem
+        # At least on android there have been some issues with not all
+        # process infos being readable. In these cases using the `ps` command
+        # worked. See the pull request at
+        # https://github.com/spotify/luigi/pull/1876
+        try:
+            with open('/proc/{0}/cmdline'.format(pid), 'r') as fh:
+                return fh.read().replace('\0', ' ').rstrip()
+        except IOError:
+            pass
```
nmandery: I just added commit 1836dd2 with a comment

Tarrasch:
Isn't it better to actually throw an exception instead of using a fallback-method when there's a file permissions error? I imagine an admin mis-configuring so that two different users run the same command. It's better to report an error early rather than always running in a `--no-lock` fashion. Have I misunderstood anything?

nmandery:
The former implementation as well as the implementation for windows systems both do not raise any errors when a pid can not be resolved. Reasons I can think of may be missing permissions or simply that the process in the meantime has exited. Instead the fallback at the end of the function is used. So there is always a string retruned giving some description on the process.

This PR simply preserves this behavior of the getpcmp function. The function tries to resolve the command of the given pid and returns the most meaningful description possible.

In case the desired behavior of `getpcmd` has changed to raise exceptions when resolving the command fails, I do not see it as sufficient to propagate the IOError from the failing access to `/proc`. The windows implementation would also be in need of a patch to also raise an exception when the pid can not be found in the output of the `wmic` command. This patch would then result in making the fallback at the end of the function non-reachable and obsolete.

Maybe you should also discuss this with @erikbern , the original author of the `getpcmd` function

Tarrasch:
@nmandery, You seem to have thought about this more thoroughly than I have. And admittedly know more about OS/procs/etc than me. :) Furthermore you just documented your thought-process in case further committers want to read up on this. Thanks. I feel confident enough to merge this. :)

Tarrasch:
@nmandery, I took the liberty to myself summarize why we did this in the squash-commit I just merged this PR with. I hope it looks good.

nmandery:
@Tarrasch Looks fine to me. Thank you for merging and the reviews.

PR #1888, opened 2016-10-19
[bigquery] add BigQueryLoadAvro that adds schema documentation to the table

ulzha:
This requires a new dependency (avro) so I put it in a separate module...


luigi/contrib/bigquery_avro.py
```diff
@@ -0,0 +1,78 @@
+"""Specialized tasks for handling Avro data in BigQuery from GCS.
+"""
+import avro
```
ulzha: I guess this would have to go in a `try` block following our current convention? (Have there been talks about longer term breaking the contrib modules into separate packages with dependencies properly declared in their `setup.py`s?)

luigi/contrib/bigquery_avro.py
```diff
@@ -0,0 +1,78 @@
...
+                reader = avro.datafile.DataFileReader(fp, avro.io.DatumReader())
+                schema.append(reader.datum_reader.writers_schema)
+            except Exception as e:
+                logger.info('%s', e)
+                return False
+            return True
+
+        input_fs.download(input_uri, 64 * 1024, read_schema).close()  # TODO check with various chunksizes (suspect the file position might matter)
+
+        return schema[0]
```
fluffypants: Possible crash if no Avro files exist in folder? Maybe accounted for elsewhere.

fluffypants:
Lacking tests?

mbruggmann:
@ulzha We would like to have this functionality as well, but I was hoping that we can get Google to implement this in the BQ avro importer instead of working around it in luigi. Have we asked them?

ulzha:
Yes we asked Google and they were reluctant. Apparently it had something to do with their protobuf preference, which doesn't propagate documentation.

luigi/contrib/bigquery_avro.py
ulzha: Yeah, a crash inside download will occur if no Avro files exist (or if the first of source_uris() doesn't point to an existing file). Will see if the error message from there is too cryptic. Perhaps it's self explanatory.

fluffypants: Did you get a chance to look into this case?

test/contrib/bigquery_gcloud_test.py
```diff
@@ -227,38 +228,49 @@ def test_run_query(self):
...
+                }, "doc": "This field shall be an inner"},
+                {"name": "col2", "type": "int", "doc": "The beautiful"}
             ]
         }""")
         writer = DataFileWriter(open("tmp.avro", "wb"), DatumWriter(), schema)
-        writer.append({'col0': 1000, 'col1': 1001})
+        writer.append({'col0': 1000, 'col1': {'inner': 1234}, 'col2': 1001})
         writer.close()
+        self.gcs_client.put("tmp.avro", self.gcs_dir_url)
         # FIXME don't need a tmp.avro; use GCS API like objects().insert that accepts file contents in a string:
```
ulzha: Remove the FIXME comment if we think it's nothing to fix...

test/contrib/bigquery_gcloud_test.py
```diff
@@ -218,3 +223,68 @@ def test_run_query(self):
...
+                }, "doc": "This field shall be an inner"},
+                {"name": "col2", "type": "int", "doc": "The beautiful"}
+            ]
+        }""")
+        writer = DataFileWriter(open("tmp.avro", "wb"), DatumWriter(), schema)
+        writer.append({'col0': 1000, 'col1': {'inner': 1234}, 'col2': 1001})
+        writer.close()
+        self.gcs_client.put("tmp.avro", self.gcs_dir_url)
+        # FIXME don't need a tmp.avro; use GCS API like objects().insert that accepts file contents in a string:
+        # https://developers.google.com/resources/api-libraries/documentation/storage/v1/python/latest/storage_v1.objects.html#insert
```
ulzha: Remove the FIXME comment if we think it's nothing to fix...

`tmp.avro` is not cleaned up as far as I can tell. Does it not pollute anything?

test/contrib/bigquery_gcloud_test.py
```diff
@@ -218,3 +223,68 @@ def test_run_query(self):
...
+
+    def setUp(self):
+        self.gcs_client = gcs.GCSClient(CREDENTIALS)
+        self.bq_client = bigquery.BigQueryClient(CREDENTIALS)
+
+        self.table_id = "avro_bq_table"
+        self.gcs_dir_url = 'gs://' + BUCKET_NAME + "/foo/tmp.avro"
+        self._produce_test_input()
+        self.addCleanup(self.bq_client.delete_dataset, bigquery.BQDataset(PROJECT_ID, DATASET_ID, EU_LOCATION))
+        self.addCleanup(self.gcs_client.remove, self.gcs_dir_url)
```
ulzha: This should come before producing test input or immediately after it, otherwise there's code in between that can crash and thus prevent adding of the cleanup.

test/contrib/bigquery_gcloud_test.py
```diff
@@ -27,23 +27,30 @@
...
 
+from contrib import gcs_test
 from nose.plugins.attrib import attr
+from helpers import unittest
 
 # In order to run this test, you should set your GCS/BigQuery project/bucket.
 # Unfortunately there's no mock
-PROJECT_ID = os.environ.get('BQ_TEST_PROJECT_ID', 'your_project_id_here')
-BUCKET_NAME = os.environ.get('BQ_TEST_INPUT_BUCKET', 'your_test_bucket_here')
+PROJECT_ID = gcs_test.PROJECT_ID
```
ulzha: Commit https://github.com/spotify/luigi/pull/1917/commits/a5291aba437c796a1a3ba080b13570bf5149e275, which introduced `BQ_` things, and which you rebased upon, had explicit intent to separate BigQuery test modules from `gcs_test` module. The resolution to include `gcs_test` again reverses it. Any good reason to couple them?

mrunesson: Tests were failing after the merge. End reason was because tests inpullrequest https://github.com/spotify/luigi/pull/1917/ did not clean up. And then did not revert this part of debugging.

tox.ini
```diff
@@ -88,6 +91,7 @@ deps =
   sqlalchemy
   Sphinx>=1.4.4,<1.5
   sphinx_rtd_theme
+  avro
```
ulzha: As far as I can tell we're adding a whole new dependency on avro for _everybody_ installing luigi (perhaps without the slightest bit of intent to use Avro). You would need to add it in setup.py too. But it is intrusive.

A better resolution would be to have the luigi.contrib.bigquery{_avro} modules in separate packages, but to be consistent with the currently consistently used approach, I suggest we import `avro` behind a `try`, same way as `oauth2client` and `googleapiclient`. Then I think in tox.ini only gcloud environment would need the dependency.

test/contrib/bigquery_gcloud_test.py
```diff
@@ -218,3 +223,68 @@ def test_run_query(self):
...
+        task.run()
+        self.assertTrue(task.complete())
+
+        table = self.bq_client.client.tables().get(projectId=PROJECT_ID,
+                                                   datasetId=DATASET_ID,
+                                                   tableId=self.table_id).execute()
+        self.assertEqual(table['description'], 'The description')
+        self.assertEqual(table['schema']['fields'][0]['description'], 'The bold')
+        self.assertEqual(table['schema']['fields'][1]['description'], 'This field shall be an inner')
+        self.assertEqual(table['schema']['fields'][1]['fields'][0]['description'], 'A inner field')
```
ulzha: Great effort overall, with all the nesting considered too... Wonder what happens if a field in a nested object is named the same as a field in an outer object. Does BigQuery Avro load become messed up then?

mrunesson: Have added test for this now.

ulzha:
LGTM!

PR #1891, opened 2016-10-20
checking significant parameter fix

ibbycurb:
## Description

This keeps passwords and such out of the visualizer.
## Motivation and Context

Having sensitive access information viewable so easily is not safe.  It's better to keep it hidden.

I'm running this right now in production.  I will not be making tests for this.  It's a simple change.


Tarrasch:
Actually there have been many patches very similar to this. And every now and then somebody comes and wants to revert the behavior (because other don't store secrets but just "insignificant" parameters). I'll try to write up some roadmap for this issue.

ibbycurb:
Thanks for the info, Tarrasch.

PR #1895, opened 2016-10-24
Hadoop streaming command options

heuvel:
Added support for getting the archives and cmdenvs parameters from the luigi configuration file and setting the environment variables for hadoop streaming jobs.

### DIFF
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -401,19 +401,20 @@ class HadoopJobRunner(JobRunner):
 
     def __init__(self, streaming_jar, modules=None, streaming_args=None,
                  libjars=None, libjars_in_hdfs=None, jobconfs=None,
-                 input_format=None, output_format=None,
+                 input_format=None, output_format=None, cmdenvs=None,
                  end_job_with_atomic_move_dir=True, archives=None):
         def get(x, default):
             return x is not None and x or default
         self.streaming_jar = streaming_jar
         self.modules = get(modules, [])
         self.streaming_args = get(streaming_args, [])
-        self.libjars = get(libjars, [])
+        self.libjars = libjars
         self.libjars_in_hdfs = get(libjars_in_hdfs, [])
         self.archives = get(archives, [])
         self.jobconfs = get(jobconfs, {})
         self.input_format = input_format
         self.output_format = output_format
+        self.cmdenvs = get(cmdenvs,[])
         self.end_job_with_atomic_move_dir = end_job_with_atomic_move_dir
         self.tmp_dir = False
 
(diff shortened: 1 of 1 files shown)


luigi/contrib/hadoop.py
```diff
@@ -582,7 +587,15 @@ class DefaultHadoopJobRunner(HadoopJobRunner):
     def __init__(self):
         config = configuration.get_config()
         streaming_jar = config.get('hadoop', 'streaming-jar')
-        super(DefaultHadoopJobRunner, self).__init__(streaming_jar=streaming_jar)
+        cmdenvs = config.get('hadoop', 'cmdenvs', default=None)
+        archives = config.get('hadoop', 'archives', default=None)
```
mfcabrera: My personal opinion: Archives options was already added as part of  https://github.com/spotify/luigi/pull/1649/files , I don't think we should force people to use the configuration file. Users can simple extended this.  `HadoopJobRunner` or `DefaultHadoopJobRunner` to get this done. In particular users already doing this programmatically will have the code broken  by this (my case).

heuvel: I agree, so I have removed that part in the new commit.

erikbern:
lgtm. maybe it should just be an argument for extra arguments

Tarrasch:
Can this be tested?

heuvel:
The use case for this is that we have Python 2.7 installed on each cluster node, but we are not allowed to install additional packages. So I create an archive of the virtual environment with the required packages (as described here [http://mfcabrera.com/python/2016/04/25/python-streaming-blog-org.html](url) by @mfcabrera) and I create a custom hadoopjobrunner to load the archive and environment variables from the configuration file. The cmdenv option of hadoop streaming is used to add the packages in the virtual environment archive to the Python path.

The HadoopJobRunner:

    class AdaHadoopJobRunner(luigi.contrib.hadoop.HadoopJobRunner):
    
        def __init__(self):
            streaming_jar = config.get('hadoop', 'streaming-jar')
            cmdenvs = config.get('hadoop', 'cmdenvs', default=None)
            archives = config.get('hadoop', 'archives', default=None)
            
            if cmdenvs:
                cmdenvs = cmdenvs.split(';')
            
            if archives:
                archives = archives.split(';')
                
            super(AdaHadoopJobRunner, self).__init__(streaming_jar=streaming_jar, cmdenvs=cmdenvs, archives=archives)


The luigi.cfg:

	...
	
	[hadoop]
	...
	cmdenvs=PYTHONPATH=venv_archive/lib/python2.7/site-packages/
	archives=venv_archive_with_packages.zip#venv_archive;

mfcabrera:
@heuvel  I create a PR with a more general version of this: #1929    Maybe you could try to adapt your code? . I think is cleaner than adding  rarely used parameters.  You will only need to add the  `extra_arguments` method to a normal JobTask.

heuvel:
@mfcabrera The more general version as proposed in #1929 will definitely work for me, so I will close this PR. Thank you!

mfcabrera:
@heuvel Cool!  can you give it a try (just checkout that branch). We actually need someone else using Streaming jobs to confirm it works and then merge into master. :)

PR #1896, opened 2016-10-24
[bigquery] add support for non-legacy SQL

ukarlsson:


luigi/contrib/bigquery.py
```diff
@@ -572,6 +572,12 @@ def udf_resource_uris(self):
         """
         return []
 
+    @property
+    def use_legacy_sql(self):
+        """Whether to use legacy SQL
+        """
+        return True
```
Tarrasch: Is using legacy default o_O?

Tarrasch:
@ukarlsson, I would love to merge this, but can you get someone else who knows BigQuery to review this?

mbruggmann:
@Tarrasch We are running something very similar internally as well. Not sure if that qualifies as "knowing BigQuery" but I'd give this a 👍

Tarrasch:
@mbruggmann, certainly good enough for me (that you run it in production). Thanks for reviewing. :)

PR #1902, opened 2016-10-28
make elasticsearch example run out of the box

miku:
## Description

Change parameters to `luigi.run` in elasticsearch indexing example.
## Motivation and Context

The `--task` parameter seems obsolete now.

> error: unrecognized arguments: --task

Also, I tested the example with the latest elasticsearch release [5.0.0](https://www.elastic.co/v5) - it works.

 

Tested locally:

```
$ brew install elasticsearch
$ elasticsearch --daemonize
$ python examples/elasticsearch_index.py
...
This progress looks :)
```

### DIFF
--- a/examples/elasticsearch_index.py
+++ b/examples/elasticsearch_index.py
@@ -112,4 +112,4 @@ class IndexDocuments(CopyToIndex):
         return FakeDocuments()
 
 if __name__ == "__main__":
-    luigi.run(['--task', 'IndexDocuments'])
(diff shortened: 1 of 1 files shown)


Tarrasch:
Actually. I would rather see the whole code block removed, and replaced with something like [this](https://github.com/spotify/luigi/blob/21f9fce674f819f0053161cbf46a90ac5162f606/examples/per_task_retry_policy.py#L4-L14)

PR #1908, opened 2016-11-02
Update config and notifs documentation

Tarrasch:
## Description

Clean up docs and notifications code.

## Motivation and Context

It started with changing 1 lines in docs. And I saw a lot of opportunity to up the code quality and out of date docs. So I decided to fix.

I have run the `TestNotificationsTask` and I get sensible error messages and I still get emails. The tests should also catch potential mistakes.

## Bonus

This also removes the extremely annoying sequence of log messages going something like this (in typical luigi logs):

```
INFO: Sending error email to arash@company.com
INFO: Oh wait, no I'm not actually because I'm running from a tty (gotcha! ;p)
```

ping @henlof who I vaguely recall once mentioning it. ^^

### DIFF
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -36,10 +36,7 @@ import luigi.task
 import luigi.parameter
 
 logger = logging.getLogger("luigi-interface")
-
-
 DEFAULT_CLIENT_EMAIL = 'luigi-client@%s' % socket.gethostname()
-DEBUG = False
 
 
 class TestNotificationsTask(luigi.task.Task):
@@ -68,7 +65,7 @@ class TestNotificationsTask(luigi.task.Task):
 class email(luigi.Config):
     force_send = luigi.parameter.BoolParameter(
         default=False,
-        description='Send e-mail even from a tty or with DEBUG set')
+        description='Send e-mail even from a tty')
     format = luigi.parameter.ChoiceParameter(
         default='plain',
         config_path=dict(section='core', name='email-type'),
@@ -254,19 +251,15 @@ def send_email_sendgrid(sender, subject, message, recipients, image_png):
     client.send(to_send)
 
 
-def _email_disabled():
+def _email_disabled_reason():
     if email().format == 'none':
-        logger.info("Not sending email when email format is none")
-        return True
+        return "email format is 'none'"
     elif email().force_send:
-        return False
+        return None
     elif sys.stdout.isatty():
-        logger.info("Not sending email when running from a tty")
-        return True
-    elif DEBUG:
-        logger.info("Not sending email when running in debug mode")
+        return "running from a tty"
     else:
-        return False
+        return None
 
 
 def send_email_sns(sender, subject, message, topic_ARN, image_png):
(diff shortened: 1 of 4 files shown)


luigi/notifications.py
```diff
@@ -254,19 +251,15 @@ def send_email_sendgrid(sender, subject, message, recipients, image_png):
     client.send(to_send)
 
 
-def _email_disabled():
+def _email_disabled_reason():
     if email().format == 'none':
-        logger.info("Not sending email when email format is none")
-        return True
+        return "Not sending email when email format is 'none'"
```
dlstadther: I'm not clear on when `email().format` will ever equal `'none'`. It's a configurable `ChoiceParameter` with a default value of `plain`. So unless someone explicitly chooses `none`, this will never run.

Seems like if you didn't want emails to send, you just wouldn't put a desired target email (rather than specifying the email type to be none.

luigi/notifications.py
```diff
@@ -254,19 +251,15 @@ def send_email_sendgrid(sender, subject, message, recipients, image_png):
...
     elif email().force_send:
-        return False
+        return ''
```
daveFNbuck: It can be a bit safer to use None here. It's compatible with your current code and makes it more clear that there is no reason.

daveFNbuck:
I'm a bit concerned that by removing the message about how to set up e-mail we're making the e-mail feature less discoverable. Perhaps we should add back an updated version of that message?

luigi/notifications.py
```diff
@@ -311,10 +304,15 @@ def send_email(subject, message, sender, recipients, image_png=None):
     }
 
     subject = _prefix(subject)
-    if not recipients or recipients == (None,) or _email_disabled():
+    if not recipients or recipients == (None,):
+        return
+
+    if _email_disabled_reason():
+        logger.info("Not sending email to %r because reason: %s",
```
daveFNbuck: The message `Not sending email to test@example.com because reason: Not sending email when running from a tty` is a bit awkward and repetitive. How about `Not sending e-mail to test@example.com: e-mail disabled when running from a tty`?

luigi/notifications.py
Tarrasch: Good point. @daveFNbuck, was `none` preserved from your previous refactoring?

Although, maybe we still want to keep this so people can pass cli parameter `--email-format-none`. But it's a bit unlikely, so I'm still okay with removing this.

luigi/notifications.py
Tarrasch: Ah right, I thought `'some_string' + None` will crash, but actually such logic won't be run. I'll fix this. :)

Tarrasch:
@daveFNbuck, as for the feature being discoverable. I did consider it, but thought that one should be able to run luigi without that feature too without being bothered all the time. There's already places in the docs mentioning you can use email.  No strong opinion though. :)

luigi/notifications.py
daveFNbuck: Yeah, I tried my best not to change anything. I'm not sure why you would ever need none. Just don't set a receiver or an smtp server. Maybe this would be useful in a situation where you have multiple configs overriding parts of eachother?

daveFNbuck:
Fair enough. LGTM

luigi/notifications.py
Tarrasch: Ok. I will try to not change the behavior, but I welcome anybody to change it in PR once this is merged. :)

Tarrasch:
Ok. I just pushed fixes to all @daveFNbuck's comments. I consider this to be ready to merge once tests pass then. :)

Tarrasch:
First time we merge a PR with the rebase strategy! :tada:

Issue #1519, opened 2016-01-18
Temporary files are messed up


erikbern:
Luigi has like 500 different inconsistent implementations of temporary files. We should clean this up.

There's lots of cases where you need temporary files. In most of those cases, you want some kind of action on `close()` or using a context manager. For instance if you are working with a `LocalTarget`, you want to be able to do both

```python
f = self.output().open('w')
# ...
f.close() # this commits the file
```

As well as 

```python
with self.output().open('w') as f:
    # ...
```

But there are also cases where you want access to the underlying filename, in which case I suggest something like this (this was originally @Tarrasch's idea)

```python
with self.output().temporary_path() as fn:
   h5py.open(fn) # or some stuff like that...
```

Do we need semantics here that does not involve a context manager? In that case you could probably do something that returns a *new* `LocalTarget`, such as

```python
t = self.output().temporary_target()
# print t.path
t.commit()
```

I'm not sure if this is necessary.

There's also a bunch of other targets that leverage temp files, like writing to a file over FTP (`RemoteTarget`). This leverages a local file as a temporary output. It would be great if RemoteTarget could simply inherit all the methods above so that you have both semantics for atomic stream opens, as well as getting an explicit local path. And you should be able to do it using a context manager or not.

Finally there's code in Luigi that just uses temporary paths in general. Running a Python mapreduce job creates a temporary directory. We should be able to leverage the internal temporary file implementation here.

Tarrasch:
This is absolutely correct. I'll doubt I'll be able to work on this myself until early March though. Thanks for creating this ticket for this recurring issue that we have.

erikbern:
If you can leave some notes here, I would appreciate it. I wanted to spend a couple of hours yesterday refactoring this but realized it was a bigger project just to understand the code. So even just outlining a strategy for refactoring would be great

Tarrasch:
Here are two ideas of what one can do that simplifies the code base and is taking us in the right direction.

## 1. Remove `is_tmp=` kwarg to targets.

Currently we have a "convenience" kwarg `is_tmp`. Works like this:

        t = luigi.LocalTarget(is_tmp=True)
        args = ['python', 'test/cmdline_test.py', 'WriteToFile', '--filename', t.path, '--local-scheduler', '--no-lock']
        self._run_cmdline(args)

See how convenient it is that you don't even need to supply a path? :)

Now why I don't like this: In luigi we encourage usage of targets like this:

```
class MyTask(luigi.Task):
[...]

PR #1909, opened 2016-11-03
targets: Implement temporary_path decorator

Tarrasch:
## Description

As has been detailed in spotify/luigi#1519, there are many ways to do
temporary files in luigi. The use case for temporary files are always
about achieving atomic writes.

This PR finally describes the problem in the luigi documentation and
tries to interlink essential parts. Also it implements a decorator
that is meant to be able to unify many of the existing implementations,
and as a new, clean and recommended way to do temporary files and atomic
moves using luigi.

There are included unit tests, although they only run with a lot mocked. I've run something similar in production for over half a year now. I intend to also try this a little bit (at least for hdfs+local disk)

### DIFF
--- a/luigi/target.py
+++ b/luigi/target.py
@@ -167,13 +167,7 @@ class FileSystem(object):
     def rename_dont_move(self, path, dest):
         """
         Potentially rename ``path`` to ``dest``, but don't move it into the
-        ``dest`` folder (if it is a folder).  This kind of operation is useful
-        when you don't want your output path to ever contain partial or
-        errinously nested data.
-
-        See `this github issue <https://github.com/spotify/luigi/pull/557>`__ and
-        `the thanksgiving bug <http://tarrasch.github.io/luigi-budapest-bi-oct-2015/#/21>`__
-        where the problem is described.
+        ``dest`` folder (if it is a folder).  This relates to :ref:`AtomicWrites`.
 
         This method has a reasonable but not bullet proof default
         implementation.  It will just do ``move()`` if the file doesn't
@@ -206,7 +200,7 @@ class FileSystemTarget(Target):
 
     A FileSystemTarget has an associated :py:class:`FileSystem` to which certain operations can be
     delegated. By default, :py:meth:`exists` and :py:meth:`remove` are delegated to the
-    :py:class:`FileSystem`, which is determined by the :py:meth:`fs` property.
+    :py:class:`FileSystem`, which is determined by the :py:attr:`fs` property.
 
     Methods of FileSystemTarget raise :py:class:`FileSystemException` if there is a problem
     completing the operation.
@@ -225,7 +219,7 @@ class FileSystemTarget(Target):
         """
         The :py:class:`FileSystem` associated with this FileSystemTarget.
         """
-        raise
+        raise NotImplementedError()
 
     @abc.abstractmethod
     def open(self, mode):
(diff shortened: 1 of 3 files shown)


Tarrasch:
@ulzha, do you mind reading the docs I included in this. Maybe there's been other experiences or best practices from Spotify you want to share?

erikbern:
looks reasonable, although consider using `contextlib.contextmanager` for this

Tarrasch:
@erikbern, I tried but `contextlib.contextmanager` doesn't really work here. The reason is because this contexmanagers is not as simple as *init-run-finalize*. `temporary_path()` will finalize conditionally, depending on if there was an exception raised, or not. 

@erikbern, does the docs section look good?

Tarrasch:
@ulzha @dlstadther @daveFNbuck, mind reviewing this? Or at least say you have not had big problems with atomic writes like I've had.

luigi/target.py
```diff
@@ -257,14 +251,62 @@ def remove(self):
...
+                    '/' if self.target._has_trailing_slash() else '')
+
+            def __enter__(self):
+                return self._temp_path
+
+            def __exit__(self, exc_type, exc_value, traceback):
+                if exc_type is None:
+                    # There were no exceptions
+                    self.target.fs.rename_dont_move(self._temp_path, self.target.path)
+                return False  # False means we don't suppress the exception
```
daveFNbuck: shouldn't we clean up the temporary file on failure?

luigi/target.py
```diff
@@ -257,14 +251,62 @@ def remove(self):
...
+        """
+        class _Manager(object):
+            target = self
+
+            def __init__(self):
+                num = random.randrange(0, 1e10)
+                self._temp_path = '{}-luigi-tmp-{:010}{}'.format(
+                    self.target.path.rstrip('/'),
+                    num,
+                    '/' if self.target._has_trailing_slash() else '')
```
daveFNbuck: The way you're handling slashes here could cause problems with targets on Windows machines.

daveFNbuck:
I like the documentation and agree with the importance of atomic writes, but I've mostly just used core code for it and not had to write any myself so I haven't had any big problems. I'm still in favor of anything that makes it easier though.

luigi/target.py
Tarrasch: I've thought about this actually. And I think it's best to not clean them up. I usually find the temporary files to be invaluable traces left when debugging. I'm however up for in an upcoming patch add a kwarg like `delete_on_failure=False`.

luigi/target.py
Tarrasch: Nice catch!

luigi/target.py
daveFNbuck: I guess that's reasonable. I have a daily cron job to clean up all the leftover tmp files from MRs so HDFS doesn't fill up, so I'm mostly ok either way.

test/target_test.py
```diff
@@ -266,3 +268,68 @@ def test_rename_dont_move_on_fs(self):
...
+        self.fs.rename_dont_move.assert_called_with(tmp_path, target_noslash.path)
+
+    def test_windowsish_dir(self):
+        target = self.target_cls(r'''C:\my\folder''' + "\\")
+        pattern = r'''C:\\my\\folder-luigi-tmp-\d{10}''' + r"\\"
+
+        with target.temporary_path() as tmp_path:
+            assert re.match(pattern, tmp_path)
+        self.fs.rename_dont_move.assert_called_once_with(tmp_path, target.path)
+
```
dlstadther: To save you the trouble of reading the Travis report, your only issue was a flake8 complaint. You included too may `\n` here.

Tarrasch: dooooh

PR #1910, opened 2016-11-04
Fixing task.flatten to allow for custom Targets inheriting from dict/MutableMapping

Lucas-C:
## Description
I added a conditional short-circuit `return` in `task.flatten`.
This adds a dependency for the `task` module to the `target` module.

## Motivation and Context
This allows for `dict`-like and iterable custom `Target` classes.

I have included a unit test.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -34,6 +34,7 @@ import re
 from luigi import six
 
 from luigi import parameter
+from luigi.target import Target
 from luigi.task_register import Register
 
 Parameter = parameter.Parameter
@@ -609,6 +610,8 @@ def flatten(struct):
     """
     if struct is None:
         return []
+    if isinstance(struct, Target):
+        return [struct]
     flat = []
     if isinstance(struct, dict):
         for _, result in six.iteritems(struct):
--- a/test/task_test.py
+++ b/test/task_test.py
@@ -43,6 +43,10 @@ class DefaultInsignificantParamTask(luigi.Task):
     necessary_param = luigi.Parameter(significant=False)
 
 
+class DummyListTarget(luigi.Target, list):
+    pass
+
+
 class TaskTest(unittest.TestCase):
 
     def test_tasks_doctest(self):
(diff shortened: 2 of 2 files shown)


erikbern:
what's the use case?

Lucas-C:
I wanted to use this class:
```
class InMemoryDictTarget(MockTarget, MutableMapping):
    """
    Relies on luigi.mock.MockTarget implementation, based on multiprocessing.Manager().dict()
    """
    def __setitem__(self, key, value):
        if os.sep in key:
            raise ValueError('File paths separator character not allowed in keys')
        all_data = self.fs.get_all_data()
        all_data[os.path.join(self.path, key)] = value

    def __getitem__(self, key):
        if os.sep in key:
            raise ValueError('File paths separator character not allowed in keys')
        all_data = self.fs.get_all_data()
        return all_data[os.path.join(self.path, key)]

    def __delitem__(self, key):
        all_data = self.fs.get_all_data()
        del all_data[os.path.join(self.path, key)]

    def __len__(self):
        all_data = self.fs.get_all_data()
        return len(all_data)

    def __iter__(self):
        all_data = self.fs.get_all_data()
        for path, _ in all_data.items():
            if os.sep not in path:
                continue
            yield os.path.basename(path)

    def __enter__(self):
        return self

    def __exit__(self, type, value, traceback):
        with super().open('w'):
            pass
```

Lucas-C:
The github preview is misleading : I did not modify the `test_tasks_doctest` test, but `test_flatten`. Check out the "full file diff".

About my usage example: the goal is to have a `Target` stored in memory, not on disk, and behaving like a standard `dict`. Does it sound absurd ? Do you have other questions I could answer about it ?

Tarrasch:
>  the goal is to have a `Target` stored in memory, not on disk, and behaving like a standard `dict`

This text is much more comprehensible to me. I'm of course a bit skeptic as it's so volatile. What real world use case do you have?

Overall I'm still negative to add some more magic to the `flatten` function.

Lucas-C:
My real world use case is simple : I want to avoid writing files on disk for metas informations.

One of the fields in this `dict` will be the filepath to the input/ouput file processed, but of this `Target` other key-values pairs will be metadata that only have value during a pipeline build.
I do not want to create temporary files just to store this metadata as I want it gone one the build ended.

Tarrasch:
Yes, I think it's too magic. In particular I can't see how the change in logic the `flatten` function relates to enabling the feature of storing metadata in memory. Now I see it if I carefully look at the code, but I want to keep the luigi codebase more approachable.

So I'm still :-1: on how this patch implements your use-case.

Lucas-C:
OK, I understand.
I can still implement an `InMemoryDictTarget` somehow without this patch, so nevermind.

Tarrasch:
Feel free to share your other approaches. Even if you don't make a PR you can include some shorter hints in this PR.

PR #1912, opened 2016-11-06
Update salesforce http header for Requests update

dlstadther:
## Description

With the release of Requests 2.11, headers were strictly required to contain string-only components. This requires the casting of `Content-Length` to the string equivalent of the length of the data.

## Motivation and Context

Fix error caused by method `check_header_validity` in `requests/utils.py` (`InvalidHeader: Header value 65 must be of type str or bytes, not <type 'int'>`).

[Requests Issue discussion](https://github.com/kennethreitz/requests/issues/3477)

Runs my jobs without error.

### DIFF
--- a/luigi/contrib/salesforce.py
+++ b/luigi/contrib/salesforce.py
@@ -474,7 +474,7 @@ class SalesforceAPI(object):
             raise Exception("Can not create a batch without a valid job_id and an active session.")
 
         headers = self._get_create_batch_content_headers(file_type)
-        headers['Content-Length'] = len(data)
+        headers['Content-Length'] = str(len(data))
 
         response = requests.post(self._get_create_batch_url(job_id),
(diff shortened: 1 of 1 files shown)


dlstadther:
I've finally gotten around to looking at testing this, but am unsure how to. The part that needs to be checked is that the header passed to `requests` is a dictionary of strings. Since the function does not simply return the response, but rather some portion of the parse response, I don't have access to that variable to check the header.

The other problem is fooling the function into thinking that I have a valid session and job_id.

Of course, I'll defer to you, @Tarrasch , if you really want me to move things around such that testing this is possible.

Tarrasch:
No I'm fine with this. You can merge this I think. :)

dlstadther:
Thanks for the approval @Tarrasch !

PR #1915, opened 2016-11-07
file.py: Rename to local_target.py

Tarrasch:
## Description

 * Rename file.py to local_target.py
 * Remove deprecated File class


Tarrasch:
Thanks for the catch @dlstadther, does it look good now? :)

PR #1916, opened 2016-11-07
Enables e-mail batching in the scheduler

daveFNbuck:
## Description
Enables sending periodic batch e-mails from the scheduler rather than sending immediate failure/disable/scheduling failure e-mails on every event.

## Motivation and Context
With a large pipeline, it's fairly easy to get hundreds of e-mails an hour when something goes wrong. This translates to thousands of e-mails per day. While this is a great resource for tracking down error messages of a particular task in your e-mail history, it's not nearly as great for understanding what the current issues are in your pipeline.

In order to make it more manageable, this groups together failures and shows sample error messages. Failures are ordered by how often they occur and it's easier to see from e-mail subject lines how many failures are occurring over time. Failure message history is preserved enough for most uses by including snippets for each task.

I have included unit tests for all the pieces of this and have been running it for the last week or so, reducing my luigi e-mail volume by two orders of magnitude and making it easier to find and fix issues, also reducing the error volume. So the feature and its usefulness have both survived testing.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -307,7 +307,7 @@ sender
 [batch_notifier]
 ----------------
 
-Parameters controlling the contents of batch notifications send from the
+Parameters controlling the contents of batch notifications sent from the
 scheduler
 
 email-interval
@@ -316,13 +316,42 @@ email-interval
   Defaults to 60.
 
 batch-mode
-  Controls how tasks are grouped together in the e-mail. Can be set to
-  "family", "all", or "unbatched_params". If "family", all tasks of the
-  same family are treated identically. If "all", only identical tasks
-  are grouped together. If "unbatched_params", tasks are grouped
-  together if they are of the same family and all parameters other than
-  batch parameters are equal.
-  Defaults to "unbatched_params"
+  Controls how tasks are grouped together in the e-mail. Suppose we have
+  the following sequence of failures:
+
+  1. TaskA(a=1, b=1)
+  2. TaskA(a=1, b=1)
+  3. TaskA(a=2, b=1)
+  4. TaskA(a=1, b=2)
+  5. TaskB(a=1, b=1)
+
+  For any setting of batch-mode, the batch e-mail will record 5 failures
+  and mention them in the subject. The difference is in how they will
+  be displayed in the body. Here are example bodies with error-messages
+  set to 0.
+
+  "all" only groups together failures for the exact same task:
+
+  - TaskA(a=1, b=1) (2 failures)
+  - TaskA(a=1, b=2) (1 failure)
+  - TaskA(a=2, b=1) (1 failure)
+  - TaskB(a=1, b=1) (1 failure)
+
+  "family" groups together failures for tasks of the same family:
+
+  - TaskA (4 failures)
+  - TaskB (1 failure)
+
+  "unbatched_params" groups together tasks that look the same after
+  removing batched parameters. So if TaskA has a batch_method set for
+  parameter a, we get the following:
+
+  - TaskA(b=1) (3 failures)
+  - TaskA(b=2) (1 failure)
+  - TaskB(a=1, b=2) (1 failure)
+
+  Defaults to "unbatched_params", which is identical to "all" if you are
+  not using batched parameters.
 
 error-lines
   Number of lines to include from each error message in the batch
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -365,8 +365,8 @@ error-messages
   Number of messages to preserve for each task group. As most tasks that
   fail repeatedly do so for similar reasons each time, it's not usually
   necessary to keep every message. This controls how many messages are
-  kept for each task or task group. Set to 0 to not include error
-  messages in the e-mails.
+  kept for each task or task group. The most recent error messages are
+  kept. Set to 0 to not include error messages in the e-mails.
   Defaults to 1.
 
 group-by-error-messages
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -233,6 +233,13 @@ no_install_shutdown_handler
   thread.
   Defaults to false.
 
+send-failure-email
+  Controls whether the worker will send e-mails on task and scheduling
+  failures. If set to false, workers will only send e-mails on
+  framework errors during scheduling and all other e-mail must be
+  handled by the scheduler.
+  Defaults to true.
+
 
 [elasticsearch]
 ---------------
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -570,6 +570,12 @@ scalding-libjars
 
 Parameters controlling scheduler behavior
 
+batch-emails
+  Whether to send batch e-mails for failures and disables rather than
+  sending immediate disable e-mails and just relying on workers to send
+  immediate batch e-mails.
+  Defaults to false.
+
 disable-hard-timeout
   Hard time limit after which tasks will be disabled by the server if
   they fail again, in seconds. It will disable the task if it fails
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -304,6 +304,51 @@ sender
   Default value: luigi-client@<server_name>
 
 
+[batch_notifier]
+----------------
+
+Parameters controlling the contents of batch notifications send from the
+scheduler
+
+email-interval
+  Number of minutes between e-mail sends. Making this larger results in
+  fewer, bigger e-mails.
+  Defaults to 60.
+
+batch-mode
+  Controls how tasks are grouped together in the e-mail. Can be set to
+  "family", "all", or "unbatched_params". If "family", all tasks of the
+  same family are treated identically. If "all", only identical tasks
+  are grouped together. If "unbatched_params", tasks are grouped
+  together if they are of the same family and all parameters other than
+  batch parameters are equal.
+  Defaults to "unbatched_params"
+
+error-lines
+  Number of lines to include from each error message in the batch
+  e-mail. This can be used to keep e-mails shorter while preserving the
+  more useful information usually found near the bottom of stack traces.
+  This can be set to 0 to include all lines. If you don't wish to see
+  error messages, instead set `error-messages` to 0.
+  Defaults to 20.
+
+error-messages
+  Number of messages to preserve for each task group. As most tasks that
+  fail repeatedly do so for similar reasons each time, it's not usually
+  necessary to keep every message. This controls how many messages are
+  kept for each task or task group. Set to 0 to not include error
+  messages in the e-mails.
+  Defaults to 1.
+
+group-by-error-messages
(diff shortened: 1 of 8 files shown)


erikbern:
seems like a great feature!

daveFNbuck:
Yeah, it makes identifying and fixing errors a lot easier.

Tarrasch:
Wow, this is a huge patch. I'll try to get to reviewing it soon :)

daveFNbuck:
I tried to keep it small, but a lot of features are required to get a handle on the amount of e-mail I was getting.

luigi/batch_notifier.py
```diff
@@ -0,0 +1,201 @@
...
+        self._update_next_send()
+
+        self._email_format = email().format
+        if email().receiver:
+            self._default_owner = set(filter(None, email().receiver.split(',')))
+        else:
+            self._default_owner = set()
+
+    def _update_next_send(self):
+        self._next_send = time.time() + 60 * self._config.email_interval
```
dlstadther: Supposed i set the interval to 5 minutes or so to accommodate for the majority of my pipelines, but i have a tiny pipeline which takes a minute or two to complete. If an error occurs during that tiny pipeline, will its emails still be delayed for that 5 minutes? Or will the final email send occur at the final attempt of execution failure?

doc/configuration.rst
```diff
@@ -297,6 +304,51 @@ sender
...
+  Defaults to 60.
+
+batch-mode
+  Controls how tasks are grouped together in the e-mail. Can be set to
+  "family", "all", or "unbatched_params". If "family", all tasks of the
+  same family are treated identically. If "all", only identical tasks
+  are grouped together. If "unbatched_params", tasks are grouped
+  together if they are of the same family and all parameters other than
+  batch parameters are equal.
+  Defaults to "unbatched_params"
```
daveFNbuck: I was trying to keep this short, but I guess it requires some examples. I'll update it to be more clear.

doc/configuration.rst
```diff
@@ -297,6 +304,51 @@ sender
   Default value: luigi-client@<server_name>
 
 
+[batch_notifier]
+----------------
+
+Parameters controlling the contents of batch notifications send from the
+scheduler
```
daveFNbuck: Right, thanks.

luigi/batch_notifier.py
daveFNbuck: The scheduler doesn't separate by pipelines here. It just collects failures and sends them all out every 5 minutes. This feature is for when you care more about reducing volume than you do about immediacy. If you want per-pipeline e-mail batches triggered at the end of each run, that can always be added as a later PR. The current setup is based around the needs of my pipelines, so I'm sure others will want to add lots of tweaks over time.

luigi/batch_notifier.py
```diff
@@ -0,0 +1,201 @@
...
+        if self._config.batch_mode == 'all':
+            return task_name
+        elif self._config.batch_mode == 'family':
+            return family
+        elif self._config.batch_mode == 'unbatched_params':
+            param_str = ', '.join('{}={}'.format(*kv) for kv in six.iteritems(unbatched_args))
+            return '{}({})'.format(family, param_str)
+        else:
+            raise ValueError('Unknown batch mode for batch notifier: {}'.format(
+                self._config.batch_mode))
```
daveFNbuck: This could happen if you add a new choice but forget to implement it. It should never happen with the code as-is, but things change.

doc/configuration.rst
```diff
@@ -297,6 +304,80 @@ sender
...
+  "unbatched_params" groups together tasks that look the same after
+  removing batched parameters. So if TaskA has a batch_method set for
+  parameter a, we get the following:
+
+  - TaskA(b=1) (3 failures)
+  - TaskA(b=2) (1 failure)
+  - TaskB(a=1, b=2) (1 failure)
+
+  Defaults to "unbatched_params", which is identical to "all" if you are
+  not using batched parameters.
```
dlstadther: Much clearer! Thanks!

luigi/batch_notifier.py
```diff
@@ -0,0 +1,201 @@
+import collections
```
daveFNbuck: That is the case, although it probably wouldn't be too hard to make it a standalone library if someone wanted to use it like that.

luigi/batch_notifier.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def _format_tasks(self, tasks):
+        lines = map(self._format_task, sorted(tasks, key=self._expl_key))
+        if self._email_format == 'html':
+            return six.u('<li>{}').format(six.u('\n<br>').join(lines))
+        else:
+            return six.u('- {}').format(six.u('\n  ').join(lines))
+
+    def _owners(self, owners):
+        return self._default_owner | set(owners or ())
```
daveFNbuck: Done. It's actually already always a list, so I just have to adjust the code a little to recognize that.

test/batch_notifier_test.py
```diff
@@ -0,0 +1,562 @@
...
+        bn.add_failure('Task(a=1)', 'Task', {'a': 1}, 'error 1')
+        bn.add_failure('Task(a=2)', 'Task', {'a': 2}, 'error 2')
+        bn.add_failure('TaskB(a=1)', 'TaskB', {'a': 1}, 'error')
+
+        bn.send_email()
+        self.check_email_send(
+            'Luigi: 3 failures in the last 60 minutes',
+            '- Task (2 failures)\n'
+            '\n'
+            '      error 2\n'
```
daveFNbuck: I updated the name of the unit test and added a short explanation to configuration.rst

test/batch_notifier_test.py
```diff
@@ -0,0 +1,562 @@
...
+        bn.add_failure('Task(a=2)', 'Task', {'a': 2}, 'error 2')
+        bn.add_failure('TaskB(a=1)', 'TaskB', {'a': 1}, 'error')
+
+        bn.send_email()
+        self.check_email_send(
+            'Luigi: 3 failures in the last 60 minutes',
+            '- Task (2 failures)\n'
+            '\n'
+            '      error 1\n'
+            '\n'
```
daveFNbuck: I think they look ok. I've been using html e-mails so I can't say too much about how readable the plain text is in practice. It's probably terrible when you have very long lines.

test/batch_notifier_test.py
```diff
@@ -0,0 +1,562 @@
...
+
+    def test_limit_expl_length(self):
+        bn = BatchNotifier(batch_mode='family', error_messages=1, error_lines=2)
+        bn.add_failure('Task(a=1)', 'Task', {'a': '1'}, 'line 1\nline 2\nline 3\nline 4\n')
+        bn.send_email()
+        self.check_email_send(
+            'Luigi: 1 failure in the last 60 minutes',
+            '- Task (1 failure)\n'
+            '\n'
+            '      line 3\n'
```
daveFNbuck: That's certainly a possibility. I hadn't thought about it too much. It might be better not to include that, as it could prevent grouping. I really find it helpful to group identical messages and have actually considered using fewer lines so more differences get cut out. If we include the number of stripped lines, I'll see something like

```
- TaskA (1 failure)

    ... 10 stripped lines ...
   mysql failure

- TaskB (1 failure)

    ... 11 stripped lines ...
    mysql failure

- TaskC

    ... 9 stripped lines ...
    mysql failure
```

instead of just

```
- TaskA (1 failure)
- TaskB (1 failure)
- TaskC (1 failure)

    mysql failure
```

because each task could have different trace lengths despite having the same underlying issue.

test/batch_notifier_test.py
```diff
@@ -0,0 +1,562 @@
...
+            '      line 4'
+        )
+
+    def test_expl_varies_by_owner(self):
```
daveFNbuck: It was a later addition after users started complaining that they weren't getting failure e-mails anymore :)

test/scheduler_api_test.py
```diff
@@ -1763,3 +1764,209 @@ def test_assistant_running_task_dont_disappear(self):
...
+            worker=WORKER, status=FAILED, task_id='T(a=5, b=6)', family='T',
+            params={'a': '5', 'b': '6'}, expl='"bad thing"')
+        self.assertEqual(1, notifications.send_error_email.call_count)
+
+    @mock.patch('luigi.scheduler.BatchNotifier')
+    def test_no_batch_notifier_without_batch_emails(self, BatchNotifier):
+        Scheduler(batch_emails=False)
+        BatchNotifier.assert_not_called()
+
+    @mock.patch('luigi.scheduler.BatchNotifier')
```
daveFNbuck: Thanks!

daveFNbuck:
@Tarrasch moving to luigi/server/batch_notifier.py would be a bit problematic because luigi.server would become ambiguous, referring to both luigi/server.py and luigi/server/. Perhaps we can move both in a separate PR?

luigi/batch_notifier.py
```diff
@@ -108,21 +113,21 @@ def _format_tasks(self, tasks):
             return six.u('- {}').format(six.u('\n  ').join(lines))
 
     def _owners(self, owners):
-        return self._default_owner | set(owners or ())
+        return self._default_owner | set(owners)
 
-    def add_failure(self, task_name, family, unbatched_args, expl, owners=None):
+    def add_failure(self, task_name, family, unbatched_args, expl, owners):
```
Tarrasch: Actually. I meant something like changing this from `owners=None` to `owners=[]`. Surely removing the default value made the tests less readable

Tarrasch:
Sure, we can do the `luigi/server`-subfolder in a separate PR. :)

luigi/batch_notifier.py
daveFNbuck: They're not that different, just a `[]` added to every call. `owners=[]` is an anti-pattern because when owners goes to the default, it's always the same list. So if you append to the list after accepting a default argument, the default is no longer empty. So it's a choice between doing `owners=None` and starting the function with

```
if owners is None:
    owners = []
```

and adding an argument in the tests. Since we don't need a default argument, I'd prefer not to do that in the actual code just to make the tests look slightly nicer.

Tarrasch: Ok. Just out of curiosity, is it still an anti-pattern to have an immutable object like `tuple()`?

daveFNbuck:
I see you approved it. Should I rebase down to fewer commits for the merge?

luigi/batch_notifier.py
daveFNbuck: Yeah, I think a tuple is fine in the same way a number is fine. You only get this unexpected outcome with mutable objects.

Tarrasch:
@daveFNbuck, I merged this with every commit because the first 4 commits were very good. The last few commits were pretty crude but it's probably not worth anyone's time to meld them back into the right commits.

PR #1917, opened 2016-11-09
Add support for BigQuery regional location

fabriziodemaria:
## Description
This PR aims at providing support for regional location for BigQuery (BQ) datasets.
More information about regional location here: https://cloud.google.com/bigquery/docs/managing_jobs_datasets_projects

The intended behaviour follows:
- If a new BQ dataset is created when uploading a BQ table, a regional location can be specified by the user for that dataset (note that all tables within the dataset inherit the same location value). BQ datasets can also have unspecified regional location.
- If the target BQ dataset is already available when uploading a new BQ table, the optional location parameter set by the user must be the same of the remote dataset or the upload will fail. If the remote dataset has unspecified regional location, no location parameter can be enforced by the user. If the client does not specify a regional location, the one of the existing dataset will be adopted.

## Motivation and Context
Users should have control over the regional location for BigQuery datasets. 

I have tested this by uploading tables on a BigQuery project setup for testing.

## TODO

- [x] Tests


mrunesson:
👍  As you say tests is good to add. Suggest one test with location and also double check that without location is covered properly in existing tests.

tox.ini
```diff
@@ -20,6 +20,7 @@ deps=
   cdh,hdp: hdfs>=2.0.4,<3.0.0
   postgres: psycopg2<3.0
   gcloud: google-api-python-client>=1.4.0,<2.0
+  gcloud: testfixtures
```
Tarrasch: Can you add an upper bound?

luigi/contrib/bigquery.py
```diff
@@ -118,14 +118,23 @@ def __init__(self, oauth_credentials=None, descriptor='', http_=None):
...
 
         try:
-            self.client.datasets().get(projectId=dataset.project_id,
-                                       datasetId=dataset.dataset_id).execute()
+            response = self.client.datasets().get(projectId=dataset.project_id,
+                                                  datasetId=dataset.dataset_id).execute()
+            if dataset.location is not None:
+                fetched_location = response.get('location', '')
+                if not fetched_location:
+                    fetched_location = 'undefined'
```
ulzha: If we want `'undefined'` as the default value then we should specify that instead of `''`...

ulzha: Using proper Python `None` would seem even simpler.

test/contrib/bigquery_gcloud_test.py
```diff
@@ -31,14 +31,20 @@
 from contrib import gcs_test
 from nose.plugins.attrib import attr
 
+from testfixtures import should_raise
```
ulzha: We can avoid the new testfixtures dependency - nose has [`assertRaises`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises)

test/contrib/bigquery_gcloud_test.py
```diff
@@ -26,19 +26,44 @@
...
+except ImportError:
+    raise unittest.SkipTest('Unable to load googleapiclient module')
 
-from contrib import gcs_test
 from nose.plugins.attrib import attr
+from testfixtures import should_raise
 
-PROJECT_ID = gcs_test.PROJECT_ID
+# In order to run this test, you should set your GCS/BigQuert project/bucket.
+# Unfortunately there's no mock
```
ulzha: It would be exciting to have automatically generated mocks to speed up local testing... In the manner of http://martinfowler.com/bliki/SelfInitializingFake.html or similar...

Tarrasch:
@ulzha, did you consider using squash-merge to keep the history cleaner? In particular I think merge-commits like https://github.com/spotify/luigi/pull/1917/commits/9d567a452565b0d389bacdc6f2d587e84e225fd2 have a pretty low valuable-history to messiness ratio.

Tarrasch:
@fabriziodemaria @mrunesson @ulzha, have you guys checked that the bulid works well after this is merged? I see some failures on the `py27-gcloud` tox environment (run `tox -e py27-gcloud`, but it requires some setups).

https://travis-ci.org/spotify/luigi/builds

ulzha:
Yes, I saw a green status. Something maybe nondeterministic? Going to have to inspect...

Yes we do run py27-gcloud locally (in a Dockerized easily usable setup... which I hope we get to opensource soon...)

Tarrasch:
@ulzha, you're right about green status. I can explain why: Fabrizio opened this pull-request from his own fabriziodemaria/luigi repo, so Travis didn't unencrypt the gcs-credentials as it would have done if the PR was sent from spotify/luigi repository. I implemented this complicated machinery to work some months before leaving spotify. See https://github.com/spotify/luigi/commit/375a47060211114675df7c5fae60b77f19cad513 :)

Also, kudos on setting up the dockerized stuff! Making luigi builds more stable would be so awesome!

Issue #1872, opened 2016-10-04
Assistant dynamic resources


fmorency:
**Use case**: Spawn new assistant instances on-the-fly (scale up/down) and connect them to the central planner
**Issue**: Resource pool is allocated when central planner is starting via configuration file. Dynamic resource pool modification is currently not possible from the Luigi client
**Solution**: It would be nice to be able to push new host-based resources pool dynamically when a new assistant is launched. For example,

Assistant A has 8 resources available
Assistant B has 6 resources available
Assistant C has 2 resources available

Task X requires 4 resources
Task Y requires 1 resources

Scheduler should be smart enough to schedule X on A or B (but not C) and Y on A, B or C. Lets spawn a new assistant

Assistant D has 12 resources available

The Luigi client should be able to notify the central planner that new resources are available and start scheduling jobs on it. The same mechanic applies when an assistant goes offline.

This is mostly food for tough and I would like the community's opinion on such a feature.

Tarrasch:
Thanks for the nice description and interesting issue @fmorency. Have you thought about leaving out the assistants concept from this? I think worker-based or host-based resource limitations is what we look for. That is it's regardless if a worker is an assistant or not. What do you think?

fmorency:
You are correct @Tarrasch. I took an example from what I'm working on at the moment, but ultimately we are looking for {host/worker}-based resources pushed via the Luigi client and without restarting the central planner.

Tarrasch:
As far as I remember, this has been requested many times by multiple parties. If you would go ahead and implement it. That would be great for sure. :)

daveFNbuck:
I missed this until now, but I have a fairly longstanding PR #1669 that deals with this.

Issue #1362, opened 2015-10-29
how to define dynamic resources according to luigi parameters?


leafjungle:
for example, x = luigi.Parameter()
if x==1:
      resource = {"r1": 1}
elif x==2:
      resource = {"r2": 1}

What can I do? redefine the **init** function?

erikbern:
not sure how to do it. maybe define a property decorator?

daveFNbuck:
I actually do this a lot and a property decorator is the answer:

```
class ExampleTask(luigi.Task):
    x = luigi.IntParameter()

    @property
    def resources(self):
        if self.x == 1:
            return {'r1': 1}
        else:
            return {'r2': 2}
```

quentinsf:
@daveFNbuck - many thanks!  That's just what I needed!

I'm updating HDF5 files, and I need to make sure that no two tasks try to write to the same file at once. Following your example of using a property, I can do (effectively):

    class HDF5TestTable(luigi.Task):
        ...

        @property
        def resources(self):
            return {self.hdf_file_name: 1}

Seems to work nicely!  I'll do a pull request to get something like this into the documentation, I think.

visjedi:
How could you ever check this? Is there a small test/hello world example to show this indeed works and the file is held by only worker at a time? I shall be glad to help if someone is interested.

daveFNbuck:
@visjedi we have unit tests for resources. Luigi can only guarantee that you won't exceed your resource limit. It's on you to set up your resources such that this means you won't have multiple jobs touching the same file at the same time.

dlstadther:
This looks already resolved

PR #1918, opened 2016-11-13
Scheduler keeps resources for each worker

TKNGUE:
Scheduler keeps resources for each worker

## Description
  Change the scheduler's behavior so that it keeps the resources for
  each worker.
  The change provide dynamically scheduling of resources.

## Motivation and Context
  Conventional implementation cannot support Dynamic resource allocation.
  Users strongly wish the feature.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -287,7 +287,8 @@ class Task(object):
         self.time_running = None  # Timestamp when picked up by worker
         self.expl = None
         self.priority = priority
-        self.resources = _get_default(resources, {})
+        self.resources = collections.defaultdict(
+            dict, _get_default(resources, {}))
         self.family = family
         self.module = module
         self.params = _get_default(params, {})
@@ -807,7 +808,8 @@ class Scheduler(object):
             task.deps.update(new_deps)
 
         if resources is not None:
-            task.resources = resources
+            task.resources[worker_id] = resources
+
 
         if worker.enabled and not assistant:
             task.stakeholders.add(worker_id)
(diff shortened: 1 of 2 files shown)


daveFNbuck:
Is there a reason you're doing this per-worker rather than per-host like in #1669?

TKNGUE:
Firstly, I tried to do so. But, It has a problem.
When new workers have launched, resource information that scheduler keeps is overwritten
due to synchronizing to task states of workers by the implementation of luigi.
I thinks it is undesirable. That means that the behaviors of running tasks can be changed by launching the new worker. 
So I keep the resource information by worker id instead of host.

daveFNbuck:
Ah, so this is actually making tasks have different resources depending on the worker, rather than having worker-based limits? This is completely different from the features desired in #1872 and #1362. Perhaps you can better explain what the purpose of this is? Why would resources used change depending on who's running it?

TKNGUE:
I'm sorry for my misunderstanding about the issues.

The PR have the mainly two purpose
Let me taking a example.

When If I implemented following, this is a dynamic resource allocating example.
```python
class TaskA(luigi.Task):
    param = luigi.Parameter(1)

    @property
    def resources(self):
          import os
          return os.uname().nodename
# ...  Define requires and run

     def run(self):
     # ... heavy process

luigi.build([TaskA(1), TaskA(2)])
```
In this example, needed resources of the task are decided at launched.
If you run the program by workers in different host, namely _host1_ and _host2_ , 
the scheduler get workers to run as my expectation, but it doesn't.

According to my analyze, this is due to override the resource information scheduler keep by newly launched worker ( This is ``add_task``'s behavior ).
My PR protected from the overwriting. (Keeping host-based limitation will cause same problem)

I thinks the PR enable task to allocate resources more dynamically.

daveFNbuck:
I think your example fits the use case of #1669 pretty well.

TKNGUE:
Okay. thanks.  I close this issue.

PR #1919, opened 2016-11-16
Make temporary_path() decorator create directories

Tarrasch:
## Description

I noticed that my program crashed in the cases where I didn't manually
create the directories. And in all cases so far I've needed for the
directories to be created.

Will I definitely think this feature should be possible to turn off, I
think it's best to have as "on by default". But I left it out for anyone
making a patch in the future.

I've added test cases and tested it very lightly in production.

### DIFF
--- a/luigi/target.py
+++ b/luigi/target.py
@@ -128,11 +128,6 @@ class FileSystem(object):
                              exist, raise luigi.target.MissingParentDirectory
         :param bool raise_if_exists: raise luigi.target.FileAlreadyExists if
                                      the folder already exists.
-
-        *Note*: This method is optional, not all FileSystem subclasses implements it.
-
-        *Note*: parents and raise_if_exists were added in August 2014. Some
-                implementations might not support these flags yet.
         """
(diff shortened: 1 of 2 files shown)


luigi/target.py
```diff
@@ -128,11 +128,6 @@ def mkdir(self, path, parents=True, raise_if_exists=False):
                              exist, raise luigi.target.MissingParentDirectory
         :param bool raise_if_exists: raise luigi.target.FileAlreadyExists if
                                      the folder already exists.
-
-        *Note*: This method is optional, not all FileSystem subclasses implements it.
-
-        *Note*: parents and raise_if_exists were added in August 2014. Some
-                implementations might not support these flags yet.
```
Tarrasch: A note for however is curious.

I think these two notes add negative value only. In general we want people to implement this method (in particular after this patch). If we leave these old (and currently mis-rendered) notes, file system implementors might not implement this method referring to this.

Tarrasch:
If nobody else complains I'll merge this tomorrow. :)

(I've verified that the Travis Issue is not this patch's fault)

PR #1039, opened 2015-07-01
Cleanup interface and parameters

Tarrasch:
This is a set of commits in preparation for a bigger re-factoring of the parameters. I doubt I'll have to finish the bigger re-factoring any time soon.

PR #1922, opened 2016-11-17
Cleanup code by removing `if __name__ == '__main__'` pattern

Tarrasch:
This patch started out as fixing the red in the current travis flake8
step.  As most occurrences was about simply removing the over-(mis)-used
`if __name__ == "__main__"` pattern that was way to prevalent in the luigi
code base.

This is kind of analogous to the other time flake8 suddenly got more
sensitive. See spotify/luigi#1786

### DIFF
--- a/doc/command_line.rst
+++ b/doc/command_line.rst
@@ -1,4 +1,4 @@
-.. highlight:: bash
+.. _CommandLine:
 
 Running from the Command Line
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/examples/elasticsearch_index.py
+++ b/examples/elasticsearch_index.py
@@ -14,7 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
 import datetime
 import json
 
--- a/examples/foo.py
+++ b/examples/foo.py
@@ -14,10 +14,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+"""
+You can run this example like this:
+
+    .. code:: console
+
+            $ rm -rf '/tmp/bar'
+            $ luigi --module examples.foo examples.Foo --workers 2 --local-scheduler
+
+"""
 
 from __future__ import print_function
-import os
-import shutil
 import time
 
 import luigi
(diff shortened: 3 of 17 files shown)


dlstadther:
@Tarrasch , Are you saying that `$ luigi --module examples.foo examples.Foo --workers 2 --local-scheduler` is the only "proper" way to execute luigi code and that `if __name__ == '__main__'` shouldn't be used? Why is this?

Tarrasch:
I believe that to be the case yes. I think there's a few advantages:

 * Less boilerplate
 * Your file is even more "declarative". I think the luigi API should have a [declarative feel](http://www.arashrouhani.com/luigi-budapest-bi-oct-2015/#/12), just like GNU make.
 * Reading a line like `python src/python/blah/blah/luigi_tasks.py ...` feels like it could do anything, having something like `luigi --module blah/blah/luigi_tasks ...` is clearer that it's about starting luigi.
 * Python community in general seem to be for "have only one way to do things". I don't like that we're currently not consistent in our docs and code.

Come to think about it. I made a similar statement when I [deprecated `luigi.run()`](https://github.com/spotify/luigi/pull/1039/commits/0dec90c16fa04dbab55c5e2d34561e151a2b61d8).

Does all this make sense to you? :)

erikbern:
I don't mind the `if __name__ == '__main__`` personally. I generally hate overly prescriptive frameworks that impose rules on how to execute stuff. But I guess this is only internal cleanup so LGTM

dlstadther:
Shoot, I'm just behind on the times then. Perhaps due to my comfort with Python, I've always used

```python
if __name__ == "__main__":
    luigi.run(['NameOfMyTopLevelTask', '--some-option', 'the_value'])
```

Since I don't print or see my non-errors, I was unaware of `luigi.run` deprecation.

I'll switch over to `luigi --module /path/to/file.py TaskName --option value`

Tarrasch:
Reviewers: Seems like the actual flake8 issues were fixed in https://github.com/spotify/luigi/commit/a540006ec802bb15791c0bde5a767d3c00c72971 already, thanks @mrunesson. :)

----

@dlstadther, I'm still not sure if `luigi.run` was a correct thing. Please feedback how using `luigi` plays out for you.

erikbern:
@dlstadther in that case there's no need to involve command line parsing. iirc you can just do `luigi.build([NameOfMyTask('banana')])`

Issue #1924, opened 2016-11-18
Luigi doesn't output FAILURE processing time, end time and start time


shaynewang:
When a task is failed the processing time is not recorded. As a result, the processing_time of a process left as initialized value.
```
{'job_id':'1479211214', 'datetime':'2016-11-15 04:12:44', 'task_id':'AdobeLoad_2016_11_14_page_summary_5281aeca7d', 'task_name':'AdobeLoad', 'task_date':'2016-11-14','task_report_name':'page_summary', 'start_time':'1479211963.7', 'end_time':'1479211964.12', 'process_time':'None', 'status':'FAILURE'}
{'job_id':'1479229552', 'datetime':'2016-11-15 09:07:09', 'task_id':'AdobeLoad_2016_11_14_page_summary_5281aeca7d', 'task_name':'AdobeLoad', 'task_date':'2016-11-14','task_report_name':'page_summary', 'start_time':'1479229554.92', 'end_time':'1479229629.66', 'process_time':'74.7349998951', 'status':'SUCCESS'}
```
Processing time for Failure events shouldn't be None or 0, therefore processing_time for FAILURE event should be calculated before exits.

Luigi let failed tasks have no processing time information. As a result, it could be hard to track failures. I think time information should be added to failed tasks.

Tarrasch:
What are does dicts? Did you create them yourself or is it a luigi future?

shaynewang:
@Tarrasch is just an output not by luigi

PR #1925, opened 2016-11-18
Added processing_time to task FAILURE

shaynewang:
Added processing time to FAILURE tasks

## Description
Added line 211 and 212 to assign processing_time to failed tasks

## Motivation and Context
If no processing time assigned to failed tasks, initialized value will be logged as processing time. If user initializes processing_time as None, they will get "None" as processing time, which creates problem if they are pushing the log data to a sql database with field type as decimal numbers.
https://github.com/spotify/luigi/issues/1924

I have included unit test, and unit tests passed.

Signed-off-by: Shayne Wang <shaynexwang@gmail.com>

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -160,6 +160,7 @@ class TaskProcess(multiprocessing.Process):
             # Need to have different random seeds if running in separate processes
             random.seed((os.getpid(), time.time()))
 
+        t0 = time.time() # Failed task start time
         status = FAILED
         expl = ''
         missing = []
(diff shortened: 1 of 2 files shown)


test/event_callbacks_test.py
```diff
@@ -148,7 +148,10 @@ def test_processing_time_handler_success(self):
 
     def test_processing_time_handler_failure(self):
         t, result = self._run_processing_time_handler(True)
-        self.assertEqual(result, [])
+        self.assertEqual(len(result), 1)
+        task, time = result[0]
+        self.assertTrue(task is t)
```
Tarrasch: there is also `self.assertIs`

shaynewang: @Tarrasch Cool, thanks

luigi/worker.py
```diff
@@ -160,6 +160,7 @@ def run(self):
             # Need to have different random seeds if running in separate processes
             random.seed((os.getpid(), time.time()))
 
+        t0 = time.time() # Failed task start time
```
chrispalmer: This is unnecessary. The declaration of `t0` on line 179 is still in scope inside the `except` block.

PR #1926, opened 2016-11-21
Include FAILED and UPSTREAM_FAILED tasks in count_pending

daveFNbuck:
## Description
Reverts #1789 and replaces it by counting FAILED tasks in count_pending

## Motivation and Context
#1789 was meant to allow workers to stay alive in order to keep working on FAILED tasks after they become PENDING again. The approach in that PR didn't really work well, as it missed cases where the root task is failed or where all leaf tasks are failed. Instead, we just directly count FAILED tasks as part of the pending count, as well as counting UPSTREAM_FAILED tasks as PENDING. UPSTREAM_DISABLED tasks are still not counted, even if only one dependency is DISABLED.

I have included unit tests.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -363,17 +363,19 @@ class Worker(object):
         if self.last_active + config.worker_disconnect_delay < time.time():
             return True
 
-    def get_pending_tasks(self, state):
+    def get_pending_tasks(self, state, include_failed=False):
         """
         Get PENDING (and RUNNING) tasks for this worker.
 
         You have to pass in the state for optimization reasons.
         """
         if len(self.tasks) < state.num_pending_tasks():
-            return six.moves.filter(lambda task: task.status in [PENDING, RUNNING],
-                                    self.tasks)
+            statuses = [PENDING, RUNNING]
+            if include_failed:
+                statuses.append(FAILED)
+            return six.moves.filter(lambda task: task.status in statuses, self.tasks)
         else:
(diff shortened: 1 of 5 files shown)


luigi/scheduler.py
```diff
@@ -363,17 +363,19 @@ def prune(self, config):
         if self.last_active + config.worker_disconnect_delay < time.time():
             return True
 
-    def get_pending_tasks(self, state):
+    def get_pending_tasks(self, state, include_failed=False):
```
Tarrasch: Is it possible to instead implement a `get_statuses([list of statuses])`? I've always found it a bit weird that `get_pending_tasks` also returns running tasks. This could be an opportunity to fix that.

test/worker_test.py
```diff
@@ -1774,7 +1793,7 @@ def requires(self):
 
                     self.assertTrue(w3.run())
                     self.assertFalse(w2.run())
-                    self.assertFalse(w1.run())
+                    self.assertTrue(w1.run())
```
Tarrasch: Just briefly, what changed here?

daveFNbuck: I'm actually not sure what happened here. I was planning to look into it but forgot.

Tarrasch:
Nice refactoring. I think the code is more readable now! :)

daveFNbuck:
ping

Tarrasch:
Thanks!

daveFNbuck:
No problem, just trying to clean up :)

PR #1927, opened 2016-11-21
Sets wait_jitter in worker_test.py

daveFNbuck:
## Description
Sets wait_jitter to 0.05 in all workers in worker_test.py where wait_interval is set to 0.1.

## Motivation and Context
Many tests in worker_test.py set wait_interval to 0.1 but leave wait_jitter at its default value of 5.0. This sets wait_jitter to 0.05 in all such tests, which maintains the expected 0.1 jitter time. This should reduce each wait by about 2.5 seconds, and I've found that it decreases the running time of the worker tests on my machine by about 15 seconds.

I ran the tests and they're 15 seconds faster.


Tarrasch:
Awesome catch :)

daveFNbuck:
Yeah, I was pretty surprised when I noticed it today!

PR #1928, opened 2016-11-22
Add docs for DataParameter for default=today()

Tarrasch:
This is one thing I've needed to look up a gazillion times, but I've always looked at other code and never consulted the docs. By documenting this, newcomers will hopefully will find this easier.


Tarrasch:
Merging already, as this is best reviewed as a rendered page on readthedocs.

PR #1944, opened 2016-12-02
Make RunAnywayTarget compatible with python 2.7

Marigold:
## Description
Make `RunAnywayTarget` compatible with python 2.7

## Motivation and Context
`exist_ok` parameter doesn't exist for function `os.makedirs` in python 2.7.12

I tested it locally with my existing pipeline (can add unit tests if necessary).

### DIFF
--- a/luigi/contrib/simulate.py
+++ b/luigi/contrib/simulate.py
@@ -101,5 +101,8 @@ class RunAnywayTarget(luigi.Target):
         logger.info('Marking %s as done', self)
 
         fn = self.get_path()
-        os.makedirs(os.path.dirname(fn), exist_ok=True)
+        try:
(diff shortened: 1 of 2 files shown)


Yoone:
Since I am being invited in the review process, it looks good to me!

`os.makedirs()` should indeed raise an `OSError` in both 2.7 and 3.x.

Marigold:
> Can you also add a test case?

The original tests were skipped in py27 environment (they were using `exist_ok` and the error was catched & skipped). I fixed them and they now test also the functionality I added (they were failing before the fix). Is there anything else or are we ready to merge this?

Tarrasch:
Looks good to me :)

PR #1946, opened 2016-12-02
Add retries when polling for the job status in BigQuery

fabriziodemaria:
## Description
Add retries when polling for the job status in BigQuery, in case of specific HTTP errors encountered.

## Motivation and Context
As suggested by documentation from Google, in case of `5xx` errors encountered while polling for the job status via `jobs.get()`, it is advised to try to poll again [1].
Without this fix, Luigi might exit with a failing Task even if the underlying loading job to BigQuery is still running.

Additionally, the way the job success is verified is changed to use `status.errorResult` instead of `status.errors` [2].

- Tested that these changes cover a success scenario in the exact same way as before.
- Tested that critical failures are handled in the same way as before, but using `errorResult`:
```
  File "/src/luigi/luigi/contrib/bigquery.py", line 332, in run_job
    raise Exception('BigQuery job failed: {}'.format(status['status']['errorResult']))
Exception: BigQuery job failed: {u'reason': u'invalid', u'message': u'The Apache Avro library failed to parse file...
```
- Unfortunately it is hard to reproduce a `5xx` http error to test the retrying mechanism  [3], but I think in the worst case Luigi would just fail instead of retrying thus handling this scenario no worse than before.

---
[1] - https://cloud.google.com/bigquery/troubleshooting-errors#backendError
[2] - https://cloud.google.com/bigquery/loading-data#loading_avro_files
[3] - https://google.github.io/google-api-python-client/docs/epy/googleapiclient.http.HttpRequest-class.html#execute

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -326,7 +326,7 @@ class BigQueryClient(object):
         job_id = new_job['jobReference']['jobId']
         logger.info('Started import job %s:%s', project_id, job_id)
         while True:
-            status = self.client.jobs().get(projectId=project_id, jobId=job_id).execute()
+            status = self.client.jobs().get(projectId=project_id, jobId=job_id).execute(num_retries=10)
             if status['status']['state'] == 'DONE':
                 if status['status'].get('errorResult'):
                     raise Exception('BigQuery job failed: {}'.format(status['status']['errorResult']))
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -328,12 +328,12 @@ class BigQueryClient(object):
         while True:
             status = self.client.jobs().get(projectId=project_id, jobId=job_id).execute()
             if status['status']['state'] == 'DONE':
-                if status['status'].get('errors'):
-                    raise Exception('BigQuery job failed: {}'.format(status['status']['errors']))
+                if status['status'].get('errorResult'):
+                    raise Exception('BigQuery job failed: {}'.format(status['status']['errorResult']))
                 return
 
             logger.info('Waiting for job %s:%s to complete...', project_id, job_id)
-            time.sleep(5.0)
+            time.sleep(5)
 
     def copy(self,
              source_table,


luigi/contrib/bigquery.py
```diff
@@ -325,15 +327,25 @@ def run_job(self, project_id, body, dataset=None):
...
-                if status['status'].get('errors'):
-                    raise Exception('BigQuery job failed: {}'.format(status['status']['errors']))
-                return
-
-            logger.info('Waiting for job %s:%s to complete...', project_id, job_id)
-            time.sleep(5.0)
+            try:
+                status = self.client.jobs().get(projectId=project_id, jobId=job_id).execute()
+                if status['status']['state'] == 'DONE':
+                    if status['status'].get('errors'):
```
TinaRanic: Errors being present doesn't imply that the job failed, presence of errorResult means that the job failed.
See https://cloud.google.com/bigquery/loading-data#loading_avro_files and example here: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/bigquery/cloud-client/load_data_from_file.py#L58

luigi/contrib/bigquery.py
```diff
@@ -325,15 +327,25 @@ def run_job(self, project_id, body, dataset=None):
...
+                    return
+
+                logger.info('Waiting for job %s:%s to complete...', project_id, job_id)
+                poll_counter = 0
+                time.sleep(5.0)
+            except HttpError as httpError:
+                if int(httpError.resp.status) == 503 and poll_counter < self.MAX_POLLING_RETRIES:
+                    poll_counter = poll_counter + 1
+                    logger.warning('Got error 503 while requesting info for job %s:%s, retrying in %d seconds...', project_id, job_id, poll_counter)
+                    time.sleep(float(poll_counter))
```
bergman: Tt doesn't have to be float, at least according to docstring of `time.sleep`. You could remove the `.0` in the other call too.

mbruggmann:
@fabriziodemaria Have you tried `request.execute(num_retries=X)`? Looks like that is a thing in the google api python  client...

fabriziodemaria:
> Have you tried request.execute(num_retries=X)? Looks like that is a thing in the google api python client...

Awesome @mbruggmann, thanks for pointing that out! I am changing the PR to make use of that instead.

luigi/contrib/bigquery.py
fabriziodemaria: Good catch, will update this to use `errorResult` instead.

luigi/contrib/bigquery.py
fabriziodemaria: +1

bergman:
👍

luigi/contrib/bigquery.py
```diff
@@ -326,14 +326,14 @@ def run_job(self, project_id, body, dataset=None):
         job_id = new_job['jobReference']['jobId']
         logger.info('Started import job %s:%s', project_id, job_id)
         while True:
-            status = self.client.jobs().get(projectId=project_id, jobId=job_id).execute()
+            status = self.client.jobs().get(projectId=project_id, jobId=job_id).execute(num_retries=10)
             if status['status']['state'] == 'DONE':
-                if status['status'].get('errors'):
-                    raise Exception('BigQuery job failed: {}'.format(status['status']['errors']))
+                if status['status'].get('errorResult'):
+                    raise Exception('BigQuery job failed: {}'.format(status['status']['errorResult']))
```
ulzha: Is `'errors'` no more in the dict at this point? I would try to make sure that the exception dumps any information available. Why not all the `status`?

ulzha:
Great work!

If we have tested that the code is capable of handling a success scenario and a failure scenario no worse than the previous one, then :shipit:

fabriziodemaria:
@uldis 
- I tested that these changes cover a success scenario in the exact same way as before.
- I tested that critical failures are handled in the same way as before, but using `errorResult`:

```
  File "/src/luigi/luigi/contrib/bigquery.py", line 332, in run_job
    raise Exception('BigQuery job failed: {}'.format(status['status']['errorResult']))
Exception: BigQuery job failed: {u'reason': u'invalid', u'message': u'The Apache Avro library failed to parse file...
```
 
- Unfortunately it is hard to reproduce a `5xx` http error to test the retrying mechanism, but I think in the worst case Luigi would just fail instead of retrying thus handling this scenario no worse than before.

bergman:
👍

Issue #1948, opened 2016-12-02
retry_count configuration documentation is incorrect


jkeung:
In https://github.com/spotify/luigi/blob/master/luigi/scheduler.py line 137-138, retry count is defined as  
```
retry_count = parameter.IntParameter(default=999999999,
                                         config_path=dict(section='scheduler', name='disable_failures'))
```

Setting `retry_count` in luigi.cfg does not work, the parameter that must be set is `disable_failures` in order for this feature to work. http://luigi.readthedocs.io/en/stable/configuration.html

PR #1949, opened 2016-12-02
Update scheduler.py ISSUE #1948

jkeung:
## Description
Renaming disable_failures to retry_count in configuration file to be consistent with the documentation.

## Motivation and Context
Currently if a job fails it will continue to run because the retry_count configuration is being referenced improperly.

I was running luigi with 2 workers having 1 job run properly and 1 job fail every minute. When I use retry_count=2, the job will continue to fail. However, when i use disable_failures, the job will stop and disable the job. I have come to the assumption that the code is wrong in scheduler.py because it is looking for disable_failures. It would be easier to change here than to redo the documentation.

The configuration for retry_count has an incorrect variable name of disable_failures. This should be retry_count.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -135,7 +135,7 @@ class scheduler(Config):
     disable_window = parameter.IntParameter(default=3600,
                                             config_path=dict(section='scheduler', name='disable-window-seconds'))
     retry_count = parameter.IntParameter(default=999999999,
-                                         config_path=dict(section='scheduler', name='disable_failures'))
+                                         config_path=dict(section='scheduler', name='retry_count'))
     disable_hard_timeout = parameter.IntParameter(default=999999999,
                                                   config_path=dict(section='scheduler', name='disable-hard-timeout'))
     disable_persist = parameter.IntParameter(default=86400,


Tarrasch:
Actually the name is automatic and based on the name of the variable name, which already is `retry_count`. Are you sure this is not a matter of old luigi version? It was renamed to `retry_count` only a couple of months ago. :)

jkeung:
I have a question. What is this line doing actually? Isn't it reading in 'disable_failures' from the configuration file?
```retry_count = parameter.IntParameter(default=999999999,
                                         config_path=dict(section='scheduler', name='disable_failures'))
```

Tarrasch:
@jkeung, good question. Please check out [this](http://luigi.readthedocs.io/en/stable/configuration.html#configuration-classes) part of the docs out, then we can work on anything in paritular that's unclear.

jkeung:
@Tarrasch Hmm. I'm using luigi 2.4.0 and my luigi.cfg file has these parameters under [scheduler]
```
[scheduler]
retry_count=5
record_task_history=True
retry-delay=60
```
I noticed that when the job fails multiple times it sends me failure emails and never disables the task.
However, when I use, the task will become disabled after 5 failed runs.
```
[scheduler]
disable_failures=5
record_task_history=True
retry-delay=60
```
btw, I am using global configuration, not Config class configuration. Is this expected behavior?

Tarrasch:
@jkeung, that's not expected behavior. But there seem to exist [tests](https://github.com/spotify/luigi/blob/6f0d175c733da94c202634aa02bfe5efab6fa6cf/test/scheduler_test.py#L120-L123) for this already. Can you try to debug to see why it doesn't work for you with a real config?

Issue #1932, opened 2016-11-24
The retcode configuration does not consider task retry


bwtakacy:
Hi.

 I use the retcode configurations as below in luigi.cfg,
```
[retcode]
already_running=10
missing_data=20
not_run=25
task_failed=30
scheduling_error=35
unhandled_exception=40
```

When an one task was once failed and then succeeded in next retry in scheduling during a whole tasks execution, I expected the return code to be zero but actually was non zero.

For example,  running `examples/foo.py`  with some code change that `Bar(num=7)`  fails at the first time run but succeeds in the second time run,  all the tasks succeeded in the end but the return code is non zero.

```
===== Luigi Execution Summary =====

Scheduled 12 tasks of which:
* 11 ran successfully:
    - 10 examples.Bar(num=0...9)
    - 1 examples.Foo()
* 1 failed:
    - 1 examples.Bar(num=7)

This progress looks :( because there were failed tasks

===== Luigi Execution Summary =====

The program exited via sys.exit(). Exit status: 30
```

It seems that the result of `Bar(num=7)` was doubly counted both in success group and failed group.

Tarrasch:
Definitely a bug I would say. Failures are normal and if we even know it got resolved we should not alarm anyone. Do you want to take a stab fixing this? :)

bwtakacy:
Thanks for your comments!
I will try to create PR.

PR #1951, opened 2016-12-04
Improve retcodes and execution summary for retry success

bwtakacy:
## Description

Let luigi to consider task retry success in return code handling and execution summary.

## Motivation and Context

Issue #1932 .

I ran test in local env.
Som tests(server_test) failed. I could not find why but they seems to be not related.

The current code does not consider the case when a task once
failed but succeeded in retry.

### DIFF
--- a/luigi/execution_summary.py
+++ b/luigi/execution_summary.py
@@ -378,10 +378,14 @@ def _summary_format(set_tasks, worker):
     smiley = ""
     reason = ""
     if set_tasks["failed"]:
-        smiley = ":("
-        reason = "there were failed tasks"
-        if set_tasks["scheduling_error"]:
-            reason += " and tasks whose scheduling failed"
+        if not (set_tasks["failed"] - set_tasks["completed"]):
+            smiley = ":)"
+            reason = "there were failed tasks but suceeded in retry"
+        else:
+            smiley = ":("
+            reason = "there were failed tasks"
+            if set_tasks["scheduling_error"]:
+                reason += " and tasks whose scheduling failed"
     elif set_tasks["scheduling_error"]:
         smiley = ":("
         reason = "there were tasks whose scheduling failed"
--- a/luigi/execution_summary.py
+++ b/luigi/execution_summary.py
@@ -278,7 +278,6 @@ _PENDING_SUB_STATUSES = set(_ORDERED_STATUSES[_ORDERED_STATUSES.index("still_pen
 _COMMENTS = set((
     ("already_done", 'present dependencies were encountered'),
     ("completed", 'ran successfully'),
-    ("ever_failed", 'ever failed'),
     ("failed", 'failed'),
     ("scheduling_error", 'failed scheduling'),
     ("still_pending", 'were left pending, among these'),
@@ -354,7 +353,7 @@ def _summary_format(set_tasks, worker):
     for status in _ORDERED_STATUSES:
         if status not in comments:
             continue
-        if status == 'ever_failed' and (set_tasks["ever_failed"]==set_tasks["failed"]):
+        if status == 'ever_failed' and (set_tasks["ever_failed"] == set_tasks["failed"]):
             continue
         str_output += '{0}'.format(comments[status])
         if status != 'still_pending':
@@ -385,7 +384,7 @@ def _summary_format(set_tasks, worker):
     if set_tasks["ever_failed"]:
         if not set_tasks["failed"]:
             smiley = ":)"
-            reason = "there were failed tasks but suceeded in retry"
+            reason = "there were failed tasks but they all suceeded in a retry"
         else:
             smiley = ":("
             reason = "there were failed tasks"
--- a/luigi/execution_summary.py
+++ b/luigi/execution_summary.py
@@ -353,8 +353,6 @@ def _summary_format(set_tasks, worker):
     for status in _ORDERED_STATUSES:
         if status not in comments:
             continue
-        if status == 'ever_failed':
-            continue
         str_output += '{0}'.format(comments[status])
         if status != 'still_pending':
             str_output += '{0}\n'.format(_get_str(group_tasks[status], status in _PENDING_SUB_STATUSES))
--- a/luigi/execution_summary.py
+++ b/luigi/execution_summary.py
@@ -43,12 +43,13 @@ def _partition_tasks(worker):
     set_tasks["completed"] = {task for (task, status, ext) in task_history if status == 'DONE' and task in pending_tasks}
     set_tasks["already_done"] = {task for (task, status, ext) in task_history
                                  if status == 'DONE' and task not in pending_tasks and task not in set_tasks["completed"]}
-    set_tasks["failed"] = {task for (task, status, ext) in task_history if status == 'FAILED'}
+    set_tasks["ever_failed"] = {task for (task, status, ext) in task_history if status == 'FAILED'}
+    set_tasks["failed"] = set_tasks["ever_failed"] - set_tasks["completed"]
     set_tasks["scheduling_error"] = {task for(task, status, ext) in task_history if status == 'UNKNOWN'}
     set_tasks["still_pending_ext"] = {task for (task, status, ext) in task_history
-                                      if status == 'PENDING' and task not in set_tasks["failed"] and task not in set_tasks["completed"] and not ext}
+                                      if status == 'PENDING' and task not in set_tasks["ever_failed"] and task not in set_tasks["completed"] and not ext}
     set_tasks["still_pending_not_ext"] = {task for (task, status, ext) in task_history
-                                          if status == 'PENDING' and task not in set_tasks["failed"] and task not in set_tasks["completed"] and ext}
+                                          if status == 'PENDING' and task not in set_tasks["ever_failed"] and task not in set_tasks["completed"] and ext}
     set_tasks["run_by_other_worker"] = set()
     set_tasks["upstream_failure"] = set()
     set_tasks["upstream_missing_dependency"] = set()
@@ -87,7 +88,7 @@ def _depth_first_search(set_tasks, current_task, visited):
         for task in current_task._requires():
             if task not in visited:
                 _depth_first_search(set_tasks, task, visited)
-            if task in set_tasks["failed"] or task in set_tasks["upstream_failure"]:
+            if task in set_tasks["ever_failed"] or task in set_tasks["upstream_failure"]:
                 set_tasks["upstream_failure"].add(current_task)
                 upstream_failure = True
             if task in set_tasks["still_pending_ext"] or task in set_tasks["upstream_missing_dependency"]:
@@ -261,6 +262,7 @@ def _get_comments(group_tasks):
 _ORDERED_STATUSES = (
     "already_done",
     "completed",
+    "ever_failed",
     "failed",
     "scheduling_error",
     "still_pending",
@@ -276,6 +278,7 @@ _PENDING_SUB_STATUSES = set(_ORDERED_STATUSES[_ORDERED_STATUSES.index("still_pen
 _COMMENTS = set((
     ("already_done", 'present dependencies were encountered'),
     ("completed", 'ran successfully'),
+    ("ever_failed", 'ever failed'),
     ("failed", 'failed'),
     ("scheduling_error", 'failed scheduling'),
     ("still_pending", 'were left pending, among these'),
@@ -351,6 +354,8 @@ def _summary_format(set_tasks, worker):
     for status in _ORDERED_STATUSES:
         if status not in comments:
             continue
+        if status == 'ever_failed' and (set_tasks["ever_failed"]==set_tasks["failed"]):
+            continue
         str_output += '{0}'.format(comments[status])
         if status != 'still_pending':
             str_output += '{0}\n'.format(_get_str(group_tasks[status], status in _PENDING_SUB_STATUSES))
@@ -377,8 +382,8 @@ def _summary_format(set_tasks, worker):
         str_output += 'Did not run any tasks'
     smiley = ""
     reason = ""
-    if set_tasks["failed"]:
-        if not (set_tasks["failed"] - set_tasks["completed"]):
+    if set_tasks["ever_failed"]:
+        if not set_tasks["failed"]:
             smiley = ":)"
             reason = "there were failed tasks but suceeded in retry"
         else:
--- a/luigi/execution_summary.py
+++ b/luigi/execution_summary.py
@@ -353,7 +353,7 @@ def _summary_format(set_tasks, worker):
     for status in _ORDERED_STATUSES:
         if status not in comments:
             continue
-        if status == 'ever_failed' and (set_tasks["ever_failed"] == set_tasks["failed"]):
+        if status == 'ever_failed':
             continue
         str_output += '{0}'.format(comments[status])
         if status != 'still_pending':
--- a/luigi/retcodes.py
+++ b/luigi/retcodes.py
@@ -82,6 +82,11 @@ def run_with_retcodes(argv):
 
     task_sets = luigi.execution_summary._summary_dict(worker)
     root_task = luigi.execution_summary._root_task(worker)
+    if root_task in task_sets['completed']:
+        completed_tasks = task_sets['completed']
+        for key in task_sets.keys():
+            if key != 'completed':
+                task_sets[key] = task_sets[key] - completed_tasks
     non_empty_categories = {k: v for k, v in task_sets.items() if v}.keys()
 
     def has(status):
--- a/luigi/retcodes.py
+++ b/luigi/retcodes.py
@@ -95,7 +95,7 @@ def run_with_retcodes(argv):
 
     codes_and_conds = (
         (retcodes.missing_data, has('still_pending_ext')),
-        (retcodes.task_failed, has('failed')),
+        (retcodes.task_failed, has('ever_failed')),
         (retcodes.already_running, has('run_by_other_worker')),
         (retcodes.scheduling_error, has('scheduling_error')),
         (retcodes.not_run, has('not_run')),
--- a/luigi/retcodes.py
+++ b/luigi/retcodes.py
@@ -82,11 +82,6 @@ def run_with_retcodes(argv):
 
     task_sets = luigi.execution_summary._summary_dict(worker)
     root_task = luigi.execution_summary._root_task(worker)
-    if root_task in task_sets['completed']:
-        completed_tasks = task_sets['completed']
-        for key in task_sets.keys():
-            if key != 'completed':
-                task_sets[key] = task_sets[key] - completed_tasks
     non_empty_categories = {k: v for k, v in task_sets.items() if v}.keys()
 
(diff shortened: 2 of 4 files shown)


bwtakacy:
Thank you for quick reviewing!

> * Can you add new test cases? It's very important for this kind of change. Then it's also see what your new code actually does.

OK. I will add.


> * Did you consider consider renaming "failed" to something intermediary-sounding "ever_failed" and then just in the end define set_tasks["failed"] = set_tasks["ever_failed"] - set_tasks["completed"]. Something like that.

Oh, I forgot about this.
I'm afraid that this may do a more wide effect because changing `_ORDERED_STATUSES` set and related codes.
But I totally agree with you and will try :)

bwtakacy:
Oops.
I will check the test failure.

luigi/execution_summary.py
```diff
@@ -377,11 +382,15 @@ def _summary_format(set_tasks, worker):
...
     reason = ""
-    if set_tasks["failed"]:
-        smiley = ":("
-        reason = "there were failed tasks"
-        if set_tasks["scheduling_error"]:
-            reason += " and tasks whose scheduling failed"
+    if set_tasks["ever_failed"]:
+        if not set_tasks["failed"]:
+            smiley = ":)"
+            reason = "there were failed tasks but suceeded in retry"
```
Tarrasch: perhaps change to  "there were failed tasks but **they all** suc**c**eeded in **a** retry"

luigi/execution_summary.py
```diff
@@ -276,6 +278,7 @@ def _get_comments(group_tasks):
 _COMMENTS = set((
     ("already_done", 'present dependencies were encountered'),
     ("completed", 'ran successfully'),
+    ("ever_failed", 'ever failed'),
```
Tarrasch: Actually, I think it's better to not print this one out. I don't think it's so relevant. We should keep the execution *summary* short. What do you think?

test/execution_summary_test.py
```diff
@@ -1086,3 +1088,87 @@ class Bar(RunOnceTask):
         self.assertIn('Luigi Execution Summary', s)
         self.assertNotIn('00:00:00', s)
         self.assertNotIn('\n\n\n', s)
+
+class ExecutionSummaryWithTaskRetryTest(LuigiTestCase):
```
Tarrasch: Do you really need a whole new class? Is it not possible to copy any of the tests above and just slightly modify them to test what you're trying?

In particular I don't see why you need to do writes to the file system and `os.sleep()`.

test/retcodes_test.py
```diff
@@ -170,3 +171,68 @@ def new_func(*args, **kwargs):
         with mock.patch('luigi.scheduler.Scheduler.add_task', new_func):
             self.run_and_expect('RequiringTask', 0)
             self.run_and_expect('RequiringTask --retcode-not-run 5', 5)
+
+class RetcodesWithTaskRetryTest(LuigiTestCase):
```
Tarrasch: The same thing here. If you compare this test with the others, it's much much harder to read and understand.

luigi/execution_summary.py
bwtakacy: Yeah, it's just notification and not so important.
I will remove this from summary output.

luigi/execution_summary.py
bwtakacy: Thanks! Will fix.

test/execution_summary_test.py
bwtakacy: Sorry for ugly test case...

I could not a smart test case with the existing one.
After struggling, I have found it can by using 

```
luigi.scheduler.Scheduler(prune_on_get_work=True, retry_delay=1)
luigi.worker.Worker(scheduler=self.scheduler, keep_alive=True)
```

and flag file which controls the target task failure and success.
By them, the task once fails, and the scheduler re-assign the task quickly and the task suceeds in second try. `os.sleep()` is just a safety for waiting task re-assign.

I will keep trying to merge them into the existing class and removing ugly trick with file system and sleep.
If there is a good trick, please let me know.

test/retcodes_test.py
bwtakacy: About this, `run_with_config` was the critical point to create a separate class, because the existing one only handle retcode configuration but i want to pass `retry-delay`.

I think these two class can merge by changing the existing side `run_with_config`.
The file system and sleep trick are needed similar to `execution_summary_test`...

bwtakacy:
@Tarrasch 
Thank you for reviewing!
I added reply for your comments.

As you pointed out, the test codes are quite ugly...
I want to add a test case with below story:
* a task once fails,
* scheduler re-assing the failed task quickly,
* the failed task succeeds in second execution.

Currently, I uses the flag file for control task failure and sucess condition and several configuration for scheduler and workers which are not used in the existing test classes.

I will keep trying to modify the test codes more similar to others.

Tarrasch:
Yes! Indeed I think you should set something like `retry_delay=1`, or perhaps `retry_delay=0` or `retry_delay=0.00001` (I'm not sure exactly what works).

Also. If you don't want to set `luigi.scheduler.Scheduler(blah blah ... , retry_delay=1, ... blah blah)`, you can simply do this nowadays (thanks to #1791):

```python
class MyTask(luigi.Task):
    retry_delay = 0
```

I don't understand why you need to store state on the file system, that seems very over kill. What about using in memory?

```python
class MyTask(luigi.Task):
    times_been_run = 0
    def run(self):
        self.times_been_run += 1
        if self.times_been_run == 1:
             raise Exception("crashy crashy")
        else:
              pass  # Lets not crash on second run
```

bwtakacy:
> Also. If you don't want to set luigi.scheduler.Scheduler(blah blah ... , retry_delay=1, ... blah blah), you can simply do this nowadays (thanks to #1791):

Wow! Nice feature:)

> I don't understand why you need to store state on the file system, that seems very over kill. What about using in memory?

Aha, you are everything right.
I was thinking too complicated.
Thanks for your suggestion!

test/retcodes_test.py
```diff
@@ -204,35 +186,21 @@ def priority(self):
...
-                    f.write('foobar')
-                    f.close()
+                self.has_run = True
+                self.run_count += 1
+                if self.num == 0 and self.run_count == 1:
                     raise ValueError()
-                time.sleep(self.num)
-                self.output().open('w').close()
+                else:
+                    pass
```
Tarrasch: You can remove this `else:` block I think. :)

test/execution_summary_test.py
```diff
@@ -1089,81 +1087,41 @@ class Bar(RunOnceTask):
...
-                self.output().open('w').close()
 
-            def output(self):
-                out_file_path = os.path.join(self.work_dir, 'FooBar')
-                return luigi.LocalTarget(out_file_path)
+            def complete(self):
+                if self.run_count == 0:
+                    return False
+                else:
+                    return True
```
Tarrasch: Why not just `return self.run_count > 0`?

test/execution_summary_test.py
```diff
@@ -1086,3 +1086,47 @@ class Bar(RunOnceTask):
         self.assertIn('Luigi Execution Summary', s)
         self.assertNotIn('00:00:00', s)
         self.assertNotIn('\n\n\n', s)
+
+    def test_statuse_with_task_retry(self):
```
Tarrasch: I have a few concerns regarding this test-case:

 * There's 3 tasks. Can we make this much easier and only have one, it's run twice (you can do `self.run_task(Foo())` twice if you want to emulate retry). All you need is a task that first fails and then succeeds. Right?
 * In the end of the test. Shouldn't you check that there are no failed tests?
 * Can you like the other tests also test that the string-output (the summary) contains the strings that we expect? Something like `self.assertIn('Luigi Execution Summary', s)`, `self.assertIn('failed but then succeeded in retry blah blah', s)`

erikbern: seems like there's a spelling mistake too. statuse?

test/retcodes_test.py
bwtakacy: OK!

test/execution_summary_test.py
bwtakacy: @Tarrasch 
Hmm, I see.
I will try to do that.

@erikbern 
Oops.
Will fix:)

test/execution_summary_test.py
bwtakacy: Ah, that's more smart.

test/execution_summary_test.py
```diff
@@ -1086,3 +1086,29 @@ class Bar(RunOnceTask):
...
+
+            def complete(self):
+                return self.run_count > 0
+
+        self.run_task(Foo())
+        self.run_task(Foo())
+        d = self.summary_dict()
+        self.assertEqual({Foo()}, d['completed'])
+        self.assertEqual({Foo()}, d['ever_failed'])
+        self.assertFalse(d['upstream_failure'])
```
Tarrasch: Shouldn't you also test `self.assertFalse(d['failed'])`?

test/retcodes_test.py
```diff
@@ -170,3 +170,35 @@ def new_func(*args, **kwargs):
...
+
+    @with_config({'worker': {'keep_alive': 'True'}})
+    def test_retry_sucess_task(self):
+        class Bar(luigi.Task):
+            num = luigi.IntParameter()
+            has_run = False
+            run_count = 0
+
+            @property
+            def priority(self):
```
Tarrasch: Why do you need priority?

test/retcodes_test.py
```diff
@@ -170,3 +170,35 @@ def new_func(*args, **kwargs):
...
+                self.run_count += 1
+                if self.num == 0 and self.run_count == 1:
+                    raise ValueError()
+
+            def complete(self):
+                return self.has_run
+
+        class Foo(luigi.Task):
+            def requires(self):
+                for i in range(3):
```
Tarrasch: Can you skip requiring completely? Is only 1 task not enough? Like in the test with execution summary.

test/retcodes_test.py
```diff
@@ -170,3 +170,35 @@ def new_func(*args, **kwargs):
         with mock.patch('luigi.scheduler.Scheduler.add_task', new_func):
             self.run_and_expect('RequiringTask', 0)
             self.run_and_expect('RequiringTask --retcode-not-run 5', 5)
+
+    @with_config({'worker': {'keep_alive': 'True'}})
+    def test_retry_sucess_task(self):
+        class Bar(luigi.Task):
+            num = luigi.IntParameter()
+            has_run = False
```
Tarrasch: Things become tricky now! Because you'll have more than one instance of this task, and `has_run = False` will be set on the class not the instance. If you just change everything to a parameter-less task like the one above. You don't need to think about these complications.

luigi/execution_summary.py
```diff
@@ -351,6 +353,8 @@ def _summary_format(set_tasks, worker):
     for status in _ORDERED_STATUSES:
         if status not in comments:
             continue
+        if status == 'ever_failed' and (set_tasks["ever_failed"] == set_tasks["failed"]):
```
Tarrasch: I think there's a bug here that can lead to crash. It seems like status can be `status=ever_failed`, and then you index in  `comments[status]`. Won't that crash? What about this case?

If you change this, please also make a test case testing this.

test/retcodes_test.py
bwtakacy: I see.
I will remove `has_run` and use `num` parameter only.

test/retcodes_test.py
bwtakacy: Oh, it is not necessary.
I will remove it.

test/retcodes_test.py
bwtakacy: Yeah, you're right.
I will modify to use only 1 task.

luigi/execution_summary.py
bwtakacy: Hmm.
I think the latter condition is no longer necessary.
Because we don't show about `ever failed` in execution summary,  we should just skip in `status == 'ever_failed'` like

```
         if status == 'ever_failed':
             continue
```

How about this?

I will add a check whether the `ever failed` task information is not exist in execution summary in  the test case I have already added.

test/retcodes_test.py
```diff
@@ -170,3 +170,27 @@ def new_func(*args, **kwargs):
         with mock.patch('luigi.scheduler.Scheduler.add_task', new_func):
             self.run_and_expect('RequiringTask', 0)
             self.run_and_expect('RequiringTask --retcode-not-run 5', 5)
+
+    @with_config({'worker': {'keep_alive': 'True'}})
+    def test_retry_sucess_task(self):
```
Tarrasch: Please add a comment here saying what you're testing

```
"""
Test that a task once crashing and then succeeding counts as no failure.
"""
```

test/retcodes_test.py
```diff
@@ -170,3 +170,27 @@ def new_func(*args, **kwargs):
...
+        class Foo(luigi.Task):
+            num = luigi.IntParameter()
+            run_count = 0
+
+            def priority(self):
```
Tarrasch: Why priority?

test/retcodes_test.py
```diff
@@ -170,3 +170,27 @@ def new_func(*args, **kwargs):
         with mock.patch('luigi.scheduler.Scheduler.add_task', new_func):
             self.run_and_expect('RequiringTask', 0)
             self.run_and_expect('RequiringTask --retcode-not-run 5', 5)
+
+    @with_config({'worker': {'keep_alive': 'True'}})
+    def test_retry_sucess_task(self):
+        class Foo(luigi.Task):
+            num = luigi.IntParameter()
```
Tarrasch: No need for this parameter.

luigi/retcodes.py
```diff
@@ -82,6 +82,11 @@ def run_with_retcodes(argv):
 
     task_sets = luigi.execution_summary._summary_dict(worker)
     root_task = luigi.execution_summary._root_task(worker)
+    if root_task in task_sets['completed']:
+        completed_tasks = task_sets['completed']
+        for key in task_sets.keys():
+            if key != 'completed':
+                task_sets[key] = task_sets[key] - completed_tasks
```
Tarrasch: Do you really need all this?

I'm pretty sure you can just remove all this, and below just check `(retcodes.task_failed, has('failed'))` instead of `has('ever_failed'). No? I thought we introduced `ever_failed` and defined `failed` to mean something that in the very end was count as failed.

luigi/retcodes.py
```diff
@@ -90,7 +95,7 @@ def has(status):
 
     codes_and_conds = (
         (retcodes.missing_data, has('still_pending_ext')),
-        (retcodes.task_failed, has('failed')),
+        (retcodes.task_failed, has('ever_failed')),
```
Tarrasch: Yea, I thought you don't need to change this line.

luigi/execution_summary.py
```diff
@@ -351,6 +353,8 @@ def _summary_format(set_tasks, worker):
     for status in _ORDERED_STATUSES:
         if status not in comments:
             continue
+        if status == 'ever_failed':
+            continue
```
Tarrasch: Actually, I'm pretty sure the previous if statement will always trigger before this. So just remove this if-statement please.

test/execution_summary_test.py
```diff
@@ -1086,3 +1086,30 @@ class Bar(RunOnceTask):
...
+        self.assertFalse(d['upstream_missing_dependency'])
+        self.assertFalse(d['run_by_other_worker'])
+        self.assertFalse(d['still_pending_ext'])
+        s = self.summary()
+        self.assertIn('Scheduled 1 task', s)
+        self.assertIn('Luigi Execution Summary', s)
+        self.assertNotIn('ever failed', s)
+        self.assertIn('\n\nThis progress looks :) because there were failed tasks but they all suceeded in a retry', s)
```
Tarrasch: This parts looks really good! :)

test/execution_summary_test.py
Tarrasch: Did you forget this part @bwtakacy? :)

test/retcodes_test.py
bwtakacy: OK!

test/retcodes_test.py
bwtakacy: Right.  I will remove it.

test/retcodes_test.py
bwtakacy: Oh, this is no longer required.
Sorry for forgetting to remove it.

luigi/retcodes.py
bwtakacy: Right.
They were no longer required becasue we have introduced `ever_failed`.
Test case we have added newly are passed without them.

luigi/retcodes.py
bwtakacy: Yeah.
I will revert this change.

luigi/execution_summary.py
bwtakacy: Hmm. Right.
This also is no longer necessary.

test/execution_summary_test.py
bwtakacy: Oops!
Sorry for forgetting to apply this.

test/execution_summary_test.py
bwtakacy: Thanks! :)

bwtakacy:
Thank you for taking a time for this reviewing.
I have updated my pull request :)

test/retcodes_test.py
```diff
@@ -170,3 +170,23 @@ def new_func(*args, **kwargs):
         with mock.patch('luigi.scheduler.Scheduler.add_task', new_func):
             self.run_and_expect('RequiringTask', 0)
             self.run_and_expect('RequiringTask --retcode-not-run 5', 5)
+
+    """
+    Test that a task once crashing and then succeeding should be counted as no failure.
+    """
+    def test_retry_sucess_task(self):
+        class Foo(luigi.Task):
+            num = luigi.IntParameter()
```
Tarrasch: I think you forgot to remove this `num` parameter

bwtakacy: OK!
Will fix.

test/retcodes_test.py
```diff
@@ -170,3 +170,22 @@ def new_func(*args, **kwargs):
...
+            def run(self):
+                self.run_count += 1
+                if self.run_count == 1:
+                    raise ValueError()
+
+            def complete(self):
+                return self.run_count > 0
+
+        self.run_and_expect('Foo --scheduler-retry-delay=0', 0)
+        self.run_and_expect('Foo --scheduler-retry-delay=0 --retcode-task-failed=5', 0)
```
Tarrasch: Oh, I didn't know about the `=` syntax. Cool

Tarrasch:
@bwtakacy, I reworded the final commit message in my own words. I hope you don't mind.

Thanks for this patch! :)

bwtakacy:
@Tarrasch 
Thank you for merging this!
And thank you for very helpful review !!

I feel very exciting to use this feature in our environments with next version luigi :)

PR #1952, opened 2016-12-05
Make luigi.task.externalize return shallow copies 

Tarrasch:
## Description

Currently, running externalize on an existing class or object will have
serious side effects. Like subsequently created objects from the
original passed in class will also be externalized. This patch changes
the behavior to return a changed copy and not having as much side-effects.

Secondly, this patch officially declares this function to work on both
classes and objects.
I see two reasons we should allow for this (possibly) overly dynamic
behavior in both accepting classes and objects. First, novice users
don't pay attention to the difference, second, it still possible and it
will *seem* to work to pass in the class instead of the object. I was at
least tempted to without being sure if it was "correct". Instead let's
try to be soft on the users for this case.

I added many test cases and I've tested this in a little bit of production code.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -669,15 +669,18 @@ def externalize(taskclass_or_taskobject):
     If you however want a task class to be external from the beginning, you're
     better off inheriting :py:class:`ExternalTask` rather than :py:class:`Task`.
 
-    This function tries to be side-effect free by creating a (shallow) copy of
-    the class or the object passed in and then modify that object. In
-    particular this code shouldn't do anything.
+    This function tries to be side-effect free by creating a copy of the class
+    or the object passed in and then modify that object. In particular this
+    code shouldn't do anything.
 
     .. code-block:: python
 
         externalize(MyTask)  # BAD: This does nothing (as after luigi 2.4.0)
     """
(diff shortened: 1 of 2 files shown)


Tarrasch:
Here's the rendered docs:

selection_073

Tarrasch:
Actually, I think this needs to return a copy of the original class and or object.

Tarrasch:
Note to self: Mutating stuff is bad >_<

Here's a new version of the docs.

selection_074

Tarrasch:
I will also try the `@luigi.util.requires(externalize(MyTask))` to see if it actually works for me in production. If it does It's a win for modularity. :blush:

Tarrasch:
The last test I added currently fails. But I believe it will stop failing once we merge #1953, once that is merged I'll rebase this.

Tarrasch:
@erikbern and @ulzha , do you want to review? I'm kinda pushing luigi towards a functional immutable style favoring framework by suggesting this change. Perhaps a consequence of being a Haskell fan and recently eaten a lof of Guava. (literally, but I've also used the Java library :))

erikbern:
yeah i think this makes a lot of sense

Issue #1950, opened 2016-12-03
task_namespace lost during inheritance


riga:
Hey guys!

When inheriting from a task class that has a `task_namespace`, imho it would be pythonic that the inheriting class has the same `task_namespace` (if not set otherwise). However, this does not seem to be the case. Is this behavior intended by design?

### Possible solution

Due to the way the task register meta class implements the `_default_namespace` fallback, the proper namespace is only set when it's contained in the `class_dict`: ([link](https://github.com/spotify/luigi/blob/master/luigi/task_register.py#L67))

```python
class Register(abc.ABCMeta):

    def __new__(metacls, classname, bases, classdict):
        if "task_namespace" not in classdict:
            classdict["task_namespace"] = metacls._default_namespace

        cls = super(Register, metacls).__new__(metacls, classname, bases, classdict)
        metacls._reg.append(cls)
```

The fix would be quite simple:

```python
class Register(abc.ABCMeta):

    def __new__(metacls, classname, bases, classdict):
        cls = super(Register, metacls).__new__(metacls, classname, bases, classdict)

        if not getattr(cls, "task_namespace", None):
            cls.task_namespace = metacls._default_namespace

        metacls._reg.append(cls)
```

This implies a minor change though: the inherited namespace would be favored over the default one, except when the inherited one is falsy.

Any thoughts?

Tarrasch:
I'm all for this change. Pretty sure this is an old old mistake lingering around. Hopefully not too many rely on this unpythonic behavior.

Tarrasch:
I reopen this issue. Because I've found the recent PR to be insufficient. The good news is that I already know a sort of a solution and I'm willing to work on it myself. Read on for details:

## Current problem since #1953 

First, the implemented solution is okay, but at least my code started to break immediately. I Had this pattern.

```
$ cat file_a.py
luigi_namespace(__module__)

class GenericBaseTask(luigi.Task)
    pass

$ cat file_b.py
import file_a
luigi_namespace(__module__)

class SpecificTask(file_a.GenericBaseTask)
    pass
```

Before #1953, I would have `file_b.SpecificTask` which I find logical. I mean it's created there and there's this `luigi_namespace(__module__)` part in that file as well.

## Proposed solution

I'll rather just implement this. I'll try to do it today. :)

Tarrasch:
Please see #1965 @riga, it supports both the case you originally wanted fixed and the case I described above. Please review it. :)

Tarrasch:
[...]

PR #1953, opened 2016-12-05
Fix task_namespace inheritance.

riga:
This PR adresses #1950 and fixes the inheritance of `task_namespace`'s.

### Description

This is just a simple change that implements exactly what is proposed in #1950.

Tested with `nosetests test/namespace_test.py`.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -125,7 +125,7 @@ class Task(object):
 
     ``Task.task_namespace``
       optional string which is prepended to the task name for the sake of
-      scheduling. If it isn't overridden in a Task, whatever was last declared
+      scheduling. If it isn't overridden in or inherited by a Task, whatever was last declared
       using `luigi.namespace` will be used.
     """
 
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -52,9 +52,9 @@ def namespace(namespace=None):
     Call to set namespace of tasks declared after the call.
 
     If called without arguments or with ``None`` as the namespace, the namespace
-    is reset, which is recommended to do at the end of any file where the
-    namespace is set to avoid unintentionally setting namespace on tasks outside
-    of the scope of the current file.
+    is reset to :py:attr:`Register._UNSET_NAMESPACE`., which is recommended to do
+    at the end of any file where the namespace is set to avoid unintentionally
+    setting namespace on tasks outside of the scope of the current file.
 
     The namespace of a Task can also be changed by specifying the property
     ``task_namespace``. This solution has the advantage that the namespace
@@ -65,6 +65,9 @@ def namespace(namespace=None):
         class Task2(luigi.Task):
             task_namespace = 'namespace2'
     """
+    if namespace is None:
+        namespace = Register._UNSET_NAMESPACE
+
     Register._default_namespace = namespace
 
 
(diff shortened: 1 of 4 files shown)


luigi/task_register.py
```diff
@@ -62,12 +62,14 @@ def __new__(metacls, classname, bases, classdict):
...
-        Set the task namespace to whatever the currently declared namespace is.
+        When the set or inherited namespace evaluates to ``None``, set the task namespace to
+        whatever the currently declared namespace is.
         """
-        if "task_namespace" not in classdict:
-            classdict["task_namespace"] = metacls._default_namespace
-
         cls = super(Register, metacls).__new__(metacls, classname, bases, classdict)
+
+        if getattr(cls, "task_namespace", None) is None:
```
erikbern: instead of a `getattr` call here, can you just set `task_namespace = None` on the base class? thanks

erikbern: yeah something like that

riga:
Ok, that change seems to break the code because the base Task does not have a `task_namespace` set.

Tarrasch:
I prefer option b. While you're at this. Can you also let the default value not be `None` or `''`. Can we have a magic `_UNSET_NAMESPACE = object()` as default? Eventually I think of making the default unset namespace to be set to the module path.

riga:
> `_UNSET_NAMESPACE = object()`

Yep, will do.

> Eventually I think of making the default unset namespace to be set to the module path.

If you're planning to use `cls.__module__` for this, I think this could lead to ambiguities depending on the way the module that contains task classes is imported - absolute vs. explicit relative vs. implicit relative (in 2.7).

riga:
In addition, I looks like `task_family` and `task_name` are used interchangeable. Maybe this should be cleaned up in another PR.

test/namespace_test.py
```diff
@@ -32,11 +34,12 @@ class FooSubclass(Foo):
 class TestNamespacing(unittest.TestCase):
 
     def test_vanilla(self):
-        self.assertEqual(Foo.task_namespace, None)
+        print(Foo.task_namespace)
```
Tarrasch: You probably forgot to remove this. :)

luigi/task.py
```diff
@@ -52,9 +52,9 @@ def namespace(namespace=None):
     Call to set namespace of tasks declared after the call.
 
     If called without arguments or with ``None`` as the namespace, the namespace
-    is reset, which is recommended to do at the end of any file where the
-    namespace is set to avoid unintentionally setting namespace on tasks outside
-    of the scope of the current file.
+    is reset to :py:attr:`Register._UNSET_NAMESPACE`., which is recommended to do
```
Tarrasch: Actually can we avoid mentioning `Register._UNSET_NAMESPACE`. I thought of it as an implementation detail of the simpler concept of "being unset". 

Perhaps the docs become more natural if we remove the  "If called without arguments or with ..." part and just say "you have to call this function without arguments at the end of any file it has been used. That is to ensure blah blah". Something like that. Just a suggestion.

riga:
> Hmm, maybe we should hold it off from default behavior as long as we have 2.7 support. But do you find any issues having it configurable?

Nope. In fact, that would be very nice.

Tarrasch:
I reworded the commit to what I found sensible. Thank you very much for this well needed patch.

By the way I just noticed there's no way currently (using public API) to "go back" to `_UNSET_NAMESPACE`. But let's keep it like this for now. I suspect it would encourage mis-usage more than good usage. I think you you've meant for a task to be "inherited" you should not set a namespace for it (as namespaces should identify parts of business rather than technology, generally)

PR #1954, opened 2016-12-06
Fix indention issue in deps_tree

bgrommes:
## Description
The last node at each level in the tree representation was indented more than the previous sibling nodes.  I fixed this by changing the level indentation strings to different characters and ensured that all nodes at the same level lined up.

## Motivation and Context
In using this tool I mistakenly thought the last task was a child of the previous task when in fact it was a sibling.  This visual cue led me to find this small issue.

I tested the change against the complex example task 'examples.Foo' in the examples.foo_complex module.

### DIFF
--- a/luigi/tools/deps_tree.py
+++ b/luigi/tools/deps_tree.py
@@ -8,19 +8,19 @@ Use this to visualize the execution plan in the terminal.
 
     $ luigi-deps-tree --module foo_complex examples.Foo
     ...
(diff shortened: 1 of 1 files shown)


Tarrasch:
Awesome! If you want to you can also add a [testcase](https://github.com/spotify/luigi/blob/6f0d175c733da94c202634aa02bfe5efab6fa6cf/test/cmdline_test.py#L305-L313)

PR #1955, opened 2016-12-06
HTCondorJobTask (submission to an HTCondor batch system)

kreczko:
## Description
Implementing the `HTCondorJobTask` to allow for submission to the [HTCondor](https://research.cs.wisc.edu/htcondor/) batch system.

## Motivation and Context
In the High-Energy Particle physics (HEP) community, HTCondor is a very popular batch system.
This PR is the first implementation of HTCondor submission for local batch systems.

Furthermore, it opens the possibility to submit workflows across data centres (e.g. [Open Science Grid (OSG)](https://www.opensciencegrid.org/) infrastructure).
Given the correct HTCondor configuration (e.g. JobRouter), this PR could be also used to submit jobs to the [Worldwide LHC Computing Grid (WLCG)](http://wlcg.web.cern.ch/).
For both the handling of X509 certificates needs to be added, which will be done in a follow-up PR.

I have included unit tests (based on SGE tests) and ran my jobs on a fresh install on my DEV machine (all-in-one install & configuration).

### DIFF
--- a/luigi/contrib/htcondor.py
+++ b/luigi/contrib/htcondor.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from __future__ import absolute_import
 import shutil
 import stat
 import glob
@@ -26,8 +27,8 @@ Adapted by Luke Kreczko (@kreczko) from
 by Jake Feala (@jfeala)
 
 HTCondor (http://research.cs.wisc.edu/htcondor/) is a job scheduler used to allocate compute resources on a
-shared cluster. Jobs are submitted using the ``condor_submit`` command and monitored
-using ``condor_q``. To get started, install luigi on all nodes.
+shared cluster. Jobs are submitted and monitored using the HTCondor python bindings.
+To get started, install luigi on all nodes.
 
 To run luigi workflows on an HTCondor cluster, subclass
 :class:`luigi.contrib.htcondor.HTCondorJobTask` as you would any :class:`luigi.Task`,
@@ -35,7 +36,7 @@ but override the ``work()`` method, instead of ``run()``, to define the job
 code. Then, run your Luigi workflow from the master node, assigning > 1
 ``workers`` in order to distribute the tasks in parallel across the cluster.
 
-The following is an example usage (and can also be found in ``sge_tests.py``)
+The following is an example usage (and can also be found in ``htcondor_test.py``)
 
 .. code-block:: python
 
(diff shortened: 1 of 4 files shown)


examples/hello_condor.py
```diff
@@ -0,0 +1,23 @@
...
+Requires HTCondor(https://research.cs.wisc.edu/htcondor) to be
+installed & configured on the machine.
+
+You can run this example like this:
+
+    .. code:: console
+
+            $ luigi --module examples.hello_condor examples.HelloCondorTask --local-scheduler
+
+If that does not work, see :ref:`CommandLine`.
```
Tarrasch: Very awesome comment! :)

But can you change the name to `hello_htcondor`?

test/contrib/htcondor_test.py
```diff
@@ -0,0 +1,116 @@
...
+    def tearDown(self):
+        for fpath in glob(os.path.join(OUTPUT_DIR, 'testfile_*')):
+            try:
+                print('removing')
+                os.remove(fpath)
+            except OSError:
+                pass
+
+
+if __name__ == '__main__':
```
Tarrasch: Can this be removed? You should run tests with tox. :)

examples/hello_condor.py
kreczko: Apologies, the 'HT' is silent ;) (condor DEV joke as it used to be just 'condor' but had to be renamed following a name clash).
Fixed now.

test/contrib/htcondor_test.py
kreczko: Of course. Old habits never die :).

kreczko:
@Tarrasch : thanks for the comments, fixed the things you pointed out.
As for the condor details, maybe @bbockelm might want to have a look :) (or recommend someone).

Tarrasch:
By the way, do you have any clue for the failing test? Also, anyone from your organization can review it too.

kreczko:
Regarding the failed tests: fixed some of the flake8  warnings and the other test seems to have timed out. Seems to be all clear now.

bbockelm:
@kreczko - Why did you utilize the CLI interface instead of the python bindings?  One major motivator of the python bindings was to avoid having to screen-scrape the outputs of things like `condor_q` as you did in this PR.

kreczko:
@bbockelm That was my original plan, but I could not get the bindings installed on Ubuntu 16.04 (see email on htcondor-users mailing list).
Also, most of our users use python >=2.7 on ScientificLinux6 which is not the default python distribution (hence no condor python bindings out of the box).
Ideas are welcome.

bbockelm:
If you don't mind helping a bit with what will be the world's weirdest `setup.py`, I think we should be able to get the native bindings going.

kreczko:
OK, that sound like a plan.
This PR should be hold off until then.

kreczko:
Apologies for the delays, this is a 0 hour project :).

After sorting out the htcondor binding installation (now available via conda*), I've changed the implementation to use the htcondor python modules.
As a result some of the code has been removed.

@bbockelm could you please have a look again?

\*
```bash
# default boost does not have boost_python, conda-forge does
wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O miniconda.sh
bash miniconda.sh -b -p <path to conda installation>
export PATH=<path to conda installation>/bin:$PATH
conda create -n condor python=2.7
source activate condor
conda install boost -c conda-forge
# match htcondor version to what is installed on the system
conda install htcondor-python=8.4.3 -c kreczko
```

luigi/contrib/htcondor.py
```diff
@@ -0,0 +1,370 @@
...
+    def _fetch_task_failures(self):
+        if not os.path.exists(self.errfile):
+            logger.info('No error file')
+            return []
+        with open(self.errfile, "r") as f:
+            errors = f.readlines()
+        return errors
+
+    def _init_local(self):
+        # Set up temp folder (trim to max filename length)
```
bbockelm: Why this strange approach to creating temporary directories?  Why not `mkdtemp` instead?

luigi/contrib/htcondor.py
```diff
@@ -0,0 +1,370 @@
...
+            'Err': self.errfile,
+            'UserLog': os.path.join(self.tmp_dir, 'job.log'),
+            'request_cpus': self.n_cpu,
+            'MemoryUsage': self.memory,
+        }
+        job_params.update(self.additional_hooks)
+        # build job description (mostly for debugging)
+        job_desc = _build_job_description(job_params)
+        logger.debug('Submitting htcondor job description: \n' + str(job_desc))
+        from htcondor import Schedd
```
bbockelm: Again, I'm not sure function-level imports are idiomatic.

luigi/contrib/htcondor.py
```diff
@@ -0,0 +1,370 @@
...
+fi
+
+"""
+
+
+def _build_job_description(job_params):
+    from classad import ClassAd
+    submit_params = DEFAULT_JOB_PARAMETERS.copy()
+    submit_params.update(job_params)
+    return ClassAd(submit_params)
```
bbockelm: I don't think you need to explicitly cast to `ClassAd` here; a dictionary should be fine on recent HTCondor versions.

luigi/contrib/htcondor.py
```diff
@@ -0,0 +1,370 @@
...
+        logger.debug(
+            "Submitted job to condor_submit with ID:\n{0}".format(self.job_id))
+
+        self._track_job()
+
+        # Now delete the temporaries, if they're there.
+        # unless transfer_output files is on.
+        if os.path.exists(self.tmp_dir) and not self.dont_remove_tmp_dir:
+            logger.info('Removing temporary directory %s' % self.tmp_dir)
+            subprocess.call(["rm", "-rf", self.tmp_dir])
```
bbockelm: Isn't there a more pythonic way to do this?  Since we know what files we created, couldn't we just delete those?

luigi/contrib/htcondor.py
```diff
@@ -0,0 +1,370 @@
...
+        known_statuses = [
+            'Unexpanded', 'Idle', 'Running', 'Removed', 'Completed', 'Held',
+            'Error', '<', '>'
+        ]
+        schedd = Schedd()
+        contraint = 'JobId =?= {0}'.format(self.job_id)
+        while True:
+            # Sleep for a little bit
+            time.sleep(self.poll_time)
+            job_ad = schedd.query(contraint)
```
bbockelm: While this is an OK approach, why not use the job log instead?  See http://htcondor-python.readthedocs.io/en/latest/scalable_job_tracking.html#event-based-tracking

That would avoid the fallback to calling `history` and having to do the `time.sleep` call.

bbockelm: Also, use `xquery` instead - it'll return an iterator that emits the resulting ads..

test/contrib/htcondor_test.py
```diff
@@ -0,0 +1,75 @@
...
+import logging
+import luigi
+from luigi.contrib.htcondor import HTCondorJobTask
+
+OUTPUT_DIR = '/tmp'
+
+logger = logging.getLogger('luigi-interface')
+
+
+def on_sge_master():
```
bbockelm: Copy/paste issue?

kreczko:
I think it would be better to let this one incubate inside one of our current projects first.
I will make another PR once this is more fleshed out and better tested.

PR #1959, opened 2016-12-09
Remove the long-deprecated BooleanParameter

Tarrasch:
## Description

That parameter has been deprecated since way over one year. I believe also before 2.0.

The correct  parameter to use is `BoolParameter`.

### DIFF
--- a/luigi/__init__.py
+++ b/luigi/__init__.py
@@ -34,7 +34,7 @@ from luigi.parameter import (
     Parameter,
     DateParameter, MonthParameter, YearParameter, DateHourParameter, DateMinuteParameter, DateSecondParameter,
     DateIntervalParameter, TimeDeltaParameter,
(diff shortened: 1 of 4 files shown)


mfcabrera:
This will break so much code. I did not even know it was long deprecated ;) - but I guess is an easy-fix for existing code-base.

Tarrasch:
@mfcabrera, sure, but then you've ignored `BooleanParameter is deprecated, use BoolParameter instead` being printed out to your console for over a year. So my sympathy is not as high. But actually it's really good for me to know that people still use the deprecated version. I wish maintainers had more time to do deprecations properly, but that takes too much time.

sbliven:
FYI, `BoolParameter` was introduced in 1.0.21. `BooleanParameter` was deprecated in 2.0.0 and removed in 2.5.0.

PR #1960, opened 2016-12-13
emails: Increase logging if email is misconfigured

Tarrasch:

Yes. I got more verbose output when testing with `luigi TestNotificationsTask --local-scheduler --email-force-send`

I got something like this:

```
ERROR: Not able to connect to smtp server: timed out
```

Instead of a mere `ERROR: Not able to connect to smtp server`.


daveFNbuck:
I think I added the same patch locally while debugging a few times. Thanks for making it a PR!

PR #1962, opened 2016-12-15
Add redshift copy support for role-based credential string

dlstadther:
Redshift copy supports key-based (with optional session token) or role-based credentials. This PR adds support for role-based credentials, making them the default choice if both role- and key-based are provided (since they are more restrictive and thus secure), and refactors implementation of optional session token for the key-based credential string.

## Description

* Added two new properties `aws_account_id` and `aws_arn_role_name`  
* Changed `aws_access_key_id` and `aws_secret_access_key` from `@abc.abstractproperty` to `@property`
    * Added possible `NotImplementedError` when no pair of credentials are provided to account for the lack of `@abc.abstractproperty` pair of credentials  
* Defaults to role-based creds if both pairs are provided. 
* Added test to ensure that `NotImplementedError` raises if no pair of creds are provided.  

## Motivation and Context

Transitioning to the use of role-based cred string for redshift copy. Maintain key-based creds. Refactor key-based session token.

Single unit test for ensuring exception is raised if no pair of creds are provided.

Actual success of use has been verified for the 4 (most likely) cases:
* role-based and key-based are provided (role is used)  
* key-based only  
* role-based only  
* neither (exception is raised)

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -77,17 +77,31 @@ class S3CopyToTable(rdbms.CopyToTable):
         """
         return None
 
-    @abc.abstractproperty
+    @property
     def aws_access_key_id(self):
         """
         Override to return the key id.
         """
         return None
 
-    @abc.abstractproperty
+    @property
     def aws_secret_access_key(self):
         """
         Override to return the secret access key.
+        """
+        return None
+
+    @property
+    def aws_account_id(self):
+        """
+        Override to return the account id.
+        """
+        return None
+
+    @property
+    def aws_arn_role_name(self):
+        """
+        Override to return the arn role name.
         """
         return None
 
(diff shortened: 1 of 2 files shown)


dlstadther:
These commits can and should be squashed prior to merge.

dlstadther:
Review request: @fordb @rantav @steenzout @ddaniels888 @moandcompany 

Thanks

dlstadther:
@Tarrasch I'm not sure what to do here. No one with Luigi Redshift experience seems to want to review. Suggestions?

dlstadther:
I'm a bit of a one-man band. So i'll wait a bit longer before merging. Thanks!

dlstadther:
It's now been three week with no reviews. I'll be merging now!

PR #1963, opened 2016-12-16
externalize: Add a testcase similar to docs

Tarrasch:
I was debugging stuff and wrote this test-case to validate my hypothesizes. Afaict there's no bug. But it's always nice to have another test case. :)


erikbern:
Lgtm

PR #1965, opened 2016-12-20
Fix luigi namespaces, again

Tarrasch:
## Description

Refactor meta magic between Task and Register in luigi. Fixes a few bugs.

## Motivation and Context

#1953 introduced a behavior change that at least broke my code. This patch not only fixes that bug. But refactors the core of luigi tasks to contain less magic.

I added tests. Also I plan to verify that whatever broke for me that made me initiate this patch now works. I'll report back on this soon.


Tarrasch:
@themalkolm, since you seem to understand metaclasses and stuff like that. Do you have any opinion on this PR?

luigi/task_register.py
```diff
@@ -16,6 +16,13 @@
 #
 """
 Define the centralized register of all :class:`~luigi.task.Task` classes.
+
+**DEPRECATION WARNING**
+-----------------------
+
+Since after Luigi 2.4.0, everything in this module, including the exception
+types are now **deprecated** in the sense that it's not public API anymore.
+This is an internal module to luigi.
```
Tarrasch: Actually I think this is too strict. I think too many rely on this. Though some things could be made `_hidden` to make the API more sensible.

doc/tasks.rst
```diff
@@ -298,6 +298,35 @@ if there is a task A with priority 1000 but still with unmet dependencies and
 a task B with priority 1 without any pending dependencies,
 task B will be picked first.
 
+.. _Task.namespaces_famlies_and_ids:
+
+Namespaces, families and ids
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In order to avoid name clashes and to be able to have an identifier for tasks.
```
riga: Typo?
```
... an identifier for tasks, luigi introduces ...
```

themalkolm:
@Tarrasch sorry, too luigi specific :D

doc/tasks.rst
Tarrasch: I changed the `.` to `,`, that's what you meant right? :)

riga:
To me, the changes look transparent and unintrusive. I just pulled the code, tested our tasks, and everything runs as expected.

Concerning the `task_register` deprecation: I also think this might me too strict. This might not be a standard use case, but we use a derived task register meta class at one or two places where we assign a dynamic task namespace depending on e.g. environment variables. So the (custom) task register is a real feature for us and not only an internal module.

Tarrasch:
@riga, wow, that's a very dynamic use case. I though I'll make a PR later hiding things I think nobody should rely on. And hopefully nobody is.

As for this PR, I take it as you approve it?

riga:
@Tarrasch Yep, looks perfectly good to me.

PR #1967, opened 2016-12-21
Make Register a normal metaclass, not ABCMeta

Tarrasch:
I couldn't find any support (in nearby commented code) for why we needed
the abc-stuff.

https://blog.ionelmc.ro/2015/02/09/understanding-python-metaclasses

I'm quite skeptical to this myself. I will try it on my setting and I suspect it will break. Don't merge until I self-approve. ^^


themalkolm:
It will break any user code (we have some) that depends on the fact that Task is subclassed from ABCMeta. Not sure what you get in return.

Tarrasch:
@themalkolm yes. I'm thinking of not merging this. Thanks for your opinion.

However, we get quite some things in return. (1) Less complex luigi and (2) no random arbitrary choice that Tasks may if they so desire ABCMeta, but all other metaclases are not allowed.

themalkolm:
I think it is released in this state for way too long and way too late to fix.

Issue #1966, opened 2016-12-20
TimeDeltaParameter can't be passed to dynamic dependencies


kierkegaard13:
I have a task with a TimeDeltaParameter, let's call it TaskA. I can run TaskA from the command line like so
```
luigi --module project.modules TaskA --buffer '1 minute'
```
TaskA has a dynamic dependency, TaskB, which I yield from the run function like so
```
def run():
    yield TaskB(self.buffer)
```
This results in the error
```
luigi.parameter.ParameterException: Invalid time delta - could not parse 0:01:00
```
It seems like passing TimeDeltaParameters to dynamic dependencies results in this issue, because they are passed as strings and then can't be parsed. Even more interesting, is when requiring TaskB from the requires function, it works just fine. Digging deeper, it looks like the parse function which throws the above error, isn't called when running dependencies from the requires function. Does anyone know how to get around this? I'm also curious as to if there's any way to set a default TimeDeltaParameter of 0 days as that seems to throw the same error as above.

Tarrasch:
I think TimeDeltaParameters were created a long time ago, probably before good coding practices were applied. I think this can be fixed by fixing the `serialize`/`parse` pair. Please take a look @kierkegaard13! :)

dlstadther:
resolved by #1968

PR #1968, opened 2016-12-21
Allow 0 time timedeltas, fix serialize to allow passing timedelta par…

kierkegaard13:
…ams to subtasks

This PR primarily fixes the serialize function for TimeDeltaParameters to allow passing the parameter to a subtask that is required or yielded in the run function. It also allows creating a parameter with 0 time, which helps with setting defaults.

## Description
This is to address problems raised in issue #1966 

## Motivation and Context
This is to address problems raised in issue #1966

I have included unit tests and tested it working on another project

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -647,13 +647,14 @@ class TimeDeltaParameter(Parameter):
     def _apply_regex(self, regex, input):
         import re
         re_match = re.match(regex, input)
-        if re_match:
+        if re_match and any(re_match.groups()):
             kwargs = {}
             has_val = False
             for k, v in six.iteritems(re_match.groupdict(default="0")):
                 val = int(v)
-                has_val = has_val or val > -1
-                kwargs[k] = val
+                if val > -1:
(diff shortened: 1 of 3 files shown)


Tarrasch:
Great job! But I see no tests for the serialization that you fixed. :)

kierkegaard13:
@Tarrasch my bad. I just added them. Hypothesis didn't have a timedelta strategy that I could see, so I wrote some manual test cases if that's okay. Let me know if there are any others you'd like to see.

Tarrasch:
@kierkegaard13. Looks good. But could you just move out the test to the same `parameter_tests.py` file? I think there are already tests (without `hypothesis`) related to serialization in there.

kierkegaard13:
@Tarrasch done.

kierkegaard13:
@Tarrasch think the build stalled. Are you able to restart it, or should I just close and open the PR?

Tarrasch:
Thanks @kierkegaard13! :)

PR #1970, opened 2016-12-23
fix status bug and add cluster parameter to ecs.py

MarkusShepherd:
* the current implementation only allows the default cluster, but often it is necessary to use a specific cluster, so this commit adds cluster as a parameter
* there was a bug in `_track_tasks()` by referencing the undefined variable `status` – this is fixed now


adamchainz:
Afraid I can't help, I only fixed the docs on this repo.

MarkusShepherd:
@outlierbio

PR #1973, opened 2016-12-28
tests: Remove `__name__ == '__main__'` guards

Tarrasch:
This is exactly like https://github.com/spotify/luigi/pull/1922, only
that I forgot to search for '__main__' with single qoutes. This patch
completes that PR.


toidi:
Since tests are run by nose, PR looks good.

Tarrasch:
Thanks for quick review @toidi. :)

PR #1975, opened 2016-12-28
externalize: Dont register the copied tasks

Tarrasch:
This fixes an implementation detail, so that luigi.task.externalize
actually does what it says it does in the documentation.  Hence I
hope pretty uncontroversial. This makes it so that calling externalize
don't register a new task. A pretty serious bug.

Now, there's still a caveat, as we're mostly hiding that we do side
affects, if anyone would actually inherit from what is returned by
externalize.  That subtask wouldn't be registered either.

This code is quite easy to get wrong somewhere. Luckily luigi is
decently tested. To make me even more confident I also added a couple of
tests. Including two tests that didn't pass before.

Further. I will try this in production too. Ensuring everything is actually working this time. :)

### DIFF
--- a/luigi/task_register.py
+++ b/luigi/task_register.py
@@ -19,7 +19,6 @@ Define the centralized register of all :class:`~luigi.task.Task` classes.
 """
 
 import abc
-from collections import OrderedDict
 
 from luigi import six
 import logging
@@ -129,23 +128,24 @@ class Register(abc.ABCMeta):
     def _get_reg(cls):
         """Return all of the registered classes.
 
-        :return:  an ``collections.OrderedDict`` of task_family -> class
+        :return:  an ``dict`` of task_family -> class
         """
         # We have to do this on-demand in case task names have changed later
-        # We return this in a topologically sorted list of inheritance: this is useful in some cases (#822)
-        reg = OrderedDict()
-        for cls in cls._reg:
-            name = cls.task_family
+        reg = dict()
+        for task_cls in cls._reg:
+            if not task_cls._visible_in_registry:
+                continue
 
-            if name in reg and reg[name] != cls and \
-                    reg[name] != cls.AMBIGUOUS_CLASS and \
-                    not issubclass(cls, reg[name]):
+            name = task_cls.get_task_family()
+            if name in reg and \
+                    (reg[name] == Register.AMBIGUOUS_CLASS or  # Check so issubclass doesn't crash
+                     not issubclass(task_cls, reg[name])):
                 # Registering two different classes - this means we can't instantiate them by name
                 # The only exception is if one class is a subclass of the other. In that case, we
                 # instantiate the most-derived class (this fixes some issues with decorator wrappers).
-                reg[name] = cls.AMBIGUOUS_CLASS
+                reg[name] = Register.AMBIGUOUS_CLASS
             else:
-                reg[name] = cls
+                reg[name] = task_cls
 
         return reg
 
--- a/test/task_test.py
+++ b/test/task_test.py
@@ -206,6 +206,26 @@ class ExternalizeTaskTest(LuigiTestCase):
         self.assertIsNotNone(MyTask.run)  # Check immutability
         self.assertIsNotNone(MyTask().run)  # Check immutability
 
+    def test_externalize_doesnt_affect_the_registry(self):
+        class MyTask(luigi.Task):
+            pass
+        reg_orig = luigi.task_register.Register._get_reg()
+        luigi.task.externalize(MyTask)
+        reg_afterwards = luigi.task_register.Register._get_reg()
+        self.assertEqual(reg_orig, reg_afterwards)
+
+    def test_can_uniquely_command_line_parse(self):
+        class MyTask(luigi.Task):
+            pass
+        # This first check is just an assumption rather than assertion
+        self.assertTrue(self.run_locally(['MyTask']))
+        luigi.task.externalize(MyTask)
+        # Now we check we don't encounter "ambiguous task" issues
+        self.assertTrue(self.run_locally(['MyTask']))
+        # We do this once again, is there previously was a bug like this.
+        luigi.task.externalize(MyTask)
+        self.assertTrue(self.run_locally(['MyTask']))
+
 
 class TaskNamespaceTest(LuigiTestCase):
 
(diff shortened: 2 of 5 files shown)


test/task_test.py
```diff
@@ -206,6 +206,24 @@ def run(self):
         self.assertIsNotNone(MyTask.run)  # Check immutability
         self.assertIsNotNone(MyTask().run)  # Check immutability
 
+    def test_dont_affects_registry(self):
```
daveFNbuck: `test_externalize_doesnt_affect_the_registry` would be more clear

test/task_test.py
```diff
@@ -206,6 +206,24 @@ def run(self):
...
+        reg_afterwards = luigi.task_register.Register._get_reg()
+        self.assertEqual(reg_orig, reg_afterwards)
+
+    def test_can_uniquely_command_line_parse(self):
+        class MyTask(luigi.Task):
+            pass
+        # This first check is just an assumption rather than assertion
+        self.assertTrue(self.run_locally(['MyTask']))
+        luigi.task.externalize(MyTask)
+        self.assertTrue(self.run_locally(['MyTask']))
```
daveFNbuck: So the idea here is that MyTask is still not external?

test/task_test.py
```diff
@@ -206,6 +206,24 @@ def run(self):
         self.assertIsNotNone(MyTask.run)  # Check immutability
         self.assertIsNotNone(MyTask().run)  # Check immutability
 
+    def test_dont_affects_registry(self):
+        class MyTask(luigi.Task):
```
daveFNbuck: why not just externalize luigi.Task in these tests?

luigi/task_register.py
```diff
@@ -135,17 +135,18 @@ def _get_reg(cls):
...
-                # The only exception is if one class is a subclass of the other. In that case, we
-                # instantiate the most-derived class (this fixes some issues with decorator wrappers).
-                reg[name] = cls.AMBIGUOUS_CLASS
-            else:
-                reg[name] = cls
+            if cls._visible_in_registry:
+                name = cls.get_task_family()
+
+                if name in reg and reg[name] != cls and \
+                        reg[name] != cls.AMBIGUOUS_CLASS and \
```
daveFNbuck: While we're looking at this code, it seems that if we register the same name three times with different classes, the second registration will turn it into cls.AMBIGUOUS_CLASS and the third registration will hit the else clause here and register itself. That doesn't seem right. Maybe this line should go?

luigi/task_register.py
```diff
@@ -135,17 +135,18 @@ def _get_reg(cls):
...
+                        not issubclass(cls, reg[name]):
```
daveFNbuck: Doesn't this cover the `reg[name] != cls` check?

luigi/task_register.py
```diff
@@ -135,17 +135,18 @@ def _get_reg(cls):
...
-            if name in reg and reg[name] != cls and \
-                    reg[name] != cls.AMBIGUOUS_CLASS and \
-                    not issubclass(cls, reg[name]):
-                # Registering two different classes - this means we can't instantiate them by name
-                # The only exception is if one class is a subclass of the other. In that case, we
-                # instantiate the most-derived class (this fixes some issues with decorator wrappers).
-                reg[name] = cls.AMBIGUOUS_CLASS
-            else:
-                reg[name] = cls
+            if cls._visible_in_registry:
```
daveFNbuck: Why didn't you just add this to the existing conditional? It would simplify the diff a bit and avoid extra nesting.

test/util_test.py
```diff
@@ -50,4 +50,31 @@ class ChildTask(luigi.Task):
...
+            def complete(self):
+                return self.class_variable == 'actuallyset'
+
+        @requires(ParentTask)
+        class ChildTask(RunOnceTask):
+            pass
+
+        return ParentTask
+
+    def test_requries_has_effect_run_child(self):
```
daveFNbuck: requires misspelled

test/util_test.py
```diff
@@ -50,4 +50,31 @@ class ChildTask(luigi.Task):
...
+        ParentTask = self._setup_parent_and_child()
+        self.assertTrue(self.run_locally_split('ChildTask --my-parameter actuallyset'))
+        self.assertEqual(ParentTask.class_variable, 'actuallyset')
+
+    def test_requries_has_effect_run_parent(self):
```
daveFNbuck: requires again

test/util_test.py
```diff
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from helpers import LuigiTestCase
```
daveFNbuck: It seems like you accidentally pulled some unrelated changes to util_test into this.

test/task_test.py
Tarrasch: Fixed.

luigi/task_register.py
Tarrasch: Great catch. I added a test case. for this. You were right, the current behavior is broken (see soon up and coming update to PR) :)

test/util_test.py
Tarrasch: fixed

test/util_test.py
Tarrasch: fixed

test/util_test.py
Tarrasch: Actually not completely unrelated. Much of the machinery here is relying on the Registry that got changed here.

I'll amend the commit message saying I intentionally added these tests for this change set.

luigi/task_register.py
```diff
@@ -129,23 +128,24 @@ def task_family(cls):
     def _get_reg(cls):
         """Return all of the registered classes.
 
-        :return:  an ``collections.OrderedDict`` of task_family -> class
+        :return:  an ``dict`` of task_family -> class
         """
         # We have to do this on-demand in case task names have changed later
-        # We return this in a topologically sorted list of inheritance: this is useful in some cases (#822)
```
daveFNbuck: Did the use case in #822 go away?

Tarrasch: Yes, since #1182, the Parameter class became (more) immutable, and a lots of these random corner-cases bugs went away. There were also test cases added with the fix for #822, and I assume green Travis means they still pass. :)

PR #1977, opened 2017-01-03
server_test.py: Fix crash in test_save_state

Tarrasch:
For whatever reasons its not running on travis but crashing when locally run the tests for luigi.


Tarrasch:
Really not much to discuss about this I assume. So I'm self-approving. :)

PR #1978, opened 2017-01-03
target_test.py: Use clean paths to avoid failures

Tarrasch:
Currently some tests will fail if you run them twice, because there's no
proper cleanup done. But that machinery already exists, we just need to
enable it.


Tarrasch:
Let me know if anyone knows a more elegant way to do this. Otherwise I'll just merge this :)

PR #1979, opened 2017-01-03
Allow a dictionary to be passed that contains the logging configuration.

jsivak:
## Description
Allow a predefined/preconfigured logging dictionary to be processed by luigi.

## Motivation and Context
I setup my logging via a YAML based configuration file and it was more convenient to be able to pass the dictionary to luigi rather than maintain an external .ini based logging config.

I have been running this in my production code. (Using Python 2.7.x)

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -56,6 +56,8 @@ def setup_interface_logging(conf_file=''):
         stream_handler.setFormatter(formatter)
 
         logger.addHandler(stream_handler)
+    elif isinstance(conf_file, dict):
+        logging.config.dictConfig(conf_file)
     else:
         logging.config.fileConfig(conf_file, disable_existing_loggers=False)
 


Tarrasch:
I mean, add docs to [this page](http://luigi.readthedocs.io/en/stable/configuration.html). Shortly describing how luigi in fact is just calling out to the existing configuration infrastructure in Python, and a small example YAML based log file.

jsivak:
I'll give it a shot an let you know when they're ready for review.

jsivak:
I've added a docstring and updated the documentation. Please review it when you have time and let me know if it looks good.

luigi/interface.py
```diff
@@ -41,6 +41,22 @@
...
+    '''
+    Luigi logging can be configured in the following ways:
+
+        * Pass in the file name of a .ini style logging configuration.
+
+          The ``conf_file`` is passed to logging.config.fileConfig()
+
+        * Pass in a dictionary containing the logging configuation format as specified
+          in the `Python 3 documentation
+          <https://docs.python.org/3/library/logging.config.html#logging-config-dictschema>`_.
```
dlstadther: This doc exists for python 2 as well. Feels weird to only call out Python 3. I'd either mention both or neither, explicitly.

jsivak: No worries; I've updated my branch with a link to the Python 2 docs as well.

jsivak:
I don't know how you would provide a dictionary of logging configuration to luigi "directly" (I am doing it with a wrapper script); I'm guessing we'd need to allow a JSON or YAML file that would have the dictionary "string representation" of the logging configuration in it.  I'm open to ideas on how to accomplish this.

Tarrasch:
I think this is much easier than you think. As both Python 2.7 and 3.2 already comes with this support built-in. See [2.7 docs](https://docs.python.org/2.7/howto/logging.html#configuring-logging) for instance.

Here's the except (from [3.6 docs](https://docs.python.org/3.6/howto/logging.html#configuring-logging))

> In Python 3.2, a new means of configuring logging has been introduced, using dictionaries to hold configuration information. This provides a superset of the functionality of the config-file-based approach outlined above, and is the recommended configuration method for new applications and deployments. Because a Python dictionary is used to hold configuration information, and since you can populate that dictionary using different means, you have more options for configuration. For example, you can use a configuration file in JSON format, or, if you have access to YAML processing functionality, a file in YAML format, to populate the configuration dictionary. Or, of course, you can construct the dictionary in Python code, receive it in pickled form over a socket, or use whatever approach makes sense for your application.

I haven't researched this thoroughly. So I might have missed something, but it seems like it's plug and play from the docs.  Please let me know of your thoughts, and thanks for your fast reply! :)

dlstadther:
@jsivak Any update here?

jsivak:
Sorry, been busy; I'll try to look at it this weekend.. based on Tarrasch's comment I was going to investigate providing a way to pass a .yaml file to the logging command line switch (and then convert the .yaml to the dictConfig that the logging system is expecting)..  I think that is what he is looking for.

jsivak:
I'm "back from the dead".. restarted work on this and wanted to ask:
Do you want to support .yaml and .json files for the "--logging-conf-file" option? I think that's the most logical way to get a "dict-like" config from a file.  I don't think we can store a dict-like config in an .ini file, so this would lead to a command line looking something like:
`luigi MyTask --my-param 5 --logging-config-file my_file.yaml`

Am I on the 'right track' on what you'd like to see with supporting a dict-based config file.

luigi/interface.py
```diff
@@ -59,6 +77,8 @@ def setup_interface_logging(conf_file='', level_name='DEBUG'):
         stream_handler.setFormatter(formatter)
 
         logger.addHandler(stream_handler)
+    elif isinstance(conf_file, dict):
```
Tarrasch: If we actually do down this path. The variable should be renamed to something like `path_or_dict`.

Tarrasch:
By the way. The docs are really good! :)

luigi/interface.py
jsivak: I think we need to stay on this path. The reason I have to pass the config dictionary to luigi.setup_interface_logging() is to ensure that all of the subprocesses that luigi creates will have the correct logging configured. If I were to only call `logging.config.dictConfig(their_dict)` in my "driver" script then the logging config is not correctly set for the worker processes.

To re-clarify: I use a python "driver" script that calls luigi.run() to do my work.  If someone is using the luigi command directly or using python -m luigi, then manually calling `logging.config.dictConfig(their_dict)` might work, but I haven't tried or tested that.

I've updated my pull request to be squashed to a single commit on the current master branch and changed the variable name as you request.

Let me know if you have any questions.

Tarrasch:
Did you consider solving this in the other way and getting rid of the `has_run` hack? Was it not possible?

jsivak:
I've never looked into why the "has_run" hack was put into luigi, so no, I've not looked at it any other way.

Tarrasch:
> I've never looked into why the "has_run" hack was put into luigi, so no, I've not looked at it any other way.

I would appreciate if you looked over it as I mentioned here https://github.com/spotify/luigi/pull/1979#pullrequestreview-51650623. This is an open source project and in order for the codebase to remain fun to maintain and hack on, we from time to time need to fix other people's mess. Thankfully it would improve the interface you propose greatly as well. :)

jsivak:
Do you know when the "has_run" hack was introduced into the code and why?  (I can hut thru the commit messages, but if you happen to know..)

jsivak:
Looks like "has_run" was added back in 2013 (commit d276d0ebb78d7e260165a684460b9b5b4e5f3a78) when the setup_interface_logging() was intially added to luigi. Is there some specific reason you want it removed?

Tarrasch:
Well my problem is that before you introduced this PR, a luigi user could simply do `logging.config.dictConfig(their_dict)` and it would work splendid afaict. The only problem is that luigi will later run `setup_interface_logging(..)` and do it's own weird thing. It would be so nice if instead of we making the function `setup_interface_logging(..)` weirder by allowing users to also pass a dictionary (this PR), we just recommend them to do the standard python way `logging.config.dictConfig(their_dict)` and then `setup_interface_logging(..)` would automatically detect that there's no need to do anything as a user has already set up logging themselves.

Would this be possible?

jsivak:
Ah, that makes things more clear to me now. Thanks for that.. I'll see what I can find.

jsivak:
Ah, that makes things more clear to me now. Thanks for that.. I'll see what I can find.

PR #1980, opened 2017-01-03
notifications.py: Handle boto3 errors when sending AWS SES error emails

ehdr:
## Description

Catch `botocore.exceptions.BotoCoreError`s (e.g. `NoCredentialsError`, `NoRegionError`) when trying to send error email notifications using AWS SES.

Problem and solution identified thanks to #1821, which solves a similar issue for SMTP.

## Motivation and Context

When a task has been retried too many times and should be set as `DISABLED`, luigi tries to send an error email (at [scheduler.py:539](https://github.com/spotify/luigi/blob/master/luigi/scheduler.py#L539)), in this case using AWS SES. If boto3 fails to find AWS credentials (e.g. AWS `region`), then it raises an exception (e.g. `NoRegionError`) in the call to `notifications.send_error_email`. Afterwards, luigi complains that

    luigi-interface ERROR    Failed connecting to remote scheduler 'http://localhost:8082'
    ...
    urllib.error.HTTPError: HTTP Error 500: Internal Server Error

and when trying to run the task again, the following is logged in `luigi-server.log`:

    tornado.application[22] ERROR: Uncaught exception POST /api/get_work (::1)

    HTTPServerRequest(protocol='http', host='localhost:8082', method='POST', uri='/api/get_work', version='HTTP/1.1', remote_ip='::1', headers={'Accept-Encoding': 'identity', 'Host': 'localhost:8082', 'Connection': 'close', 'User-Agent': 'Python-urllib/3.4', 'Content-Type': 'application/x-www-form-urlencoded', 'Content-Length': '251'})
      File ".../lib/python3.4/site-packages/tornado/web.py", line 1467, in _execute
      File ".../lib/python3.4/site-packages/luigi/server.py", line 77, in get
      File ".../lib/python3.4/site-packages/luigi/scheduler.py", line 1087, in get_work
        reply = self.count_pending(worker_id)
      File ".../lib/python3.4/site-packages/luigi/scheduler.py", line 956, in count_pending
        more_info = {'task_id': task.id, 'worker': str(other_worker)}
    TypeError: __str__ returned non-string (type NoneType)

    tornado.access[22] ERROR: 500 POST /api/get_work (::1) 1.75ms

This change catches `BotoCoreError`s which prevents this from happening.

Unit test included, and the change solves the issue as observed above.


ehdr:
Unfortunately I couldn't find a way to avoid adding `boto3` as a dependency (here and here). Feedback appreciated!

Also, I guess any other exception raised from `notifications.send_email` would cause similar problems. So maybe a better solution would be to catch all `Exception`s in [`notifications.send_email`](https://github.com/spotify/luigi/blob/master/luigi/notifications.py#L327)?

Tarrasch:
> Unfortunately I couldn't find a way to avoid adding boto3 as a dependency in test. Feedback appreciated!

That's OK I think.

> Also, I guess any other exception raised from notifications.send_email would cause similar problems. So maybe a better solution would be to catch all Exceptions in notifications.send_email?

Interesting. Though at the very moment maybe we don't need to re-factor this. Catching the more specific exception is a clean solution, although easy to get wrong too.

ehdr:
Installing boto3 apparently had an unintended consequence in [ecs_test.py](https://github.com/spotify/luigi/blob/master/test/contrib/ecs_test.py#L39) (which is why CI failed)... I'll have a look at it (hopefully) tonight!

ehdr:
So one option for handling `ecs_test.py` could wrapping the [boto3 stuff](https://github.com/spotify/luigi/blob/master/test/contrib/ecs_test.py#L37-L41) in something like:

    try:
        from luigi.contrib.ecs import ECSTask, _get_task_statuses
        import boto3
        client = boto3.client('ecs')
    except botocore.exceptions.BotoCoreError:
        raise unittest.SkipTest...

However it seems that running this test will modify things in your AWS account, which might not be what you expect if you run the luigi test suite and just happen to have AWS credentials set up.

So I'm tempted to suggest adding something like this to the top of `ecs_test.py` instead:

    if 'LUIGI_ECS_TEST' not in os.environ:
        raise unittest.SkipTest('ecs_test.py will only run if LUIGI_ECS_TEST environment variable is set')

What do you think? @jfeala? Or is there a better pattern for handling tests like these, that depend on external infrastructure?

jfeala:
@ehdr sorry for the delayed response. Good point that you might not want to have the tests talking to your AWS account without asking first. Ok with me if you want to add that to the top of `test_ecs.py` in this PR, if it will take care of your build issue as well.

ehdr:
Thanks @jfeala and no worries! I had almost forgot that I implemented a potential solution a while ago, that might be more in line with how dependencies on other external resources (e.g. hadoop clusters) is handled(?). Just pushed it as #2026.

Also @Tarrasch, what do you think about discussing nose attribute vs. environment variable there, and reconsider this PR once that's settled?

Tarrasch:
@ehdr sounds good to me.

ehdr:
Added a commit that skips `ecs_test` unless `LUIGI_ECS_TEST` environment variable is set, according to discussion in #2026. (also rebased)

Unfortunately, `ImportTest.import_test` still fails, because importing `ecs` raises a `NoRegionError: You must specify a region` exception if boto fails to find AWS credentials.

Issue #1897, opened 2016-10-24
Plugins support


ulzha:
I see a plugin system's primary purpose in providing standardized extension points for operators of Luigi-based scheduling infrastructure.

Some of the primary purposes for plugins:
* Register logging handlers
* Register [event](https://github.com/spotify/luigi/blob/master/luigi/event.py) handlers
* Initialize and clean up ad-hoc resource management
* Register utility Tasks to be available during CLI parsing before the user-provided `--module` has been parsed out and imported? (Unsure, maybe `--module`(s) should be parsed in a separate pass and utilities that the user uses should be explicitly imported in the user code...)

Currently the approach we use for achieving the above is to wrap `luigi` in a proprietary wrapper executable, which is confusing and doesn't compose.

Related issues have been filed previously:
* https://github.com/spotify/luigi/issues/1088

ulzha:
I imagine quite minimal functionality, that is, in luigi.cfg specify modules that contain plugins, and have the CLI import them, using some industry standard plugin API.

I looked through Google results (http://wehart.blogspot.se/2009/01/python-plugin-frameworks.html, https://www.reddit.com/r/Python/comments/1qaepq/very_basic_plugin_system/ ...) about Python plugin systems.
* [plugnplay](https://github.com/daltonmatos/plugnplay) 2 contributors :
* [pluginbase](http://pluginbase.pocoo.org/) looks nice, but few contributions after 2014
* [yapsy](http://yapsy.sourceforge.net/) Last Update: 2015-09-12
* [stevedore](https://github.com/openstack/stevedore) - promising :white_check_mark:

PR #1982, opened 2017-01-09
Add slack notifications module

labianchin:
## Description

This module configures and listen for task event handlers to send slack
notifications. The configuration is done through `luigi.cfg` file.

## Motivation and Context

The goal is to have slack notifications on luigi event transitions: on START, SUCCESS and FAILURE. Using this module and a slack channel for the notifications is very easy to get status and some basic info of the luigi tasks.

It also support some basic configuration and templating. In this way the user can send some task `@property` or attribute in the slack message.

I am using in production in some of the pipelines here in Spotify. The reason for this PR is that this is a generic feature that any other team could benefit from.

- Unit tests covering all the code are included
- Running in production for over a month

### DIFF
--- a/luigi/contrib/slack_notifications.py
+++ b/luigi/contrib/slack_notifications.py
@@ -1,0 +1,180 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2017 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+''' Supports sending slack notifications for different task events.
+
+Usage:
+
+Create an incomming webhook on slack: https://api.slack.com/incoming-webhooks
+
+Setup the slack integration with the following code.
+This can be done in the same module where the luigi tasks are defined.
+
+.. code-block:: python
+
+    from luigi.contrib.slack_notifications import configure_slack
+    configure_slack()
+
+Use a configuration like the following on luigi.cfg
+
+.. code-block:: ini
+
+    [slack]
+    webhook = https://hooks.slack.com/services/something
+    events = ["START", "SUCCESS", "FAILURE"]
+    template = Luigi task {task_id} event {event}: {args} {kwargs}
+
+
+Alternatives:
+
+Another plugin with similar purpouse is luigi-slack (https://github.com/bonzanini/luigi-slack). The major differences between this and luigi-slack is that:
+
+- Inspired by email notifications (notifications.py): simple configuration though luigi.cfg (no need for python configuration)
+- Minimal changes to python code (only call configure_slack()), no need create a new wrapper around luigi.run.
+- This only depends on requests module (does not depend on slackclient).
+- Work both for python 2 and 3
+- Configurable template, per event, that can expose specific task properties
+
+'''
+
+
+import logging
+import luigi
+import json
+
+logger = logging.getLogger(__name__)
+
+try:
+    import requests
+except ImportError:
+    logger.warning("Module %s requires the python package 'requests'." % __name__)
+
+
+class slack(luigi.Config):
+    webhook = luigi.parameter.Parameter(
+        default='',
+        description='Webhook URL to send the notifications')
+    channel = luigi.parameter.Parameter(
+        default='',
+        description='Channel to send slack notifications to')
+    username = luigi.parameter.Parameter(
+        default='',
+        description='Slack username originating the message')
+    icon_emoji = luigi.parameter.Parameter(
+        default='',
+        description='Slack bot icon emoji')
+    events = luigi.parameter.ListParameter(
+        default=['START', 'FAILURE', 'SUCCESS'],
+        description='List of events to be handled as slack notifications')
+    template = luigi.parameter.Parameter(
+        default='Luigi task {task_id} event {event}: {args} {kwargs}',
+        description='Template message for event handling slack notifications')
+    event_templates = luigi.parameter.DictParameter(
+        default={},
+        description='Custom template message per event')
+
+
+class SafeGetAttr(object):
+    """
+    Wraps an object returning a default_value for undefined attributes
+    """
+    def __init__(self, obj, default_value=None):
+        self._obj = obj
+        self._default_value = default_value
+
+    def __getattr__(self, attr):
+        return getattr(getattr(self, '_obj'), attr, getattr(self, '_default_value'))
+
+
+class SlackNotifyHandling():
+    slack_hook_url = None
+
+    def __init__(self, slack_config):
+        self.slack_hook_url = slack_config.webhook
+        self.channel = slack_config.channel
+        self.events = slack_config.events
+        self.template = slack_config.template
+        self.username = slack_config.username
+        self.icon_emoji = slack_config.icon_emoji
+        self.event_templates = slack_config.event_templates
+
+    @property
+    def is_configured(self):
+        """
+        Check if the slack web hook is configured so that slack messages can be sent
+        """
+        return bool(self.slack_hook_url)
+
+    def _make_data(self, text):
+        data = dict(text=text)
+        if self.channel:
+            data['channel'] = self.channel
+        if self.username:
+            data['username'] = self.username
+        if self.icon_emoji:
+            data['icon_emoji'] = self.icon_emoji
+        return json.dumps(data)
+
+    def send_slack_message(self, text):
+        """
+        Send a string as a slack message
+        """
+        r = requests.post(self.slack_hook_url, data=self._make_data(text))
+        if not r.status_code == 200:
+            logger.error('Could not notify slack, got error: %s ' % r.text)
+            return False
+        return True
+
+    def _handler_for_template(self, template, event):
+        def handle(task, *args, **kwargs):
+            msg = template.format(
+                event=event,
+                args=args,
+                kwargs=kwargs,
+                task_id=task.task_id,
+                task=SafeGetAttr(task, '')
+            )
+            self.send_slack_message(msg)
+        return handle
+
+    _is_set_handlers = False
+
+    def set_handlers(self):
+        """
+        Configure all luigi.task.event_handler if slack notifications ar e configured
+        To avoid setting up event handler multiple times, this only runs once
+        """
+        if not self.is_configured or self.__class__._is_set_handlers:
+            return
+        logger.info('Configuring slack notifications event handling...')
+        event_aliases = {
+            'START': luigi.Event.START,
+            'SUCCESS': luigi.Event.SUCCESS,
+            'FAILURE': luigi.Event.FAILURE
+        }
+        for event in self.events:
+            luigi_event = event_aliases.get(event, event)
+            template = self.event_templates.get(event, self.template)
+            luigi.Task.event_handler(luigi_event)(self._handler_for_template(template, event))
+        self.__class__._is_set_handlers = True
+
+
+def configure_slack():
+    """
+    Setup task event handlers to send slack notifications
+    """
+    SlackNotifyHandling(slack()).set_handlers()
--- a/test/contrib/slack_notifications_test.py
+++ b/test/contrib/slack_notifications_test.py
@@ -1,0 +1,175 @@
+
+
+from helpers import (unittest, with_config)
+import mock
+import json
+import luigi
+from luigi.contrib.slack_notifications import SlackNotifyHandling, slack, configure_slack
+
+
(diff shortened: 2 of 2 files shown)


luigi/contrib/slack_notifications.py
```diff
@@ -0,0 +1,148 @@
...
+    [slack]
+    webhook: https://hooks.slack.com/services/...
+    events: ["START", "SUCCESS", "FAILURE"]
+    template: Luigi task {task_id} event {event}: {args} {kwargs}
+
+
+In the python module, setup the slack integration with:
+.. code-block:: python
+    from luigi.contrib.slack_notifications import configure_slack
+    configure_slack()
```
Tarrasch: Cool! Can this be done from anywhere in your python code?

I'm kinda hinting that the wording "`the python module`" is not super clear to me.

luigi/contrib/slack_notifications.py
```diff
@@ -0,0 +1,148 @@
...
+        default=['START', 'FAILURE', 'SUCCESS'],
+        config_path=dict(section='slack', name='slack-events'),
+        description='List of events to be handled as slack notifications')
+    template = luigi.parameter.Parameter(
+        default='Luigi task {task_id} event {event}: {args} {kwargs}',
+        config_path=dict(section='slack', name='slack-template'),
+        description='Template message for event handling slack notifications')
+    event_templates = luigi.parameter.DictParameter(
+        default={},
+        config_path=dict(section='slack', name='slack-event-templates'),
```
Tarrasch: I think you misunderstood the `config_path`, if you just remove all the `config_path` lines, the

    [slack]
    webhook: https://hooks.slack.com/services/...
    events: ["START", "SUCCESS", "FAILURE"]
    template: Luigi task {task_id} event {event}: {args} {kwargs}

part will still work. The `config_path` code will make it so you also support:

    [slack]
    slack_webhook: https://hooks.slack.com/services/...
    slack_events: ["START", "SUCCESS", "FAILURE"]
    slack_template: Luigi task {task_id} event {event}: {args} {kwargs}

Is that intentional?

Tarrasch:
I'm pretty sure @daveFNbuck and friends also already made something like this. There's also https://github.com/bonzanini/luigi-slack already. It would be great if this PR could be made into it's own repository (to decrease maintenance of this megarepo), or you collaborate with the other existing repo, if that's possible.

luigi/contrib/slack_notifications.py
labianchin: It was intentional in the sense I was trying to replicate the `class email` from `notifications.py`. Now I am not sure, should I remove the `config_path`?

luigi/contrib/slack_notifications.py
labianchin: Yes. I am thinking how to improve this wording.

This is where I mostly unsure, and I would like to explain my reasoning. This code needs to execute on luigi initialization or before the tasks run. I really did not find a nice place for this to execute and for now I am recommending to use leave at the end of the python module.

Alternatively, I would like call this at the `interface.py` or `worker.py`, as that does not have any side effect if slack is not configured. But then I think it is too much change.

The goal here is not to create a new `luigi_cmd` wrapping `luigi.run` (as https://github.com/bonzanini/luigi-slack) and also calling `configure_slack()`.

labianchin:
@Tarrasch thanks for taking the time to look at this. I will ask somebody else in Spotify to take a look.

> There's also https://github.com/bonzanini/luigi-slack already.

Yes, I saw it, but need something different. The major differences between this and `luigi-slack` is that:

- Inspired by email notifications (`notifications.py`): configuration though `luigi.cfg` (no need for python configuration)
- Minimal changes to python code (for now only call `configure_slack()`), no need create a new wrapper around `luigi.run`.
- This only depends on `requests` module (does not depend on `slackclient`).
- Work both for python 2 and 3

luigi/contrib/slack_notifications.py
Tarrasch: Yep. You should remove it. `[email]` is like that only to keep backwards compatibility. :)

luigi/contrib/slack_notifications.py
```diff
@@ -0,0 +1,168 @@
...
+
+Use a configuration like the following on luigi.cfg
+
+.. code-block:: ini
+
+    [slack]
+    webhook = https://hooks.slack.com/services/something
+    events = ["START", "SUCCESS", "FAILURE"]
+    template = Luigi task {task_id} event {event}: {args} {kwargs}
+
```
Tarrasch: Please add to the docs here a reference so the other existing luigi-slack-plugin, mentioning the pros/cons.

luigi/contrib/slack_notifications.py
```diff
@@ -0,0 +1,168 @@
...
+            'FAILURE': luigi.Event.FAILURE
+        }
+        for event in self.events:
+            luigi_event = event_aliases.get(event, event)
+            template = self.event_templates.get(event, self.template)
+            luigi.Task.event_handler(luigi_event)(self._handler_for_template(template, event))
+        self.__class__.is_set_handlers = True
+
+
+def configure_slack():
```
Tarrasch: Maybe add some documentation as this is the top-level function to be called.

labianchin:
Thanks for the review! I made the changes to take care of that. 

I would also be happy to have comments from @bonzanini and @daveFNbuck.

> Hey. Can you also _Hide all classes/functions not meant to be publicly used? For example like is done in this module

Can you clarify this one? `SlackNotifyHandling.is_configured`, `SlackNotifyHandling.send_slack_message()`, `SlackNotifyHandling.set_handlers()` and `configure_slack()` are meant to be public. Now I only changed `is_set_handlers` to `_is_set_handlers`. I also added docs to all public methods.

luigi/contrib/slack_notifications.py
```diff
@@ -0,0 +1,181 @@
...
+    events = ["START", "SUCCESS", "FAILURE"]
+    template = Luigi task {task_id} event {event}: {args} {kwargs}
+
+
+Alternatives:
+
+Another plugin with similar purpouse is luigi-slack (https://github.com/bonzanini/luigi-slack). The major differences between this and luigi-slack is that:
+
+- Inspired by email notifications (notifications.py): simple configuration though luigi.cfg (no need for python configuration)
+- Minimal changes to python code (only call configure_slack()), no need create a new wrapper around luigi.run.
```
daveFNbuck: Why require changes to Python code at all? Shouldn't this be called automatically if Slack notifications are configured? If this is going to be part of Luigi, don't treat it like a separate package.

luigi/contrib/slack_notifications.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def __getattr__(self, attr):
+        return getattr(getattr(self, '_obj'), attr, getattr(self, '_default_value'))
+
+
+class SlackNotifyHandling():
+    slack_hook_url = None
+
+    def __init__(self, slack_config):
+        self.slack_hook_url = os.getenv('SLACK_HOOK_URL', slack_config.webhook)
```
daveFNbuck: why are you overriding Luigi settings with environment variables?

luigi/contrib/slack_notifications.py
```diff
@@ -0,0 +1,181 @@
...
+    @property
+    def is_configured(self):
+        """
+        Check if the slack web hook is configured so that slack messages can be sent
+        """
+        return bool(self.slack_hook_url)
+
+    def _make_data(self, text):
+        data = dict(text=text)
+        if self.channel:
```
daveFNbuck: Shouldn't channel be required? I think username is required too...

luigi/contrib/slack_notifications.py
labianchin: Yes, that is a good point. In that case this function would have to be called in the `interface.py` or `worker.py`. For now I was trying to keep this isolate of those modules, avoiding any effect for people that would not use this. Still, this is pretty safe to call if slack isn't configured. @Tarrasch do you have any opinions here?

luigi/contrib/slack_notifications.py
labianchin: Good question. As the webhook url contains a token, some users would prefer to manage this as a secret outside the luigi config.

luigi/contrib/slack_notifications.py
labianchin: No. When setting up a [slack incoming hook](https://api.slack.com/incoming-webhooks) you can define default ones. The reason to have this here is so that you could reuse the same hook with some different configuration.

labianchin:
> I think events like SUCCESS and FAILURE are way too common to warrant Slack notifications.

That depends on the use case. It can be configured, with defaults as START, SUCCESS and FAILURE. I see a lot of value of using at least FAILURE, that is what you get when you setup failure notification e-mails, a slack alternative with other states is nice to have.

> If you're going to do Slack notifications, they might benefit from using attachments (https://api.slack.com/docs/message-attachments) to improve formatting and add color-coding.

luigi/contrib/slack_notifications.py
Tarrasch: I always have on my mind that I want to move out stuff from the `luigi.contrib` kitchensink to separate packages. Particularly as I'm the maintainer every added package is an added maintenance burden for me. So if this could be a separate github repository and python package that would be good.

-----

There is actually one thing that I'm pretty impressed about when I see this package, and that is how you've managed to make the initialization to only be about running a function from this module. If that could be done everywhere (but I doubt it), we hardly need any plugin system for luigi anymore (see #1897)

luigi/contrib/slack_notifications.py
daveFNbuck: How are environment variables more secret than config files? If this is something people want to do, we should probably make changes to the configuration code so we can do it for other secrets in the config like smtp credentials rather than doing it as a one-off just for this.

bonzanini:
Thanks for reaching out, here are my two cents. This looks a bit more interesting than the luigi-slack package especially with py2/py3 explicit support and the minimal code changes required. On the latter, I agree with @daveFNbuck : if you're going for a configuration via luigi.cfg, why requiring code changes at all? 

On the topic of different event types to report, what motivated luigi-slack at the time was the lack of email notification support for the case of all successful tasks (to the best of my knowledge): in this case, a single notification, to report that the whole pipeline has run successfully, was all I needed (hence the message grouping and the use of a context manager you can see in luigi-slack)

luigi/contrib/slack_notifications.py
labianchin: Yes, good points. I guess what would be better is to reference a file with secrets (similar to [.pg_password](https://www.postgresql.org/docs/9.4/static/libpq-pgpass.html) or [kubernetes secrets](https://www.postgresql.org/docs/9.4/static/libpq-pgpass.html)). But yes, I just removed reading from env variables as there is no benefit other than flexibility.

luigi/contrib/slack_notifications.py
labianchin: Yes. I started this for an internal pipeline, and later thought that simple and nice slack notifications might be useful for other pipelines, as e-mail notifications has been useful. The motivation to PR here in luigi is that in this way no external dependencies is needed, just some minimal configuration. But yes, I am aware that this can be a burden to maintain, and so I tried to keep it simple, well documented and tested.

About `configure_slack()`, I believe for now users will have to make that call in their task definition modules. Ideally it would be interesting if we could _configure_ luigi to run some initialization functions (even as a very simple plugin system). Then, all configuration here could be done though `luigi.cfg`.

labianchin:
@bonzanini, thanks for your comments!

> On the latter, I agree with @daveFNbuck : if you're going for a configuration via luigi.cfg, why requiring code changes at all?

I can understand that. But at this moment, it is not clear where in luigi would be a good place to run this initialization. I don't believe this is final and I would like to revisit this in the near future.

> On the topic of different event types to report, what motivated luigi-slack at the time was the lack of email notification support for the case of all successful tasks (to the best of my knowledge): in this case, a single notification, to report that the whole pipeline has run successfully, was all I needed (hence the message grouping and the use of a context manager you can see in luigi-slack)

Good to hear that. Message grouping and defining a number o max_events are very good feature on `luigi-slack`. Here I was trying to keep it simple and I favoring to send a message for each event, in this way you can get a feedback of the task "progress".

luigi/contrib/slack_notifications.py
Tarrasch: @labianchin, would you consider creating a new repo like github.com/spotify/luigi-slack-requests (`-requests` to distinguish it from the existing repo)?

Also, I think it makes sense if users explicitly have to add `luigi-slack-blah` as a package dependency.

-----

Come to think about it, I'm not sure why this is the first time I'm suggesting this in a PR. Clearly it applies to pretty much everything already existing in `contrib`. :)

labianchin: > would you consider creating a new repo like github.com/spotify/luigi-slack-requests (-requests to distinguish it from the existing repo)?

Maybe. We already have an internal repo with other tools (where users install that package and get slack + other things). The goal to PR here is to make it easily available to others. I work on setting up a new repo, it will take a few more days to do that.

dlstadther:
@labianchin  Where did we land here? Continuing PR? Or keeping as external module functionality?

If the latter, I'll close.

labianchin:
Neither. I was about to create an external repo, but I am not using this that much anymore. I think we can close this.

PR #1986, opened 2017-01-12
Keep datatable state in the browser history.

nmb10:
Keep tasks datatable state in the browser history.

## Description
Allows to refresh page without tasks datatable state missing. Also added abilities to add filtered/ordered tasks datatable to bookmarks and links exchange.

## Motivation and Context
Sometimes we need to add filtered table with tasks to bookmarks or refresh the page.

Selenium tests added for entries amount, order and filter (I used selenium + phantomjs driver because I needed 'wait for elements' feature which is too complicated for phantom.js)

### DIFF
--- a/luigi/static/visualiser/js/visualiserApp.js
+++ b/luigi/static/visualiser/js/visualiserApp.js
@@ -866,7 +866,6 @@ function visualiserApp(luigi) {
                 if (data.length) {
                     params.length = data.length;
                 }
-                console.log(data.order);
 
                 var uri = URI().query(params);
                 history.pushState(data, 'task-table', uri.toString());
--- a/luigi/static/visualiser/js/visualiserApp.js
+++ b/luigi/static/visualiser/js/visualiserApp.js
@@ -850,6 +850,53 @@ function visualiserApp(luigi) {
         });
 
         dt = $('#taskTable').DataTable({
+            stateSave: true,
+            stateSaveCallback: function(settings, data) {
+                // Convert data table state to query and save to the browser history.
+                var params = {};
+
+                if (data.search.search) {
+                    params.search__search = data.search.search;
+                }
+
+                if (data.order && data.order.length) {
+                    params.order = '' + data.order[0][0] + ',' + data.order[0][1];
+                }
+
+                if (data.length) {
+                    params.length = data.length;
+                }
+                console.log(data.order);
+
+                var uri = URI().query(params);
+                history.pushState(data, 'task-table', uri.toString());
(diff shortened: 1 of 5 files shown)


luigi/static/visualiser/js/visualiserApp.js
```diff
@@ -850,6 +850,53 @@ function visualiserApp(luigi) {
...
+                if (uri.order) {
+                    order = [uri.order.split(',')];
+                }
+                var o = {
+                    order: order,
+                    length: uri.length,
+                    start: 0,
+                    time: new Date().getTime(),
+                    columns: [
+                        {visible: true, search: {}},
```
Tarrasch: What are the 6 rows here? Can you comment on explaining this magic?

nmb10:
Hi, @Tarrasch 
 > What's a "datatable"? :)

It's js library you used to display table with tasks :) - https://datatables.net/

> Butif we keep the state after a refresh, doesn't that kill the purpose of refreshing? I mean I want to see the new latest values. No?

No. Values (tasks) are not stored. History keeps order, filter and amount of entries on page (Entries select box value).

> I'm not sure I fully understand the part about

An example: you selected to display 50 entries on page, you filtered by 'Uber' and ordered by name. Then you refreshed the page. After page refresh entries, filter and order stay the same (50, 'Uber', name).

> Did you mean you implement so that you can send urls like http://myluigi.com:8082/q=MyTask???

Yes. This is copy-paste from local instance - http://127.0.0.1:8082/static/visualiser/index.html?search__search=Uber&order=1,asc&length=10

> To clarify what's done here, can you add any screenshots or example urls to this PR to clarify what it adds? If it's not too much work.

http://127.0.0.1:8082/static/visualiser/index.html?search__search=Uber&order=1,asc&length=50 . But usually you do not need to construct such an url - just copy-paste it from browser.

> What are the 6 rows here? Can you comment on explaining this magic?

Ok.

Tarrasch:
Ok. It's all clear to me now. I've wanted this thing for like forever. But I let someone else do the code review. :)

Tarrasch:
@daveFNbuck, can you review this? :)

daveFNbuck:
Looks good to me. Would be nice if we also saved the state of the graph buttons for easy sharing.

Tarrasch:
@nmb10, thanks a lot, I'll try this out soon. :)

nmb10:
> Would be nice if we also saved the state of the graph buttons for easy sharing.
I'll check that.

PR #1987, opened 2017-01-16
Set default log level via config

riga:
## Description

This PR allows for the configuration of the default log level, which is [`DEBUG` by default](https://github.com/spotify/luigi/blob/b92e09b6564e22351a3984dc3a3b9a7823614a6c/luigi/interface.py#L50).

## Motivation and Context

It might be a first world problem, but sometimes it would be nice to control the default log level without the need to create (and deploy) a logging config file, but rather by a simple environment variable. The value of this variable must match the name of a [python logging level](https://docs.python.org/2/library/logging.html#logging-levels). Usage example:

```shell
LUIGI_DEFAULT_LOG_LEVEL=INFO luigi --module my_module MyTask --x 123
```

An other option would be to add `default_log_level` to the core config and to prefer the environment variable if set (at least that's what most libs do).

## Tests

Too trivial to test :P 

## Docs

I couldn't find a place in the docs for this feature, but I sure can add a sentence or two to a section of your choice.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -114,6 +114,12 @@ history-filename
   configuration where no history is stored in the output directory by
   Hadoop.
 
+log_level
+  The default log level to use when no logging_conf_file is set. Must be
+  a valid name of a `Python log level
+  <https://docs.python.org/2/library/logging.html#logging-levels>`_.
+  Default is ``DEBUG``.
+
 logging_conf_file
   Location of the logging configuration file.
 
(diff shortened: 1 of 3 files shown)


Tarrasch:
Overall awesome patch. I have wanted this too sometimes. :)

Oh, bonus points if you assert that if there's a config set to read the logging-config, then luigi shouldn't allow one to set this new config variable as it will not take affect.

riga:
Like that?

Tarrasch:
Exactly like that. :)

Tarrasch:
Thanks @riga! :)

Issue #1423, opened 2015-11-24
Pickle crashing when trying to pickle "update_tracking_url" in luigi.worker?


lediur:
This is on Windows, Luigi latest.
I'm trying to run some tasks with `--workers 8`, but this is causing Luigi to crash every time:

```
INFO: Done scheduling tasks
INFO: Running Worker with 8 processes
DEBUG: Asking scheduler for work...
DEBUG: Pending tasks: 1
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Anaconda\lib\multiprocessing\forking.py", line 380, in main
INFO: Worker Worker(salt=938199001, workers=8, host=ai-fieldtest, username=azureuser, pid=4552) was stopped. Shutting do
wn Keep-Alive thread
    prepare(preparation_data)
  File "C:\Anaconda\lib\multiprocessing\forking.py", line 488, in prepare
    assert main_name not in sys.modules, main_name
ERROR: Uncaught exception in luigi
Traceback (most recent call last):
  File "C:\Anaconda\lib\site-packages\luigi\retcodes.py", line 61, in run_with_retcodes
    worker = luigi.interface._run(argv)['worker']
  File "C:\Anaconda\lib\site-packages\luigi\interface.py", line 237, in _run
    return _schedule_and_run([cp.get_task_obj()], worker_scheduler_factory)
  File "C:\Anaconda\lib\site-packages\luigi\interface.py", line 194, in _schedule_and_run
    success &= w.run()
  File "C:\Anaconda\lib\site-packages\luigi\worker.py", line 862, in run
    self._run_task(task_id)
  File "C:\Anaconda\lib\site-packages\luigi\worker.py", line 693, in _run_task
    p.start()
  File "C:\Anaconda\lib\multiprocessing\process.py", line 130, in start
    self._popen = Popen(self)
  File "C:\Anaconda\lib\multiprocessing\forking.py", line 277, in __init__
    dump(process_obj, to_child, HIGHEST_PROTOCOL)
  File "C:\Anaconda\lib\multiprocessing\forking.py", line 199, in dump
    ForkingPickler(file, protocol).dump(obj)
  File "C:\Anaconda\lib\pickle.py", line 224, in dump
    self.save(obj)
  File "C:\Anaconda\lib\pickle.py", line 331, in save
    self.save_reduce(obj=obj, *rv)
  File "C:\Anaconda\lib\pickle.py", line 419, in save_reduce
    save(state)
[...]

PR #1992, opened 2017-01-19
add TaskStatusReporter class to fix Windows pickle issue

davemt:
Windows uses a different method when forking, which caused issues with
the task status updating functions that are passed to TaskProcess
(they need to be pickle-able). Basically running with >1 worker was broken
on Windows. See #1423

This fixes the problem by introducing a new pickle-able class to encapsulate task status
update methods.

- There was existing test coverage for code that was refactored (e.g. `test/task_status_message_test.py`)
- I ran a test task tree on a Windows machine with `--workers 10` before the change to reproduce the issue (got pickle error), then after and the test succeeded -- tested with py2.7

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -110,22 +110,21 @@ class TaskProcess(multiprocessing.Process):
 
     Mainly for convenience since this is run in a separate process. """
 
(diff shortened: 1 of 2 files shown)


Tarrasch:
Unless anybody objects or have other review-points. I'll merge this in soon.

PR #1993, opened 2017-01-19
Control worker processes via UI (scheduler → worker communication)

riga:
## Description

This PR adds a message queue for remote workers that is fetched via the scheduler ping and processed by the actual worker. As a usage example, this PR also adds a feature to the UI that allows users to control the number of processes of a worker.

## Motivation and Context

During our daily work, the scheduling of tasks can take quite some time (even though we enabled parallel scheduling) as we often have thousands of required tasks whose output (several GB or TB) is distributed over a number of remote resources. Sometimes, it takes ~30min - 1h only to build the dependency tree. After starting such a task with e.g. `--workers 1` it might happen that we like to add worker processes to that task (don't ask me why, high-energy physics infrastructure is funny) as if we called it with e.g. `--workers 10`. To avoid the overhead of creating the dependency tree again, I added the feature to control the number of worker processes via the UI.

Screenshot:

For this feature to work (commit 2), I had to introduce communication scheduler → worker that is essentially piggybacked by the ping polling. Besides using `get_work`(which imho might not be the best way to exchange data that is not related to, well, new work/tasks to run), the keepalive polling seems like a good place for that (commit 1).

Also, this bidirectional-ish communication can be used to add further interactive features to the UI. 

I tested the features with these [tasks](https://dl.dropboxusercontent.com/u/1021570/tasks.py) with a single worker process via

```shell
python tasks.py TaskB
```

and added more processes using the UI.

If this is a desired feature, I can also contribute actual tests.


luigi/worker.py
```diff
@@ -1063,3 +1076,39 @@ def run(self):
...
+
+    @message_callback
+    def set_worker_processes(self, n, diff):
+        # determine the new number of worker processes
+        worker_processes = self.worker_processes
+        if diff:
+            worker_processes += n
+        else:
+            worker_processes = n
+        worker_processes = max(0, worker_processes)
```
Tarrasch: I thought we should have at least `1` not `0`. At least comment why 0 and not 1.

luigi/worker.py
```diff
@@ -1063,3 +1076,39 @@ def run(self):
...
+
+        # set it
+        self.worker_processes = worker_processes
+
+        # tell the scheduler
+        self._scheduler.add_worker(self._id, {"workers": worker_processes})
```
Tarrasch: Cool that you didn't forget this ^^

luigi/worker.py
```diff
@@ -1063,3 +1076,39 @@ def run(self):
...
+        if not callable(func):
+            logger.info("Worker %s has no function %s" % tpl)
+        elif not getattr(func, "is_message_callback", False):
+            logger.info("Worker %s function %s is not availale as message callback" % tpl)
+        else:
+            logger.info("Worker %s successfully dispatched message to function %s" % tpl)
+            func(*args)
+
+    @message_callback
+    def set_worker_processes(self, n, diff):
```
Tarrasch: Can't you make it into two different calls? One for `set_worker_processes_increment` and one for `set_worker_processes`. Both only taking one argument.

Or on second thought. I think it's better to not even having the `diff` version. I see these problems:

 * Not needed, humans trivially can make this calculation anyway.
 * If the UI is unresponsive humans are likely to set it twice, getting undesired effect.
 * Also, this is a new feature, let's start simple? :)

Altough, I'm also fine with including it. I don't have any strong opinions, I'll let you decide. :)

luigi/worker.py
```diff
@@ -368,11 +369,22 @@ def run(self):
...
                 except:  # httplib.BadStatusLine:
                     logger.warning('Failed pinging scheduler')
 
+                # handle messages
+                if response is not None:
+                    for message in response["messages"]:
+                        self._message_callback(message)
+
+
+def message_callback(fn):
```
Tarrasch: Thanks for thinking about security from day one. I really like this decorator. :)

luigi/worker.py
```diff
@@ -368,11 +369,22 @@ def run(self):
...
             with fork_lock:
+                response = None
                 try:
-                    self._scheduler.ping(worker=self._worker_id)
+                    response = self._scheduler.ping(worker=self._worker_id)
                 except:  # httplib.BadStatusLine:
                     logger.warning('Failed pinging scheduler')
 
+                # handle messages
+                if response is not None:
```
Tarrasch: Let's be pythonic and just do `if response:`. :)

luigi/worker.py
```diff
@@ -1063,3 +1075,23 @@ def run(self):
             self._handle_next_task()
 
         return self.run_succeeded
+
+    def _handle_message(self, message):
+        logger.debug("Worker %s got message %s" % (self._id, message))
+
+        # the message is a tuple where the first element is an string that defines
+        # the function to call and the remaining elements are its arguments
```
riga: Sure, sounds good.

luigi/worker.py
riga: I removed the diff flag ;)

luigi/scheduler.py
```diff
@@ -878,6 +887,10 @@ def disable_worker(self, worker):
         self._state.disable_workers({worker})
 
     @rpc_method()
+    def set_worker_processes(self, worker, n, diff):
```
riga: I left it as `worker` for the moment as all the rpc methods above that one also use `worker`. But in general, this would really clarify the code.

luigi/scheduler.py
```diff
@@ -1121,7 +1130,8 @@ def get_work(self, host=None, assistant=False, current_tasks=None, worker=None,
     @rpc_method(attempts=1)
     def ping(self, **kwargs):
         worker_id = kwargs['worker']
-        self._update_worker(worker_id)
+        worker = self._update_worker(worker_id)
+        return {"messages": worker.fetch_messages()}
```
riga: Looks like the [`@rpc_method`](https://github.com/spotify/luigi/blob/master/luigi/scheduler.py#L94) decorator uses [`RemoteScheduler._request`](https://github.com/spotify/luigi/blob/master/luigi/rpc.py#L141) which json-dumps the return value, so this should be fine.

If cross-language support is desired, JSON-RPC 2.0 specs would be awesome. https://github.com/riga/jsonrpyc (sorry for the self-promotion ;))

riga:
I'll try to add tests later this day.

Tarrasch:
This looks good. Only tests needed. :)

riga:
@Tarrasch I hope I added enough tests for the beginning ;)

Tarrasch:
@riga, I tried to fix the merge conflicts for you. Let's see if the tests pass. It would be nice if you could quickly review my last merge merge commit. I believe it was just about matching opening and closing brackets for javascript functions...

Tarrasch:
Note, the visualiser tests passed locally for me.

riga:
Should be alright now =)

Tarrasch:
Merged. I'll check this out and see if it works for me too or not.

But after you refresh the page, you'll see that the workers is still the old value (probably because the workers hasn't pinged yet). That doesn't look so good. Can you add a visual hint like "This worker has 13 unread messages" or something like that?

PR #1994, opened 2017-01-22
introduce rpc-attempt retry count and wait time in configuration

zwotzie:
## Description
I find the retries in this short time (3x30s=90 seconds) 
nearly useless. So i changed the rpc.py in this way
that i can do handle the retry attempts and the wait
time by configuration.

## Motivation and Context
We got "to often" the 
"RPCError: Errors (3 attempts) when connecting to remote scheduler"

## Tested?
I ran some testjobs with it and it works for me.

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -104,7 +104,7 @@ class RemoteScheduler(object):
             connect_timeout = config.getfloat('core', 'rpc-connect-timeout', 10.0)
         self._connect_timeout = connect_timeout
 
-        self._rpc_retry_attemps = config.getint('core', 'rpc-retry-attempts', 3)
+        self._rpc_retry_attempts = config.getint('core', 'rpc-retry-attempts', 3)
         self._rpc_retry_wait = config.getint('core', 'rpc-retry-wait', 30)
 
         if HAS_REQUESTS:
@@ -120,10 +120,10 @@ class RemoteScheduler(object):
         full_url = _urljoin(self._url, url_suffix)
         last_exception = None
         attempt = 0
-        while attempt < self._rpc_retry_attemps:
+        while attempt < self._rpc_retry_attempts:
             attempt += 1
             if last_exception:
-                logger.info("Retrying %r/%r" %(attempt, self._rpc_retry_attemps))
+                logger.info("Retrying attempt %r of %r (max)" % (attempt, self._rpc_retry_attempts))
                 self._wait()  # wait for a bit and retry
             try:
                 response = self._fetcher.fetch(full_url, body, self._connect_timeout)
@@ -136,7 +136,7 @@ class RemoteScheduler(object):
         else:
             raise RPCError(
                 "Errors (%d attempts) when connecting to remote scheduler %r" %
-                (self._rpc_retry_attemps, self._url),
+                (self._rpc_retry_attempts, self._url),
                 last_exception
             )
         return response
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -104,22 +104,26 @@ class RemoteScheduler(object):
             connect_timeout = config.getfloat('core', 'rpc-connect-timeout', 10.0)
         self._connect_timeout = connect_timeout
 
+        self._rpc_retry_attemps = config.getint('core', 'rpc-retry-attempts', 3)
+        self._rpc_retry_wait = config.getint('core', 'rpc-retry-wait', 30)
+
         if HAS_REQUESTS:
             self._fetcher = RequestsFetcher(requests.Session())
         else:
             self._fetcher = URLLibFetcher()
 
     def _wait(self):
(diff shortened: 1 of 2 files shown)


luigi/rpc.py
```diff
@@ -104,22 +104,26 @@ def __init__(self, url='http://localhost:8082/', connect_timeout=None):
...
+        self._rpc_retry_wait = config.getint('core', 'rpc-retry-wait', 30)
+
         if HAS_REQUESTS:
             self._fetcher = RequestsFetcher(requests.Session())
         else:
             self._fetcher = URLLibFetcher()
 
     def _wait(self):
-        time.sleep(30)
+        logger.info("Wait for %d seconds" % self._rpc_retry_wait)
```
Tarrasch: Does this line really work, I thought you should replace ` %` with `,`.

luigi/rpc.py
```diff
@@ -104,22 +104,26 @@ def __init__(self, url='http://localhost:8082/', connect_timeout=None):
...
+        time.sleep(self._rpc_retry_wait)
 
-    def _fetch(self, url_suffix, body, log_exceptions=True, attempts=3):
+    def _fetch(self, url_suffix, body, log_exceptions=True):
         full_url = _urljoin(self._url, url_suffix)
         last_exception = None
         attempt = 0
-        while attempt < attempts:
+        while attempt < self._rpc_retry_attemps:
```
Tarrasch: variable misspelled

luigi/rpc.py
```diff
@@ -104,22 +104,26 @@ def __init__(self, url='http://localhost:8082/', connect_timeout=None):
...
             attempt += 1
             if last_exception:
-                logger.info("Retrying...")
+                logger.info("Retrying %r/%r" %(attempt, self._rpc_retry_attemps))
```
Tarrasch: Not a human readable error message. Can you come up with something that can be understood on it's own?

luigi/rpc.py
```diff
@@ -141,7 +145,7 @@ def _request(self, url, data, log_exceptions=True, attempts=3, allow_null=True):
         body = {'data': json.dumps(data)}
 
         for _ in range(attempts):
-            page = self._fetch(url, body, log_exceptions, attempts)
```
Tarrasch: What? did the old code retry 3*3 times?? I see both an outer `for _ in range(attempts)` loop and a loop inside the `_fetch` function. Strange...

luigi/rpc.py
```diff
@@ -104,22 +104,26 @@ def __init__(self, url='http://localhost:8082/', connect_timeout=None):
...
         else:
             self._fetcher = URLLibFetcher()
 
     def _wait(self):
-        time.sleep(30)
+        logger.info("Wait for %d seconds" % self._rpc_retry_wait)
+        time.sleep(self._rpc_retry_wait)
 
-    def _fetch(self, url_suffix, body, log_exceptions=True, attempts=3):
+    def _fetch(self, url_suffix, body, log_exceptions=True):
```
Tarrasch: Hey, if you remove the kwarg `attempts` here, won't you affect other surrounding code that assumes it to be `3`??

zwotzie: the code didn't expect it is 3. It is set hardcoded inwith rpc.py to this number. And the default remains to 3...

luigi/rpc.py
zwotzie: I don't deepdive into the code, but when the central scheduler is not working the logging seems to explain a retry of 3

luigi/rpc.py
zwotzie: yes it does

zwotzie:
To your general questions, why i think 90 seconds of retry time is useless.

There are a lot of scenarios which could have an outage of more then 90 seconds. If something went wrong with the scheduler or overload/crashed or the connectivity is more or less bad then why so optimistic that everything will be ok whithin this short time? When it last more then that i must do something by hand...

But i understand the trade off. When your retry eternally your monitoring has to be more complex...

luigi/rpc.py
Tarrasch: Oh, and there's nowhere it is called? If so then this is fine. :)

Tarrasch:
@zwotzie, I believe they picked 90 seconds, because if things haven't got good in that short time, it's perhaps desired to just "die quickly" and then get restarted by cron an hour later, rather than letting another python thread dingle around doing sleeps.

I'm not convinced to change the default. But I do like that this patch makes it configurable.

luigi/rpc.py
Tarrasch: Actually I meant as "doesn't flake8 complain?". I think you can give the `self._rpc_retry_wait` as a parameter to the logging framework, and letting it do interpolating as a part of it's display step.

Tarrasch:
Thanks!

PR #1997, opened 2017-01-24
Move stuff to contrib

Tarrasch:
## Description

While it's of course a bit subjective that what is core and not. I
settled on choosing these modules.  I found the rest either to be kind
of core or very hard to move and eventually create seperate packages
for.

I conciously did not create an `luigi/mrrunner.py` importing contrib, as
it will not do as one expects as you typically will use `.__file__` on
it to ship it with your map reduce job.

## Motivation and Context

This is a step in the direction of defining something that is later core. Eventually `luigi.contrib` can be split into modules.

Nope. I tried to edit the files carefully, and then there will be Travis testing this out.


Tarrasch:
@dlstadther, well, they'll only get an deprecation warning in the beginning. The temporary backward compatibility modules will eventually be deleted tough. So they have time to upgrade between versions.

PR #2000, opened 2017-01-25
Allow namespaces to be set to python module

Tarrasch:
This have been a problem I've thought about for years. And also
consciously gradually refactored the code base in previous patches
towards making this relatively short path.

Coming up with the final syntax and implementation also took me a while.
Initially I imagined there would be a [config] parameter one could set,
or specify by command line. But it turned out to be quite troublesome:

 * It's not clear how much it should affect. Should it also affect the
   things luigi exports?
 * It's hard to transition into it. Imagining a company with independent
   teams, the `auto_namespace` function can be applied package by
   package, luigi-version by luigi-version to whoever chooses to apply
   this new convention.

This solution came out to be pretty clean, and with readable tests and
detailed and a bit opinionated documentation.

I'm going for a week-long no computer vacation soon. After that I'll try running this in production.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -57,10 +57,10 @@ def namespace(namespace=None, scope=''):
     It is often desired to call this function with the keyword argument
     ``scope=__name__``.
 
-    The ``scope`` keyword makes so this call is only affective for task classes
-    with a matching [*]_ ``__module__``. The default value for ``scope`` is the
-    empty string, which means all classes. Multiple calls with the same scope
-    simply replace each other.
+    The ``scope`` keyword makes it so that this call is only effective for task
+    classes with a matching [*]_ ``__module__``. The default value for
+    ``scope`` is the empty string, which means all classes. Multiple calls with
+    the same scope simply replace each other.
 
     The namespace of a :py:class:`Task` can also be changed by specifying the property
     ``task_namespace``.
@@ -70,7 +70,7 @@ def namespace(namespace=None, scope=''):
         class Task2(luigi.Task):
             task_namespace = 'namespace2'
 
-    This explicit setting takes priority over whatever is set in
+    This explicit setting takes priority over whatever is set in the
     ``namespace()`` method, and it's also inherited through normal python
     inheritence.
 
@@ -79,7 +79,7 @@ def namespace(namespace=None, scope=''):
     *New since Luigi 2.6.0:* ``scope`` keyword argument.
 
     .. [*] When there are multiple levels of matching module scopes like
-           ``a.b`` vs ``a.b.c``, the more specific one wins.
+           ``a.b`` vs ``a.b.c``, the more specific one (``a.b.c``) wins.
     .. seealso:: The new and better scaling :py:func:`auto_namespace`
     """
     Register._default_namespace_dict[scope] = namespace or ''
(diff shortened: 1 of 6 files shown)


Tarrasch:
Rendered docs screenshots:

selection_097

selection_098

Tarrasch:
@ulzha @sisidra, I know you're one of the biggest Spotify users and that you've wanted something similar to this for a long time. What do you think about this patch?

Tarrasch:
Anyone wants to review? @erikbern @dlstadther

erikbern:
lgtm but are namespaces widely used? idk

luigi/task.py
```diff
@@ -47,27 +47,72 @@
...
     """
     Call to set namespace of tasks declared after the call.
 
-    It is best practice to call this function without arguments at the end of any file it has been
-    used in. That is to ensure that subsequent tasks have the default namespace again.
+    It is often desired to call this function with the keyword argument
+    ``scope=__name__``.
+
+    The ``scope`` keyword makes so this call is only affective for task classes
+    with a matching [*]_ ``__module__``. The default value for ``scope`` is the
```
dlstadther: English correction (if understand correctly):

The ``scope`` keyword makes __*it*__ so __*that*__ this call is only __*effective*__ for task classes with a matching [*]_``__module__``.

luigi/task.py
```diff
@@ -47,27 +47,72 @@
...
-    ``task_namespace``. This solution has the advantage that the namespace
-    doesn't have to be restored.
+    ``task_namespace``.
 
     .. code-block:: python
 
         class Task2(luigi.Task):
             task_namespace = 'namespace2'
 
+    This explicit setting takes priority over whatever is set in
```
dlstadther: over whatever is set in __*the*__

luigi/task.py
```diff
@@ -47,27 +47,72 @@
...
+
+
+def auto_namespace(scope=''):
+    """
+    Same as :py:func:`namespace`, but instead of a constant namespace, it will
+    be set to the ``__module__`` of the task class. This is desirable for these
+    reasons:
+
+     * Two tasks with the same name will not have conflicting task families
+     * It's more pythonic, as modules are pythons recommended way to
```
dlstadther: __*Python's*__

luigi/task.py
```diff
@@ -47,27 +47,72 @@
...
+       do namespacing.
+     * It's traceable, when you see the full name of a task, you can immedietly
```
dlstadther: It's traceable. __*When*__ you see the full name of a task, you can __*immediately*__

luigi/task.py
```diff
@@ -47,27 +47,72 @@
...
+       identify where it is defined.
+
+    We recommend to call this function from your package's outermost
```
dlstadther: We recommend __*calling*__ this function...

Tarrasch:
@erikbern, that's a really good concern actually. I'm not sure either if people care/use namespaces. But here's some related facts and thoughts:

 * At Spotify I thought it was hard to know where tasks are defined when there's no namespace. Compare the task family `IngestUsersToDbTask` vs `discover_weekly.ingestions.IngestUsersToDbTask`. Also the problem when two different teams have a task named `IngestUsersToDbTask`, the luigi server cannot distinguish the task families then.
 * I've recently improved documentation about what namespaces is.
 * In an follow up PR I'll try to make it so that instead of `luigi --module hello.world hello.world.MyTask --my-param 123` you'll just write `luigi hello.world.MyTask --my-param 123`. The increased readability and enforced consistency I find to be great wins. Also I tend to like frameworks with conventions so that I don't need to think for myself.
 * There have also been a few others sending PRs about namespaces, so I think some people care.

-----

@dlstadther, thanks. I'll fix these lingual issues right away. :)

erikbern:
sounds good!

PR #2006, opened 2017-01-27
Fixed link to HDFS API docs in README

snorfalorpagus:
Minor fix to the URL of the link to the HDFS documentation in the README.

I'm not familiar with the history of the API, but it seems that the hdfs module is currently located in contrib.


Tarrasch:
Thanks @snorfalorpagus :)

PR #2007, opened 2017-01-28
use compress_response by default

FxIII:
## Description
Tornado.web.Application has the ability to compress responses if clients add Accept-Encoding in their requests

## Motivation and Context
Setting compress_response allows to cut down network traffic particularly when asking for dep_graph on (very) complex workflows. 
I got JSON responses as big as 19MB on huge workflows composed by a relative small number of Task Families, cut down to 200KB.

I changed my site-package's luigi version on my preproduction systems, ran my job successfully, then manually changed the code here. 
I don't know how to trigger the travis tests from the github web interface, sorry.

### DIFF
--- a/luigi/server.py
+++ b/luigi/server.py
@@ -193,6 +193,7 @@ class RootPathHandler(BaseTaskHistoryHandler):
 def app(scheduler):
     settings = {"static_path": os.path.join(os.path.dirname(__file__), "static"),
                 "unescape": tornado.escape.xhtml_unescape,
+                "compress_response": True,
                 }
     handlers = [
         (r'/api/(.*)', RPCHandler, {"scheduler": scheduler}),


Tarrasch:
Cool that you even benchmarked it from "19MB" to "200KB". Does it actually load faster too? I mean, do we even kill the bottleneck with this patch? :)

Tarrasch:
Nobody have reviewed for 5 days, and this seems pretty harmless so I'll just accept this, as @FxIII seem to have tested it pretty well.

Tarrasch:
Big big big thanks! :)

Tarrasch:
pinging @FxIII, how much faster did luigi get "IRL" from this?

FxIII:
Well, we had an huge workflow with a thousands of split/gater tasks. Those tasks where basically the same task with just a single parameter changing amongst them, I think that is somewhat an extreme case.
We moved from a couple of minutes or so to something less than 30" IIRC

Tarrasch:
Great. That means it at least sometimes was a bottle neck! :)

PR #2009, opened 2017-02-03
Set authors to "The Luigi Authors"

Tarrasch:
As of there being close to 300 commiters to luigi, and today's main
contributors to the code base are not the two original founders anymore
(which is a good sign of success :)). It makes more sense to mention
a generic term as the authors of the software.

I took the name from the Tornado docs (obviously replacing "Tornado"
with "Luigi"). The original authors are still mentioned in the bottom of
the README.md.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -1,7 +1,7 @@
 Source: luigi
 Section: libs
 Priority: extra
-Maintainer: Elias Freider <freider@spotify.com>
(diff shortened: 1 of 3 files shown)


Tarrasch:
@erikbern @freider

erikbern:
LGTM!

PR #2011, opened 2017-02-03
PyMySql support

### LINKED ISSUE #1529: Problems with mysql.connector dependencies installation
Recently I started to have problems with installation of `mysql-connector-python` package which is needed for `lugi.contrib.mysqldb.py`

For more details see:
https://bitbucket.org/pypa/pypi/issues/365/many-packages-have-empty-list-of-links

My question is if that wouldn't be better to replace this package for example by https://pypi.python.org/pypi/MySQL-python/1.2.5?

Would you agree with this change or would you prefer "forcing" people to install `mysql-connector-python` from custom non standard pip address via `pip install https://cdn.mysql.com/Downloads/Connector-Python/mysql-connector-python-1.0.12.tar.gz`

## Note

I was wondering that travis has not found that out, but the problem seems to me copying cached packages. How often are you regenerating the cache and trying to install all the packages from scratch on the travis machine? Would'nt be better to install all the packages whenever the test is run (I guess that problem might be very long installation of all the required packages)?

sbliven:
## Description

Support the pure-python PyMySql library, while maintaining support for mysql-connector-python. This fixes #1529.

## Motivation and Context

The mysql-connector-python package depends on externally hosted files and so doesn't install from pypi (#1529). While a third-party replacement is available (mysql-connector-python-rf), a pure python replacement ([PyMySql](https://github.com/PyMySQL/PyMySQL)) is available and actively maintained. Both libraries use nearly the same API, so luigi can support both with minimal effort.

## Testing
 

A unit test for the mysqldb module is at `test/_mysqldb_test.py`. This requires the database parameters to be set to valid values, so it is disabled by default. To run it, I created new tox environments depending on either `pymysql>=0.7` or `mysql-connector-python-rf >= 2.0.0`. The tests pass with either library, using py27 and mysql 5.5.

Currently the environments are not included in tox.ini, but I could easily add them if desired.


sbliven:
@ziky90

sbliven:
All tests passed except the py35-cdh environment, which timed out. I find it very unlikely that my code changes are responsible for the test failure on only that one environment. Can someone restart travis?

themalkolm:
@sbliven I think closing and reopening this pull request should trigger travis to rebuild.

PR #2012, opened 2017-02-03
Fixed problem that the retry policy per task does not work.

mtoriumi:
## Description
Fixed an issue where "retry policy per task" is not given as a parameter of add_task in _handle_next_task.


interskh:
I was wondering why my retry policy per task doesn't work and hasn't got a chance to fix it. Thank you very much for the fix!

Tarrasch:
This looks correct. But this requires test cases before it can be merged. Also I would like to know for which code paths it does work, and which new code paths this patch solves.

mtoriumi:
Previously, it worked normally if the task failed twice, and it failed normally when it failed three times. Therefore, it was fixed as a test case when the task failed three times.

Tarrasch:
Thanks!

PR #2013, opened 2017-02-03
Show number of unread worker messages in UI

riga:
This PR adds a feature as proposed in #1993.

In the UI, the number of unread messages per worker is shown near the top right tool box.

Changes are quite trivial and should require any additional tests =)


erikbern:
hm 13 commits for 10 lines of code? :)

luigi/scheduler.py
```diff
@@ -1328,6 +1328,7 @@ def worker_list(self, include_running=True, **kwargs):
                 started=worker.started,
                 state=worker.state,
                 first_task_display_name=self._first_task_display_name(worker),
+                unread_messages=len(worker.rpc_messages),
```
Tarrasch: Call this `num_unread_messages`? Or even better, `num_unread_rpc_messages`. Would be nice if a grep-search for `rpc_messages` turned out all relevant bits of code for the RPC messages.

luigi/static/visualiser/index.html
```diff
@@ -229,6 +229,12 @@
             margin-left: 4px;
           }
 
+          #workerList .box-tools > span.label-unread-worker-messages {
+            margin-right: 6px;
+            font-style: italic;
+            vertical-align: middle;
```
Tarrasch: What color do you have it in? I was thinking of maybe making it red, as it's something that under normal conditions quickly disappears.

luigi/scheduler.py
riga: Yep, sounds good. Done.

luigi/static/visualiser/index.html
riga: Also done ;)

riga:
@erikbern Sorry for that. I merged with some changes from @Tarrasch on the master (d0dcac6). Should have opened a new branch...

Tarrasch:
@erikbern @riga, thankfully GitHub now has squashmerge, so we can merge this without fixing the strange-looking history in this PR.

Awesome job @riga, thanks for fast fix. :)

PR #2015, opened 2017-02-06
(SF) Ensure bulk api output is in requested order

dlstadther:
After [4 years of request](https://success.salesforce.com/ideaView?id=08730000000h6C2AAI), Salesforce now allows Foreign Key Relationships within their Bulk API. However, the csv response provided does not always return in the same order as requested. This PR ensures that the output csv is written in the same order as it was requested.

## Description

For CSV content type, output is written in the same order as the fields of the input SOQL query.

## Motivation and Context

Of my 8 daily Salesforce SOQL Queries, 4 of them do not return with the same column order as was requested in the SOQL. This PR ensures that.

I have run this with my tasks and it works for me. I have specified this check only for CSV content type as it is the only one I use and cannot speak to the validity of return types for other approved SF content types.

### DIFF
--- a/luigi/contrib/salesforce.py
+++ b/luigi/contrib/salesforce.py
@@ -23,7 +23,7 @@ from collections import OrderedDict
 import re
 import csv
 import tempfile
-from StringIO import StringIO
+from luigi.six.moves import StringIO
 
 import luigi
 from luigi import Task
--- a/luigi/contrib/salesforce.py
+++ b/luigi/contrib/salesforce.py
@@ -23,6 +23,7 @@ from collections import OrderedDict
 import re
 import csv
 import tempfile
+from StringIO import StringIO
 
 import luigi
 from luigi import Task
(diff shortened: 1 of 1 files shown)


dlstadther:
Hey @dacort , could you review (as you're the only other Salesforce contributor)?

dacort:
My only thought is regarding UTF-8 output. Since the original code just wrote the raw file, I wonder if there's anything different that needs to be done to ensure this change also consistently writes UTF-8 files. 

@dlstadther Is this something you can confirm locally? Example using the current code:
```shell
› file data/salesforce_account.csv
data/salesforce_account.csv: UTF-8 Unicode text
```

Otherwise, looks good to me!

dlstadther:
@dacort I'm not sure i understand your concern.

There wasn't previously any guarantee that the output would be utf-8. Are you saying that guarantee should be added here?

dacort:
I'm saying that a non-UTF-8 file shouldn't be written in the event that it has UTF-8 characters in it. I'm not sure if the CSV module handles this intelligently or not.

dlstadther:
From python2.7 csv doc, `The csv module doesn’t directly support reading and writing Unicode`. However, python3 doesn't mention any explicit issues with unicode, but does say it can handle any sequence type (which includes unicode).

dacort:
Hm, ok. Will try to evaluate this week to verify that the change works as expected.

dlstadther:
@colemanja91, mind taking a look since you're also a SF user? Thanks

colemanja91:
@dlstadther just tried it with a few different queries - returns all columns in expected order. Code looks good to me!

dlstadther:
I no longer use nor have access to Salesforce anymore. So I'll leave it up to @dacort whether i should resolve the conflicts and merge, or just close this PR.

PR #2016, opened 2017-02-06
Fix key-based credentials for redshift copy

dlstadther:
Fixes breaking changes to key-based credentials from PR #1962 

## Description

Mistakenly merged PR using word `aws_secret_key` rather than `aws_secret_access_key`. Caused Redshift `InternalError` when using key-based credentials during Redshift copy.

## Motivation and Context

Fixes breaking PR

Ran fine for me this morning.


Tarrasch:
Thanks! :)

PR #2020, opened 2017-02-09
Update outdated mainClass to mainJarFileUri

pkova:
The current dataproc API does not accept mainClass as a parameter, the correct parameter name is now mainJarFileUri.

## Description
Changed outdated REST API parameter in the method submit_spark_job from mainClass to mainJarFileUri

## Motivation and Context
The current code is broken and dataproc can not find the .jar-files specified in mainClass.

Yes, the Spark jobs now run as they should.

### DIFF
--- a/luigi/contrib/dataproc.py
+++ b/luigi/contrib/dataproc.py
@@ -62,12 +62,12 @@ class DataprocBaseTask(_DataprocBaseTask):
             },
             "sparkJob": {
                 "args": job_args,
-                "mainClass": main_class,
+                "mainJarFileUri": main_class,
                 "jarFileUris": jars
             }
         }}
         self.submit_job(job_config)
-        self._job_name = os.path.basename(self._job['sparkJob']['mainClass'])
+        self._job_name = os.path.basename(self._job['sparkJob']['mainJarFileUri'])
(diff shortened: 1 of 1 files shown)


constantijn:
Hi Pyry,

If you look over here:
https://cloud.google.com/dataproc/docs/reference/rest/v1/projects.regions.jobs#SparkJob

You'll see that there's 2 ways to get your main class to spark. One is to submit a normal jar using "jarFileUris" and then tell spark which class in your jar file to run using the "mainClass" parameter. The other one is to build a jar file that has some extra meta info that specifies it's own main class. If you go that way you can submit it with "mainJarFileUri".

This api has been in place since the v1 release of Dataproc and hasn't changed since then. The current Luigi code only supports the way of submitting jar files and the mainClass separately. (And it very much works because I'm using it.) If you'd like to make a pull request that adds the functionality to use the "mainJarFileUri" way of submitting jobs (along with an extra test case) that'd be very useful, however the pull request as is breaks the existing functionality of submitting jars and the main class separately.

Thanks,
Constantijn

P.S.
If you submit your current jar using "jarFileUris" and then also fill in the "mainClass" parameter to pint at the fully qualified name of your main class i believe it should also work with the current Luigi code and your current jar file.

pkova:
Allright, thanks for the info!

PR #2021, opened 2017-02-09
Adds Task.on_external_failure method to provide custom error messages for failed External Tasks

sklarsa:
## Description
This PR adds an `on_external_failure` method to the `Task` class.  This method is called in the `TaskProcess.run` method when an external task fails, and enables the user to provide a custom failure message for external tasks.  I also add 2 relevant tests to cover the additional codepaths.

## Motivation and Context
My pipeline consists of a fair amount of external tasks, many of which have overridden `Task.complete` methods.  One thing that I'd like to be able to do is provide a more informative explanation for why an external task failed in the luigi dashboard.   Currently, there is no way to do this, and I can't repurpose the existing `on_failure` method because no Exception was raised.

I added the method directly to the `Task` class in case users use the `externalize` method instead of inheriting from the `ExternalTask` class.

Included 2 additional unit tests that pass.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -562,6 +562,19 @@ class Task(object):
         Default behavior is to send an None value"""
         pass
 
+    def on_external_failure(self):
+        """
+        Override for custom error handling for failed external tasks.
+
+        This method gets called if a task is an :py:class:`luigi.task.ExternalTask`, or if a
+        :py:class:`luigi.task.Task`'s :py:meth:`run` method is not implemented.
+
+        The returned value is json encoded and sent to the scheduler as the `expl` argument.
+
+        Default behavior is to send a None value
+        """
+        pass
+
     @contextmanager
     def no_unpicklable_properties(self):
         """
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -186,8 +186,10 @@ class TaskProcess(multiprocessing.Process):
                     status = DONE
                 else:
                     status = FAILED
-                    expl = 'Task is an external data dependency ' \
-                        'and data does not exist (yet?).'
+                    expl = self.task.on_external_failure()
+                    if not expl:
+                        expl = 'Task is an external data dependency ' \
+                            'and data does not exist (yet?).'
             else:
                 new_deps = self._run_get_new_deps()
                 status = DONE if not new_deps else PENDING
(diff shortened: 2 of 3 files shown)


luigi/task.py
```diff
@@ -562,6 +562,19 @@ def on_success(self):
...
+        Override for custom error handling for failed external tasks.
+
+        This method gets called if a task is an :py:class:`luigi.task.ExternalTask`, or if a
+        :py:class:`luigi.task.Task`'s :py:meth:`run` method is not implemented.
+
+        The returned value is json encoded and sent to the scheduler as the `expl` argument.
+
+        Default behavior is to send a None value
+        """
+        pass
```
Tarrasch: Did you consider just do `return Task is an external data dependency and data does not exist (yet?).'`? Less `None`'s and `ifs`.

sklarsa:
@Tarrasch, I like the suggestion of moving the `"Task is an external data dep..."` text to the `on_external_failure` method body, which I implemented in the latest diff.

Even though dynamic nature of python gives you the flexibility to return strings from the `complete()` method, I think it could confuse users (and perhaps introduce unintentional bugs) if I implemented that functionality.  Since it is such a core component of the luigi workflow, I personally prefer that `complete()` only return boolean values.

Tarrasch:
Can you paste some example of how you would implement the `on_external_failure` method?

-----

Also, I think `on_external_failure()` isn't the best name and it doesn't need to be specific to external tasks. Why not call it `reason_not_complete()`? It could be implemented for both external and normal tasks. It could even have an intelligent default returning which files are missing (if you depend on tasks with only FileSystemTargets).

sklarsa:
Sorry about the delay on this, I'm just working through whether or not it would be a good idea to take care of @erikbern's TODO in `luigi/worker.py:182` which should resolve #814.  Unfortunately, changing the logic in worker.py to:
```
if self.task.complete():
    status = DONE
 elif _is_external(self.task):
    status = FAILED
    expl = self.task.on_external_failure()
else:
``` yields additional test failures in `test/worker_multiprocess_test.py` 

Just want to provide a status update

kalvdans:
I agree with Tarrasch that `on_external_failure()` is a bad name and `reason_not_complete()` is a better name.

Issue #2024, opened 2017-02-10
[visualizer] Wrong links to task graph in worker tab.


riga:
In the workers tab, when clicking on the "Root task" link of a worker, or in the "View graph" action in the task list of that worker, I'm getting redirected to the "tasks" tab again.

This bug was probably introduced by #2002.

It seems like the links in the "workers" tab have the old URL format (`.../static/visualiser/index.html#myTaskId_a785cb2ad6`), while the links in the "tasks" tab have the new, correct format (`.../static/visualiser/index.html#tab=workers&taskId=myTaskId_a785cb2ad6`)).

riga:
Fixed in #2025.

PR #2002, opened 2017-01-26
UI field values tracking improved.

nmb10:
## Description
Keep Dependency graph tab fields state in the hash. Task list fields tracking moved to hash.

## Motivation and Context
Fix fields values reset on page refresh.
Allow to pass direct links with tab and fields states.

## Have you tested this? If so, how?
I included functional tests (selenium + phantomJS)

## Cases covered:
1. Open tasks page by direct link.

    Steps:

       Open http://127.0.0.1:8080/static/visualiser/index.html#tab=tasks&length=50&search__search=er&filterOnServer=1&order=1,desc

    Result:
        Task list tab is opened;
        'Show entries' value = 50;
        "Filter table" value = er;
        "Filter On Server" is checked;
        Table is ordered by Name column.

2. Open graph tab by direct link.

    Steps:
        http://127.0.0.1:8082/static/visualiser/index.html#tab=graph&taskId=23&invert=1&hideDone=1&visType=svg

    Result:
        Dependency graph tab is opened;
        Task Id input value = 23;
       "Show Upstream Dependencies" is checked;
       "Hide Done" is checked;
       "Visualization Type" = SVG

3. Keep task list fields state after changing tabs.

    Steps:

        Open task list tab at http://127.0.0.1:8082/static/visualiser/index.html

        Change "Show entries" to 50

        Enter "er" to "Filter Table"

        Check "Filter On Server"

        Order table by Name column

        Switch to Dependency graph tab

        Switch to Task list tab

    Result (all fields should keep their values.):

        Task list tab is opened;

        "Show entries" value = 50;

        "Filter table" value = er;

        "Filter On Server" is checked;

        Table is ordered by Name column.

4. Keep dependency graph fields state after changing tabs.

    Steps:

        Open task list tab at http://127.0.0.1:8082/static/visualiser/index.html

        Switch to Dependency Graph tab

        Change Task ID to 13

        Click "Show task details"

        Check "Show Upstream Dependencies"

        Check "Hide Done"

        Change "Visualization Type" to SVG

        Switch to Task list tab

        Switch to Dependency Graph tab

    Result (all fields should keep their values.):

        Dependency graph tab is opened;

        Task Id input value = 13;

        "Show Upstream Dependencies" is checked;

        "Hide Done" is checked;

        "Visualization Type" = SVG

PR #2025, opened 2017-02-10
Fix links to task graphs in visualizer.

riga:
## Description

This PR fixes the links to task graphs in the visualizer (workers tab) that were introduced in PR #2002.

The problem is described in issue #2024.


riga:
Yep =)
The updated line are just copy&paste changes of working links that were missing in 2 or 3 places, so nothing too dramatic.

PR #2026, opened 2017-02-12
Only run ecs_test.py if nose attribute awsecs is set

ehdr:
## Motivation and Context

The `test/contrib/ecs_test.py` requires AWS credentials and an AWS account
with a running ECS cluster to execute properly. Previously, it's been
prevented from running by being skipped unless `boto3` has been manually
installed in the test environment.

This was triggered by #1980, which adds `boto3` as a dependency in test.

## Description

To be able to include `boto3` in the default test environment, and still
prevent `ecs_test.py` from running e.g. in CI, this change sets an `awsecs`
[nose attribute][1] on the test, so it will only execute e.g. if
`NOSE_ATTR=awsecs` is in the shell environment, or if tox is invoked
with the `-awsecs` tox-environment:

    tox -e py27-awsecs

Specifically, it should not be executed by the Travis CI engine (even if
`boto3` is installed).

[1]: https://nose.readthedocs.io/en/latest/plugins/attrib.html

It has been tested locally using tox.

### DIFF
--- a/test/contrib/ecs_test.py
+++ b/test/contrib/ecs_test.py
@@ -33,6 +33,8 @@ Written and maintained by Jake Feala (@jfeala) for Outlier Bio (@outlierbio)
 
 import unittest
 
+from nose.plugins.attrib import attr
+
 import luigi
 from luigi.contrib.ecs import ECSTask, _get_task_statuses
 
@@ -73,6 +75,7 @@ class ECSTaskOverrideCommand(ECSTaskNoOutput):
         return [{'name': 'hello-world', 'command': ['/bin/sleep', '10']}]
 
 
+@attr('awsecs')
 class TestECSTask(unittest.TestCase):
 
     def setUp(self):
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,33,34,35}-{cdh,hdp,nonhdfs,gcloud,postgres,unixsocket}, visualiser, pypy-scheduler, docs, flake8
+envlist = py{27,33,34,35}-{awsecs,cdh,hdp,nonhdfs,gcloud,postgres,unixsocket}, visualiser, pypy-scheduler, docs, flake8
 skipsdist = True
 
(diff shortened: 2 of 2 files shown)


kwilcox:
Is the longer term/cleaner solution to use something similar to https://github.com/spulec/moto to avoid communicating with AWS directly or do the innards of ECS actually require a legitimate connection?

ehdr:
@kwilcox Could probably be interesting, though @jfeala is more familiar with this specific test.

jfeala:
Yes, mocking aws is a better long-term solution than these tests, which are nice for integration testing but not really unit tests. Maybe even better would be to opt-in to run the tests, via an environment variable, since only a small slice of Luigi users will care about this extension.

Another important note is that Amazon has introduced its [Batch](https://aws.amazon.com/batch/) service, which is a more convenient layer on top of ECS for job queues on scalable compute environments. I expect that many users of Luigi and ECS will be migrating to Batch at some point, and I'm going to shift my own effort to a Luigi Batch extension.

ehdr:
Thanks! So should we go for this solution, or instead check for an environment variable (e.g. `LUIGI_ECS_TEST`) and skip the test if it's not set?

jfeala:
My choice would be the latter, maybe with a note in the ecs.py docstring about setting the environment variable if you want to run an invasive test on your own AWS account

ehdr:
Great, thanks. Pushed a commit with that change in #1980, so closing this PR.

jingpengwu:
@jfeala I am looking to use luigi with AWS Batch too, but did not find any documentation! How do you get started?

jfeala:
Hi @jingpengwu, if you're already set up with Batch, it's easy to wrap the jobs in `luigi.Task`s. I have code for this that I was planning to submit a PR when finished, but here's a preview that works for me currently:
https://gist.github.com/jfeala/edd68921d49f2daf86a31017c50c0be7

jingpengwu:
@jfeala tks, pretty helpful!

PR #2032, opened 2017-02-13
Introduce Python3 compatibility in example

amarco90:
The example in `examples/foo_complex.py` does not work in Python3 due to the division operator.

## Description
I fixed the error by importing the division operator from `__future__` and using the integer division operator (`//`).

## Motivation and Context
When running the example using Python3 I was getting the following error:

      File "/Users/amarco/Desktop/luigi/examples/foo_complex.py", line 64, in requires
        for i in range(valor / max_depth):
    TypeError: 'float' object cannot be interpreted as an integer

I ran the example task and it completed successfully:

    Scheduled 51 tasks of which:
    * 51 ran successfully:
        - 50 examples.Bar(num=0,1,10,11,12,13,14,15,16,17,18,...)
        - 1 examples.Foo()

    This progress looks :) because there were no failed tasks or missing external dependencies

### DIFF
--- a/examples/foo_complex.py
+++ b/examples/foo_complex.py
@@ -23,6 +23,7 @@ You can run this example like this:
             $ luigi --module examples.foo_complex examples.Foo --workers 2 --local-scheduler
 
 """
+from __future__ import division
 import time
 import random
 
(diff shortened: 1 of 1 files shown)


Tarrasch:
Thanks! :)

Issue #2028, opened 2017-02-13
Failed building wheel for luigi


GunasekaranR:
Running setup.py bdist_wheel for luigi ... error
  Complete output from command c:\python27\python.exe -u -c "import setuptools, tokenize;__file__='c:\\users\\treselle\\appdata\\local\\temp\\pip-buil
d-ql05v6\\luigi\\setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exe
c'))" bdist_wheel -d c:\users\treselle\appdata\local\temp\tmpa7equkpip-wheel- --python-tag cp27:
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build\lib
  creating build\lib\luigi
  copying luigi\batch_notifier.py -> build\lib\luigi
  copying luigi\cmdline.py -> build\lib\luigi
  copying luigi\cmdline_parser.py -> build\lib\luigi
  copying luigi\configuration.py -> build\lib\luigi
  copying luigi\date_interval.py -> build\lib\luigi
  copying luigi\db_task_history.py -> build\lib\luigi
  copying luigi\event.py -> build\lib\luigi
  copying luigi\execution_summary.py -> build\lib\luigi
  copying luigi\file.py -> build\lib\luigi
  copying luigi\format.py -> build\lib\luigi
  copying luigi\interface.py -> build\lib\luigi
  copying luigi\local_target.py -> build\lib\luigi
  copying luigi\lock.py -> build\lib\luigi
  copying luigi\mock.py -> build\lib\luigi
  copying luigi\notifications.py -> build\lib\luigi
  copying luigi\parameter.py -> build\lib\luigi
  copying luigi\postgres.py -> build\lib\luigi
  copying luigi\process.py -> build\lib\luigi
  copying luigi\retcodes.py -> build\lib\luigi
  copying luigi\rpc.py -> build\lib\luigi
  copying luigi\s3.py -> build\lib\luigi
  copying luigi\scheduler.py -> build\lib\luigi
  copying luigi\server.py -> build\lib\luigi
  copying luigi\six.py -> build\lib\luigi
  copying luigi\target.py -> build\lib\luigi
  copying luigi\task.py -> build\lib\luigi
  copying luigi\task_history.py -> build\lib\luigi
  copying luigi\task_register.py -> build\lib\luigi
  copying luigi\task_status.py -> build\lib\luigi
[...]

PR #2033, opened 2017-02-14
Repaired luigi build

Aaronhrndz:
## Description
Changed URI.js route

## Motivation and Context
Luigi does not build properly, referencing issue #2028

The project has passed the travis tests, luigid is running properly.


luigi/static/visualiser/index.html
```diff
@@ -21,11 +21,10 @@
         <script src="lib/jquery.slimscroll.min.js"></script>
         <script src="lib/AdminLTE/js/app.min.js"></script>
         <script src="lib/datatables/js/jquery.dataTables.min.js"></script>
-        <script src="lib/URI.js/1.18.2/URI.js"></script>
+        <script src="lib/URI.js"></script>
```
Tarrasch: Wasn't the issue the `URI.js` matching some regex? Do we not have the same issue here?

Aaronhrndz: In setup.py it copies the static files with this function:

def get_static_files(path):
    return [os.path.join(dirpath.replace("luigi/", ""), ext)
            for (dirpath, dirnames, filenames) in os.walk(path)
            for ext in ["*.html", "*.js", "*.css", "*.png",
                        "*.eot", "*.svg", "*.ttf", "*.woff", "*.woff2"]]


The URI folder was called URI.js, and the above function fails when it gets this folder extension

dlstadther:
It's been requested and (somewhat) supported by @nmb10 and @kwilcox to leave the version number within the path. So just change `luigi/static/visualiser/lib/URI.js/1.18.2./URI.js` to `luigi/static/visualiser/lib/URI/1.18.2/URI.js`.

luigi/static/visualiser/index.html
```diff
@@ -21,7 +21,7 @@
         <script src="lib/jquery.slimscroll.min.js"></script>
         <script src="lib/AdminLTE/js/app.min.js"></script>
         <script src="lib/datatables/js/jquery.dataTables.min.js"></script>
-        <script src="lib/URI.js"></script>
+        <script src="lib//URI/1.18.2/URI.js"></script>
```
dlstadther: `lib/URI/1.18.2/URI.js`

one too many `/`

dlstadther:
@Tarrasch @nmb10 Are you good with this?

nmb10:
Hi, @dlstadther . Yes.

miku:
> Can anyone else verify this works?

I believe this PR would fix the installation problem, that I encountered:

```
$ pip install luigi
....
error: can't copy 'luigi/static/visualiser/lib/URI.js': doesn't exist or not a regular file
```

In setup.py, [get_static_files](https://github.com/spotify/luigi/blob/7701d9fb4ad2ec859d208fde9fd475bbc43c64a1/setup.py#L20) will return (among other globs) `'static/visualiser/lib/*.js',`, which will match [URI.js](https://github.com/spotify/luigi/tree/master/luigi/static/visualiser/lib/URI.js), which is not a regular file, as setup seems to expect.

dlstadther:
Thanks!

Lucas-C:
I have the same error when `pip install`ing luigi 2.6.0. I cannot use this version in its current state.
Could you please include this fix in a 2.6.1 release ?

dlstadther:
@Lucas-C I don't have the ability to push luigi releases. That'll have to wait until @Tarrasch returns.

shahabedinh:
I had issue 

/opt/cloudera/parcels/Anaconda-4.0.0/bin/pip install luigi

....
reating build/lib/luigi/static/visualiser/lib
    copying luigi/static/visualiser/lib/mustache.js -> build/lib/luigi/static/visualiser/lib
    copying luigi/static/visualiser/lib/jquery-1.10.0.min.js -> build/lib/luigi/static/visualiser/lib
    error: can't copy 'luigi/static/visualiser/lib/URI.js': doesn't exist or not a regular file


then I download and then pip it.  

wget https://github.com/spotify/luigi/archive/master.zip
unzip master.zip
cd luigi-master/
sudo /opt/cloudera/parcels/Anaconda/bin/pip install .

it works for me

PR #2034, opened 2017-02-19
update contrib.salesforce to use 'wb' when opening output file

colemanja91:
## Description
In `open()` calls, changed option `'w'` to `'wb'` for forward-compatibility with Python3.x

## Motivation and Context
I am attempting to run a QuerySalesforce export task on Python3.5 and am running into an error described in #1647. I tried the provided solution, `format=luigi.format.Nop`, but it does not help because `'w'` appears to be hard-coded into this package.

I successfully ran a QuerySalesforce task in both Python2.7 and Python3.5.

### DIFF
--- a/luigi/contrib/salesforce.py
+++ b/luigi/contrib/salesforce.py
@@ -213,7 +213,7 @@ class QuerySalesforce(Task):
         """
         Merges the resulting files of a multi-result batch bulk query.
         """
-        outfile = open(self.output().path, 'wb')
+        outfile = open(self.output().path, 'w')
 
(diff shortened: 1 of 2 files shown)


dlstadther:
I don't have any problem with this, but please try to resolve Travis' errors.

colemanja91:
Got it passing locally, but had to revert back to `'w'` on `merge_batch_results`; the unit test for this one uses `MockTarget`, which has `'w`' hard-coded. I don't know enough about that part of the codebase yet to be comfortable modifying.

Tarrasch:
Thanks!

PR #2038, opened 2017-02-24
Load os.environ into the config singleton

hugobast:
## Description

`LuigiConfigParser` is a derived class of python's `ConfigParser`. When there is a need to have environment variable interpolation support in the cfg/ini file we can pass `os.environ` to the initializer to make it work. This PR looks for bringing that support to the project.

## Motivation and Context

I think it would be neat if luigi.cfg could interpolate some environment
variables. I think this is the minimum required amount of work to make
this happen but I'm looking for guidance.

I've tested this in the way of overriding `luigi.configuration.get_config`:

```
luigi.configuration.get_config = lambda: luigi.configuration.LuigiConfigParser.instance(os.environ)
```

Example luigi.cfg file:

```
[DatabaseConfiguration]
host=%(DATABASE_HOST)s
database=%(DATABASE_NAME)s
user=%(DATABASE_USER)s
password=%(DATABASE_PASSWORD)s
```

### DIFF
--- a/luigi/configuration.py
+++ b/luigi/configuration.py
@@ -128,4 +128,4 @@ def get_config():
     """
     Convenience method (for backwards compatibility) for accessing config singleton.
     """
-    return LuigiConfigParser.instance()
+    return LuigiConfigParser.instance(os.environ)


hugobast:
@gpoulin requesting comment on this PR when you get a chance, thanks!

gpoulin:
I haven't worked on Luigi for a while, but based on the docstring, this method is deprecated. We probably shouldn't change the behaviour of a deprecated method.

cabouffard:
@dlstadther Hey! I'm unsure if you're the right person to ask but I was wondering if something could be done with this PR as this is a great  behavior to have!

I've been able to see this issue requesting this behavior: https://github.com/spotify/luigi/issues/2354 and there are 2 persons that have thumbs-up the PR.

In the story linked above, @Tarrasch  mention that this behavior is something offered by ConfigParser by default, but it isn't. 

Looking around from different posts and thread resolution on the internet, the fix is to pass the `os.environ` value to the ConfigParser so that it has access to the environment variable.

Let me know if you need additional informations fron us, as suspect this additions to cause no harm and that even though the function says it's deprecated, this the the main entry point to load the configuration object throughout the Luigi codebase.

Please let us know if there is another way to load environment variables into the configuration file that we may not be aware of.

Thanks!

dlstadther:
@cabouffard I don't have much experience working with environment variables within Python. I can understand the usefulness of this functionality.

1) Does this PR solve the problem (@hugobast , @cabouffard )?
2) Can tests be added here, @hugobast ?
3) Can some documentation be added regarding this functionality?

hugobast:
@dlstadther thanks for getting back, I'll gladly do all test and documentation.

>Does this PR solve the problem (@hugobast , @cabouffard )?

Absolutely it does solve the problem

Issue #2036, opened 2017-02-23
Expose Central Scheduler metrics for easy ingestion


chrispalmer:
It would be extremely helpful to be able to get some metrics from the Central Scheduler via an RPC call. Will the visualizer gives you a view of what has been happening recently, it only shows a fixed amount of data (depending on your `remove-delay` config setting). Being able to continuously collect metrics would enable us to track overall health of our pipelines and the central scheduler.

Our particular use case would be to pull those metrics with Prometheus on a regular basis (although personally I'm still trying to understand how Prometheus works). But a more general solution not tailored to a particular metric collection tool, would still be helpful.

I've started exploring this and I'm still getting to grips with exactly what state the central scheduler stores. So I don't know what the complete list of potential metrics might be, but for starters they could include things like:

- Number of workers connected (per host)
- Number of tasks in each state (pending, running, failed etc) (per host where applicable)
- Average runtime of completed tasks

chrispalmer:
I'd be very happy to work on adding this solution, as I've wanted to find something I could contribute back to this project for a while. Figured there might be some discussion about the correct way to build this.

colemanja91:
This is a cool idea. I've been using Prometheus for about 7 months now; I was planning to add success/failure handlers to my tasks, but this would make it easier. I might also be interested in exploring metric collection for New Relic, as there's a chance I'll be migrating there in the near future. 

Output for Prometheus might look something like this: 
```
# HELP luigi_tasks_failed Count of tasks failed
# luigi_tasks_failed gauge
luigi_tasks_failed{task="task1"} 5
```

Tarrasch:
I would love this too. I think you could start with a draftish PR (perhaps without tests) and ask for comments. Then you could iterate on the feedback from that and later turn it into a proper PR.

Do you see this as two parts? One to create the `/api/statistics` call and one to create a "Statistics" tab in the visualizer UI?

chrispalmer:
One of the challenges with this is how much this wants to be integrated into the scheduler code. In my mind you could go one of two ways

1. Have an endpoint that when accessed, calculates/collects all the desired metrics and reports them back. In this case the endpoint would probably just get passed a reference to the scheduler, and it would go inspect `scheduler._state` and count various things.
2. Have the statistics being constantly collected and the endpoint simply reports back the collected metrics. This would require a persistent object of some sort to collect the metrics in and then add calls to the scheduler code to update various metrics as events happen.

[...]

PR #2044, opened 2017-02-27
Initial pass at metrics collection for Central Scheduler

chrispalmer:
## Description
This is an initial pass at adding metric collection functionality to the Central Scheduler, and is still a work in progress. I'm looking for feedback on the general approach.

Summary of changes:

- Dummy `MetricsCollector` class (luigi/metrics.py) with no-op methods, but can be subclassed for a specific metric collection tool.
- `PrometheusMetricsCollector` class (luigi/contrib/prometheus.py) that subclasses the above, as an example implementation.
- New `scheduler` config option/parameter to select the metric collection tool.
- changes to `SimpleTaskState` to identify and trigger metric updates.
- Added `Scheduler.remove_worker` rpc_method to related `Worker._remove_worker` method to allow workers to tell the scheduler when they are finishing and disconnecting. This allows us to distinguish between deliberate disconnects and worker failures.

## Motivation and Context
This is an attempt to provide functionality desired in [issue 2306](https://github.com/spotify/luigi/issues/2036)

Does not yet include any tests.
I have run it against some very simple pipelines and it seems to do as expected.

### DIFF
--- a/luigi/metrics.py
+++ b/luigi/metrics.py
@@ -1,0 +1,14 @@
+
+
+class MetricsCollector(object):
+    """
+    Dummy MetricsCollecter base class that can be replace by tool specific implementation
+    """
+    def __init__(self, scheduler):
+        self._scheduler = scheduler
+
+    def handle_task_status_change(self, task, status):
+        pass
+
+    def handle_worker_status_change(self, worker, status):
+        pass
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -147,6 +147,8 @@ class scheduler(Config):
 
     prune_on_get_work = parameter.BoolParameter(default=False)
 
+    metrics_collection = parameter.Parameter(default=None)
+
     def _get_retry_policy(self):
         return RetryPolicy(self.retry_count, self.disable_hard_timeout, self.disable_window)
 
@@ -354,7 +356,7 @@ class Worker(object):
         self.disabled = False
         self.rpc_messages = []
 
-    def add_info(self, info):
+    def update_info(self, info):
         self.info.update(info)
 
     def update(self, worker_reference, get_work=False):
@@ -431,6 +433,7 @@ class SimpleTaskState(object):
         self._status_tasks = collections.defaultdict(dict)
         self._active_workers = {}  # map from id to a Worker object
         self._task_batchers = {}
+        self._metrics_collector = None
 
     def get_state(self):
         return self._tasks, self._active_workers, self._task_batchers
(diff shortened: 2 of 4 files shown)


luigi/scheduler.py
```diff
@@ -354,7 +356,7 @@ def __init__(self, worker_id, last_active=None):
         self.disabled = False
         self.rpc_messages = []
 
-    def add_info(self, info):
+    def update_info(self, info):
```
chrispalmer: I changed this method name, because I feel like it fits much better with what is actually happening.

luigi/scheduler.py
```diff
@@ -642,7 +663,16 @@ def _remove_workers_from_tasks(self, workers, remove_stakeholders=True):
     def disable_workers(self, worker_ids):
         self._remove_workers_from_tasks(worker_ids, remove_stakeholders=False)
         for worker_id in worker_ids:
-            self.get_worker(worker_id).disabled = True
+            worker = self.get_worker(worker_id)
+            worker.disabled = True
+            self._update_metrics_worker_status_change(worker, 'disabled')
+
+    def _update_metrics_task_status_change(self, task, status):
+        if status != UNKNOWN:
```
chrispalmer: Not sure this is the correct place for this check, but tasks can get added to the state before we actually know anything about them. I wish there was a better way to identify when new tasks are being added.

Tarrasch:
I just came back from computer-less vacation. I'll try to look at this soon :)

luigi/scheduler.py
```diff
@@ -621,17 +628,31 @@ def get_active_workers(self, last_active_lt=None, last_get_work_gt=None):
...
+
     def get_worker_ids(self):
         return self._active_workers.keys()  # only used for unit tests
 
     def get_worker(self, worker_id):
         return self._active_workers.setdefault(worker_id, Worker(worker_id))
 
-    def inactivate_workers(self, delete_workers):
+    def remove_worker(self, worker_id):
+        self.inactivate_workers([worker_id], reason='disconnected')
```
Tarrasch: Maybe `"gracefully_stopped"` is clearer than `disconnected`?

luigi/metrics.py
```diff
@@ -0,0 +1,14 @@
+
```
Tarrasch: I think you need to copy paste the license-blah blah from the other files, updating the year.

luigi/scheduler.py
```diff
@@ -147,6 +147,8 @@ class scheduler(Config):
 
     prune_on_get_work = parameter.BoolParameter(default=False)
 
+    metrics_collection = parameter.Parameter(default=None)
```
Tarrasch: Maybe add a description?

luigi/scheduler.py
Tarrasch: Yes! Those edits are great actually. :)

PR #2045, opened 2017-02-27
Added cluster support for ECS and corrected status debug message

vanhove:
## Description

- Added luigi.Parameter for cluster to be able to use another cluster then the default one.

- Fixed an issue with the status in debug message when following up running ECS tasks.

## Motivation and Context

I needed support to run ECS tasks on other clusters then the default one.
 

### DIFF
--- a/luigi/contrib/ecs.py
+++ b/luigi/contrib/ecs.py
@@ -128,6 +128,9 @@ class ECSTask(luigi.Task):
                 ]
             }
 
+    :param cluster: str defining the ECS cluster to use.
+        When this is not defined it will use the default one.
+
     """
 
     task_def_arn = luigi.Parameter(default=None)
--- a/luigi/contrib/ecs.py
+++ b/luigi/contrib/ecs.py
@@ -135,7 +135,7 @@ class ECSTask(luigi.Task):
 
     task_def_arn = luigi.Parameter(default=None)
     task_def = luigi.Parameter(default=None)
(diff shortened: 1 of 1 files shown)


luigi/contrib/ecs.py
```diff
@@ -133,6 +132,7 @@ class ECSTask(luigi.Task):
 
     task_def_arn = luigi.Parameter(default=None)
     task_def = luigi.Parameter(default=None)
+    cluster = luigi.Parameter(default=None)
```
dlstadther: Could you update the above docstring to include this parameter and a description? (I don't use ECS, so perhaps this is obvious and unnecessary)

dlstadther:
Also, can you confirm that this replaces PR #1717 ?

adamchainz:
Afraid I'm useless as a reviewer, I barely know what luigi is

luigi/contrib/ecs.py
vanhove: I updated the docstring

jfeala:
Please add a `default='default'` to the Parameter, to prevent breaking for users running on their default cluster, as is done in PR #1717. Also, it doesn't completely replace #1717 because there are other improvements in that PR (e.g., checking exit codes)

vanhove:
@jfeala Thanks for pointing that out. I pushed the change.

dlstadther:
@vanhove Thanks for your first github contribution! Sorry for such delay in merging!

PR #2049, opened 2017-03-02
Specify default port to comply with change in init for psycopg 2.7

smehan:
## Description
Updated psycopg to 2.7 and encountered a fault in the init for postgres.py. Previously, we could define parameters for db connection to postgres with no separate port defined and no port in the host name. If that it attempted in 2.7, it attempts to create a connection with port=None.

Changing the default init value of port from None to int:5432 allows for postgres.py to continue to work with 2.7 as before.

I ran my jobs with this code and it worked as before the 2.7 update.


Tarrasch:
Thanks! :)

PR #2056, opened 2017-03-07
add Weebly to list of Luigi users

jasonmay:
## Description
Addition to documentation

## Motivation and Context
This solves the issue that Weebly is a cool company but people don't know they use Luigi.

I clicked 'Preview' and it looked fine.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,6 +154,7 @@ Some more companies are using Luigi but haven't had a chance yet to write about 
 * `Squarespace <https://www.squarespace.com/>`_
 * `OAO <https://adops.com/>`_
 * `Grovo <https://grovo.com/>`_
+* `Weebly <https://www.weebly.com/>`_
 
 We're more than happy to have your company added here. Just send a PR on GitHub.
 


dlstadther:
Do you have any blog or PR post about Weebly's use of Luigi? I believe the preference is to link to useful applications of Luigi, more so than company homepages.

jasonmay:
We don't have any posts at the moment, if that's a requirement, I'd be happy to discuss with my team about publishing something.

dlstadther:
I'm not sure it's a requirement, but it's preferred. In general, it would be awesome to see more luigi blog posts out there though!

erikbern:
yes would prefer to link to something!

erikbern:
wonder if we should drop the requirement a bit and link to anyone that uses luigi, presentation or not.

Tarrasch:
Not sure. [Airflow](https://github.com/apache/incubator-airflow#who-uses-airflow) have that kind of list, it's surely a lot bigger, but I'm not sure if it's providing value. I maybe prefer Luigi's list, as sometimes linked presentations are useful, but sometimes it's just links to websites or other repositories, so the quality and consistency is a bit low right now...

dlstadther:
@Tarrasch & @erikbern ,

I think this depends on the image we're trying to convey here. If we just want to show that a lot of companies are/have using/used Luigi, then we go the Airflow route and accept links to homepages. However, if the goal is to provide interesting use cases/implementations of luigi then we go the latter.

As @Tarrasch noted, we currently have some of both. So perhaps we split the middle and accept links to homepages, but add references to blogs after the company name. Something like the following:

[Spotify](https://www.spotify.com) ([presentation](https://www.slideshare.net/erikbern/luigi-presentation-nyc-data-science))
[Schibsted](http://www.schibsted.com/)
[Example Company](#) ([presentation](#)) ([blog](#))

If we agree on this, i'm happy to submit the PR to modify the current README for this change.

erikbern:
yeah i like @dlstadther's idea. i was thinking along the same lines. i would also put all companies with presentations or blog posts above the companies without

erikbern:
can you rebase on latest changes now? thanks!

dlstadther:
@jasonmay Please rebase on the latest changes. We'd be happy to merge your company into our list! Thanks!

jasonmay:
It's rebased now; thanks for reviewing!

dlstadther:
Thanks @jasonmay !

Issue #2050, opened 2017-03-02
LocalTarget and temporary_path


adaitche:
FileSystemTarget has the method `temporary_path`, which returns a context manager. On exit the context manager renames the temporary path to the final path by means of the method `rename_dont_move` of the attribute fs (of type FileSystem).

When `temporary_path` is used on LocalTarget, on exit from the context manager the following warning is printed

```
lib/python3.5/site-packages/luigi/target.py:171: UserWarning: File system LocalFileSystem client doesn't support atomic mv.
  warnings.warn("File system {} client doesn't support atomic mv.".format(self.__class__.__name__))
```

This is because LocalFileSystem does not implement the `rename_dont_move` method and the default implementation is used. However LocalFileSystem has essentially implemented this functionality in the `move` method, but unfortunately it is not used by temporary_path.

Is there a particular reason for not implementing `rename_dont_move` on LocalFileSystem? If no, one could make `rename_dont_move` just a reference to the `move` method of LocalFileSystem, which already implements the desired functionality.

I am happy to submit a pull request, if the above suggestion makes sense.

rbdixon:
If you put this pull request in I'll test it... I just hit the same issue.

Tarrasch:
Ah, yes. The `temporary_path` path feature is new and this is a rough edge I didn't notice.  Please send a PR. :)

PR #2057, opened 2017-03-07
Add method rename_dont_move to LocalFileSystem

adaitche:
## Description
See issue #2050

## Motivation and Context
When the method `temporary_path` is called on LocalTarget, a warning is
given because the default implementation of the method `rename_dont_move`
is used. However LocalFileSystem has already a more sophisticated
implementation of the desired functionality in the `move` method. Here
`rename_dont_move` is implemented by means of this `move` method.

Additionally the exception raised by `move` is changed from RuntimeError
to FileAlreadyExists, as the latter is expected by the test of
`rename_dont_move`.

Run the tests in test/local_target_test.py

### DIFF
--- a/luigi/local_target.py
+++ b/luigi/local_target.py
@@ -99,7 +99,7 @@ class LocalFileSystem(FileSystem):
         but cannot be guaranteed.
         """
         if raise_if_exists and os.path.exists(new_path):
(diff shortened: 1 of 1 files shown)


Tarrasch:
Thanks!

PR #2058, opened 2017-03-07
Make temporary_path work with paths not containing a directory

adaitche:
## Motivation and Context
The context manager returned by FileSystemTarget.temporary_path creates
the parent directory of the target path. However this fails when the
path does not contain a directory and an exception is raised. To be able
to use paths without a directory (e.g. "foobar.txt"), the creation of
the parent directoy is skipped in this case.

Verified that no exception is raises when the target path does not contain a directory.


Tarrasch:
This would require a test case I think. :)

adaitche:
Test case has been added (to the same commit).

Tarrasch:
Great! Thank you! :)

PR #2061, opened 2017-03-12
Fix history display

amarco90:
Some of the JS/CSS imports in the history page are broken. This leads to a couple of wrong-displayed elements (see screenshots).

Moreover I think they should be integrated in the visualizer (as a new tab?). In that way the history feature would be more visible and possibly lead to a more extended use. Probably this is not the proper place to discuss this last point but I don't know where can be discussed otherwise, sorry about that.

## Description
I fixed the paths on the imports and now bootstrap and jQuery are imported properly. Moreover I used the default layout more extensively and removed a, now, unused file. The use of the layout page is important since it creates a container div in which the other elements are placed.

## Motivation and Context
Currently there are some issues on how the history pages are displayed.`localhost:8082/tasklist`:

screen shot 2017-03-12 at 3 58 03 pm

And `localhost:8082/history/by_id/1` (note that a part of the text is even outside of the screen!):

screen shot 2017-03-12 at 4 32 49 pm

After the changes:

screen shot 2017-03-12 at 3 58 39 pm

screen shot 2017-03-12 at 4 34 59 pm

I ran the central scheduler like so:

    luigid --pidfile pidfile --logdir log/ --state-path state/state

I ran one of the example tasks:

    luigi --module examples.foo examples.Foo --workers 2

This is how my `luigi.cfg` looks like:

    [scheduler]
    record_task_history = True
    state-path = /Users/name/develop/luigi/luigi-state.pkl

    [task_history]
    db_connection = sqlite:////Users/name/develop/luigi/luigi-task-hist.db

After accessing localhost:8082/history the pages are displayed properly.

### DIFF
--- a/luigi/templates/header.html
+++ b/luigi/templates/header.html
@@ -1,4 +1,0 @@
-<!doctype html>
-<title>Luigi Task History</title>
-<link href="/static/visualiser/lib/bootstrap3/css/bootstrap.min.css" rel="stylesheet">
-<script src="/static/visualiser/lib/bootstrap3/js/bootstrap.min.js"></script>
--- a/luigi/templates/layout.html
+++ b/luigi/templates/layout.html
@@ -82,10 +82,10 @@
       .lead { margin-top: -17px; margin-bottom: 13px; }
 
     </style>
-    <link href="{{ static_url("visualiser/lib/bootstrap/css/bootstrap.css") }}" rel="stylesheet">
+    <link href="{{ static_url("visualiser/lib/bootstrap3/css/bootstrap.min.css") }}" rel="stylesheet">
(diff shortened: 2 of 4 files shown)


amarco90:
@Tarrasch it would be great if you could take a look at this :)

Tarrasch:
When the build is red. you should mention why it's still OK (like this time it was unrelated error). I'll just merge this as I don't know this part, and nobody else have reviewed.

tashrifbillah:
Hi @amarco90 , do you know how to get `1` to use in `localhost:8082/history/by_id/1` for a task? At task level, there is a `self.task_id` different from the `1`. The `1` is in fact the column id in the history `db` file.

PR #2062, opened 2017-03-13
Add MongoDB Target Support

MathiasDesch:
## Description

Add support for MongoDB target with unit testing
* Add `MongoCellTarget` targeting a field in a single document
* Add `MongoRangeTarget` targeting a field in a list of documents

## Motivation and Context

Use MongoDB as a backend for task execution

* I have included unit testing with local MongoDB database. `tox -e py27-nonhdfs mongo_test.py` looks good. It needs  a local mongo instance running.
* We use a sightly different implementation in production since almost 2 years for our tasks. Everything works just fine.

## Possible Issue and implementation choice
* The mongo queries used [here](https://github.com/spotify/luigi/compare/master...short-edition:master#diff-7b8ff4750a32408e0dbaa0015994b2f9R123) and [here](https://github.com/spotify/luigi/compare/master...short-edition:master#diff-7b8ff4750a32408e0dbaa0015994b2f9R157) might be too heavy. A solution is be to make them execute as batch but I was not sure it is needed in this PR.
* I choosed to pass the pymongo client as a parameter intead of instanciating it in the constructor. This allow the client to choose how to handle the connection to mongoDB. Having a new instance at each constructor might lead to creating too many connection pools.


luigi/contrib/mongodb.py
```diff
@@ -0,0 +1,165 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2017 Big Datext Inc
```
dlstadther: Shouldn't this be Spotify?

MathiasDesch: I took that from [Twitter](https://github.com/spotify/luigi/blob/2dcfeaa5773d54582ad9d8546394685636ce4783/test/contrib/bigquery_test.py) contribution, but I can change that, we do not mind at all.

I do not know the policy about that, anyway it is open source :)

Tarrasch: To be honest I have no idea what should go in there. I typically wrote the name of my current institution.

MathiasDesch: @Tarrasch Then, I will keep it like that.

MathiasDesch:
@dlstadther 

The commit 23534be fix the Flake8 and doc syntax. 
I also followed the pattern in [Postgres Test](https://github.com/spotify/luigi/blob/master/test/contrib/postgres_with_server_test.py) in order to skip the unit test if there is not local mongo db instance running. That should make all the travis test pass.

Unfortunately, I can not use [mongomock](https://github.com/mongomock/mongomock) cause it does not fully support `$project` yet.

Concerning the reviewers, I do not know anyone using luigi and mongo except my coworkers whom already internally reviewed this code.

dlstadther:
@MathiasDesch I agree that i'm not sure the current py34-cdh error is caused by your code - i'd still like to get some insight from @Tarrasch or someone regarding this error.

Regarding reviewers, since this is a contrib module which doesn't currently exist (and thus community impact is minimal), I have no problems merging if no one else reviews in the next few days.

MathiasDesch:
@dlstadther Sounds good to me! :)

Tarrasch:
LGTM. Ignore the py34-cdh error.

MathiasDesch:
@dlstadther No reviewer seems to show up. I guess we can merge.

PR #2063, opened 2017-03-14
Fix worker tool box in visualizer

riga:
## Description

The worker toolbox in the visualizer shows some buggy information in the phase between adding a worker to the scheduler and actual task processing (so while the dependency tree is built). This looks like:

bug

The reason is that `workers` is both used for the list of worker objects in the visualizer for rendering and the number of worker processes of a worker. See [here](https://github.com/spotify/luigi/blob/00d637f553f814b75dec717a77d7b26ced90b7e1/luigi/static/visualiser/index.html#L325-L341) (first and last line in the marked area). Initially, there's no 'worker' field in the worker object so Mustache falls back to the list of workers which renders to the example above.

## Motivation and Context

A simple renaming of the variable that holds the list of worker objects for rendering fixes the bug (2nd commit). In addition, the worker toolbox is removed now entirely when a worker is disabled (1st commit). Previously, the "disable" button was removed while the "add workers" menu was still visible.

### DIFF
--- a/luigi/static/visualiser/index.html
+++ b/luigi/static/visualiser/index.html
@@ -322,7 +322,7 @@
                 </div>
               </div>
             </div>
-            {{#workers}}
+            {{#workerList}}
             {{#is_disabled}}
             <div class="box box-solid box-default">
             {{/is_disabled}}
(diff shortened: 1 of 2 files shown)


Tarrasch:
How was this tested by the way? Looks good too me.

Tarrasch:
Whoups, I wanted to use the rebase merge strategy. Oh well. Not too big deal.

riga:
I created a short test task that just did a `time.sleep(n)` to delay the dependency tree building. Wasn't sure how this could be covered by a test case.

Issue #2060, opened 2017-03-10
Avoid exception handling for PostgresTarget.create_marker_table() with "CREATE TABLE IF NOT EXISTS"


ant1j:
I think it would be more elegant to add `IF NOT EXISTS` in the query for creating the `marker_table` in `luigi.contrib.postgres` module, instead of dealing with an exception to handle the case where the table already exists.

```sql
CREATE TABLE IF NOT EXISTS {marker_table} {
update_id TEXT PRIMARY KEY,
target_table TEXT,
inserted TIMESTAMP DEFAULT NOW())
```

`IF NOT EXISTS` is available since postgres 9.1 (released more than 5 years ago) and it seems that it is also available in many RDBMS.

So if the method should be rewrote I guess it would be as follows:

```python
    def create_marker_table(self):
        """
        Create marker table if it doesn't exist.

        Using a separate connection since the transaction might have to be reset.
        """
        connection = self.connect()
        connection.autocommit = True
        cursor = connection.cursor()
        if self.use_db_timestamps:
            sql = """ CREATE TABLE IF NOT EXISTS {marker_table} (
                      update_id TEXT PRIMARY KEY,
                      target_table TEXT,
                      inserted TIMESTAMP DEFAULT NOW())
                """.format(marker_table=self.marker_table)
        else:
            sql = """ CREATE TABLE IF NOT EXISTS {marker_table} (
                      update_id TEXT PRIMARY KEY,
                      target_table TEXT,
                      inserted TIMESTAMP);
                  """.format(marker_table=self.marker_table)
        try:
            cursor.execute(sql)
        except psycopg2.ProgrammingError as e:
                # in case there is another exception, i.e. other than a duplicated table name, raise
                raise
        connection.close()
```

Again I am not experienced enough to make a PR, so if any good soul can manage to do it...

Thanks

Tarrasch:
> Again I am not experienced enough to make a PR, so if any good soul can manage to do it...

Not correct. Anyone can make a PR. All previous, current and future contributors are humans and typically make mistakes. We would be happy to accept your bugfix. :)

ant1j:
Here you go.
You were true, anyone can make a PR!

dlstadther:
resolved by #2064

PR #2064, opened 2017-03-14
Add 'IF NOT EXISTS' to prevent duplicate table exception

ant1j:
PR for issue #2060: Avoid exception handling for PostgresTarget.create_marker_table() with "CREATE TABLE IF NOT EXISTS"

## Description
adding Postgresq `IF NOT EXISTS` clause in order to avoid having to rely on a exception handling when the `marker_table` already exists

## Motivation and Context
Fix issue #2060 

Not tested directly, but the SQL query is valid for Postgresql since version 9.1

### DIFF
--- a/luigi/contrib/postgres.py
+++ b/luigi/contrib/postgres.py
@@ -219,7 +219,7 @@ class PostgresTarget(luigi.Target):
                       inserted TIMESTAMP);
                   """.format(marker_table=self.marker_table)
 
-        cursor.execute(sql) 
+        cursor.execute(sql)
         connection.close()
 
     def open(self, mode):
--- a/luigi/contrib/postgres.py
+++ b/luigi/contrib/postgres.py
@@ -207,13 +207,13 @@ class PostgresTarget(luigi.Target):
         connection.autocommit = True
         cursor = connection.cursor()
         if self.use_db_timestamps:
-            sql = """ CREATE TABLE {marker_table} (
+            sql = """ CREATE TABLE IF NOT EXISTS {marker_table} (
                       update_id TEXT PRIMARY KEY,
                       target_table TEXT,
                       inserted TIMESTAMP DEFAULT NOW())
-                """.format(marker_table=self.marker_table)
+                  """.format(marker_table=self.marker_table)
         else:
-            sql = """ CREATE TABLE {marker_table} (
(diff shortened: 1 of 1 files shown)


luigi/contrib/postgres.py
```diff
@@ -207,24 +207,21 @@ def create_marker_table(self):
...
         else:
-            sql = """ CREATE TABLE {marker_table} (
+            sql = """ CREATE TABLE IF NOT EXISTS {marker_table} (
                       update_id TEXT PRIMARY KEY,
                       target_table TEXT,
                       inserted TIMESTAMP);
                   """.format(marker_table=self.marker_table)
         try:
             cursor.execute(sql)
         except psycopg2.ProgrammingError as e:
```
dlstadther: Flake8 doesn't like the assignment, but non-use of `e` here.

I think we can just remove the try/except here for `cursor.execute(sql)`.

ant1j:
I have updated the code; it seems that the PR is automatically updated and tests are run also, which means I have nothing more to do... is that right?

Tarrasch:
Can you find somebody to review this code? Somebody who knows about the technologies (postgres?) this PR touches upon.

ant1j:
Isn't the `luigi/test/postgres_with_server_test.py` file supposed to do so?

I am more fluent in (postgre)sql than Python, so I know the query is working. I crafted quickly 2 sqlfiddle here and they do not return any error:
- http://sqlfiddle.com/#!15/d1587 : `CREATE TABLE IF NOT EXISTS` with values inserted => OK
- http://sqlfiddle.com/#!15/b33421 : (`CREATE TABLE IF NOT EXISTS` twice and then values inserted => OK

I don't know if it is enough as a code review; maybe some people using Redshift could also confirm that, as Redshift as the same feature (see http://docs.aws.amazon.com/fr_fr/redshift/latest/dg/r_CREATE_TABLE_NEW.html)

Tarrasch:
I think it helps to have real humans look at code. Also you'll learn more if you get feedback from peers. :)

If however noone volunteers for a few days or a week. I'm fine with just merging this.

smehan:
I'll have a look at it today.

-s.

dlstadther:
If no one else reviews, I'll merge this soon. Thanks

PR #2066, opened 2017-03-19
Do not raise HDFSCliError inside snakebite client.

jkukul:
Re 94e1b9f - In one of snakebite client's methods, `HDFSCliError` is re-raised. I don't think there's any reason for this - it simply has been there ever since snakebite client was "extracted" out of the command line client. When the `HDFSCliError` is re-raised the actual snakebite stacktrace is swallowed and makes debugging more difficult.

+ Two tiny code improvements in `contrib/hive` (ef6440a, 54764bb).


Tarrasch:
Awesome! Thanks!

PR #2067, opened 2017-03-20
add Red Hat to list of Luigi users

colemanja91:
## Description
Adding Red Hat to list of Luigi users

## Motivation and Context
Would like to share our use case with others

Link is working


dlstadther:
Just finished reading this post. Super cool @colemanja91 . Thanks for sharing!

PR #2068, opened 2017-03-20
REDSHIFT: standardize credential usage

zsalzbank:
## Description
Obtaining the credentials for a redshift connection is now uniform
accross all of the redshift tasks. This has the added benefit of
allowing role-based credentials in all of the tasks as well.

## Motivation and Context
I need to use role-based credentials for all of my redshift tasks, not just some of them.

I ran this against the current unit tests for the redshift files and everything passed. If anything else is needed, please let me know.

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -21,11 +21,6 @@ import logging
 import time
 import os
 
-try:
-    from ConfigParser import NoSectionError
-except:
-    from configparser import NoSectionError
-
 import luigi
 from luigi import postgres
 from luigi.contrib import rdbms
@@ -95,10 +90,7 @@ class _CredentialsMixin():
     def _get_configuration_attribute(self, attribute):
         config = luigi.configuration.get_config()
 
-        try:
-            value = config.get(self.configuration_section, attribute)
-        except NoSectionError:
-            value = None
+        value = config.get(self.configuration_section, attribute, default=None)
 
         if not value:
             value = os.environ.get(attribute.upper(), None)
@@ -124,7 +116,9 @@ class _CredentialsMixin():
             )
         else:
             raise NotImplementedError("Missing Credentials. "
-                                      "Override one of the following pairs of auth-args: "
+                                      "Ensure one of the pairs of auth args below are set "
+                                      "in a configuration file, environment variables or by "
+                                      "being overridden in the task: "
                                       "'aws_access_key_id' AND 'aws_secret_access_key' OR "
                                       "'aws_account_id' AND 'aws_arn_role_name'")
 
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -21,6 +21,11 @@ import logging
 import time
 import os
 
+try:
+    from ConfigParser import NoSectionError
+except:
+    from configparser import NoSectionError
+
 import luigi
 from luigi import postgres
 from luigi.contrib import rdbms
@@ -37,60 +42,26 @@ except ImportError:
                    "Will crash at runtime if postgres functionality is used.")
 
 
-class RedshiftTarget(postgres.PostgresTarget):
+class _CredentialsMixin():
     """
-    Target for a resource in Redshift.
-
-    Redshift is similar to postgres with a few adjustments
-    required by redshift.
+    This mixin is used to provide the same credential properties
+    for AWS to all Redshift tasks. It also provides a helper method
+    to generate the credentials string for the task.
     """
-    marker_table = luigi.configuration.get_config().get(
-        'redshift',
-        'marker-table',
-        'table_updates')
-
-    use_db_timestamps = False
-
-
-class S3CopyToTable(rdbms.CopyToTable):
-    """
-    Template task for inserting a data set into Redshift from s3.
-
-    Usage:
-
-    * Subclass and override the required attributes:
-
-      * `host`,
-      * `database`,
-      * `user`,
-      * `password`,
-      * `table`,
-      * `columns`,
-      * `aws_access_key_id`,
-      * `aws_secret_access_key`,
-      * `s3_load_path`.
-    """
-
-    @abc.abstractmethod
-    def s3_load_path(self):
-        """
-        Override to return the load path.
-        """
-        return None
 
     @property
     def aws_access_key_id(self):
         """
         Override to return the key id.
         """
-        return None
+        return self._get_s3_configuration_attribute('aws_access_key_id')
 
     @property
     def aws_secret_access_key(self):
         """
         Override to return the secret access key.
         """
-        return None
+        return self._get_s3_configuration_attribute('aws_secret_access_key')
 
     @property
     def aws_account_id(self):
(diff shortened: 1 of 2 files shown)


luigi/contrib/redshift.py
```diff
@@ -37,6 +42,85 @@
...
+        Override to return the secret access key.
+        """
+        return self._get_s3_configuration_attribute('aws_secret_access_key')
+
+    @property
+    def aws_account_id(self):
+        """
+        Override to return the account id.
+        """
+        return None
```
dlstadther: If `aws_access_key_id` and `aws_secret_access_key` auto fill from config, why doesn't `aws_account_id` and `aws_arn_role_name`?

luigi/contrib/redshift.py
```diff
@@ -37,6 +42,85 @@
...
+        """
+        Override to return the session token.
+        """
+        return None
+
+    def _get_s3_configuration_attribute(self, attribute):
+        config = luigi.configuration.get_config()
+
+        try:
+            value = config.get('s3', attribute)
```
dlstadther: Since these are general AWS creds, I'm not sure it makes sense to restrict their config location to `s3`, particularly if they're being used here for Redshift.

luigi/contrib/redshift.py
```diff
@@ -37,6 +42,85 @@
...
+        except NoSectionError:
+            value = None
+
+        if value is None or value == '':
```
dlstadther: This can just be `if not value:` because `None` and `''` equate to False

luigi/contrib/redshift.py
```diff
@@ -37,6 +42,85 @@
...
+            value = os.environ.get(attribute.upper(), None)
+
+        return value
+
+    def _credentials(self):
+        """
+        Return a credentials string for the provided task. If no valid
```
dlstadther: `Return a credential string` (not credentials)

dlstadther:
I neglected to write a summary review, but in general this is very much something I've wanted to implement myself (just never had the time). All my comments are minor.

Thanks for contributing!

luigi/contrib/redshift.py
zsalzbank: I was really trying to mimic the existing functionality for those fields, but would be happy to add that in if you think it is the right thing to do.

Right now, I'm not to sure that the implementation (current and previous) was even really so correct, because it is using the information from the `s3` section as opposed to a `redshift` section in the configuration file.

luigi/contrib/redshift.py
zsalzbank: Haha - just mentioned that above, but kept the current implementation. Happy to change to a `redshift` section, which I think makes more sense, but it wouldn't be backwards compatible.

luigi/contrib/redshift.py
zsalzbank: I learned something new!

zsalzbank:
Made a couple of changes. Let me know what you'd like to do regarding `s3`/`redshift` section and auto-loading the credentials, and I'll be happy to change it as well.

luigi/contrib/redshift.py
dlstadther: S3 gets its config [here](https://github.com/spotify/luigi/blob/master/luigi/contrib/s3.py#L551-L564). So if we were to change this value (in the cred mixin), someone who uses both s3 and redshift (which is most redshift users) would need to include their creds two places in their config. This isn't ideal.

What are your thoughts on having a named `section` parameter somewhere with a default value of `'redshift'` (where the user can override) which gets passed to `_get_s3_configuration_attribute()`?

luigi/contrib/redshift.py
```diff
@@ -37,6 +42,85 @@
...
+        return None
+
+    @property
+    def aws_session_token(self):
+        """
+        Override to return the session token.
+        """
+        return None
+
+    def _get_s3_configuration_attribute(self, attribute):
```
dlstadther: Also just noticed that the name of this method includes `s3`. Seem odd given explicit redshift usage.

luigi/contrib/redshift.py
dlstadther: I definitely think both role-based and key-based creds should be capable of pulling from config.

I agree that there is likely a better way to approach this implementation. I'll keep thinking on it.

luigi/contrib/redshift.py
zsalzbank: Sounds doable.

zsalzbank:
Made some more changes.

dlstadther:
@zsalzbank Can you confirm that this code works for you?

zsalzbank:
I tested getting the values from the environment variables. I don't use the luigi configuration files in my setup, so I didn't really test that part. But I can if you point me in the right direction for how to use them.

dlstadther:
You can create a `luigi.cfg` file in the same directory as your `.py` luigi files. 

It would look like

```
[redshift]
aws_access_key_id: <your id here>
aws_secret_access_key: <your secret here>

[s3]
<same as redshift above just to check that you can change the section to something other than the default>
```

zsalzbank:
Works for me. I made one minor change to take advantage of the configuration object supporting a default that I noticed when debugging.

If this all looks good, I'll squash all my fixups.

zsalzbank:
I also updated the error message to say where the variables can be set when there are no valid pairs of credentials.

dlstadther:
Let's get at least one more review before worrying about squashing or merging just yet.

If no one else reviews, @Tarrasch are you familiar enough with S3's use of accessing config variables to be able to provide a review here?

dlstadther:
Also, @zsalzbank thanks for such quick turn around edits!

zsalzbank:
No problem at all. I like giving back.

dmohns:
I like this a lot! 
Two questions:
1.) Just double-checking as I am not 100% familiar with Mixins: Will backwards compatibility be preserved? E.g. will we still be able to provide Redshift credentials by overriding not using the Mixin or cfg?
2.) Isn't the comment `You must also override the attributes provided by the CredentialsMixin.` a bit misleading? As far as I understand the main benefit of this PR is that we do *not* have to override attributes but can simply provide entries in the cfg.

zsalzbank:
1. Yes, at least as I understand your question. You can still do `aws_access_key_id = '...'` in the class definition or:

```python
@property
def aws_access_key_id(self):
    return '...'
```

2. Clarified.

luigi/contrib/redshift.py
```diff
@@ -52,7 +138,7 @@ class RedshiftTarget(postgres.PostgresTarget):
     use_db_timestamps = False
 
 
-class S3CopyToTable(rdbms.CopyToTable):
+class S3CopyToTable(rdbms.CopyToTable, _CredentialsMixin):
```
dmohns: Is this the correct order of inheritance? Based on this http://stackoverflow.com/questions/825945/abstract-class-mixin-multiple-inheritance-in-python it should be ordered the other way around. I couldn't test it yet.

zsalzbank: Works for me as it is, but I'm happy to change it if you think that is the right way to go.

zsalzbank: Should I?

dmohns: Unfortunately I don't have the time to test it. I was doing something similar some time back and remember it failed because the ABC's from the base class where still abstract on instantiation. 
As this is just based on my memory your code might be correct already. Maybe someone else can help out?

dlstadther: So the issue identified in that SO question seems to be around the Mixin implementing an abstractmethod which is supplied by an earlier inherited class.

Perhaps it's worth swapping the order of inheritance for a future case where `_CredentialsMixin` overrides an abstractmethod which is added to `rdbms.CopyToTable`.

So long as the overall functionality still works, it doesn't hurt anything.

THanks,

dmohns: Let me re-phrase my concern: From my understanding the current implementation should not work unless you override the abstract properties in sub class. (Which is exactly what you **not** want to do, when you try to read credentials from `luigi.cfg`.)
Currently this is not covered by any unittest, as all the test subclass and override. So, my pragmatic suggestion would be: Can we create a test for reading credentials from `luigi.cfg` (or even ENV) ? If this test goes through successfully my comment is decrepit.

zsalzbank: Sounds doable. Might take me a little bit though, as I have to focus on some other stuff right now.

dlstadther:
@zsalzbank Any update here? Really looking forward to merging and using this PR!

zsalzbank:
Sorry, been busy, but I'll try and get to it next week.

zsalzbank:
Added test cases - pass locally with the existing mixin order, but will confirm that is the case with the full CI suite.

luigi/contrib/redshift.py
dlstadther: Where did we land on this thread? Do your tests cover this (potential) issue?

zsalzbank: This is what my most recent commit is for. I test loading the credentials from the different available sources. Let me know if you want to see other tests.

dlstadther:
@dmohns Could you also review again? If approved, i'll merge.

luigi/contrib/redshift.py
dmohns: This looks good. Just to clarify: What I missed earlier is the fact that within this PR the `aws` credentials are no longer abstract. Potentially missing credentials are now handled explicitly in the mixin. Therefore order of inheritance is indeed irrelevant here. 👍

zsalzbank:
Thanks!

PR #2069, opened 2017-03-23
Remove assert that verifies if tables were marked 

cabouffard:
A summary of this would be that having multiple workers and multiple
insert/select on the same table causes a lot of issues. Please refer
yourself to the issue below to have a better idea.

fixes glossier#2


cabouffard:
Wrong repo. -- To delete.

PR #2072, opened 2017-03-24
Update format of luigi users and their public content

dlstadther:
## Description

Separate company sites from public luigi content. Date of publication is provided is I could find it. Those with external content were included higher in the list.

## Motivation and Context

See [PR comment](https://github.com/spotify/luigi/pull/2056#issuecomment-289057904)

Looks fine in [github](https://github.com/dlstadther/luigi/blob/readme_update/README.rst#who-uses-luigi)


erikbern:
great!!

PR #2076, opened 2017-03-27
Split up the list of users in two lists

erikbern:
## Description

Break up the long list of users in two lists, one for blogs/presentation, one without

## Motivation and Context

encourage people to add presentations/blogs, but also not lose out on companies that don't


Tarrasch:
Great job you two!

PR #2077, opened 2017-03-28
luigi.worker.Worker(retry_external_tasks=True) has no effect

kalvdans:
## Description
Use the retry_external_tasks parameter local to the worker instead of the global settings.

## Motivation and Context
The change is required when starting a worker from Python. It is not needed when starting a worker using the luigi binary.

I have included unit tests. The test don't run to completion without the change to the code.

### DIFF
--- a/test/worker_external_task_test.py
+++ b/test/worker_external_task_test.py
@@ -147,6 +147,27 @@ class WorkerExternalTaskTest(unittest.TestCase):
 
         self.assertGreaterEqual(test_task.dependency.times_called, 5)
 
+    def test_external_dependency_bare(self):
+        """
+        Test ExternalTask without altering global settings.
+        """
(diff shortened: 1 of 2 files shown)


test/worker_external_task_test.py
```diff
@@ -147,6 +147,27 @@ def test_external_dependency_worker_is_patient(self):
 
         self.assertGreaterEqual(test_task.dependency.times_called, 5)
 
+    def test_external_dependency_bare(self):
+        """
+        Test ExternalTask without altering global settings.
+        """
+        assert luigi.worker.worker().retry_external_tasks is False
```
Tarrasch: Awesome assertion :sunglasses:

Tarrasch:
Thanks! :)

PR #2078, opened 2017-03-28
Avoid warnings in unit tests.

kalvdans:
This change removes some warnings that appear when running the unit tests:
<pre>
luigi/luigi/file.py:26: DeprecationWarning: luigi.file module has been moved to luigi.local_target
luigi/luigi/parameter.py:259: UserWarning: Parameter 1 is not of type string.
luigi/luigi/parameter.py:259: UserWarning: Parameter 10 is not of type string.
</pre>

It also allows the luigi binary to set self.complete_after to integer type, when using the --complete-after command-line flag.

### DIFF
--- a/test/worker_external_task_test.py
+++ b/test/worker_external_task_test.py
@@ -13,7 +13,7 @@
 # the License.
 
 import luigi
-from luigi.file import LocalTarget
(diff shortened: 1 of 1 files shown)


Tarrasch:
Thanks! :)

PR #2079, opened 2017-03-28
Defer S3 Client instantiation until needed

### LINKED ISSUE #2048: Delay instantiating S3Client when generating S3Targets
I was profiling luigi code that had recursive dependencies (i.e., Task A grabs outputs from Task B which grabs outputs from Task C, etc) I noticed that there was a big performance hit when generating too many S3Targets (because boto has to hit the filesystem to check all config files and then a connection object [I believe] is generated to AWS).  This also means that you open a ton of file descriptors (so you can hit a ulimit if you fork too many times ).

I am happy to put up a PR implementing this change, but wanted to see if anybody had feedback on this before putting up a PR.

Rather than instantiating `S3Client` in the `__init__` method, I'd like to suggest changing `self.client` to a cached property that only instantiates the client at the moment we need it.

Pros:

- outputs() method is faster
- prereq for using a connection pool for communicating with S3
- caching S3Targets doesn't surprisingly leak file descriptors

Cons:

- validation errors on client kwargs only come out once you start trying to access targets, (currently you see it as soon as S3Target is instantiated)

jtratner:
## Description
Delays instantiation of the S3Client until we need to contact S3, speeding the `output()` method and allowing nested task graphs to use fewer connections.  Fixes #2048.

## Motivation and Context
Previously the S3 Client was instantiated in the `__init__` method of
S3Target, which meant that if you wanted to traverse a large task
graph using `output()`, quickly creating and throwing away targets, you'd end up
incurring a lot of wasted connections and slowness.

Now we create the client only when we want to make a remote request.

This does not change any of the semantics of how the target is
instantiated (still can override `s3` property and the `Key` property),
except that making an STS connection is also deferred until the moment
it's actually used.

I tested this with our workflows (which use Openstack Swift as an S3Target) and they continued to work.  I also assume unit tests cover this code as well. (had some issues running the test suite locally).  I also checked that connections were not being generated solely from instantiating S3Targets.

### DIFF
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -124,6 +124,15 @@ class S3Client(FileSystem):
                                                    **options)
         return self._s3
 
+    @s3.setter
+    def s3(self, value):
+        self._s3 = value
+
+    @s3.deleter
+    def s3(self):
+        if hasattr(self, '_s3'):
+            del self._s3
+
     def exists(self, path):
         """
         Does provided path exist on S3?
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -73,6 +73,8 @@ class S3Client(FileSystem):
     boto-powered S3 client.
     """
 
+    _s3 = None
+
     def __init__(self, aws_access_key_id=None, aws_secret_access_key=None,
                  **kwargs):
         from boto.s3.key import Key
@@ -94,7 +96,7 @@ class S3Client(FileSystem):
 
         options = dict(self._options)
 
-        if getattr(self, '_s3', None):
+        if self._s3:
             return self._s3
 
         aws_access_key_id = options.get('aws_access_key_id')
@@ -127,11 +129,6 @@ class S3Client(FileSystem):
     @s3.setter
     def s3(self, value):
         self._s3 = value
-
-    @s3.deleter
-    def s3(self):
-        if hasattr(self, '_s3'):
-            del self._s3
 
     def exists(self, path):
         """
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -75,13 +75,31 @@ class S3Client(FileSystem):
 
     def __init__(self, aws_access_key_id=None, aws_secret_access_key=None,
                  **kwargs):
+        from boto.s3.key import Key
+        options = self._get_s3_config()
+        options.update(kwargs)
+        if aws_access_key_id:
+            options['aws_access_key_id'] = aws_access_key_id
+        if aws_secret_access_key:
+            options['aws_secret_access_key'] = aws_secret_access_key
+
+        self.Key = Key
+        self._options = options
+
+    @property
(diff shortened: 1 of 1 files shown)


luigi/contrib/s3.py
```diff
@@ -75,13 +75,31 @@ class S3Client(FileSystem):
...
         # only import boto when needed to allow top-lvl s3 module import
         import boto
         import boto.s3.connection
-        from boto.s3.key import Key
 
-        options = self._get_s3_config()
-        options.update(kwargs)
+        options = dict(self._options)
+
+        if getattr(self, '_s3', None):
```
kalvdans: Please use self._s3 = None in the constructor, so that this check will be simply `if self._s3: return self._s3`

luigi/contrib/s3.py
```diff
@@ -97,22 +115,23 @@ def __init__(self, aws_access_key_id=None, aws_secret_access_key=None,
...
+                                                   aws_secret_access_key,
+                                                   security_token=aws_session_token,
+                                                   **options)
+        return self._s3
+
+    @s3.setter
+    def s3(self, value):
+        self._s3 = value
+
+    @s3.deleter
```
kalvdans: No need to provide a deleter, since no code in luigi deletes a member from an instance.

luigi/contrib/s3.py
jtratner: sure - do you prefer setting in `__init__` to setting `_s3 = None` in the class definition?

kalvdans: I personally would prefer to initialize instance variables in __init__, but it is not a strong opinion.

luigi/contrib/s3.py
```diff
@@ -97,22 +117,18 @@ def __init__(self, aws_access_key_id=None, aws_secret_access_key=None,
...
-        self.Key = Key
+        self._s3 = boto.s3.connection.S3Connection(aws_access_key_id,
+                                                   aws_secret_access_key,
+                                                   security_token=aws_session_token,
+                                                   **options)
+        return self._s3
+
+    @s3.setter
+    def s3(self, value):
+        self._s3 = value
```
kalvdans: You can as well remove the setter, since the `s3` member is not assigned to anywhere in the code (as far as I can see). With this change, I am okay with the changes, but I am not a core developer and not using the s3 stuff.

dlstadther:
@jtratner Could you check on your build failures? Thanks!

luigi/contrib/s3.py
jtratner: I'd like to leave the setter in because it makes it easier to test out / mock interactions here

kalvdans: Ok, that is fine with me. Personally I keep my good-to-have-in-the-future code in my local repo only.

jtratner:
Thanks for reviewing @kalvdans!

dlstadther:
Thanks @jtratner !

PR #2082, opened 2017-03-31
Document that Parameter() now must be a string.

kalvdans:
In #1607 it was decided that Parameter must be serializable and thus can no longer be an untyped object. This pull request changes the documentation to reflect this.


Tarrasch:
Great!

Issue #2080, opened 2017-03-28
Error with redshift does_table_exist()


dlstadther:
Redshift schema and table names are case insensitive (`select * from TableTest;` == `select * from tabletest;` == `select * from TaBlEtEsT;`). However, both queries in [does_table_exist](https://github.com/spotify/luigi/blob/master/luigi/contrib/redshift.py#L327-L346) compare case sensitive table names.

These should be updated to case insensitive comparisons.

The `%s` string parameters need to be surrounded with `lower()`

It may be a few days before i can submit the PR, so feel free to beat me to it.

Thanks,

dlstadther:
Resolved.

PR #2086, opened 2017-04-05
Update check for table existence to be case insensitive

dlstadther:
## Description

Redshift schema/table/column names are case-insensitive. So if you create table `MyTable` in schema `MySchema`, redshift saves those values in `pg_table_def` and `information_schema.tables` in all lowercase. Therefore, the check for table existence in `does_table_exist()` will fail.

This PR just surrounds the table and schema parameters in `lower()`.

From [Redshift documentation](http://docs.aws.amazon.com/redshift/latest/dg/r_names.html):

> ASCII letters in standard and delimited identifiers are case-insensitive and are folded to lower case.

## Motivation and Context

Resolves Issue #2080 

Works for me.

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -332,11 +332,11 @@ class S3CopyToTable(rdbms.CopyToTable):
         if '.' in self.table:
             query = ("select 1 as table_exists "
                      "from information_schema.tables "
-                     "where table_schema = %s and table_name = %s limit 1")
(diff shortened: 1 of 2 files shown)


dlstadther:
@Tarrasch All tests passed except the two which haven't been running for anyone.

Tarrasch:
Cool, also feel free self-merge these changes to an area where you have expertise.

dlstadther:
@Tarrasch I would have, but my write access to this repo was removed.

Thanks for merging!

PR #2089, opened 2017-04-10
Change ambiguous name 'Config'

emmjaykay:
Luigi on Python 3.6.0 will throw an exception about the 'Config' name.

## Description
Change 'Config' to 'MyConfig'

## Motivation and Context
While looking at dynamic requirements, I ran this code and found it wouldn't run unless I changed the name of the 'Config' class.

I ran the code with `python dynamic_requirements.py Dynamic --local-scheduler --workers=10` and it looks like it ran fine.

### DIFF
--- a/examples/dynamic_requirements.py
+++ b/examples/dynamic_requirements.py
@@ -21,7 +21,7 @@ import time
 import luigi
 
 
-class Config(luigi.Task):
+class MyConfig(luigi.Task):
     seed = luigi.IntParameter()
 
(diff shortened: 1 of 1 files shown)


dlstadther:
@Tarrasch Looks like a Travis error "The job exceeded the maximum time limit for jobs, and has been terminated.". Possibly just needs to be restarted

Tarrasch:
Sttill failing..., I bet something else is wrong :-/

Issue #2035, opened 2017-02-20
force reschedule from web app


FxIII:
It would be nice to allow failed tasks to be rescheduled by user interface especially on workflows executed in keep-alive mode.

FxIII:
I tried to exploit the add_task function, but it asserts for a non-None worker and I don't want to pass a fake one.
I added a @rpc_method force_reschedule on Scheduler that basically does the reschedule stuff that add_task does; I don't know if it is the way to go.


with regard to the graphical interface, while I was able to add a new reschedule button near the show_error one, I can't figure out how to update the status of the selected entry when the server responds, probably I missed something...

Tarrasch:
You have a good use case, but I'm afraid "reschedule" could be misinterpreted. If the worker disconnect, the task won't be run anywhere. Instead. I think you could:

 * Just configure your worker to `--worker-keep-alive`, so it doesn't quit and runs the task after the cooldown.
 * Just lower the time-out for your sensitive task (or all tasks globally if you desire)
 * You could add something like you say here, but it should rather be called "Forgive failures", which is more clear than "reschedule". It basically turns a red task into yellow, failed into pending. (this one has to be implemented, similar to what you say here :))

Will this work?

FxIII:
I totally agree (I don't know if is better to forgive or to forget :)), and this will work.
I can't figure out how to update the presentation (correctly) in the user interface from red to yellow once I have the ok from the scheduler.

dckc:
bump. this looks ready to close as not-a-bug or whatever.

(are these triage comments helpful? Is there another way to help?)

Tarrasch:
Yep, seems like #2091 fixed this!

It definetly is @dckc :)

PR #2091, opened 2017-04-15
Forgive Failures

FxIII:
## Description
Let the user to forgive failures for failed tasks using the web interface.

## Motivation and Context
(issue #2035)
Failed tasks need to stay in their failed status for a while before being allowed to be run again if the workers are kept alive. Adding the forgive_failures feature to the scheduler, the user has the opportunity to acknowledge the error, perhaps fix it, and let the task to be run again.

I have extended the scheduler api unit test to check that forgive_failure allows a failed task to be retried and that is basically idempotent


FxIII:
schermata 2017-04-16 alle 11 43 47

Tarrasch:
Hehe, gotta love that cute ambulance icon. Anyway, the icon only appears on failed tasks, right? This looks goods to me. If nobody objects we can merge.

Tarrasch:
@FxIII, I see you find errors, please also add test cases fixing stuff. In addition, maybe you want to run this version locally for 1 week and then we merge?

FxIII:
@Tarrasch I simply removed an accidental paste fallen in a sector of code that is unrelated to my changes. I noticed that by inspecting the changeset (I try to keep it tiny for that reason). 
I didn't notice that in production, because it happen to be related to the d3 visualization (the way errors are shown) and we didn't use d3 visualization that much. 
The fact that such error didn't broke the regression test probably could fit the scope of another issue. Right now I can't figure out how to write down such a test (one should be able to simulate a click to a failed task in the d3 graph through phantomjs).

Tarrasch:
@FxIII, okay. Thanks for looking into this. I'm ready to merge this. Is this PR ready to merge don your side too?

FxIII:
@Tarrasch it is ok for me. Thanks

mbelmadani:
How do I enable this feature? And is it only available in the github version of luigi or would the pip package also have it? Also would be nice to have some documentation on this too if this a configuration setting.

Looks like a great feature otherwise!

Tarrasch:
It's auto-enabled but it's not released. Let's see if I can release it now.

Tarrasch:
@mbelmadani it's now on pypi :)

bmaggard:
My 2 cent suggestion (sorry it's late, I just stumbled on this feature): rename to "Retry" or "Reschedule".  To me, "Forgive failures" seems more like "mark Done"  (we are circumventing "retry_delay", after all).

Tarrasch:
@bmaggard yes! That's a great idea to avoid introducing new terminology like "forgiving" since we already have "retry" meaning this. Can you send a PR with this small change?

Issue #2039, opened 2017-02-24
Raise FileExistsError when running multiple workers


elnikkis:
When running multiple workers, mkdir in FileSystemTarget(LocalTarget).temporary_path() raises FileExistsError sometimes.

My output files of parallel tasks have same parent directory.
It seems that "raise_if_exists=False" of LocalFileSystem.mkdir does not work properly when running multiple workers.

test_luigi_bug.py:

```import os.path
from subprocess import run
import luigi


class TestTask1(luigi.Task):
    param = luigi.IntParameter()

    def requires(self):
        return None

    def output(self):
        return luigi.LocalTarget(os.path.join('data', 'test_data', 'task1', 'sample{}.txt'.format(self.param)))

    def run(self):
        with self.output().temporary_path() as temp_output_path:
            run(['touch', temp_output_path])


class TestTask2(luigi.WrapperTask):
    def requires(self):
        return [TestTask1(param=i) for i in range(10)]
```

I execuded this command:

`PYTHONPATH='.' luigi --local-scheduler --module test_luigi_bug TestTask2 --workers 4 > luigi_bug.txt 2>&1`

and I got this output:
[luigi_bug.txt](https://github.com/spotify/luigi/files/798728/luigi_bug.txt)

kzuberi:
I am also encountering this (luigi 2.6.1), looks like a race [here](https://github.com/spotify/luigi/blob/d2694c9c07f8601deb0716228ea339137bfbc8f4/luigi/local_target.py#L74) when the earlier check for path existence had returned false but another task creates the path first, os.makedirs() call fails due to the leaf folder existing.

Tarrasch:
Ah, yes, the bad things with non-atomicity. Luckily this isn't super-severe as next luigi run the tasks will be retried. Anyhow, if we can make it atomic it would be great. :)

kalvdans:
In Python 3.2 and above, we can use the exist_ok=True flag to os.makedirs(). In Python 2.7 I guess we have to catch the error, check errno, and reraise if errno != EEXIST.

PR #2093, opened 2017-04-20
Fix os.makedirs issue in multithreading (#2039)

elnikkis:
## Description

There is a problem that os.makedirs is not atomic in multithreading.

## Motivation and Context

This problem is reported in issue #2039.

Since exist_ok was added in Python 3.2 or later, it cannot be used in Python 2.7, so we need to check OSError.errno that we caught.
But in Python 3.x, os.makedirs raise FileExistsError instead of OSError.
FileExistsError is a subclass of OSError but does not have errno attribute.

In my environment, it seems to work well.

### DIFF
--- a/luigi/local_target.py
+++ b/luigi/local_target.py
@@ -71,12 +71,17 @@ class LocalFileSystem(FileSystem):
                 return
 
         if parents:
-            error_to_catch = getattr(__builtins__, 'FileExistsError', OSError)
+            # for Python 2 compatibility
+            try:
+                FileNotExistsError
+            except NameError:
+                FileNotExistsError = OSError
+
             try:
                 os.makedirs(path)
-            except error_to_catch as err:
+            except FileNotExistsError as err:
                 # somebody already created the path
-                if getattr(err, 'errno', 0) != errno.EEXIST:
+                if err.errno != errno.EEXIST:
                     raise
         else:
             if not os.path.exists(os.path.dirname(path)):
--- a/luigi/local_target.py
+++ b/luigi/local_target.py
@@ -71,7 +71,13 @@ class LocalFileSystem(FileSystem):
                 return
 
         if parents:
-            os.makedirs(path)
+            error_to_catch = getattr(__builtins__, 'FileExistsError', OSError)
+            try:
+                os.makedirs(path)
+            except error_to_catch as err:
+                # somebody already created the path
+                if getattr(err, 'errno', 0) != errno.EEXIST:
+                    raise
         else:
             if not os.path.exists(os.path.dirname(path)):
                 raise MissingParentDirectory()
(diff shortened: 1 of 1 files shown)


kalvdans:
In Python 3.5 on Linux, FileExistsError has the errno attribute. So you can simplify the logic to always catch OSError and unconditionally access err.errno attribute.

Test performed in Ipython:
<pre>
n [3]: try: os.makedirs("/tmp")
   ...: except Exception as e:
   ...:     a = e
   ...:     

In [4]: a
Out[4]: FileExistsError(17, 'File exists')

In [5]: a.errno
Out[5]: 17
</pre>

Tarrasch:
Not sure if this is the same thing, but there's a similar compatibility-hack here: https://github.com/spotify/luigi/blob/master/luigi/lock.py#L125-L129

kalvdans:
Yes, that hack should also be changed to check errno, to correctly propagate other OSError errors in Python 2.

elnikkis:
FileExistsError has a errno attribute.
It was my missunderstand.
Thank you.

luigi/local_target.py
```diff
@@ -71,7 +71,18 @@ def mkdir(self, path, parents=True, raise_if_exists=False):
                 return
 
         if parents:
-            os.makedirs(path)
+            # for Python 2 compatibility
+            try:
+                FileNotExistsError
+            except NameError:
+                FileNotExistsError = OSError
```
kalvdans: To reduce complexity, please use OSError on Python3 as well.

luigi/local_target.py
```diff
@@ -71,7 +71,18 @@ def mkdir(self, path, parents=True, raise_if_exists=False):
...
+
+            try:
+                os.makedirs(path)
+            except FileNotExistsError as err:
+                # somebody already created the path
```
kalvdans: This comment is very helpful explaining why we can ignore the error. Thanks for adding it.

luigi/local_target.py
```diff
@@ -71,7 +71,12 @@ def mkdir(self, path, parents=True, raise_if_exists=False):
                 return
 
         if parents:
-            os.makedirs(path)
+            try:
+                os.makedirs(path)
+            except OSError as err:
+                # somebody already created the path
+                if err.errno != errno.EEXIST:
```
dlstadther: Typo on `EEXIST`. Should be `EXIST`

elnikkis: I think it's correct. https://docs.python.org/3.6/library/errno.html#errno.EEXIST

dlstadther: Well I didn't expect that. I recant.

dlstadther:
If no one else reviews, i'll merge this week.

dlstadther:
Sorry for such delay @elnikkis ! Thanks for your first Luigi contribution!

PR #2099, opened 2017-04-26
Updated Kubernetes task to recieve KubeConfig object

eamontaaffe:
## Description

I have updated the `KubernetesJobTask` in `luigi.contrib.kubernetes` to receive a pykube KubeConfig object to instantiate the task.

## Motivation and Context

Currently, there the only two options to configure a kubernetes task is by using a config file or using service accounts. 

The `KubernetesJobTask` is a wrapper around the pykube library. Within this library, `KubeConfig` objects are used to configure the client in a number of different ways. By allowing the `KubernetesJobTask` to receive `KubeConfig` objects for initialisation, we can expand the number of use cases dramatically.

In my current use case, I want to configure Kubernetes using environment variables rather than a file. Using this patch I am able to instantiate a `KubeConfig` object using envirnoment variables then pass it to the KubernetesTask.

### DIFF
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -68,7 +68,9 @@ class KubernetesJobTask(luigi.Task):
     def _init_kubernetes(self):
         self.__logger = logger
         self.__logger.debug("Kubernetes auth method: " + self.auth_method)
-        if(self.auth_method == "kubeconfig"):
+        if(self.kube_config.__class__ == KubeConfig):
+            self.__kube_api = HTTPClient(self.kube_config)
+        elif(self.auth_method == "kubeconfig"):
             self.__kube_api = HTTPClient(KubeConfig.from_file(self.kubeconfig_path))
         elif(self.auth_method == "service-account"):
             self.__kube_api = HTTPClient(KubeConfig.from_service_account())


dlstadther:
@eamontaaffe Could you get some other Luigi K8s users or contributors to review? We appreciate the PR!

dlstadther:
@mcapuccini As the only other Luigi K8s contributor, could you review? Thanks in advance!

mcapuccini:
@eamontaaffe I think that the Luigi interface shouldn't be tide to a specific python library, so in the future if we need to replace pykube with something else is going to be much smoother. In my opinion you should instead add an additional `auth_method` that tells the constructor to read the authentication variables from the environment.

eamontaaffe:
@mcapuccini: I agree, it isn't ideal to have it tied to a specific library. Especially since there are now other options such as the officially supported [client-python](https://github.com/kubernetes-incubator/client-python) package.

However, this current implementation is very much tied to pykube already with bits like this:

https://github.com/kubernetes-incubator/client-python

To be honest, I'm not really too worried about merging this pr anymore.

dlstadther:
@eamontaaffe I apologize for the delay here with reviewing and merging. If this feature is still desired with the current pykube dependency, I can merge.

mcapuccini:
@eamontaaffe In fact I am talking about the interface, not about the implementation, that can be changed later on. By interface I mean the constructor signature, that would have an addition parameter (at least this is my understanding of it). If you don't need it anymore I would discourage merging as it is now.

dlstadther:
@eamontaaffe Do i understanding that there isn't a desire to merge this anymore?

PR #2107, opened 2017-05-04
Adding an optional `name` parameter in Parameter.serialize in order to make warnings more helpful

Lucas-C:
## Description
I added an optional `name` Parameter in https://github.com/voyages-sncf-technologies/luigi/blob/master/luigi/parameter.py#L247 in order to improve warnings.
I also passed this new parameter in many place where `Parameter.serialize` is invoked (not all, only when possible).
Finally I updated existing tests and added new ones.

## Motivation and Context
When this `name` parameter is passed to `Parameter.serialize`, warnings raised will now include the parameter name, helping to diagnose the issue.
This is especially helpful in https://github.com/spotify/luigi/blob/master/luigi/task.py#L473 this is called to validate `luigi.cfg`.

I have included unit tests and I used the patched version in our own luigi pipelines.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -252,9 +252,13 @@ class Parameter(object):
 
         :param x: the value to serialize.
         """
-        if not isinstance(x, six.string_types) and self.__class__ == Parameter:
-            warnings.warn("Parameter {0} is not of type string.".format(str(x)))
         return str(x)
+
+    def _warn_on_wrong_param_type(self, param_name, param_value):
+        if self.__class__ != Parameter:
+            return
+        if not isinstance(param_value, six.string_types):
+            warnings.warn('Parameter "{}" with value "{}" is not of type string.'.format(param_name, param_value))
 
     def normalize(self, x):
         """
@@ -692,8 +696,6 @@ class TimeDeltaParameter(Parameter):
 
         :param x: the value to serialize.
         """
-        if not isinstance(x, datetime.timedelta) and self.__class__ == TimeDeltaParameter:
-            warnings.warn("Parameter {0} is not of type timedelta.".format(str(x)))
         weeks = x.days // 7
         days = x.days % 7
         hours = x.seconds // 3600
@@ -701,6 +703,12 @@ class TimeDeltaParameter(Parameter):
         seconds = (x.seconds % 3600) % 60
         result = "{} w {} d {} h {} m {} s".format(weeks, days, hours, minutes, seconds)
         return result
+
+    def _warn_on_wrong_param_type(self, param_name, param_value):
+        if self.__class__ != TimeDeltaParameter:
+            return
+        if not isinstance(param_value, datetime.timedelta):
+            warnings.warn('Parameter "{}" with value "{}" is not of type timedelta.'.format(param_name, param_value))
 
 
 class TaskParameter(Parameter):
--- a/test/task_test.py
+++ b/test/task_test.py
@@ -17,6 +17,8 @@
 
 import doctest
 import pickle
+import six
+import warnings
 
 from helpers import unittest, LuigiTestCase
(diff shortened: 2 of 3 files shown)


README.rst
```diff
@@ -155,6 +155,7 @@ Some more companies are using Luigi but haven't had a chance yet to write about
 * `OAO <https://adops.com/>`_
 * `Grovo <https://grovo.com/>`_
 * `Weebly <https://www.weebly.com/>`_
+* `Voyages-Sncf <https://voyages-sncf.com>`_
```
dlstadther: Could you separate this into a independent PR? Thanks

Lucas-C: Sure. Done

luigi/parameter.py
```diff
@@ -244,16 +244,17 @@ def _parse_list(self, xs):
...
         Opposite of :py:meth:`parse`.
 
         Converts the value ``x`` to a string.
 
         :param x: the value to serialize.
+        :param name: the parameter name, for reporting purposes in case of an error.
         """
         if not isinstance(x, six.string_types) and self.__class__ == Parameter:
-            warnings.warn("Parameter {0} is not of type string.".format(str(x)))
+            warnings.warn("Parameter {0} is not of type string.".format(name or x))
```
Tarrasch: Is this the only place where you use `name`?

Tarrasch:
Just to clarify, I added a different way of solving this, because I think this is a great idea, and I could have used it myself many times. But I think this implementation could be better, and that in the long run it saves time for everyone: Users, you (the contributor) and me (the maintainer).

luigi/parameter.py
Tarrasch: also, was removing `str()` on purpose? :)

Lucas-C: In this function yes. And also in the overriden version in TimeDeltaParameter.serialize

Lucas-C: Yes, the removal was on purpose, I think it was unneeded

Lucas-C:
Alright, this is a good suggestion as it will reduce the impact of this change.
However IHMO adding a new function slightly adds in complexity.
I'll update this PR tomorrow.

Lucas-C:
New version pushed. It is definitively cleaner I think.

Lucas-C:
Tested with:
```
tox -e py34-nonhdfs --run-command "python test/runtests.py -v test/task_test.py"
```

test/task_test.py
```diff
@@ -106,6 +109,28 @@ def test_no_unpicklable_properties(self):
         message = task.set_status_message('message')
         self.assertEqual(message, 'message')
 
+    def test_no_warn_if_param_types_ok(self):
+        with warnings.catch_warnings(record=True) as w:
+            DummyTask(**DUMMY_TASK_OK_PARAMS)
+        self.assertEqual(len(w), 0, msg='No warning should be raised when correct parameter types are used')
+
+    if six.PY3:  # assertWarnsRegex was introduced in Python 3.2
```
Tarrasch: Nice :)

test/task_test.py
```diff
@@ -106,6 +109,28 @@ def test_no_unpicklable_properties(self):
...
+        def test_warn_on_non_str_param(self):
+            params = dict(**DUMMY_TASK_OK_PARAMS)
```
Tarrasch: Can't you just use `DUMMY_TASK_OK_PARAMS` directly? :)

Lucas-C: This test checks that an integer value (42) given to the "param" parameter (that expects a string) triggers a warning.
Hence I need to modify a copy of `DUMMY_TASK_OK_PARAMS`.

Tarrasch:
Can you fix the flake issue please? Then I think we can merge. :)

Lucas-C:
Done

PR #2108, opened 2017-05-04
Progress bar in the status message modal + auto-refresh of the modal

fvinas:
## Description
Added the ability to display a progress bar in the status message modal using `task.set_progress_percentage()` + automatically refreshing the modal (for both status message and the progress bar) to follow progress dynamically while it's open.

image

## Motivation and Context

- More visual & practical way to track the progress of running tasks **inside Luigi** (i.e. not using an external tracking URL). Status message can now be dedicated to text messages instead of progress tracking.

- Brings dynamic view on a running task when the status message modal is open (elsewhere the UI remains static for now)

Included unit tests, seems to work well in my environment.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -275,7 +275,7 @@ class OrderedSet(collections.MutableSet):
 
 class Task(object):
     def __init__(self, task_id, status, deps, resources=None, priority=0, family='', module=None,
-                 params=None, tracking_url=None, status_message=None, retry_policy='notoptional'):
+                 params=None, tracking_url=None, status_message=None, progress_percentage=None, retry_policy='notoptional'):
         self.id = task_id
         self.stakeholders = set()  # workers ids that are somehow related to this task (i.e. don't prune while any of these workers are still active)
         self.workers = OrderedSet()  # workers ids that can perform task - task is 'BROKEN' if none of these workers are active
@@ -301,6 +301,7 @@ class Task(object):
         self.failures = Failures(self.retry_policy.disable_window)
         self.tracking_url = tracking_url
         self.status_message = status_message
+        self.progress_percentage = progress_percentage
         self.scheduler_disable_time = None
         self.runnable = False
         self.batchable = False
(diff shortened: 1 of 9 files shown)


test/task_progress_percentage_test.py
```diff
@@ -0,0 +1,41 @@
...
+import luigi.scheduler
+import luigi.worker
+
+luigi.notifications.DEBUG = True
+
+
+class TaskProgressPercentageTest(LuigiTestCase):
+
+    def test_run(self):
+        percentage = 30
```
Tarrasch: I think avoiding a variable makes it more readable. :)

test/task_progress_percentage_test.py
```diff
@@ -0,0 +1,41 @@
...
+# limitations under the License.
+#
+
+from helpers import LuigiTestCase
+
+import luigi
+import luigi.scheduler
+import luigi.worker
+
+luigi.notifications.DEBUG = True
```
Tarrasch: Remove this line

luigi/scheduler.py
```diff
@@ -1454,6 +1456,23 @@ def get_task_status_message(self, task_id):
...
+                for batch_task in self._state.get_batch_running_tasks(task.batch_id):
+                    batch_task.progress_percentage = progress_percentage
+
+    @rpc_method()
+    def get_task_progress_percentage(self, task_id):
+        if self._state.has_task(task_id):
+            task = self._state.get_task(task_id)
+            return {"taskId": task_id, "progressPercentage": task.progress_percentage}
+        else:
+            return {"taskId": task_id, "progressPercentage": None}
```
Tarrasch: Wait a second. Why don't we return `None`/`{}` here or some sort of failure indication, since the task id didn't exist right?

test/task_progress_percentage_test.py
fvinas: Hi @Tarrasch, yes we're using it in production for a while and it's pretty smooth.
I try to take some time over the week-end to do the changes and I let you know.

dlstadther:
@fvinas Any update here? Or response to @Tarrasch 's comment/question?

Tarrasch:
Thanks @fvinas, I reworded the final commit a bit. I wasn't sure what "modal" means here so I removed it before merging. :)

Anyway expect a release with this patch soon!

PR #2109, opened 2017-05-08
Allow callable default parameter values

lsglick:
## Description
Allow a the default to be specified for a `Parameter` as a `default_callable`. The callable will be executed and the value locked in when a task is initialized, but initializing another task could yield a different value.

## Motivation and Context
Prior to this change, default parameter values are only evaluated when a task class is initially loaded. For defaults such as `datetime.datetime.utcnow()`, this might not always provide the desired flexibility, since running the task again would result in the same datetime value.

In our case, this is a problem because we trigger tasks in a loop with `luigi.build()`, but often find ourselves duplicating parameter `default` definitions in our executor script so we can recalculate it every time and match the behavior of running on the command line. It would be nice to avoid the duplication.

I imagine there may also be cases where a long running worker wouldn't want to be stuck with the same default value.

I have included unit tests.

### DIFF
--- a/test/parameter_test.py
+++ b/test/parameter_test.py
@@ -321,9 +321,8 @@ class ParameterTest(LuigiTestCase):
         self.assertEqual(CallableDefaultParameterTask().x, 'abc')
 
     def test_both_default_and_default_callable_param(self):
-        self.assertRaises(
-            ParameterException,
-            lambda: luigi.parameter.Parameter(default='abc', default_callable=lambda: 'xyz'))
+        with self.assertRaises(ParameterException):
+            luigi.parameter.Parameter(default='abc', default_callable=lambda: 'xyz')
 
 
 class TestParametersHashability(LuigiTestCase):
--- a/test/parameter_test.py
+++ b/test/parameter_test.py
@@ -158,6 +158,19 @@ class ParameterTest(LuigiTestCase):
     def test_default_param(self):
         self.assertEqual(WithDefault().x, 'xyz')
 
+    def test_default_callable_param(self):
+        default_value = 'xyz'
+
+        class CallableDefaultParameterTask(luigi.Task):
+            x = luigi.Parameter(default=lambda: default_value)
+
+        task = CallableDefaultParameterTask()
+        self.assertEqual(task.x, 'xyz')
+
+        default_value = 'abc'
+        self.assertEqual(task.x, 'xyz')
+        self.assertEqual(CallableDefaultParameterTask().x, 'abc')
+
     def test_missing_param(self):
         def create_a():
             return A()
--- a/test/parameter_test.py
+++ b/test/parameter_test.py
@@ -158,19 +158,6 @@ class ParameterTest(LuigiTestCase):
     def test_default_param(self):
         self.assertEqual(WithDefault().x, 'xyz')
 
-    def test_default_callable_param(self):
-        default_value = 'xyz'
-
-        class CallableDefaultParameterTask(luigi.Task):
-            x = luigi.Parameter(default=lambda: default_value)
-
-        task = CallableDefaultParameterTask()
-        self.assertEqual(task.x, 'xyz')
-
-        default_value = 'abc'
-        self.assertEqual(task.x, 'xyz')
-        self.assertEqual(CallableDefaultParameterTask().x, 'abc')
-
     def test_missing_param(self):
         def create_a():
             return A()
@@ -319,6 +306,24 @@ class ParameterTest(LuigiTestCase):
     def test_parse_list_as_tuple(self):
         param = luigi.IntParameter(batch_method=tuple)
         self.assertEqual((7, 17, 5), param._parse_list(['7', '17', '5']))
+
+    def test_default_callable_param(self):
+        default_value = 'xyz'
+
+        class CallableDefaultParameterTask(luigi.Task):
+            x = luigi.Parameter(default_callable=lambda: default_value)
+
+        task = CallableDefaultParameterTask()
+        self.assertEqual(task.x, 'xyz')
+
+        default_value = 'abc'
+        self.assertEqual(task.x, 'xyz')
+        self.assertEqual(CallableDefaultParameterTask().x, 'abc')
+
+    def test_both_default_and_default_callable_param(self):
+        self.assertRaises(
+            ParameterException,
+            lambda: luigi.parameter.Parameter(default='abc', default_callable=lambda: 'xyz'))
 
 
 class TestParametersHashability(LuigiTestCase):
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -117,8 +117,8 @@ class Parameter(object):
         """
         :param default: the default value for this parameter. This should match the type of the
                         Parameter, i.e. ``datetime.date`` for ``DateParameter`` or ``int`` for
-                        ``IntParameter``. By default, no default is stored and
-                        the value must be specified at runtime.
+                        ``IntParameter``, or be a callable that returns the appropriate type. By
+                        default, no default is stored and the value must be specified at runtime.
         :param bool significant: specify ``False`` if the parameter should not be treated as part of
                                  the unique identifier for a Task. An insignificant Parameter might
                                  also be used to specify a password or other sensitive information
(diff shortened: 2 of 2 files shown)


lsglick:
It just occurred to me that this will break `TaskParameter` and possibly other custom parameters since classes are callable.

I'll reimplement with a `default_callable` param and assert only one of `default` and `default_callable` are specified.

lsglick:
@Tarrasch et al, is there anything in particular I need to do to get feedback and/or a review? Thanks!!

test/parameter_test.py
```diff
@@ -307,6 +307,24 @@ def test_parse_list_as_tuple(self):
...
+
+        task = CallableDefaultParameterTask()
+        self.assertEqual(task.x, 'xyz')
+
+        default_value = 'abc'
+        self.assertEqual(task.x, 'xyz')
+        self.assertEqual(CallableDefaultParameterTask().x, 'abc')
+
+    def test_both_default_and_default_callable_param(self):
+        self.assertRaises(
```
Tarrasch: there's a nice decorator syntax for these now. :)

Tarrasch:
I'm fine with this being merged, maybe somebody else wants to review too?

test/parameter_test.py
```diff
@@ -307,6 +307,24 @@ def test_parse_list_as_tuple(self):
...
+            ParameterException,
+            lambda: luigi.parameter.Parameter(default='abc', default_callable=lambda: 'xyz'))
```
daveFNbuck: `assertRaises` should take the exception, then the function, then the arguments to the function. So this should be `self.assertRaises(ParameterException, luigi.Parameter.Parameter, default='abc', default_callable=lambda: 'xyz')`

daveFNbuck:
@Tarrasch it's still true that `MyTask(**mydict).x == MyTask(**mydict).x` because the constructor will look these values up in the cache. This should be equivalent to setting the default like `DEFAULT_VALUE = datetime.datetime.now()` at the module level and reading that as your script default, it just gets set a little later. The only difference will be if you have multiple different instances of the same task using the default. So if `x` isn't in `dict1` or `dict2` you're no longer guaranteed that `MyTask(**dict1).x == MyTask(**dict2).x`.

I don't quite see the point in making this change given how easy it is to declare the default at the module level, but I can't see a downside either. The only potential issue I see is that a Parameter class you've passed a lambda function to won't be picklable. But you shouldn't need to pickle the parameter class anyway.

test/parameter_test.py
Tarrasch: Actually both works. But better to just the decorator syntax I think. :)

Tarrasch:
@daveFNbuck, I think the use case was to daily, from the _same_ python process (without restarting), get the values recalculated.

-----

I get seconds thoughts about this. @lsglick, can't you just pass along the appropriate parameter throughout your chain of Tasks? What you pass to `build()` could instead be constructed with the current date, so much more determinism? I'm afraid adding this patch encourages people to follow an anti-pattern.

test/parameter_test.py
daveFNbuck: It also works as a context manager, which I think is usually a bit more clear than the decorator version as it restricts the scope.

Tarrasch: Oh sorry. I also meant the context manager. Yes, please use the context manager. :)

lsglick:
Hey @Tarrasch, @daveFNbuck, thanks for the feedback and sorry for the long reply! I do understand there being concerns, but I honestly can't think of another way to do this without duplicating the code for default parameters. Here's an example of what I currently have that I'm trying to avoid:

```
class MyTask(luigi.WrapperTask):
    """Perform task on outputs since from_date."""
    from_date = luigi.DateParameter(default=datetime.date.today() - datetime.timedelta(weeks=2))
    ...

# Create one per task I need to schedule because I have to re-specify the defaults for each type of task.
def run_my_task():
    task = MyTask(from_date=datetime.date.today() - datetime.timedelta(weeks=2))  # Duplicate of default
    luigi.build(tasks=[task])

def run_other_task():
    task = OtherTask(date=datetime.date.today() - datetime.timedelta(days=1))  # Duplicate of default
    luigi.build(tasks=[task])

def main():
    schedule.every().hour.do(run_my_task)
    schedule.every(2).hours.do(run_other_task)
    while True():
        schedule.run_pending()
        time.sleep(1)
```

With this PR, I'd do something like:
```
class MyTask(luigi.WrapperTask):
    """Perform task on outputs since from_date."""
    from_date = luigi.DateParameter(default_callable=lambda: datetime.date.today() - datetime.timedelta(weeks=2))
    ...

# One generic task runner can be reused for all task types.
def run_task(task_cls):
    luigi.build(tasks=[task_cls()])  # Uses task defaults

def main():
    schedule.every().hour.do(run_task, MyTask)
    schedule.every(2).hours.do(run_task, OtherTask)
    while True():
        schedule.run_pending()
        time.sleep(1)
```

`MyTask` needs to run both in the `main` loop as well as from the command line (e.g. `luigi MyTask`). But right now for the `main` case I'm stuck creating helper functions to recalculate the default params because they're otherwise only calculated at load time.

I suppose I could create helper methods at the module level to help with duplication like:
```
def my_task_default_from_date():
    return datetime.data.today() - datetime.timedelta(weeks=2)

class MyTask(luigi.WrapperTask):
    """Perform task on outputs since from_date."""
    from_date = luigi.DateParameter(default=my_task_default_from_date())
    ...

# Create one per task I need to schedule because I have to re-specify the defaults for each type of task.
def run_my_task():
    task = MyTask(from_date=my_task_default_from_date())
    luigi.build(tasks=[task])

def run_other_task():
    task = OtherTask(date=other_task_default_date())
    luigi.build(tasks=[task])

def main():
    schedule.every().hour.do(run_my_task)
    schedule.every(2).hours.do(run_other_task)
    while True():
        schedule.run_pending()
        time.sleep(1)
```

But that would be quite confusing since task specific values now live outside the task, and if there are multiple tasks in a module, the code would get messy. Also, it wouldn't help with the fact that I'd still have to create one helper per task for the `main` loop since I'd have to call the appropriate default value function.

The only other pattern I could think of would be to have `MyTask.from_date` default to `None` and then handle a default -2 weeks in `__init__()`, but `DateParameter` gets errors with `None` values so that isn't currently possible either.

I'm certainly open to other suggestions! I was just getting stuck trying to do it any other way!

And in case you're wondering why we're using `schedule` library in a python script instead of system cron, it hugely simplifies configuration and logging when running within docker containers.

lsglick:
Oh, and @Tarrasch it still isn't so simple as just taking the current date and calculating the offset later, because any default value for the current date still wouldn't be changing as time progresses.

daveFNbuck:
After seeing your example and testing it, it seems that the instance cache will pick up the defaults each time when it calls get_param_values so it's possible that `MyTask(**mydict).x != MyTask(**mydict).x`. You should probably add a test for that so it's more clear that this can and should happen.

Now that I understand the use case better I think it's reasonable for root tasks but dangerous for tasks you might depend on. If you depended on something with a parameter that changed every time you instantiated it, you would run one task, then check whether a different task is complete in your dependency check. I'd probably use it for some of my root tasks, but then I'd be worried that someone else at my company would see my example and use it in a non-root task without knowing the limitation.

lsglick:
@daveFNbuck thanks for the feedback. Do you think updating the docstring would be sufficient to highlight the concerns about depending on dynamic default tasks?

Seems like in general though it seems risky to depend on defaults throughout your graph if you're expecting consistent values? For example if I were running the task from 2 different workers, wouldn't I have a similar problem where the two workers, running the same task, might have some dependent tasks that initialized to a different default value?

Also, can you clarify what the suggested test should cover? That if the default_callable changes between two task instantiations, you won't get the cached version? If you manually specify the parameter or default_callable doesn't change, though, you'd still get the cached version?

Tarrasch:
@lsglick, I really don't feel convinced. I echo the concern of @daveFNbuck that this could be quite misused. In your case, you could easily come up with helpers with reasonable syntax I think. :)

lsglick:
@Tarrasch Ah bummer. Well thanks for the feedback anyways. I honestly didn't expect this to be so controversial since it essentially just mimicked the same behavior as if I were repeatedly calling a `luigi` command (such as in a cronjob) except without having to exit the python process.

Still open to suggestions on what reasonable syntax would be for our case, though :).

PR #2113, opened 2017-05-14
Make getpaths type-aware, add getpaths test.

riga:
## Description

This PR slightly updates `luigi.task.getpaths` which is used to replace tasks in the structured object returned by `Task.requires` with their respective outputs. The change makes that procedure type-aware, so e.g. when the struct's class is a subclass of dict, the returned struct has the same class.

## Motivation and Context

We often rely on the order in which we define our task requirements, e.g. via an `OrderedDict`. However, when looping over `self.input()`, that order might not be preserved at the moment, because the returned object type is a plain `dict`.

## Tests

I added `task_test.TaskTest.test_getpaths` and checked for several types.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -817,18 +817,18 @@ def getpaths(struct):
     if isinstance(struct, Task):
         return struct.output()
     elif isinstance(struct, dict):
-        r = {}
+        r = struct.__class__()
         for k, v in six.iteritems(struct):
             r[k] = getpaths(v)
         return r
+    elif isinstance(struct, (list, tuple)):
+        return struct.__class__(getpaths(r) for r in struct)
     else:
-        # Remaining case: assume r is iterable...
+        # Remaining case: assume struct is iterable...
         try:
-            s = list(struct)
+            return [getpaths(r) for r in struct]
         except TypeError:
             raise Exception('Cannot map %s to Task/dict/list' % str(struct))
-
-        return [getpaths(r) for r in s]
 
 
 def flatten(struct):
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -817,10 +817,7 @@ def getpaths(struct):
     if isinstance(struct, Task):
         return struct.output()
     elif isinstance(struct, dict):
-        r = struct.__class__()
-        for k, v in six.iteritems(struct):
-            r[k] = getpaths(v)
-        return r
+        return struct.__class__((k, getpaths(v)) for k, v in six.iteritems(struct))
     elif isinstance(struct, (list, tuple)):
         return struct.__class__(getpaths(r) for r in struct)
     else:
--- a/test/task_test.py
+++ b/test/task_test.py
@@ -82,6 +82,29 @@ class TaskTest(unittest.TestCase):
         task = load_task("luigi", "ExternalTask", {})
         assert(isinstance(task, luigi.ExternalTask))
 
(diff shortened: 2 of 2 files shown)


riga:
Seems like the `TOXENV=py36-nonhdfs` failed, but I don't have a clue why...

riga:
Could you give me a hint how to check what went wrong?

Tarrasch:
> Could you give me a hint how to check what went wrong?

I just needed to rerun travis on that. Sorry never your fault.

luigi/task.py
```diff
@@ -817,19 +817,19 @@ def getpaths(struct):
     if isinstance(struct, Task):
         return struct.output()
     elif isinstance(struct, dict):
-        r = {}
+        r = struct.__class__()
```
riga: Yep. If it's a subtype of dict it should always be mutable.

riga:
> Can you write a test case using an in immutable type please?

Should be covered with the `tuple` test, right?
https://github.com/spotify/luigi/pull/2113/files#diff-7fe10b1a62e793e1535a8477614c1b30R96

I can add more tests if you want.

test/task_test.py
```diff
@@ -82,6 +82,29 @@ def test_external_tasks_loadable(self):
...
+        reqs["generator"] = (t for _ in range(10))
+
+        struct = luigi.task.getpaths(reqs)
+        self.assertIsInstance(struct, dict)
+        self.assertIsInstance(struct["bare"], luigi.Target)
+        self.assertIsInstance(struct["dict"], dict)
+        self.assertIsInstance(struct["OrderedDict"], collections.OrderedDict)
+        self.assertIsInstance(struct["list"], list)
+        self.assertIsInstance(struct["tuple"], tuple)
+        self.assertIsInstance(struct["generator"], list)
```
Tarrasch: You should probably only check for being `iterable` here.

luigi/task.py
```diff
@@ -817,19 +817,16 @@ def getpaths(struct):
     if isinstance(struct, Task):
         return struct.output()
     elif isinstance(struct, dict):
-        r = {}
-        for k, v in six.iteritems(struct):
-            r[k] = getpaths(v)
-        return r
+        return struct.__class__((k, getpaths(v)) for k, v in six.iteritems(struct))
```
riga: Here, I assumed that even immutable mappings (such as a "frozendict"-like mapping) shares the functionality of a standard dict to take a list/generator of tuples in the constructor.

Tarrasch: I believe this would crash if you use a [`collections.Counter`](https://docs.python.org/3.6/library/collections.html#collections.Counter). But if you use that for this purpose, you totally deserve it. ;p

To be clear. This code is better, there's no need for a fix. :)

riga: > This code is better, there's no need for a fix. :)

You mean
```python
for k, v in six.iteritems(struct):
    r[k] = getpaths(v)
    return r
```
? Sure, I can revert that commit.

Tarrasch: No, I think the current code is better. If I have to choose between supporting custom frozen dicts or `collections.Counter`. I would pick the former any day (for this instance)

Tarrasch:
@riga, I merged this, but took the liberty to greatly reword the commit message. I hope you don't mind.

Thanks for this patch! :)

Issue #2111, opened 2017-05-11
RemoteFileSystem of luigi.contrib.ftp does not implement the listdir method.


dmohns:
Pretty much as the title states: Currently RemoteFileSystem of `luigi.contrib.ftp` does not implement the `listdir` method. 
As FTP basically provides access to an ordinary filesystem this should be perfectly do-able and would very helpful in certain situations.

dlstadther:
@dmohns Please feel free to submit a PR!

OstrichProjects:
@dmohns @dlstadther This feature request can be marked as resolved.

PR #2114, opened 2017-05-19
add listdir method to ftp remote file system

OstrichProjects:
## Description
This adds the `listdir` method to `luigi.contrib.ftp.RemoteFileSystem`.

## Motivation and Context
Resolves issue #2111.

I tested against FTP and SFTP servers running in Docker containers locally with Python 2.7.10 and Python 3.6.1 and the method worked as expected.


OstrichProjects:
@Tarrasch I could work on setting up FTP/SFTP tests, but I'm not sure if they'll be worth the effort.

Issue #2094, opened 2017-04-20
DictParameter fails when containing empty lists


sguermond:
Related to [#1657](https://github.com/spotify/luigi/issues/1657#issuecomment-217066024), #1780 

The following error occurs at the end of the run when running a task with a DictParameter whose default is a (key -> empty list) pair or a (key -> empty list). 

```
Uncaught exception in luigi
Traceback (most recent call last):
  File "luigi/retcodes.py", line 74, in run_with_retcodes
    worker = luigi.interface._run(argv)['worker']
  File "luigi/interface.py", line 238, in _run
    return _schedule_and_run([cp.get_task_obj()], worker_scheduler_factory)
  File "luigi/interface.py", line 198, in _schedule_and_run
    logger.info(execution_summary.summary(worker))
  File "luigi/execution_summary.py", line 418, in summary
    return _summary_wrap(_summary_format(_summary_dict(worker), worker))
  File "luigi/execution_summary.py", line 356, in _summary_format
    str_output += '{0}\n'.format(_get_str(group_tasks[status], status in _PENDING_SUB_STATUSES))
  File "luigi/execution_summary.py", line 147, in _get_str
    params = _get_set_of_params(tasks)
  File "luigi/execution_summary.py", line 190, in _get_set_of_params
    params[param] = {getattr(task, param[0]) for task in tasks}
  File "luigi/execution_summary.py", line 190, in <setcomp>
    params[param] = {getattr(task, param[0]) for task in tasks}
  File "luigi/parameter.py", line 793, in __hash__
    self.__hash = functools.reduce(operator.xor, hashes, 0)
TypeError: unhashable type: 'list'
```

The following two tasks will raise an error at the end of the run, even when completing successfully:
``` python
class TaskA(luigi.Task):
    dict_a = luigi.DictParameter(default = {"a": []})
```
``` python
class TaskB(luigi.Task):
    dict_b = luigi.DictParameter(default = {"b": [""]})
```

However, a non-empty list inside the default parameter prevents the error from happening:
``` python
class TaskC(luigi.Task):
    dict_c = luigi.DictParameter(default = {"c": ["c"]})
```
As will adding any other item to the dictionary, including a second empty list:
``` python
class TaskD(luigi.Task):
    dict_d = luigi.DictParameter(default = {"d": [], "e": []})
```

[...]

PR #2115, opened 2017-05-19
Recursively normalize DictParameter

OstrichProjects:
## Description
This recursively calls normalize on `Mapping`s and `list`s when calling `DictParameter.normalize`.  Also, makes `OrderedFrozenDict` and `DictParameter.DictParamEncoder` private, as requested by @Tarrasch.

## Motivation and Context
Fixes issue #2094.

Added a test, but am open to suggestions of better tests.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -857,11 +857,18 @@ class DictParameter(Parameter):
                 return obj.get_wrapped()
             return json.JSONEncoder.default(self, obj)
 
+    def _normalize_value(self, v):
+        if isinstance(v, Mapping):
+            return self.normalize(v)
+        elif isinstance(v, list):
+            return ListParameter().normalize(v)
+        return v
+
     def normalize(self, value):
         """
         Ensure that dictionary parameter is converted to a _FrozenOrderedDict so it can be hashed.
         """
-        return _FrozenOrderedDict(value)
+        return _FrozenOrderedDict({k: self._normalize_value(v) for k, v in value.items()})
 
     def parse(self, s):
         """
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -782,7 +782,7 @@ class EnumParameter(Parameter):
         return e.name
 
 
-class FrozenOrderedDict(Mapping):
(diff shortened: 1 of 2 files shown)


luigi/parameter.py
```diff
@@ -848,20 +848,27 @@ def run(self):
...
-            if isinstance(obj, FrozenOrderedDict):
+            if isinstance(obj, _FrozenOrderedDict):
                 return obj.get_wrapped()
             return json.JSONEncoder.default(self, obj)
 
+    def _normalize_value(self, v):
+        if isinstance(v, Mapping):
+            return self.normalize(v)
+        elif isinstance(v, list):
+            return ListParameter().normalize(v)
```
Tarrasch: Will this work if a list contains a dict or a list?

Tarrasch:
By the way, it seems like the failed test weren't just flukes. I think either code or testcode needs to be updated/fixed.

OstrichProjects:
Yea, I'm trying to reproduce that behavior but 3.4 and 3.5 are behaving as expected for me.  I'll break the method out to be `_recursively_freeze`.

luigi/parameter.py
```diff
@@ -917,7 +928,7 @@ def normalize(self, x):
         :param str x: the value to parse.
         :return: the normalized (hashable/immutable) value.
         """
-        return tuple(x)
+        return _recursively_freeze(x)
```
Tarrasch: this should still be `tuple()` I think, or do we say anywhere a "bigger" structure is allowed?

OstrichProjects: [The documentation says `<JSON string>`](https://github.com/spotify/luigi/blob/master/luigi/parameter.py#L905)

OstrichProjects: It is sort of overkill for a very specific use case though.

Tarrasch: Why on earth is it a JSON Object ... oh well then. Looks good to me.

Tarrasch:
Big thanks! Such an improvement.

But As for issue #2094, I'm not 100% confident this solved that issue. Let's ping the issue author in that issue.

PR #2117, opened 2017-05-19
Remove superfluos projectId parameter in BigQuery job definition

### LINKED ISSUE #2116: Superfluos project_id parameter in BigQuery job definition
The parameter `project_id` of most tasks in `luigi.contrib.bigquery` is superfluous as per https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs. Indeed, if set to arbitrary values jobs still run successfully.

It doesn't really *hurt* but can be confusing at first sight as it suggest some additional dependency.

dmohns:
## Description
Remove parameter `projectId` from `luigi.contrib.bigquery` tasks.

## Motivation and Context
- Resolves #2116

I tested it using a job based on `BigQueryRunQueryTask` against live data.

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -348,7 +348,6 @@ class BigQueryClient(object):
         """
 
         job = {
-            "projectId": dest_table.project_id,
             "configuration": {
                 "copy": {
                     "sourceTable": {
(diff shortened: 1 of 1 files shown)


dmohns:
@Tarrasch I also created the issue. So that should be fine.

dmohns:
@mikekap Being the original author would you mind to review?

PR #2120, opened 2017-05-20
Don’t use cached autoconfig client in config tests

daveFNbuck:
## Description
In hdfs_test.ConfigurationTest, calls to hdfs.get_autoconfig_client return the version cached in the call to setUp. Because this is called before the config is set by the helpers.with_config decorator, this means that the tests aren’t actually testing different configurations.

Luckily, get_autoconfig_client takes an optional argument to specify its cache. When passing a fresh cache, these will actually test what they’re supposed to. This PR passes a fresh cache to get_autoconfig_client in each test.

## Motivation and Context
While working on #2119 I tried copying code from the config tests to get a snakebite client and found that I was still getting the same responses as expected from a hadoop cli client. This is just applying my solution to that problem from #2119 to the config tests so they'll do what they're supposed to.

I ran the unit tests with the `isinstance` check added but not the fix and the two that didn't use hadopcli failed. With the fix, they all succeed.

### DIFF
--- a/test/contrib/hdfs_test.py
+++ b/test/contrib/hdfs_test.py
@@ -19,6 +19,7 @@ import functools
 import re
 from helpers import unittest
 import random
+import threading
 import pickle
 
 import helpers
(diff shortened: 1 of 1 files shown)


Tarrasch:
Nice catch. From the days Luigi didn't care for python 3 as much I suppose. :)

PR #2121, opened 2017-05-20
AWS_SESSION_Token read from ENV

ouanixi:
Currently, Luigi offers the option to start the process of role assumption if `aws_role_arn` and `aws_role_session_name` are provided in the config. In production, sometimes the `sts:assumeRole` is given to the ec2/ecs resources and the access key, secret and session token are available as an env variables, instead of config.

The current setup forces the session token to be `None` which doesn't allow for this scenario:
- assume-role has been called from outside luigi and the temporary auth keys and session token are available as env variables.

### DIFF
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -106,7 +106,7 @@ class S3Client(FileSystem):
         role_arn = options.get('aws_role_arn')
         role_session_name = options.get('aws_role_session_name')
 
-        aws_session_token = None
+        aws_session_token = os.environ.get('AWS_SESSION_TOKEN')
 
         if role_arn and role_session_name:
             from boto import sts


ouanixi:
Thanks, so the documentation is a little bit scattered across their website so here are two links that allow the understanding of AWS_SESSION_TOKEN (I'll summarize here)

http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html
This page gives an overview of what temporary credentials are. They are generated by the STS service to give access to resources for a limited amount of time. When queried, it returns a temporary `access key id`, `security id` as well a `session token` that are valid for a predefined amount of time. (See [here](https://github.com/spotify/luigi/blob/master/luigi/contrib/s3.py#L115) for how sts is queried in Luigi) Now to use temporary credentials, the session token must be passed to the `S3Connection` constructor.

See boto documentation as well, they mention how SESSION_TOKEN is mandatory when using temporary credentials.
http://boto.cloudhackers.com/en/latest/ref/sts.html?highlight=SESSION%20TOKEN#boto.sts.credentials.Credentials


Because the aws_session_token variable is forced to be None in the current code, we are only left with one option in terms of temporary credentials; ask luigi to assume the role. And even with that, it wouldn't work if the sts:assumeRole requires MFA identification (this can be another MR potentially)

The scenario that I'm trying to allow here is for those who store the temporary credentials as env variables (a lot of production code really).

ouanixi:
@Tarrasch Can we please put this on hold for a short while ? I need to investigate changes that boto made. I think the env var needs to be called `AWS_SECURITY_TOKEN` instead of `AWS_SESSION_TOKEN` and this potentially means that I need to rename the variable in my env and things should work OK.

https://github.com/99designs/aws-vault/pull/35/files

If that's the case, this MR is redundant.

ouanixi:
@Tarrasch Hi again, yeah the MR is redundant. I don't think this change is a useful one. boto's S3Connection reads the env var and looks for `AWS_SECURITY_TOKEN` already. So users just need to be aware of this and use the appropriate naming.

Although I think it's a bit of a shame that all the AWS documentation refers to the token as `AWS_SESSION_TOKEN` but then boto expects it to be `AWS_SECURITY_TOKEN`

Closing the MR now. Sorry for the noise.

PR #2092, opened 2017-04-17
Fix "Show Upstream Dependencies" in visualiser

xoob:
## Description

Fixes the "Show Upstream Dependencies" checkbox which did not have any effect on the displayed chart.

## Have you tested this? If so, how?
 

I tested the UI with my jobs on my local installation.

PR #2124, opened 2017-05-24
Url fixes

daveFNbuck:
## Description
Fix a few bugs with how state is stored in the URL:
- Check the correct value for inverting the graph so the graph will actually invert when the box is checked
- Update tasks when the server side checkbox is clicked instead of just updating the URL so it works without having to refresh
- Remove the search query from the url when it's deleted so it matches the current state

## Motivation and Context
I just merged the trunk into my branch and noticed these tiny bugs all related to the same updates.

Ran locally and in production, these bugs are gone now.


Tarrasch:
Great! This doesn't collide with #2092 right?

daveFNbuck:
They share an identical change so I guess it's ok.

PR #2125, opened 2017-05-24
Set hideDone on in graph links

daveFNbuck:
# Description
We used to have hideDone set by default, as this default can prevent large graphs from freezing the visualiser. This went away when we moved to storing the checkbox state in the URL bar, but we have an easy fix by adding it to the link.

## Motivation and Context
I tend to have large graphs where most items are done, so this feature is the difference between a near-immediate response and a delay of tens of seconds or minutes. We used to have this feature, it seems to have been an accidental removal.

I used this locally and in production and it works.

### DIFF
--- a/luigi/static/visualiser/index.html
+++ b/luigi/static/visualiser/index.html
@@ -239,7 +239,7 @@
 
         <script type="text/template" name="actionsTemplate">
             <div class="span2">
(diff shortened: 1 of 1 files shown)


Tarrasch:
Sounds reasonable to me! :)

PR #2127, opened 2017-05-24
Remember family and status

daveFNbuck:
## Description
Remember status and task family selectors from the tasks page in the visualiser in the url hash.

## Motivation and Context
Now that the search query is stored, it would be nicer if the rest of the filters could be sent in links and survive refreshes too.

It's working for me in production right now. Tested on chrome and firefox. I'm not 100% sure about this code but I tried my best to find any corner cases where it might break.

### DIFF
--- a/luigi/static/visualiser/index.html
+++ b/luigi/static/visualiser/index.html
@@ -513,7 +513,7 @@
                 </nav>
             </div>
             <div class="main-sidebar">
-                <div class="sidebar">
+                <div class="sidebar" id="familySidebar">
                 </div>
             </div>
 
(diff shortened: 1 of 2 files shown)


daveFNbuck:
We have unit tests for javascript now? Can you point me at how to use them?

Tarrasch:
I know there where some efforts at some point in time. Maybe [this](https://github.com/spotify/luigi/tree/master/test/visualiser)?

Tarrasch:
Thanks for this patch!

daveFNbuck:
No problem, I've been really looking forward to getting the url hash stuff!

PR #2128, opened 2017-05-24
Remove length from state when it equals default

daveFNbuck:
## Description
Delete the length from the url hash when it's default.

## Motivation and Context
We previously just didn't set the default value, but this would result in a non-default value being preserved in the hash after you set to the default value.

Ran this locally and in production.


Tarrasch:
Awesome! :)

PR #2130, opened 2017-05-25
Clean up unittest deprecations and warnings

xoob:
This PR cleans up a number of warnings and notices in the tests. Detailed descriptions are in the commit messages.

## Motivation and Context

Needed to get aquatinted with the test infrastructure. Aside from that, general maintenance and keeping things neat.

Tested locally with `tox -e py36-nonhdfs` and `tox -e py27-nonhdfs`.

### DIFF
--- a/luigi/lock.py
+++ b/luigi/lock.py
@@ -26,8 +26,6 @@ import os
 import sys
 from subprocess import Popen, PIPE
 
-from luigi import six
-
 
 def getpcmd(pid):
     """
--- a/luigi/lock.py
+++ b/luigi/lock.py
@@ -63,7 +63,7 @@ def getpcmd(pid):
         # https://github.com/spotify/luigi/pull/1876
         try:
             with open('/proc/{0}/cmdline'.format(pid), 'r') as fh:
-                return fh.read().replace('\0', ' ').rstrip()
+                return fh.read().replace('\0', ' ').decode('utf8').rstrip()
         except IOError:
             # the system may not allow reading the command line
             # of a process owned by another user
--- a/luigi/lock.py
+++ b/luigi/lock.py
@@ -26,6 +26,8 @@ import os
 import sys
 from subprocess import Popen, PIPE
 
+from luigi import six
+
 
 def getpcmd(pid):
     """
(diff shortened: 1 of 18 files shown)


luigi/lock.py
```diff
@@ -65,12 +81,7 @@ def get_info(pid_dir, my_pid=None):
         my_pid = os.getpid()
 
     my_cmd = getpcmd(my_pid)
-
-    if six.PY3:
-        cmd_hash = my_cmd.encode('utf8')
-    else:
-        cmd_hash = my_cmd
-
+    cmd_hash = my_cmd.encode('utf8')
```
xoob: Hmm, now this causes breakage on Linux, py27: `UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 4: ordinal not in range(128)`.

Would this change to [line 52](https://github.com/spotify/luigi/pull/2130/commits/c676395b4eeb1cf14b04aec95848faccfd1e2197#diff-6e699ca1b8953337c23f81c42066b0bbL52) be appropriate?

```patch
-return fh.read().replace('\0', ' ').rstrip()
+return fh.read().replace('\0', ' ').decode('utf8').rstrip()
```

Tarrasch:
Thank you so much!!!

Can you just fix the flake8 error? (also check why py27 fails)

Tarrasch:
@xoob, this is a really good PR which I would like to merge, can you just fix the flake8 error?

xoob:
I'll take a shot at it this weekend. 





On Fri, Jun 30, 2017 at 3:08 AM -0700, "Arash Rouhani" <notifications@github.com> wrote:












@xoob, this is a really good PR which I would like to merge, can you just fix the flake8 error?

erikbern:
looks awesome!

xoob:
@Tarrasch @erikbern All tests and flake8 now pass.

Unrelated to the changes in this PR, the `py27-cdh` and `py36-cdh` suites time out after 50 minutes. This seems to happen on master too.

Tarrasch:
The commit message quality was quite mixed, but overall I thought it might be useful to keep the whole commit history as some commits were particularly useful to be able to introspect in isolation. So did a rebase-merge here. :)

Thanks a lot for this contribution. This greatly increases the code quality in luigi!

xoob:
@Tarrasch Which commit messages worked for you and how would you improve the other ones?

Tarrasch:
@xoob good question! These are my opinion and they are by no means better than anyone else's:

 * I really liked the first one. As it has a clear description and useful details in the body
 * Actually I definitely liked most of your commits
 * I found one to have >20 lines of text in body. Which might feel not so scroll friendly in some views, and I doubt anybody is going to read all that, and it is maybe obvious from the actual code change.
 * I prefer flake8 fixes to not be separate, but I do it myself too, because it's so hard to amend them in later.
 * The last 3 commits seem to say the same thing and should probably have been only 1 (unless I'm mistaken).

Are any of these thoughts useful as feedback? :)

xoob:
@Tarrasch Very useful, thanks. I could have amended and force-pushed the last three commits (and possibly the flake8 changes too), which I'll do next time. Appreciate the feedback.

Tarrasch:
> I could have amended and force-pushed the last three commits

Not everyone is so responsive and to iterate on their commits. Thanks for your cooperation. ^^

PR #2131, opened 2017-05-25
Resource url hash

daveFNbuck:
## Description
Add expanded resources from the resource tab to the URL bar and fix the graph links on the resource page.

## Motivation and Context
Now that we have the luxury of the task list and graph tab states being stored in the URL, I wanted it on every page. I like to refresh the resource page a lot and always need to re-expand the resources I'm interested in. Not anymore! I also updated the graph links in the expanded resource boxes so they actually work again.

Ran this locally and in production. I'm willing to add unit tests if possible.


Tarrasch:
Awesome! :)

PR #2132, opened 2017-05-25
Preserve options in svg graph links

daveFNbuck:
## Description
Include options from url hash in svg graph node links.

## Motivation and Context
When clicking on the nodes in an svg graph, a graph will load with that node as the root but with hide done and show upstream dependencies unset. This can be very annoying when you suddenly have to load a much larger graph or your graph flips orientation.

In order to preserve these and any future options, we pass the hash with taskId stripped out of it into the dependency graph so it can use that as a basis for the node links. This preserves the options present when the graph was created, which should match the current options on the
page.

D3 graphs also suffer from the same bug, but they are not addressed by this change.

Ran locally and in production. Willing to write tests once I learn how.

### DIFF
--- a/luigi/static/visualiser/js/graph.js
+++ b/luigi/static/visualiser/js/graph.js
@@ -33,7 +33,7 @@ Graph = (function() {
             name: task.name,
             taskId: task.taskId,
             status: task.status,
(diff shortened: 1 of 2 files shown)


Tarrasch:
Thanks!

PR #2134, opened 2017-05-26
Add ExtractTask to luigi.contrib.BigQuery module

dmohns:
The extract job is currently not implemented in `contrib.bigquery` module. Scope of this PR is to implement it according to https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs .

## Motivation and Context
Without the extract we can only assemble pipelines **into** BigQuery but not from BQ to elsewhere.

## Description
- Fill job configuration according to https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs for extract jobs.
- Add assertitions for the jobs to be sane.
- Add basic unittest for extracts
- Add Integration test with basic configuration

- Added basic unit and integration tests
- A version of this runs in our production environment.

## Questions
Being relatively new to luigi ecosystem I am not 100% certain I correctly followed the luigi *style* everywhere. Happy to receive feedback.

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -71,6 +71,22 @@ class FieldDelimiter(object):
     PIPE = "|"
 
 
+class PrintHeader(object):
+    TRUE = True
+    FALSE = False
+
+
+class DestinationFormat(object):
+    AVRO = 'AVRO'
+    CSV = 'CSV'
+    NEWLINE_DELIMITED_JSON = 'NEWLINE_DELIMITED_JSON'
+
+
+class Compression(object):
+    GZIP = 'GZIP'
+    NONE = 'NONE'
+
+
 class Encoding(object):
     """
     [Optional] The character encoding of the data. The supported values are UTF-8 or ISO-8859-1. The default value is UTF-8.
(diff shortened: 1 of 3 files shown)


miike:
This would be great to have as we're after something similar. Is it worth checking if the `table_exists` in `run`?

luigi/contrib/bigquery.py
```diff
@@ -668,6 +684,91 @@ class ExternalBigQueryTask(MixinBigQueryBulkComplete, luigi.ExternalTask):
...
+        job = {
+            'configuration': {
+                'extract': {
+                    'sourceTable': {
+                        'projectId': input.table.project_id,
+                        'datasetId': input.table.dataset_id,
+                        'tableId': input.table.table_id
+                    },
+                    'destinationUris': destination_uris,
+                    'printHeader': self.print_header,
```
miike: Does BigQuery ignore `printHeader` and `fieldDelimiter` if the destinationFormat is not CSV?

dmohns: Good catch! I never used with BigQuery with anything other than `csv` and am therefore not aware of the behavior. I know from other cases than superfluous job parameter are ignored so I would _assume_ same is true for `printHeader` and `fieldDelimiter`.

Regardless it would be more consistent to set `printHeader` and `fieldDelimiter` conditionally. I will add as soon as find some time.

dmohns:
The way the Task is currently designed I think it would be unnecessary overhead to add a dedicated `exists` check. The ExtractTask asserts to have a singleton `BigQueryTarget` as requires (which's exists method is essentially the `table_exists`). The main reason behind is separation responsibility. The ExtractTask should take care of unloading the table while it is luigi's job to take care of dependency checking and resolution. 
In case the table is non-existent at run-time it will result in a BigQuery exception which I think is appropriate behavior.

luigi/contrib/bigquery.py
miike: Yeah - I'm not sure if it's still an issue but I only noticed because Airflow has a [conditional check](https://github.com/apache/incubator-airflow/blob/ffe6162f2b9e61f686b1fe993eb95e43e60c7dd5/airflow/contrib/hooks/bigquery_hook.py#L297) on this.

dmohns:
@miike Your were right: non-CSV extracts indeed fail when provided with `FieldDelimiter` or `PrintHeader` options. They are now set conditionally as suggested.

dlstadther:
@dmohns @miike Is this PR approved and ready to merge?

dmohns:
@dlstadther From my point of view this can be merged, I am waiting for some potential feedback and/or reviews.

dlstadther:
@dmohns Thanks for contributing! Sorry again for the long delay between submission and merge

PR #2136, opened 2017-05-26
Added BigQuery chunking logic for huge uploads

BartoszJanota:
## Motivation and Context
Some people are uploading huge amounts of data (15+ TB) and they are hitting BQ Timeouts. The easiest solution is to chunk data and upload partially.

## Description
We added two new Luigi Parameters:
 - Boolean `chunking-enabled` - when True (default is False) it will chunk your source data and upload synchronously with a default chunk size (1 TB)
- Int `chunk-size-gb` - user can adjust (optionally) chunk size 

If chunking is disabled, then the upload logic is not changed. If chunking is enabled, we will upload chunks to separated temp tables, then we will atomically merge them and store result in one final table. All the temp tables are removed after successful run.

We have a heuristic calculating the size of the input data to be chunked, so chunk size can vary. We are also assuming that files within a a folder to be chunked are of a similar size. So we can handle let's say 22000 files 90MB each or 500 files 17GB each. 

We have included unit tests and did some stress and manual tests in house.

We will be cleaning up some unnecessary printouts, we are looking for some architectural and technical feedback.

Thanks!

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -574,9 +574,6 @@ class BigQueryLoadTask(MixinBigQueryBulkComplete, luigi.Task):
 
                     chunk_intervals, uris_per_chunk = self.calculate_chunk_intervals(gcs_client, uris, self.chunk_size_gb)
 
-                    print("About to schedule " + str(len(chunk_intervals)) + " chunks, each max "
-                          + str(uris_per_chunk) + " uris")
-
                     for chunk_num in chunk_intervals:
                         chunk_uris = uris[chunk_num:chunk_num+uris_per_chunk]
                         # min func to handle the very last chunk from the list, usually shorter then "uris_per_chunk"
@@ -586,7 +583,6 @@ class BigQueryLoadTask(MixinBigQueryBulkComplete, luigi.Task):
                         submit_load_job(chunk_uris, partial_table_id)
                 else:
                     # we don't chunk for non-wildcard source_uri
-                    print("Non-wildcard path for: " + source_uri)
                     table_id = "TEMP_" + output.table.table_id + "_" + str(source_counter) + "_" + str(int(load_start_time))
                     submit_load_job([source_uri], table_id)
 
@@ -597,7 +593,6 @@ class BigQueryLoadTask(MixinBigQueryBulkComplete, luigi.Task):
             self.remove_tables(bq_client, partial_table_ids)
 
         load_end_time = time.time()
-        print("UPLOAD DONE, it took: " + str(load_end_time - load_start_time) + " seconds")
 
     @staticmethod
     def calculate_avg_blob_size(gcs_client, uris):
@@ -620,7 +615,6 @@ class BigQueryLoadTask(MixinBigQueryBulkComplete, luigi.Task):
         #  todo: what if avg_blob_size_bytes * num_of_uris < chunk_size_gb ??
         #  todo: then we will upload this data in one chunk, but at the end we will still merge this one table
         #  todo: so this step could be skipped
-        print("Summary avg blob size is: " + str(avg_blob_size_bytes) + " bytes")
         chunk_size_bytes = chunk_size_gb * GB_TO_BYTES
         uris_per_chunk = min(MAX_SOURCE_URIS, max(1, chunk_size_bytes / avg_blob_size_bytes))
         chunk_intervals = range(0, len(uris), uris_per_chunk)
@@ -646,20 +640,16 @@ class BigQueryLoadTask(MixinBigQueryBulkComplete, luigi.Task):
         }
         if self.schema:
             merge_job['configuration']['query']['schema'] = {'fields': self.schema}
-        print("Starting merge")
         start_merge = time.time()
         bq_client.run_job(project_id, merge_job, dataset=output.table.dataset)
         end_merge = time.time()
-        print("Merge done, it took: " + str(end_merge - start_merge) + " seconds")
 
     @staticmethod
     def remove_tables(bq_client, table_ids):
-        print("Starting cleaning up")
         for t in table_ids:
             table_params = t.split(".")
             bq_client.client.tables().delete(projectId=table_params[0], datasetId=table_params[1],
                                              tableId=table_params[2]).execute()
-        print("Cleaning up is done, removed " + str(len(table_ids)) + " tables")
 
 
 class BigQueryRunQueryTask(MixinBigQueryBulkComplete, luigi.Task):
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -520,7 +520,6 @@ class BigQueryLoadTask(MixinBigQueryBulkComplete, luigi.Task):
         project_id = output.table.project_id
 
         bq_client = output.client
-        gcs_client = GCSClient()
 
         source_uris = self.source_uris()
         assert all(x.startswith('gs://') for x in source_uris)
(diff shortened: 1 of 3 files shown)


luigi/contrib/bigquery.py
```diff
@@ -330,6 +337,19 @@ def run_job(self, project_id, body, dataset=None):
             logger.info('Waiting for job %s:%s to complete...', project_id, job_id)
             time.sleep(5)
 
+    def submit_job(self, project_id, body, dataset=None):
+        """Submits a BigQuery "job". Returns job id. See the documentation for the format of body.
```
ulzha: Which "the documentation"?

luigi/contrib/bigquery.py
```diff
@@ -371,9 +391,11 @@ def copy(self,
 
 
 class BigQueryTarget(luigi.target.Target):
-    def __init__(self, project_id, dataset_id, table_id, client=None, location=None):
+    def __init__(self, project_id, dataset_id, table_id, client=None, location=None, enable_chunking=False, chunk_size_gb=1000):
```
ulzha: These should be `BigQueryLoadTask` parameters, not `BigQueryTarget` constructor's parameters.

The purpose of a `Target` is identifying and locating an output, and help in achieving some properties 
 like idempotency. Better not couple the aspects of the output's _production_ in it, as producing output is what a `Task` does.

luigi/contrib/bigquery.py
```diff
@@ -330,6 +337,19 @@ def run_job(self, project_id, body, dataset=None):
             logger.info('Waiting for job %s:%s to complete...', project_id, job_id)
             time.sleep(5)
 
+    def submit_job(self, project_id, body, dataset=None):
```
ulzha: Where is this method used?

luigi/contrib/bigquery.py
ulzha: Also please document what reasons would people possibly have to not enable chunking. Otherwise why even have it toggleable if chunking works fine for all use cases?

Isn't setting chunk_size_gb to "infinity" the same as disabling chunking? Then also perhaps the `enable_chunking` parameter is superfluous.

luigi/contrib/bigquery.py
```diff
@@ -507,40 +529,155 @@ def run(self):
...
+                job['configuration']['load']['allowQuotedNewlines'] = self.allow_quoted_new_lines
+
+            if self.schema:
+                job['configuration']['load']['schema'] = {'fields': self.schema}
+
+            print "Job :" + str(job)
+
+            bq_client.run_job(output.table.project_id, job, dataset=output.table.dataset)
+            if output.enable_chunking:
+                partial_table_ids.append(output.table.project_id + "." + output.table.dataset_id + "." + table_id)
```
ulzha: Astonishing conditional and side effect. Should just `return` that string? Or there could be a separate function to build the partial table id altogether, as this line doesn't use anything that was computed in this function's body. There would be one `if` less then.

test/contrib/bigquery_gcloud_test.py
```diff
@@ -224,6 +227,49 @@ def test_run_query(self):
...
+                            table=self.table.table_id,
+                            enable_chunking=True,
+                            chunk_size_gb=500)
+        task.run()
+        self.assertTrue(self.bq_client.dataset_exists(self.table))
+        self.assertTrue(self.bq_client.table_exists(self.table))
+
+    def test_avg_blob_size(self):
+        uris = [
+            "gs://luigi-travistestenvironment/176348453/contrib.gcs_test.GCSTargetTest.test_writelines",
```
ulzha: Should use `BUCKET_NAME` from above...

luigi/contrib/bigquery.py
```diff
@@ -507,40 +529,155 @@ def run(self):
...
+                    num_of_uris = len(uris)
+
+                    chunk_intervals, uris_per_chunk = self.calculate_chunk_intervals(gcs_client, uris, output.chunk_size_gb)
+
+                    print "About to schedule " + str(len(chunk_intervals)) + " chunks, each max " \
+                          + str(uris_per_chunk) + " uris"
+
+                    for chunk_num in chunk_intervals:
+                        chunk_uris = uris[chunk_num:min(chunk_num+uris_per_chunk, num_of_uris)]
+                        # min func to handle the very last chunk from the list, usually shorter then "uris_per_chunk"
```
ulzha: You don't need min: `[1, 2, 3, 4][1:131231] = [2, 3, 4]`

luigi/contrib/bigquery.py
```diff
@@ -507,40 +529,155 @@ def run(self):
...
+            if blob is not None:
+                blob_size_bytes = blob['size']
+                total_sampled_blobs_bytes += int(blob_size_bytes)
+            else:
+                raise Exception("BigQueryLoadTask failed: Object + " + uri + " doesn't exist")
+        return total_sampled_blobs_bytes / population_size
+
+    def calculate_chunk_intervals(self, gcs_client, uris, chunk_size_gb):
+        avg_blob_size_bytes = self.calculate_avg_blob_size(gcs_client, uris)
+        #  todo: what if avg_blob_size_bytes * num_of_uris < MAX_UPLOAD_SIZE_BYTES ??
```
ulzha: What's MAX_UPLOAD_SIZE_BYTES?

luigi/contrib/bigquery.py
```diff
@@ -507,40 +529,155 @@ def run(self):
...
+        #  todo: then we will upload this data in one chunk, but at the end we will still merge this one table
+        #  todo: so this step could be skipped
```
ulzha: Perhaps when `partial_table_ids` has length of 1 one can optimize the merge into a rename.

luigi/contrib/bigquery.py
```diff
@@ -507,40 +529,155 @@ def run(self):
...
+        print "Summary avg blob size is: " + str(avg_blob_size_bytes) + " bytes"
+        chunk_size_bytes = chunk_size_gb * GB_TO_BYTES
+        uris_per_chunk = min(MAX_SOURCE_URIS, max(1, chunk_size_bytes / avg_blob_size_bytes))
```
ulzha: :star: chunking on the MAX_SOURCE_URIS limit is value added. Perhaps that's another case in point that we should just only support the "chunked" route - the "non chunked" route misses out on this fix.

luigi/contrib/bigquery.py
Tarrasch: Also: `[1, 2, 3, 4][1:] = [2, 3, 4]`

dmohns:
One general comment on the scope of this PR: 
I struggle to see which exact problem this PR is trying to solve. Loading data into separate tables and merging them together feels like a highly customized solution for a very specific problem.
If timeouts are a problem that should be addressed shouldn't we rather look at resumable uploads as per https://cloud.google.com/bigquery/loading-data-post-request#resumable ?

luigi/contrib/bigquery.py
```diff
@@ -507,40 +517,139 @@ def run(self):
...
         source_uris = self.source_uris()
         assert all(x.startswith('gs://') for x in source_uris)
 
-        job = {
+        partial_table_ids = []
+        load_start_time = time.time()
+
+        def submit_load_job(uris, table_id):
+            job = {
+                'projectId': output.table.project_id,
```
dmohns: This line is not needed. See https://github.com/spotify/luigi/issues/2116

luigi/contrib/bigquery.py
```diff
@@ -507,40 +517,139 @@ def run(self):
         output = self.output()
         assert isinstance(output, BigQueryTarget), 'Output must be a BigQueryTarget, not %s' % (output)
 
+        project_id = output.table.project_id
+
         bq_client = output.client
+        gcs_client = GCSClient()
```
dmohns: This is only needed in the chunked case . Should we hence move it into the if-else block to remove computational overhead in the non-chunked version? (If I am not mistaken instantiating the GCSClient runs the full google-auth procedure?)

PR #2138, opened 2017-05-30
Move visualiser graph type default to a constant

daveFNbuck:
## Description
Refactors the way the visType default is handled so it's controlled by a constant defined at the top of visualiserApp.js

This also cleans up a few places where the default was being set
unnecessarily:
  - don't set it in html since we set it in javascript instantly anyway
  - remove code to set buttons to checked since this has no meaning

## Motivation and Context
In my fork of Luigi, I use svg as the graph type default. This results in annoying merges when javascript related to handling the defaults changes. In order to make different defaults easier to maintain, this uses a constant in visualiserApp.js to determine which graph type is default rather than hard-coding it everywhere the default value is used.

I ran it locally with both d3 and svg as defaults and tried loading pages. It seems to work.

### DIFF
--- a/luigi/static/visualiser/index.html
+++ b/luigi/static/visualiser/index.html
@@ -646,7 +646,7 @@
                       <div class="form-group col-md-3">
                         <label>Visualisation Type</label>
                         <div id="toggleVisButtons" class="btn-group" data-toggle="buttons">
-                            <label class="btn btn-default active">
+                            <label class="btn btn-default">
                                 <input type="radio" name="vis-type" value="d3"/> D3
                             </label>
                             <label class="btn btn-default">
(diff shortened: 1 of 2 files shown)


daveFNbuck:
While working on #2139 I discovered that the current default graph type is actually svg, despite code that attempts to make it d3. It seems we default the radio buttons to d3 but the graph behavior to svg. To minimize the effects of this change on the end user, I've added a commit that changes the default to svg. These commits now combine to fix a bug where by default you have an svg graph showing under a clicked d3 button, and you can't click the already clicked d3 button to change it.

daveFNbuck:
Note: this failed travis due to a timeout. Nothing to do with this PR as it has no Python code changes.

Tarrasch:
Finally a red diff! :)

daveFNbuck:
Hopefully I'll be done with the javascript stuff soon and I can get back to red-heavy Python diffs :)

Issue #2140, opened 2017-05-31
cdh tests too slow


daveFNbuck:
I see three recent builds failed at least partially because travis timed out while running py27-cdh or py36-cdh.
- https://travis-ci.org/spotify/luigi/builds/237706572
- https://travis-ci.org/spotify/luigi/builds/236148084
- https://travis-ci.org/spotify/luigi/builds/235193403

Tarrasch:
I'll just disable it.

daveFNbuck:
That seems a bit extreme, there are some important tests in there.

Tarrasch:
@daveFNbuck ok. I don't have time for doing any much bigger changes. But I would be fine with a pretty hack-and-slashy removal of existing tests, or any other measure to get down the time to run the tests. :)

Tarrasch:
I've been too much bothered by these tests for now. I'm writing up a completely new solution idea.

erikbern:
nice to factor out

PR #2141, opened 2017-05-31
Travis: Remove minicluster builds

Tarrasch:
This is my proposed fix for #2140.

This should greatly speed up results of Travis builds as minicluster tests were the bottleneck.
Moreover, we don't waste the nice Travis people's resources. :)


Tarrasch:
I hope nobody blames me for this, but it bothered me once again in #2130.

daveFNbuck:
Hopefully we can bring these back with #2170

PR #2148, opened 2017-06-02
Prevent unlimited resource usage

daveFNbuck:
## Description
- ensure that all tasks returned as part of a batch run have the same resources
- preserve resources of tasks at time of get_work to prevent changes from avoiding constraints

## Motivation and Context
After changing the resources for one of my jobs and rescheduling it, I saw dozens of copies of the task running at once when the resource constraint should have limited it to one copy only. After digging into the issue, I discovered that one worker would call get_work, obtain a task that uses resource_1, then another worker would call add_task and change the resource for that task to resource_2. By repeating this, we never have a task using resource_1 when we call get_work, so we can always grab another task. However, we have dozens of tasks using resource_2, well beyond its constraint.

In order to fix this, we simply store what the resources were when we applied the limit. So if we scheduled a task because it used resource_1, we remember that and consider it to be using resource_1 until it's done running. Now we won't be able to get another task that uses this resource. While we're scheduling, we will eventually pick up a task that uses resource_2, but we'll still only get one of those. So instead of running every task all at once in this scenario, we only end up with 2 tasks running.

While figuring out how this would work with batch tasks, I also decided we need to enforce a limitation that all tasks in a batch group have the same resources so we can apply that to the runner. Otherwise, it's unclear which resource the task that runs batch jobs should use.

I added unit tests and have been using this in production since October.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -1069,7 +1069,7 @@ class Scheduler(object):
             if (best_task and batched_params and task.family == best_task.family and
                     len(batched_tasks) < max_batch_size and task.is_batchable() and all(
                     task.params.get(name) == value for name, value in unbatched_params.items()) and
-                    self._schedulable(task)):
+                    task.resources == best_task.resources and self._schedulable(task)):
                 for name, params in batched_params.items():
                     params.append(task.params.get(name))
                 batched_tasks.append(task)
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -515,6 +515,7 @@ class SimpleTaskState(object):
         self.set_status(task, BATCH_RUNNING)
         task.batch_id = batch_id
         task.worker_running = worker_id
+        task.resources_running = task.resources
         task.time_running = time.time()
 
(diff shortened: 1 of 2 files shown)


daveFNbuck:
scenario 1
- worker 1 decides that Task has resource A
- you decide to change the code so Task has resource B instead
- you deploy the code and start worker 2 to schedule Task with resource B
- while this scheduling is happening, your assistant worker picks up an instance of Task, then has worker 2 change it to use resource B while it's already running, freeing up resource A for the assistant to get another task
- by repeating this, your assistant can end up running arbitrarily many instances of Task, ignoring your resource limitation

scenario 2
- worker 1 assigns resource A to Task, which is batch runnable
- you change the code so it's resource B
- you deploy fresh assistants but do not start scheduling yet
- the assistant is able to pick up batches of tasks with resource A but reports back a task with resource B
- because resource A is never used and nothing in the scheduler is limited by resource B, this can be repeated to create arbitrarily many batches of Task

daveFNbuck:
Changing resource constraints are still bad, this just limits the badness.

PR #2149, opened 2017-06-04
S3Client to use Boto3

ouanixi:
## Description
This work is to move away from boto to start using Boto3 for the S3Client.

## Motivation and Context
Boto being no longer supported, Luigi should move away from it and use the current Boto3. This would solve a number of issues for example:

- More stable downloads for big files
- Built-in support for multipart uploads
-  Encryption with KMS 
- .... 
One of the main motivation from my part is the lack of support for aws task roles in boto. As more people are using this AWS functionality, It would make sense to move the S3Client to use boto3.
More reasons can be found here: https://github.com/spotify/luigi/issues/1344

This is Work In Progress. I'm trying my best to stick to the original tests but sometimes change is inevitable (Happy to chat for more details)

## Note
This is my very first contribution so feedback and suggestions are more than welcome.


dlstadther:
Super excited for this! Thanks for taking a stab at the update, @ouanixi !

luigi/contrib/s3.py
```diff
@@ -163,350 +161,217 @@ def remove(self, path, recursive=True):
...
+        if 'encrypt_key' in kwargs:
+            raise DeprecatedBotoClientException(
+                'encrypt_key deprecated in boto3. Please refer to boto3 documentation for encryption details.')
+
         (bucket, key) = self._path_to_bucket_and_key(destination_s3_path)
 
         # grab and validate the bucket
-        s3_bucket = self.s3.get_bucket(bucket, validate=True)
+        if not self.validate_bucket(bucket):
+            self.s3.create_bucket(Bucket=bucket)
```
dlstadther: Do we really want to create a new bucket here? I'd think not.

dlstadther: If that functionality is desired, i'd lobby for an option parameter which is off by default.

luigi/contrib/s3.py
```diff
@@ -163,350 +161,217 @@ def remove(self, path, recursive=True):
...
-    def get(self, s3_path, destination_local_path):
-        """
-        Get an object stored in S3 and write it to a local path.
-        """
-        (bucket, key) = self._path_to_bucket_and_key(s3_path)
 
         # grab and validate the bucket
-        s3_bucket = self.s3.get_bucket(bucket, validate=True)
+        if not self.validate_bucket(bucket):
+            self.s3.create_bucket(Bucket=bucket)
```
dlstadther: Same concern here. (see comment within `put()`

luigi/contrib/s3.py
```diff
@@ -163,350 +161,217 @@ def remove(self, path, recursive=True):
...
         # grab and validate the bucket
-        s3_bucket = self.s3.get_bucket(bucket, validate=True)
+        s3_bucket = self.s3.Bucket(bucket)
 
         key_path = self._add_path_delimiter(key)
         key_path_len = len(key_path)
-        for item in s3_bucket.list(prefix=key_path):
-            last_modified_date = time.strptime(item.last_modified, "%Y-%m-%dT%H:%M:%S.%fZ")
+        for item in s3_bucket.objects.filter(Prefix=key_path):
+            last_modified_date = item.last_modified.date()
```
dlstadther: This removes the ability to limit dir contents by datetime.

When I added this functionality, i only needed it for dates. But perhaps someone may need it for more granular times.

ouanixi:
@dlstadther Yeah they've been removed on purpose, I'll add them again later. Of course the point is to maintain the same interface as we had before. Sorry for rushing with an MR but it was more to let you guys know someone is working on the update. 

Would it be best for me to ping you when I believe the code is ready for reviewing ? Or are you happy to feedback on the go like this ??

dlstadther:
@ouanixi Cool. Just wanted to make sure they weren't forgotten. I didn't see any explicit mention of them forthcoming in the PR description. But glad to know they will be included when complete.

I'll likely review as you push changes (and as I have time).

Thanks!

ouanixi:
@dlstadther so far, most of the functionality provided by the old interface has been tested an implemented. The remaining work is:

- implement and test `copy()` and `__copy_multipart()` (although the latter might not be needed if boto3 offers automatic multi-part like it does with upload)
- limit dir contents by datetime rather than date only (as per your comment)

Questions I have are:
- `get_key()` used to return a `Key` object which doesn't exist in boto3. Would we be OK with returning the key string representation ? Or would you prefer an `ObjectSummary` object instead ?
- The new S3Client won't have a `Key` object as attribute, would this cause any issues ?

ouanixi:
Note that I removed the S3Target tests. I'll obviously put them back once I'm finished with S3Client

dlstadther:
@ouanixi There are some cases where we only need to get back the string representation of the key and other times (i.e. list with returned key) where we need to get back the key's metadata. It'd be awesome if boto3 could still return key metadata which included size and created/last modified dates.

ouanixi:
@dlstadther Thanks for getting back to me. I've been trying to stick to the current interface we have so we stay backward compatible but AWS's replacement of the concept of Key might be problematic. A key in boto3 simply means a full path to an `Object`. In order to get metadata (more specifically the size of the object) we need to return an ObjectSummary from `get_key`. This doesn't matter much until we look at `ReadableS3File` that takes in a `Key` object (that is iterable). The closest we can get to they Key interface is a `StreamingBody` that exposes a `read` and `close` methods but it's unfortunately not iterable.
Any suggestions as to what we could do here ?

luigi/contrib/s3.py
```diff
@@ -604,27 +527,20 @@ def __init__(self, path, s3_client, **kwargs):
...
-        self.s3_client.put_multipart(self.tmp_path, self.path, **self.s3_options)
+        self.s3_client.put_multipart(
+            self.tmp_path, self.path, **self.s3_options)
 
 
 class ReadableS3File(object):
     def __init__(self, s3_key):
-        self.s3_key = s3_key
+        # This is a botocore StreamingBody object
+        self.s3_key = s3_key.get()['Body']
```
ouanixi: @dlstadther 
http://botocore.readthedocs.io/en/latest/reference/response.html#botocore.response.StreamingBody
This is not iterable :(

dlstadther:
Sorry @ouanixi for such delay getting back to you!

The currently luigi implementation only requires Key metadata for `S3Client` `list`/`listdir` methods (also used by `copy`). I'm not sure that it's a problem that `ReadableS3File` can't return this metadata.

I'm not sure i follow the need for an iterable key object for `ReadableS3File`. Do you mean fbo of updating its `__iter__` dunder method? It appears that that for iterating over the contents of the file. If boto3 doesn't offer the ability to iterate over the contents of the file by line, we deprecate `ReadableS3File`.

Personably, I don't use it. But looks like @ddaniels888 added [this functionality back in late 2013](https://github.com/spotify/luigi/commit/49b8be6449375e16c5eff9dddb50339408993d4c). Maybe he can weigh in on this matter.

ddaniels888:
Hey @dlstadther — I added the ability to iterate line-by-line for parity with the way other file-like objects behaved in luigi. I don't think we're using it personally, but I'm not sure who else uses ReadableS3File (lots of AWS luigi users around).

ouanixi:
@dlstadther Hey. Sorry for going quiet until now. It appears the guys down in botocore are working on making the `StreamingBody` object  to be iterable. I think it's worth keeping an eye on their MR for a while before jumping on making bigger changes ??? At least we won't need to deprecate anything once they release the change. What do you think ??

https://github.com/boto/botocore/pull/1195/files

ouanixi:
Hey @dlstadther This this is now ready for reviewing. I apologize there's a lot of changes but in general the main points you should be aware of are the following:

1 - S3Client has no attribute Key anymore.
2- S3Client's `get_key` returns an `ObjectSummary` object https://boto3.readthedocs.io/en/latest/reference/services/s3.html#objectsummary
3 - As a consequence, the ReadableS3File takes in an 'ObjectSummary' object as argument in the constructor.

Also as this is my first contribution, I'm not sure why the CI is failing. All of the tests pass in my local environment !!!

dlstadther:
@ouanixi I'll review this as i have time this week. 

Regarding the Travis failures, @Tarrasch merged a PR recently that allows Travis to run properly. You'll need to rebase with master to get this update.

ouanixi:
@dlstadther thank you.
Moved this back to [WIP] as I'm not sure what's happening with the CI!! Random tests are getting errors now; namely:
- `contrib.ecs_test.TestECSTask: botocore spitting `NoCredentialsError: Unable to locate credentials`
- `redshift_test.TestRedshiftManifestTask`: botocore spitting `Missing required parameter in input: "Bucket"`

I think the problem is with my understanding of tox and travis though !!
I'll wait for your input.

Cheers

brianestlin:
@ouanixi @dlstadther Any timeline on this? I would like to use it! :)

@ouanixi Did you see this comment from Dillon about the CI failures?
> You'll need to rebase with master to get this update.

ouanixi:
@brianestlin hey I did see the comments indeed. That didn't help unfortunately :(
Waiting on @dlstadther for reviewing :)

dlstadther:
Sorry guys; i haven't had time to review this. @brianestlin do you feel comfortable reviewing too?

brianestlin:
@dlstadther Maybe -- I'll try to take a look this week and let you know.

brianestlin:
@ouanixi Regarding the `NoCredentialsError`, see https://github.com/spotify/luigi/blob/master/test/contrib/ecs_test.py#L24 and L39 -- prior to your change, boto3 wasn't installed so this test was being skipped. Now that you're requiring boto3, the test is being run and tripping up on the lack of credentials in the environment. Or that's my hunch, anyway. I am not sure the right approach to make this work, whether there is some way to configure aws credentials on travis or whether we just need another way to opt this test out (maybe it should check for the existence of the credentials in that try block in L39?).

Regarding `TestRedshiftManifestTask`, I believe that is a legit test failure because the value of `client.s3` here https://github.com/spotify/luigi/blob/master/test/redshift_test.py#L64 is now a boto3 object with a different API than before.

Hope this helps.

ouanixi:
@brianestlin Thanks for your feedback.
Regarding the `NoCredentialsError`:
boto3 can get the credentials from env vars ? Maybe there's a way to expose them in Travis ??

Regarding `TestRedshiftManifestTask` The guys there are accessing the s3 property from within the S3Client, which is now a boto3 object so yeah you're right that's why it's failing. I think we should make the s3 property private so people know not to use it ?? I think it should have been from the start anyway.
I feel that the best thing to do is to edit the test not to use `S3Client` to create buckets but use boto3 directly ?? I don't think exposing internal boto3 stuff in `S3Client` is a good idea anyway !!

Don't know how @dlstadther feels about this ?

dlstadther:
I would think Travis can receive env vars through [its config](https://github.com/spotify/luigi/blob/master/.travis.yml). See [travis documentation](https://docs.travis-ci.com/user/environment-variables/).

Assuming there's no valid reason to directly access the s3 property, i'm cool making it private. (personally, i've never needed to access it directly).

brianestlin:
Hey @ouanixi , are you stuck on the travis failures again or just busy with other stuff?

ouanixi:
hey @brianestlin a bit of both to be honest :( 
I'll find some time to work on this today.

ouanixi:
Sorry for all these commits, I now setup travis for a separate branch and i will update the MR once everything passes there first.

dlstadther:
@ouanixi Any update here? (Thanks for your hard work to migrate s3client to boto3!)

ouanixi:
Hey @dlstadther sorry for the delay in replying I've had very little time lately to look into this. I have a little bit of time today so I'll try and fix the issues I'm having with Travis.
I'll give you an update by the end of today.

ouanixi:
Hey @dlstadther @brianestlin 
I just fixed the ECS tests to be mocked and they pass fine.
Did the same for test/redshift_test.py which pass fine locally but not on Travis !! I skipped them for the build and now the test coverage isn't enough :)
Could someone help out as to why redshift tests fail on Travis ???

dlstadther:
Why do we have a `test/redshift_test.py` and a `test/contrib/redshift_test.py`? The former shouldn't exist. Regardless if the test passes or not, could you move it to the contrib dir and combine with the existing test file?

ouanixi:
Hi @dlstadther this is ready to be reviewed. Any idea what the code coverage fail is about ? The commit is over the minimum coverage threshold! Can it be related to this https://github.com/codecov/support/issues/180 ?

tmattio:
@ouanixi Thank you for taking the time to do this!

Any news on when this could be reviewed/merged? I am currently trying to use the module with fake-s3 and digital ocean, and most of my problems would actually be solved by this PR :)

luigi/contrib/s3.py
```diff
@@ -163,236 +181,157 @@ def remove(self, path, recursive=True):
...
-
+        s3_bucket = self.s3.Bucket(bucket)
         # root
         if self._is_root(key):
-            raise InvalidDeleteException('Cannot delete root of bucket at path %s' % path)
-
-        # grab and validate the bucket
-        s3_bucket = self.s3.get_bucket(bucket, validate=True)
+            raise InvalidDeleteException(
+                'Cannot delete root of bucket at path %s' % path)
```
dlstadther: Reason for line break here? Is it over 80?

luigi/contrib/s3.py
```diff
@@ -163,236 +181,157 @@ def remove(self, path, recursive=True):
...
-            s3_bucket.delete_key(s3_key)
+        if self._exists(bucket, key):
+            self.s3.meta.client.delete_object(Bucket=bucket, Key=key)
             logger.debug('Deleting %s from bucket %s', key, bucket)
             return True
 
         if self.isdir(path) and not recursive:
-            raise InvalidDeleteException('Path %s is a directory. Must use recursive delete' % path)
+            raise InvalidDeleteException(
+                'Path %s is a directory. Must use recursive delete' % path)
```
dlstadther: Line break reason?

luigi/contrib/s3.py
```diff
@@ -163,236 +181,157 @@ def remove(self, path, recursive=True):
...
 
-        delete_key_list = [
-            k for k in s3_bucket.list(self._add_path_delimiter(key))]
+        delete_key_list = []
+        for obj in s3_bucket.objects.filter(Prefix=self._add_path_delimiter(key)):
+            delete_key_list.append({'Key': obj.key})
```
dlstadther: Why is this no longer a list comprehension statement?

luigi/contrib/s3.py
```diff
@@ -163,236 +181,157 @@ def remove(self, path, recursive=True):
...
 
         # delete the directory marker file if it exists
-        s3_dir_with_suffix_key = s3_bucket.get_key(key + S3_DIRECTORY_MARKER_SUFFIX_0)
-        if s3_dir_with_suffix_key:
-            delete_key_list.append(s3_dir_with_suffix_key)
+        if self._exists(bucket, key + S3_DIRECTORY_MARKER_SUFFIX_0):
+            delete_key_list.append({'Key': key + S3_DIRECTORY_MARKER_SUFFIX_0})
```
dlstadther: Can we update the string combination here? (to not use `+`)

luigi/contrib/s3.py
```diff
@@ -163,236 +181,157 @@ def remove(self, path, recursive=True):
...
+            max_concurrency=threads, multipart_chunksize=part_size)
         total_keys = 0
 
-        copy_pool = ThreadPool(processes=threads)
-
         if self.isdir(source_path):
-            # The management pool is to ensure that there's no deadlock between the s3 copying threads, and the
-            # multipart_copy threads that monitors each group of s3 copy threads and returns a success once the entire file
-            # is copied. Without this, we could potentially fill up the pool with threads waiting to check if the s3 copies
-            # have completed, leaving no available threads to actually perform any copying.
```
dlstadther: Does boto3's `copy` remove this concern?

luigi/contrib/s3.py
```diff
@@ -403,110 +342,109 @@ def copy(self, source_path, destination_path, threads=100, start_time=None, end_
...
         # grab and validate the bucket
-        s3_bucket = self.s3.get_bucket(bucket, validate=True)
+        s3_bucket = self.s3.Bucket(bucket)
 
         key_path = self._add_path_delimiter(key)
         key_path_len = len(key_path)
-        for item in s3_bucket.list(prefix=key_path):
-            last_modified_date = time.strptime(item.last_modified, "%Y-%m-%dT%H:%M:%S.%fZ")
+        for item in s3_bucket.objects.filter(Prefix=key_path):
+            last_modified_date = item.last_modified
```
dlstadther: Is `item.last_modified` returned as a string or a date?

luigi/contrib/s3.py
```diff
@@ -578,7 +472,8 @@ def _get_s3_config(self, key=None):
                 pass
         if key:
             return config.get(key)
-        section_only = {k: v for k, v in config.items() if k not in defaults or v != defaults[k]}
+        section_only = {k: v for k, v in config.items(
+        ) if k not in defaults or v != defaults[k]}
```
dlstadther: Line break?

luigi/contrib/s3.py
ouanixi: Yes it would be 90 without the line break.

ouanixi: Oops I just saw you're allowing lines up to 160. Getting rid of the line break now.

luigi/contrib/s3.py
ouanixi: Fixing now

luigi/contrib/s3.py
ouanixi: Thought it would be quite long and changed it for readability. Changing to list comprehension now.

luigi/contrib/s3.py
ouanixi: Fixing now

luigi/contrib/s3.py
ouanixi: this is a datetime object
http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Object.last_modified

luigi/contrib/s3.py
ouanixi: It is indeed. Since boto3 now handles transfers and threads for us, the idea is to delegate the work to it. I have kept the interface the same (users can choose number of threads if they want to) by creating the appropriate TransferConfig object (which is used by boto3's copy)

http://boto3.readthedocs.io/en/latest/reference/customizations/s3.html#boto3.s3.transfer.TransferConfig 

http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.copy

ouanixi:
@dlstadther Submitted the requested changes. Let me know if there's anything else you want changed. 
Ta

ouanixi:
@dlstadther what's happening with `contrib.docker_runner_test.TestDockerTask` ?
I don't remember seeing these tests before! I pulled a recent master to make sure no conflicts are created with my branch and things went mad haha

luigi/contrib/s3.py
```diff
@@ -109,21 +124,29 @@ def s3(self):
...
-        self._s3 = boto.s3.connection.S3Connection(aws_access_key_id,
-                                                   aws_secret_access_key,
-                                                   security_token=aws_session_token,
-                                                   **options)
+
+        # At this stage, if no credentials provided, boto3 would handle their resolution for us
+        # For finding out about the order in which it tries to find these credentials
+        # please see here details
+        # http://boto3.readthedocs.io/en/latest/guide/configuration.html#configuring-credentials
+        logger.debug('calling boto3 resource with access_key={}'.format(aws_access_key_id))
```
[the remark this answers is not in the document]
ouanixi: Sorry that's just something I was doing during dev. I'd say we don't need to log anything at that stage! Will remove now.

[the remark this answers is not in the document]
dlstadther: I think this could be meaningful. Particularly if an EC2 instance has a role assigned, but i'm trying to use a more restrictive role or iam.

luigi/contrib/s3.py
```diff
@@ -109,21 +124,29 @@ def s3(self):
...
+        self._s3 = boto3.resource('s3',
+                                  aws_access_key_id=aws_access_key_id,
+                                  aws_secret_access_key=aws_secret_access_key,
+                                  aws_session_token=aws_session_token,
+                                  **options)
```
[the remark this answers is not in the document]
ouanixi: Hmm not sure it's a good idea to check for args that need to be removed. You came across `calling_format` being absent from boto3 but there may be more!
How about surrounding that bit with try/catch and capture `TypeError: resource() got an unexpected keyword argument` error and throw a more meaningful one instead (`DeprecatedBotoClientException with a message telling people we're now on boto3 and they need to review what they're passing`) ?

luigi/contrib/s3.py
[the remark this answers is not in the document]
ouanixi: Ok. I'll make sure we're logging what's happening there then :)

luigi/contrib/s3.py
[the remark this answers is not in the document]
dlstadther: I definitely agree that removing args is not a good idea. I was just writing my thought process.

Personally, I don't have experience deprecating code like this. So i'll defer to your judgement or we can get @Tarrasch 's thoughts, given he's the executor of Luigi.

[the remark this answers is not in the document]
ouanixi: I went ahead and changed this to throw a deprecation exception. Open for suggestions if anyone thinks we should do it differently.

dlstadther:
Finally getting back to verifying this PR!

Now verified `exist` and cross-region access (boto2 with OrdinaryCallingFormat without explicitly specifying an aws region causes issues)

I'm now performing some maintenance on my s3 tasks which use these various methods. So i should (hopefully) complete this review soon!

Thanks again @ouanixi !

luigi/contrib/s3.py
```diff
@@ -553,16 +416,57 @@ def mkdir(self, path, parents=True, raise_if_exists=False):
...
+        """
+        (bucket, key) = self._path_to_bucket_and_key(path)
+
+        # grab and validate the bucket
+        s3_bucket = self.s3.Bucket(bucket)
+
+        key_path = self._add_path_delimiter(key)
+        key_path_len = len(key_path)
+        for item in s3_bucket.objects.filter(Prefix=key_path):
+            last_modified_date = item.last_modified
```
dlstadther: Mostly a note for myself - and maybe worth updating the docstring... `last_modified_date` is now a datetime with timezone. So `start_time`/`end_time` must now also be offset-aware datetimes (or `None`).

honnix:
I will take a look in depth later. Since you have reviewed and verified I don’t see any major problem. Maybe fix the docstring you pointed out and Travis (didn’t check why it is still red).

dlstadther:
Thanks for the quick response @honnix 

Looks like batch.py might also need a boto3 update? Could you investigate further @ouanixi ? Perhaps start with updating branch with master (not sure if there have been relevant updates).

```
  File "/home/travis/build/spotify/luigi/test/contrib/batch_test.py", line 103, in setUp
    self.bc = batch.BatchClient(poll_time=10)
  File "/home/travis/build/spotify/luigi/luigi/contrib/batch.py", line 94, in __init__
    self._queue = self.get_active_queue()
  File "/home/travis/build/spotify/luigi/luigi/contrib/batch.py", line 100, in get_active_queue
    queues = {q['jobQueueName']: q for q in self._client.describe_job_queues()['jobQueues']
  File "/home/travis/build/spotify/luigi/.tox/py27-nonhdfs/lib/python2.7/site-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/travis/build/spotify/luigi/.tox/py27-nonhdfs/lib/python2.7/site-packages/botocore/client.py", line 612, in _make_api_call
    raise error_class(parsed_response, operation_name)
ClientError: An error occurred (UnrecognizedClientException) when calling the DescribeJobQueues operation: The security token included in the request is invalid.
```

honnix:
Another thing I didn’t check yet. Is this a breaking change? Does it require user code change? Mostly thinking about what we need to highlight in release note.

dlstadther:
@honnix The only breaking change for me in passing datetimes with timezones to `listdir` instead of `time_struct`s. Since I'm the one who added that feature, i'm not even sure if anyone else uses it.

I'm not intimately aware of any other breaking changes

ouanixi:
Hi @dlstadther thanks for your feedback.

I've had a little bit of time to look into this earlier on and found that the batch tests aren't mocking the batch client. They pass fine because they skipOnTravis when `boto3` cannot be imported but of course now with my changes, it can!

This is not the first time this happened (I've had to change tests for redshift, ecs, and now batch) and I feel maybe we should be stricter when accepting PR's using this pattern of testing ?

I'll try and find some time to fix the failing batch tests this weekend hopefully :)

Thanks again for your help

luigi/contrib/s3.py
```diff
@@ -403,145 +352,59 @@ def copy(self, source_path, destination_path, threads=100, start_time=None, end_
...
-        :param key_size: size of the key to copy in bytes
-        :param part_size: Part size in bytes. Must be >= 5MB and <= 5 GB.
-        :param kwargs: Keyword arguments are passed to the boto function `initiate_multipart_upload`
-        """
+            copy_source = {
+                'Bucket': src_bucket,
+                'Key': src_key
+            }
+            self.s3.meta.client.copy(
+                copy_source, dst_bucket, dst_key, Config=transfer_config, ExtraArgs=kwargs)
```
dlstadther: I've been using this for a couple days and it is taking drastically longer to copy lots of small files.

I'm trying to understand this multi-threaded logic for boto3.

Previously, we were assigning async threads to copy individual files. This new version appears to be assigning possibly multiple threads to copy portions of a single file. Am I reading this correctly?

dlstadther: https://github.com/boto/boto3/issues/787

dlstadther: I'm not sure if i'm possibly trying to supply too many thread. So i'm decreasing my thread count and will report back with results when i have them.

dlstadther: Regardless of thread count, i'm getting 6.67 files copied per second.

I'm skeptical that the current threading implementation provides any benefit for the copying of lots of small files.

dlstadther: Re-enabled `ThreadPool` logic for using `apply_async` and actually multithreaded copy works again.

Will submit a PR.

PR #2151, opened 2017-06-07
Update resources from the visualiser

daveFNbuck:
## Description
Update individual resource counts from the visualiser, like we can do with number of workers now.

This works exactly like updating the worker counts in the workers tab, except that the counts are referenced in the progress bar rather than on the button itself. This is because we already have the counts in the progress bar, so adding them to a button would only complicate things.

The only other difference between this and the worker counts is that this allows resources to be set to 0, effectively pausing certain kinds of jobs.

Note that when you update the resource counts using the visualiser, it only affects the currently running scheduler. If you restart the scheduler it will reload the resource counts from the config.

Screenshots:

image

image

image

## Motivation and Context
Sometimes you need to quickly tweak a resource in response to your jobs falling behind or your cluster getting overloaded. In order to make such tweaks a little easier, this adds the option to change the limits for all resources stored in the scheduler right in the visualiser.

Use case 1: you notice that your pipeline is backed up and is currently being constrained by the mapreduce resource. You realize that the mapreduce resource was set a little arbitrarily so you want to try raising its maximum count. You can now quickly do this without having to restart the scheduler.

Use case 2: there is a production fire where the redis cluster keeps going down. While debugging, the backend team requests that you stop all redis uploads from the data pipeline. Now you can do that without having to pause the whole pipeline or restart the scheduler.

I added a test, verified the js both locally on this branch and in production since last week.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -928,6 +928,13 @@ class Scheduler(object):
             self._resources = {}
         self._resources.update(resources)
 
+    @rpc_method()
+    def update_resource(self, resource, amount):
+        if not isinstance(amount, int) or amount < 0:
+            return False
+        self._resources[resource] = amount
+        return True
+
     def _generate_retry_policy(self, task_retry_policy_dict):
         retry_policy_dict = self._config._get_retry_policy()._asdict()
         retry_policy_dict.update({k: v for k, v in six.iteritems(task_retry_policy_dict) if v is not None})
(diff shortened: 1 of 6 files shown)


daveFNbuck:
Added some screenshots. The UI does not indicate what happens if you put 0, but I think it's fairly intuitive that it should mean no resources. I didn't consider skipping the incr and decr buttons because I think consistency is as important as simplicity. Incrementing and decrementing by 1 are probably the most common use cases anyway, and it's nice to have a shortcut that skips a the extra step of opening a dialog box.

Tarrasch:
Sorry for merging this a bit late. This is a nice addition. :)

daveFNbuck:
No worries, we all get busy.

PR #1892, opened 2016-10-21
docs: Add a roadmap

Tarrasch:
## Description

Add a roadmap. A description of the roadmap is given inside the roadmap markdown file.

PR #2157, opened 2017-06-14
Exclude insignificant parameters of tasks in error email

henrytxz:
## Description
1 line change to exclude insignificant parameters from error emails.

## Motivation and Context
Some Luigi tasks have password as parameters, I would have 
password = luigi.Parameter(significant=False) but the notifications module is emailing out insignificant params.

### DIFF
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -426,7 +426,7 @@ def format_task_error(headline, task, command, formatted_exception=None):
         {traceback}
         ''')
 
-        str_params = task.to_str_params()
+        str_params = task.to_str_params(only_significant=True)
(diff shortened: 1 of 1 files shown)


henrytxz:
@Tarrasch, 1 line change when you get a chance. Thanks!

dlstadther:
The idea here seems reasonable to me.

Is there any situation where you may want to email an insignificant parameter? If so, can we have this be configurable?

Tarrasch:
I link to this [parapraph](https://github.com/spotify/luigi/pull/1892/files#diff-f5f076fb9da1107cb1221ae176bbc955R91). There's been many similar attempts to "solve" this. But the community hasn't committed to any one solution yet

PR #2165, opened 2017-06-27
Serialize statuses array for URL hash

daveFNbuck:
## Description
When updating the status hash during tab switching, use JSON to stringify the status array so it works properly.

## Motivation and Context
When switching tabs away from Task List and back with statuses selected, the url hash ends up with something like `statuses=FAILED` rather than `statuses=%5B%22FAILED%22%5D`, which is the json-serialized array. This causes a json parsing error when trying to deserialize the expected json.

Switched back and forth between tabs with statuses selected. Before the change, this would always break the visualiser. After the change, it works as expected.


Tarrasch:
Thanks! :)

Issue #2174, opened 2017-07-10
Luigi master is falling down due to JSON serialisation errors in task state output.


mrjj:
When task error output containing non JSON-serializable symbols Luigi master is failing.

We are getting following error:

finished run
number of examples per pass = 133
passes used = 4
weighted example sum = 532.000000
weighted label sum = -4.000000
average loss = 0.678752 h
best constant = -0.015038
best constant's loss = 0.693119
total feature number = 368952
WARNING: model file has set of {-q, --cubic, --interactions} settings stored, but they'll be OVERRIDEN by set of {-q, --cubic, --interactions} settings from command line.
XXX:/var/log$ cat ./XXX/run.log.1 | grep -C 10 json\.dump
Traceback (most recent call last):
  File "/var/lib/XXX", line 277, in <module>
    success = luigi.build([MainTask(dt)], workers=10, scheduler_port=int(config.LUIGID_PORT))
  File "/usr/share/XXX/local/lib/python2.7/site-packages/luigi/interface.py", line 263, in build
    return _schedule_and_run(tasks, worker_scheduler_factory, override_defaults=env_params)['success']
  File "/usr/share/XXX/local/lib/python2.7/site-packages/luigi/interface.py", line 196, in _schedule_and_run
    success &= worker.run()
  File "/usr/share/XXX/local/lib/python2.7/site-packages/luigi/worker.py", line 1054, in run
    self._handle_next_task()
  File "/usr/share/XXX/local/lib/python2.7/site-packages/luigi/worker.py", line 950, in _handle_next_task
    expl=json.dumps(expl),
  File "/usr/lib/python2.7/json/__init__.py", line 243, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 201, in encode
    return encode_basestring_ascii(o)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xb4 in position 7091: invalid start byte
--- SOME OUTPUT CONTAINING "BAD" SYMBOLS ---

I propose to make controlled exception catch around JSON serialisation with fallback to the latin1 encoding that is completely free from this kind of error.

Tarrasch:
Can you please:

 * fix the formatting of your issue (otherwise we'll have to close it per recent [policy](https://raw.githubusercontent.com/spotify/luigi/master/ISSUE_TEMPLATE.md))
 * add a runnable small example reproducing this?

Also remove any unnecessary information to this issue please.

PR #2175, opened 2017-07-10
#2174

mrjj:
Task metadata json serialisation failure due invalid unicode symbols is fixed.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -971,10 +971,18 @@ class Worker(object):
                     self.add(t)
                 new_deps = [t.task_id for t in new_req]
 
+            try:
+                expl_json = json.dumps(expl)
+            except UnicodeDecodeError:
+                logger.info('Unicode decode error occur during %s task metadata serialisation, '
+                            'falling back to latin1 encoding.', task_id)
+
+                expl_json = json.dumps(expl, encoding='latin1')
+
             self._add_task(worker=self._id,
                            task_id=task_id,
                            status=status,
-                           expl=json.dumps(expl),
+                           expl=expl_json,
                            resources=task.process_resources(),
                            runnable=None,
                            params=task.to_str_params(),


Tarrasch:
Can you accompany this with a test case that failed before your patch, but now passes?

luigi/worker.py
```diff
@@ -971,10 +971,18 @@ def _handle_next_task(self):
                     self.add(t)
                 new_deps = [t.task_id for t in new_req]
 
+            try:
+                expl_json = json.dumps(expl)
+            except UnicodeDecodeError:
+                logger.info('Unicode decode error occur during %s task metadata serialisation, '
+                            'falling back to latin1 encoding.', task_id)
+
+                expl_json = json.dumps(expl, encoding='latin1')
```
ikalnytskyi: Hm.. I'm not sure that proposed solution is good enough to go with due to number of reasons.

* First, according to [RFC 7159](https://tools.ietf.org/html/rfc7159#page-9), JSON text SHALL be encoded in UTF-8, UTF-16, or UTF-32. Encoding it in `latin1` may break consumers one day.

* Second, what do you expect to see if you pass Python 2 `str` (which is bytes) to `json.dumps()`? Obviously, bytes may contain unserializable characters. Instead, I think it would make more sense to craft a `unicode` (in Python 2) object instead. Can we figure out why exactly `expl` is `str` in py2?

mrjj: Ihor, this is a fallback, accidentally any dirt is possible in task metadata respecting any RFC or not, but this should not be the reason of DOS Luigi master.

json.dumps of the string = string, i don't clearly understand what other luigi contributors who made making expl handling are expecting so output processed exactly the same way as before except sustaining to the bad symbols and patch designed to be as non-invasive as possible.

Tarrasch: Perhaps keep the the idea of not crashing by simply returning a string like `"Internal Parsing Error"` as the expl string instead of doing something like assuming latin1 which I suppose can also crash.

Issue #1984, opened 2017-01-11
Allow schema to be specified for target table in sqla.CopyToTable


sd2k:
This could be supported by passing overriding the `schema` attribute of the task, and passing this to the `sqlalchemy.MetaData` constructor in `sqla.CopyToTable.create_table`.

A consideration: should we also allow a schema to be specified for the marker table? My current use case requires them to be different.

If this sounds OK I'll write it up later and create a PR for review.

dlstadther:
@sd2k Any update here?

sd2k:
@dlstadther I've been using a subclass of `CopyToTable` which overrides `create_table`. I'll send a PR.

PR #2176, opened 2017-07-10
Allow schema to be specified for target table in sqla.CopyToTable

sd2k:
## Description
When using an RDBMS that supports schemas, this allows target tables to
exist in a non-default schema, specified using the 'schema' attribute of
the task class.

## Motivation and Context
See #1984.

I have been running my jobs with this code for several months and it has worked for me.
The unit tests use SQLite which doesn't support schemas, so I haven't included any unit tests.

### DIFF
--- a/luigi/contrib/sqla.py
+++ b/luigi/contrib/sqla.py
@@ -273,6 +273,8 @@ class CopyToTable(luigi.Task):
     Usage:
 
     * subclass and override the required `connection_string`, `table` and `columns` attributes.
+    * optionally override the `schema` attribute to use a different schema for
+      the target table.
     """
     _logger = logging.getLogger('luigi-interface')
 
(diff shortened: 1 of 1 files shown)


luigi/contrib/sqla.py
```diff
@@ -328,15 +334,37 @@ def construct_sqla_columns(columns):
...
+                    row = curs.fetchone()
+                    if row is not None:
+                        self.logger.info(
+                            'Schema {} already exists, not creating'.format(
+                                self.schema,
+                            )
+                        )
+                    if not row:
+                        self.logger.info('Creating schema {}'.format(self.schema))
+                        engine.execute(sqlalchemy.schema.CreateSchema(self.schema))
```
dlstadther: Does a situation exist where you would intend for this to create a new schema? I can understand table creation in an existing schema, but cannot think of a case where I'd intentionally specify a schema name in my ETL for which one does not yet exist.

I worry that this may undesirably fail silently where someone has a typo in their schema name.

sd2k: Yes, it seems quite reasonable to ask users to create the schema separately to avoid this kind of silent failure. I'll remove that part and add a note.

Tarrasch:
Thanks @sd2k!

PR #2181, opened 2017-07-12
breaking: Remove deprecated contrib redirections

Tarrasch:
## Description

After this commit the following imports will no longer work

    import luigi.s3
    import luigi.postgres

Instead use

    import luigi.contrib.{s3,postgres}

This is a continution on spotify/luigi#1997

## Motivation and Context

This will make it clear that these modules should not be considered core luigi. 

I ran some tox-tests locally. But I mainly rely on Travis here.


Tarrasch:
>  I think the merge of this PR should result in release 2.7. Thoughts?

Sounds good to me. I guess people will be more careful upgrading to a minor bump. Plus we've had many other nice contributions since last release. :)

dlstadther:
@Tarrasch Thoughts on when to merge this? (and release a new version)

dlstadther:
@Tarrasch Following up here regarding merging this PR (and maybe a few other small ones) and releasing the next large Luigi Version. Thoughts?

Tarrasch:
@dlstadther totally agree, I've just been a bit lazy. Sorry about that. :)

Issue #520, opened 2014-11-03
mkdir error when using HdfsClientCdh3


imazor:
Hi, 

I am using the following config:

[hadoop]
version: apache1
streaming-jar: /opt/mapr/hadoop/hadoop-2.4.1/share/hadoop/tools/lib/hadoop-streaming-2.4.1-mapr-1408.jar
python-executable: /usr/bin/python

[hdfs]
client: hadoopcli

When executing the following task:

class Streams(Task):
    ''' Faked version right now, just generates bogus data.
    '''
    date = DateParameter()

```
def run(self):
    with self.output().open('w') as output:
        for i in xrange(1000):
            output.write('{a} {b} {c}\n'.format(
                a=random.randint(1, 999),
                b=random.randint(1, 999),
                c=random.randint(1, 999)))

def output(self):
    return hdfs.HdfsTarget(self.date.strftime('/user/user_name/hadoop/learn_luigi/data/streams_%Y_%m_%d_faked.tsv'))
```

I am getting the following error:

Runtime error:
Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/luigi/worker.py", line 76, in run
    self.task.run()
  File "top_artists.py", line 19, in run
    with self.output().open('w') as output:
  File "/usr/lib/python2.6/site-packages/luigi/hdfs.py", line 708, in open
    return self.format.pipe_writer(HdfsAtomicWritePipe(self.path))
  File "/usr/lib/python2.6/site-packages/luigi/hdfs.py", line 602, in **init**
    mkdir(parent_dir, parents=True, raise_if_exists=False)
TypeError: mkdir() got an unexpected keyword argument 'parents'

looks like based on my config, the hdfs client is HdfsClientApache1() which inherits from 
HdfsClientCdh3.

the mkdir function defined for the parent class taking only 1 arg : path
however while trying to open hdfs target (using HdfsAtomicWritePipe), the **init** try to execute the following:
mkdir(parent_dir, parents=True, raise_if_exists=False)
as a result the error.

Also would be nice if  -p option will be added to the mkdir in HdfsClientCdh3.

BTW, is there preferred config for the hdfs client (in the config file) for using Mapr ?

Thanks.

mfcabrera:
I'm closing this issue. 

- [x] It has been inactive for +4 months.
- [ ] It's not about luigi core, so not as many users are affected about this.
- [ ] The change seems quite big, it's unlikely to be sporadically picked up.
[...]

PR #2184, opened 2017-07-17
Update mkdir method signature for CDH3 client.

anjackson:
## Description
As of 2.6.2 `mkdir` fails when running on CDH3 and using the `hadoopcli` client.

## Motivation and Context
We're currently running Luigi on CDH3 and just hit this very old issue: #520 

Yes, this change has been patched in locally and our tasks now run as expected.


Tarrasch:
Thanks!

PR #2186, opened 2017-07-18
MongoDB target items count

kritchie:
## Description
Added item count to the MongoDB target using aggregate method.
https://docs.mongodb.com/manual/reference/method/db.collection.count/#behavior

## Motivation and Context
I had a use case where I store data from a CSV file in a mongodb collection for later use. 
The CSV as a very specific number of lines and I don't need to redo the pre-processing each time I re-run the pipeline.

### DIFF
--- a/luigi/contrib/mongodb.py
+++ b/luigi/contrib/mongodb.py
@@ -161,3 +161,30 @@ class MongoRangeTarget(MongoTarget):
         )
 
         return set(self._document_ids) - set([doc['_id'] for doc in cursor])
+
+
+class MongoCountTarget(MongoTarget):
+
+    """ Target for a level 0 field in a range of documents """
+
+    def __init__(self, mongo_client, index, collection, target_count):
+        """
+        :param target_count: Value of the desired item count in the target
+        :type field: int
+        """
+        super(MongoCountTarget, self).__init__(mongo_client, index, collection)
+
+        self._target_count = target_count
+
+    def exists(self):
+        """
+        Test if the target has been run
+        Target is considered run if the number of items in the target matches value of self._target_count
+        """
+        return self.read() == self._target_count
+
+    def read(self):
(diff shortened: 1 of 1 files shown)


Tarrasch:

kritchie:
Sure, looks like @MathiasDesch is the original contributor of this target. Mathias if you have a chance to look this up, would greatly appreciate !

MathiasDesch:
Hi @kritchie,
the overall PR looks good to me. Just wondering  : 

Is there any specific reason to `return -1` if the collection does not exist?

I would rather see the logic of testing if the collection exists in charge of the client of this class. I feel that this task should only check the count and raise an exception if the collection is not present.

kritchie:
Hi @MathiasDesch , thanks for taking time to look this up. 

The only reason I used the `return -1` was that when a collection doesn't exists, an exception would have been thrown, thus stopping the whole pipeline together because of that exception. This was unfortunate because my task was the one creating the said collection in the run method. So it was never going to be ran because of that. 

That being said, I get your point, it makes total sense. How about that, I will remove the check for the existence of the collection from the MongoCountTarget and I will add a new target in the file which would be a MongoCollectionTarget which would have the responsibility to verify if a collection exists or not. It would contain the logic to verify that a collection exists the same way I was doing it inside the MongoCountTarget.

With that new target I could add a prior task that will create the collection if it doesn't exist or simply skip if the collection already exists.

What do you think?

MathiasDesch:
Perfect ! 👍

kritchie:
@MathiasDesch I applied the changes and I've tested it with my tasks. It working fine on my side. If you can confirm on your side that would be awesome !

MathiasDesch:
Everything looks good to me. @Tarrasch I think you can merge.
Thanks @kritchie !

PR #2188, opened 2017-07-20
Fix CDH3 mkdir implementation

anjackson:
## Description
Complete the implementation of work started under #2184.

## Motivation and Context
After updating the method signature in #2184, while running in production it became obvious I'd neglected to implement the `raise_if_exists` option. Running `mkdir` on an existing directory was always failing, even when `raise_if_exists=False`.

Yes, this version has been run on our production system.


Tarrasch:
Thanks! :)

PR #2189, opened 2017-07-21
Add worker config check_unfulfilled_deps.

riga:
## Description

When a worker is about to run a task, it checks whether its dependencies are complete. This PR adds a configuration parameter which makes that behavior optional.

## Motivation and Context

In our workflow, we use remote targets that are stored in various locations on the [WLCG](http://wlcg.web.cern.ch/). Some of our tasks have O(1000) dependencies with O(10) output targets each. When running these tasks, the dependency tree is built by checking if the targets exist, which results in network load. But before a worker actually *runs* a task, [it checks its dependencies again](https://github.com/spotify/luigi/blob/master/luigi/worker.py#L167). By skipping this, we could significantly reduce the network load. Therefore, I added a `[worker]` config parameter `check_unfulfilled_deps ` to alter this behavior.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -111,7 +111,7 @@ class TaskProcess(multiprocessing.Process):
     Mainly for convenience since this is run in a separate process. """
 
     def __init__(self, task, worker_id, result_queue, status_reporter,
-                 use_multiprocessing=False, worker_timeout=0):
+                 use_multiprocessing=False, worker_timeout=0, check_unfulfilled_deps=True):
         super(TaskProcess, self).__init__()
         self.task = task
         self.worker_id = worker_id
@@ -121,6 +121,7 @@ class TaskProcess(multiprocessing.Process):
             worker_timeout = task.worker_timeout
         self.timeout_time = time.time() + worker_timeout if worker_timeout else None
         self.use_multiprocessing = use_multiprocessing or self.timeout_time is not None
+        self.check_unfulfilled_deps = check_unfulfilled_deps
 
     def _run_get_new_deps(self):
         self.task.set_tracking_url = self.status_reporter.update_tracking_url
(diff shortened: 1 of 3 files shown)


luigi/worker.py
```diff
@@ -910,7 +914,8 @@ def _create_task_process(self, task):
         return TaskProcess(
             task, self._id, self._task_result_queue, reporter,
             use_multiprocessing=bool(self.worker_processes > 1),
-            worker_timeout=self._config.timeout
+            worker_timeout=self._config.timeout,
+            check_unfulfilled_deps=self._config.check_unfulfilled_deps
```
Tarrasch: Can you add a trailing comma here? Next line addition will look nicer in the diffs. :)

riga: Yep, done.

riga:
Kind of. Depending on the current load at some computing and storage facilities, the completeness checks can be slow. But usually, the number of requests is the bottle neck.

A typical use case for this feature is a tree of M tasks which all have the same N dependencies. When the completeness check of those N deps is load-intensive (say 1k network requests per dep), this N x 1k requests are executed once for tree building and repeated M times which we like to skip by disabling the check. Of course, we have to make sure that the targets of the deps don't change during the processing of the tree.

btw: Do you know what's wrong with travis?

Tarrasch:
I think this patch looks reasonable. Two thoughts:

 * Would it make sense to include a warning in docs saying that one should only use it when the bottleneck is known to be the complete checks?
 * Would it make sense to only enable this per tasks rather than a Worker?

> btw: Do you know what's wrong with travis?

I think they changed Travis, and we having magic caused failures.

riga:
> Would it make sense to include a warning in docs saying that one should only use it when the bottleneck is known to be the complete checks?

Yes, that makes sense. The more docs the better ;) I'll add one or two sentences.

> Would it make sense to only enable this per tasks rather than a Worker?

In our use case, we would use the worker config, but other users might prefer a per-task setting. What about using both? Of course, the task setting would have precedence.

Tarrasch:
By the way, since I merged https://github.com/spotify/luigi/pull/2190, maybe the build issues are resolved.  :)

dlstadther:
@Tarrasch even after manually restarting Travis, build continues to exit

Tarrasch:
> @Tarrasch even after manually restarting Travis, build continues to exit

@dlstadther, that's expected if you didn't rebase upon master. As the Travis reads the config from the checked in `.travis.yml` file.

riga:
> However, I do kindly request to add at least one test case before we merge this. :)

Done :)

There's still one failing travis job, but this is unrelated to my code, right?

Tarrasch:
Thanks @riga :)

PR #2191, opened 2017-07-29
Fix emailing of unexpected error in `Worker._announce_scheduling_failure`.

Stranger6667:
## Description
This PR fixes a small issue with error reporting.

## Motivation and Context
The code after ``raise`` is unreachable and ``_email_unexpected_error`` will never be executed in this case. It seems like a typo.

This code branch is not tested (as I see in coverage info), so it should not affect any builds. But I can include a test case for this if needed.

Kind Regards

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -572,9 +572,9 @@ class Worker(object):
                 owners=task._owner_list(),
             )
         except Exception:
-            raise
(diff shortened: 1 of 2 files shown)


Stranger6667:
Done! :)

Tarrasch:
Thanks! :)

PR #2196, opened 2017-08-01
Allow spark task to load config from different version

interskh:
## Description
Allow spark task to load config from different version

## Motivation and Context
In our use cases, we want to support multiple version of spark, or have one config for spark and one for pyspark. This change allows us to read config from different section when `spark_version` is set

We have ran this on our production for some time.


dlstadther:
I see nothing wrong with this update. @interskh , is this still running well in production for you?

Tarrasch:
I like the cleverness going on here. :)

PR #2197, opened 2017-08-02
Support cluster mode in PySpark

interskh:
## Description
Support cluster mode in PySpark

## Motivation and Context
We want to use cluster mode for pyspark like spark tasks.

We run it on production for a couple of months.

### DIFF
--- a/luigi/contrib/spark.py
+++ b/luigi/contrib/spark.py
@@ -237,8 +237,6 @@ class PySparkTask(SparkSubmitTask):
 
     # Path to the pyspark program passed to spark-submit
     app = os.path.join(os.path.dirname(__file__), 'pyspark_runner.py')
-    # Python only supports the client deploy mode, force it
-    deploy_mode = "client"
 
     @property
     def name(self):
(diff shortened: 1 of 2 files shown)


luigi/contrib/spark.py
```diff
@@ -237,8 +237,6 @@ class PySparkTask(SparkSubmitTask):
 
     # Path to the pyspark program passed to spark-submit
     app = os.path.join(os.path.dirname(__file__), 'pyspark_runner.py')
-    # Python only supports the client deploy mode, force it
-    deploy_mode = "client"
```
dlstadther: Was this an intentional deletion? Why not just allow overwrite of `deploy_mode`?

interskh: `deploy_mode = "client"` overwrites `deploy_mode` in `SparkSubmitTask`. It was here to force deploy mode to be `client` since previously cluster deploy mode wasn't supported. Now that we do support it, there is no need to pin it to `client` only.

interskh:
@jthi3rry @ivannotes @ntim How do you like this PR? I tagged you here because you have contributed to pyspark class :)

interskh:
@dlstadther @Tarrasch I remember there used to be some bot that automatically tag ppl for reviews. What happened to it?

ntim:
👍 Looking forward ti try it!

jthi3rry:
LGTM!

Tarrasch:
Thanks @interskh!

PR #2205, opened 2017-08-05
Added support to specify the number of processes for parallel scheduling

chris-zen:
## Description
This PR adds support to specify the number of processes to use for parallel scheduling, by introducing a new configuration parameter under the `[core]` section called `parallel-scheduling-processes`. It allows to specify the number of processes to use. If it is not specified, then the number of CPUs available will be used (previous behaviour).

## Motivation and Context
The Luigi scheduler is able to compute complete functions of tasks in parallel using a [multiprocessing Pool](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.pool.multiprocessing.Pool) when the configuration file specifies `parallel-scheduling` as `True` in the `[core]` section. Such Pool allows to specify the number of processes to use, but if it is not specified it takes the number of CPUs available from [cpu_count()](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.cpu_count). The Luigi scheduler doesn't specify any value for the number of processes when instantiating the Pool, nor allows to configure it from the configuration file. Most of the time the complete logic is IO bound (instead of CPU bound), and in pipelines having thousands of tasks running in containers with a low number of CPUs this can be a real problem.
For example we saw that a pipeline that use to take 20 minutes to check for the completed tasks in a host with 16 CPUs, took several hours when moved to a Mesos cluster where the containers have 1 CPU by default. Given that the hosts have a maximum number of CPUs of 8 (the more CPUs selected the less chances for the pipeline to be scheduled in the cluster on time), and the pipeline is scheduled every hour, our only chance to run it is by being able to use a number of processes greater than the actual number of CPUs available in the container.

In `interfaces_test.py` a test suite is added to make sure that the new configuration parameter behaves as expected for different scenarios (specified by the user, or not specified)
In `worker_parallel_scheduling_test.py` a test checks that the number of processes specified to the worker `add()` method is used when instantiating the `multiprocessing.Pool`.
Checked in a development environment for a real case, and it decreased significantly the time spent on the scheduling using a container with a single CPU.


dlstadther:
@Tarrasch Do you also approve of this PR?

luigi/interface.py
```diff
@@ -122,6 +122,10 @@ class core(task.Config):
     parallel_scheduling = parameter.BoolParameter(
         default=False,
         description='Use multiprocessing to do scheduling in parallel.')
+    parallel_scheduling_processes = parameter.IntParameter(
+        default=None,
```
Tarrasch: I think this will cause a warning, can you set the default to be the value to be zero (`0`)?

chris-zen:
@Tarrasch thanks for the feedback. I just came back from holidays and will start working on your suggestions this week.

About having a new scheduling section, it seems to me that it would be out of scope for this PR. I propose to introduce the config param first (this PR), and then later on introduce the new scheduling section (which may involve backward incompatibilities and minor version change if I understood it well).

chris-zen:
@Tarrasch please could you review ? I changed the default value from `None` to `0`.
I also checked the feature for a real project (with 10600 tasks) in a development environment and everything worked fine.

dlstadther:
In @Tarrasch 's busyness, I will merge. Sorry for such delay in merging. Thanks and congrats on your first Luigi contribution!

Tarrasch:
> About having a new scheduling section, it seems to me that it would be out of scope for this PR. I propose to introduce the config param first (this PR), and then later on introduce the new scheduling section (which may involve backward incompatibilities and minor version change if I understood it well).

Good point! Keep it simple and get the PRs merged is indeed right way. Good job!

PR #2207, opened 2017-08-10
Prevent worker from running same task repeatedly

daveFNbuck:
## Description
When get_work returns a task_id already in the _running_tasks list, don't start running a new task.

## Motivation and Context
If the scheduler keeps giving the worker the same task_id from get_work, the worker will just overwrite the same entry in _running_tasks and never think that it is running too many processes.

I've had to reboot one of my worker machines almost daily due to hundreds of worker processes being spawned to run the same task. This seems to be mostly happening with batch tasks, probably because they require more rounds of communication before the scheduler understands
that the task is running. Since this tends to happen only in short bursts, I've also added a sleep to give the scheduler a chance to recover.

This is still a bit worrisome because the scheduler could potentially give the task to other workers, but it at least patches over the bug enough that my pipelines can run smoothly again. My hope is that the issue is only with the scheduler re-issuing task_ids that it already assigned to that worker but the worker isn't including in its current_tasks list in get_work, in which case the scheduler will not send the task to other workers.

I simulated a buggy scheduler both by tweaking the scheduler RPC functions and by tweaking the worker's RPC calls to trigger this bug reliably for batch tasks by not registering the batch runner. In both cases, this fix does the trick.

I've also been running this in production for a couple of days and have included unit tests.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -896,6 +896,11 @@ class Worker(object):
         )
 
     def _run_task(self, task_id):
+        if task_id in self._running_tasks:
+            logger.debug('Got already running task id {} from scheduler, taking a break'.format(task_id))
+            next(self._sleeper())
+            return
+
         task = self._scheduled_tasks[task_id]
 
         task_process = self._create_task_process(task)
--- a/test/worker_test.py
+++ b/test/worker_test.py
@@ -1543,6 +1543,26 @@ class MultipleWorkersTest(unittest.TestCase):
             self.assertTrue(is_running())
         self.assertFalse(is_running())
 
+    @mock.patch('luigi.worker.time')
+    def test_no_process_leak_from_repeatedly_running_same_task(self, worker_time):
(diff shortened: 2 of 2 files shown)


daveFNbuck:
Yeah I'm still investigating the cause.

daveFNbuck:
I've found a somewhat satisfying explanation for this. It seems to occur when exceptions are thrown. I had a few common uncaught exceptions happening in the scheduler (PRs upcoming). If exceptions are being thrown in the scheduler during get_work, it makes sense that the scheduler might end up with a weird internal state.

Since fixing the uncaught exceptions, I've only seen this error happen twice. Both times, it occurred at about the same time that the keep-alive thread raised a KeyError trying to look up `response["rpc_messages"]`. Not sure how this can happen, maybe some wires are getting crossed and responses are being read by the wrong thread? I'll try to catch the error and log the response in case it happens again.

Tarrasch:
Thanks @daveFNbuck

PR #2208, opened 2017-08-11
Clear tasks on disabled workers

daveFNbuck:
## Description
Clear worker task lists on disable to ensure a consistent internal state.

## Motivation and Context
One of the uncaught exceptions I was seeing in my scheduler recently was caused by an exception trying to peek at the worker list of a task with empty worker list during count_pending. I solved this in my fork by ignoring the exception, but I think I've now found the real cause.

After a worker is marked disabled, it will call get_work or count_pending one last time before stopping. This call can fail on count_pending when it has a unique task, as that task will now have no
workers but we try to look up the workers from each task.

Included unit tests, but I haven't tested this particular fix in production. Should we catch the exception too to be safe?

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -643,7 +643,9 @@ class SimpleTaskState(object):
     def disable_workers(self, worker_ids):
         self._remove_workers_from_tasks(worker_ids, remove_stakeholders=False)
         for worker_id in worker_ids:
-            self.get_worker(worker_id).disabled = True
+            worker = self.get_worker(worker_id)
+            worker.disabled = True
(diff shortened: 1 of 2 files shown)


daveFNbuck:
Looks like one of the e-mail tests might be flaky :(

daveFNbuck:
I think this is ok to merge now. I did add it to production with some extra logging to help figure out if there are other causes for this and one other bug I still don't know any cause for. Hopefully I'll have at least one more fix on Monday.

Tarrasch:
Thanks @daveFNbuck ! :)

PR #2210, opened 2017-08-15
Set utf-8 encoding in the MIMEText constructor

daveFNbuck:
## Description
Pass the e-mail encoding into the MIMEText constructor rather than 

## Motivation and Context
If this is not set during the constructor, an exception is thrown in Python 2 when the e-mail body contains unicode.

Added a unit test, deployed to production but it'll take a while before I see any unicode e-mails.

### DIFF
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -141,8 +141,7 @@ def generate_email(sender, subject, message, recipients, image_png):
 
     msg_root = MIMEMultipart('related')
 
-    msg_text = MIMEText(message, email().format)
-    msg_text.set_charset('utf-8')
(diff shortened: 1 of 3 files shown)


daveFNbuck:
This is now tested in production, and I'm getting e-mails with unicode in them. Turns out R scripts use unicode single quotes around package names in error messages.

dlstadther:
@daveFNbuck Feeling good about this? Ready to merge?

daveFNbuck:
Yes, it was ready.

Tarrasch:
Oh, forgot to say. Thanks for this PR @daveFNbuck  :)

PR #2217, opened 2017-08-24
minor inconsistency fix for the month value

MihailRussu:
see the title, a month can't be 15th & the value "5" was used before and after anyway


Tarrasch:
Thanks @MihailRussu ! :)

PR #2222, opened 2017-08-28
README: avoid _here_ links; ack readthedocs

dckc:
## Motivation and Context

> Screen reader users often navigate from link to link, skipping the text in between
>
> Implication 1: Links should make sense out of context. Phrases such as "click here," "more," "click for details," and so on are ambiguous when read out of context. -- http://webaim.org/techniques/hypertext/

https://www.w3.org/TR/WCAG10-HTML-TECHS/#link-text

I expect tweaking README text has no testing impact.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,12 +25,12 @@ Getting Started
 ---------------
 
 Run ``pip install luigi`` to install the latest stable version from `PyPI
-<https://pypi.python.org/pypi/luigi>`_. Documentation for the latest release is
-hosted `here <https://luigi.readthedocs.io/en/stable/>`__.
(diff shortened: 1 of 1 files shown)


Tarrasch:
Wow, I really love how you pay attention to detail and users who use screen readers here!

dckc:
Thanks.

I've been having fun with luigi for a little while now... https://github.com/kumc-bmi/grouse/tree/master/etl_i2b2

I use that link from the README to get to the documentation pretty regularly, and it bugged me a little bit each time... and github makes it so easy to fix... so today I did.

PR #2224, opened 2017-08-29
Update to handle complex avro types for Avro Schema Definition copying to Big Query Schema Definitions

Mandalorian007:
Update to handle complex avro types for Avro Schema Definition copying to Big Query Schema Definitions

## Description
These Changes handle Complex Avro Schema documentation being copied into Big Query Schema descriptions

## Motivation and Context
Allow most Avro documentation fields to be copied into Big Query Schema Descriptions

I ran many different test avro schema scenarios to manage complex nested schema documentation


luigi/contrib/bigquery_avro.py
```diff
@@ -86,10 +86,64 @@ def get_fields_with_description(bq_fields, avro_fields):
...
+                        bq_map_value_field[u'description'] = avro_map_value.doc
+
+                        #This is jumping into the map value directly and working with that
+                        bq_map_value_field[u'fields'] = get_fields_with_description(bq_map_value_field[u'fields'], avro_map_value.fields_dict)
+
+                    if value_field_type is avro.schema.ArraySchema:
+                        bq_map_value_field[u'description'] = avro_map_value.items.doc
+                        bq_map_value_field[u'fields'] = get_fields_with_description(bq_map_value_field[u'fields'], avro_map_value.items.fields_dict)
+
+                    #Unions nested in maps are not supported
```
ulzha: The comment repeats what the code says and doesn't explain "why". Is it because we leave implementation for later? Then we could explain exactly that, or say "TODO" instead of "not supported". Is it for another reason? It's hard to tell for a new reader of the code.

luigi/contrib/bigquery_avro.py
```diff
@@ -86,10 +86,64 @@ def get_fields_with_description(bq_fields, avro_fields):
...
+
+                #Primitive and top level complex type support
                 field[u'description'] = avro_field.doc
-                if field[u'type'] == u'RECORD' and hasattr(avro_field.type, 'fields_dict'):
-                    field[u'fields'] = \
-                        get_fields_with_description(field[u'fields'], avro_field.type.fields_dict)
+
+                #Enum Support
+                if field_type is avro.schema.EnumSchema:
+                    field[u'description'] = avro_field.type.doc
```
ulzha: Does this and the following assignments overwrite the `field[u'description'] = avro_field.doc` above intentionally? I wonder if the type documentation should really take precedence like that. The field documentation is presumably more specific to the concrete usage and the dataset we're loading, thus more interesting to the user than the type documentation, isn't it? Conceivably both could be useful... Concatenate? :)

luigi/contrib/bigquery_avro.py
```diff
@@ -86,10 +86,64 @@ def get_fields_with_description(bq_fields, avro_fields):
...
+                #Union Support
+                if type(avro_field.type) is avro.schema.UnionSchema:
+                    for schema in avro_field.type.schemas:
+                        if type(schema) is avro.schema.RecordSchema:
+                            field[u'description'] = schema.doc
+                            field[u'fields'] = get_fields_with_description(field[u'fields'], schema.fields_dict)
+
+                #Map Support
+                if field_type is avro.schema.MapSchema:
+                    #directly bypass the key and value attributes in BigQuery Schema
```
ulzha: Sorry, I don't understand what "bypass" means in a context like this. We _are_ doing things with `...value...` on the following lines, so what was bypassed? Is this trying to explain the `-1`?

luigi/contrib/bigquery_avro.py
```diff
@@ -86,10 +86,64 @@ def get_fields_with_description(bq_fields, avro_fields):
...
+                        #This is jumping into the map value directly and working with that
+                        bq_map_value_field[u'fields'] = get_fields_with_description(bq_map_value_field[u'fields'], avro_map_value.fields_dict)
+
+                    if value_field_type is avro.schema.ArraySchema:
+                        bq_map_value_field[u'description'] = avro_map_value.items.doc
+                        bq_map_value_field[u'fields'] = get_fields_with_description(bq_map_value_field[u'fields'], avro_map_value.items.fields_dict)
+
+                    #Unions nested in maps are not supported
+                    if value_field_type is avro.schema.UnionSchema:
+                        logger.info("Unions inside of maps are not supported for documentation.")
```
ulzha: Probably no need to make info logs noisy with this. Just explain they're not supported in the docstring of this task, https://github.com/spotify/luigi/pull/2224/files#diff-ce7b46ced67552a1fc5e2bdc4ccb11faL22

ulzha:
Some inconsistent style issues. In the surrounding code there's usually a space after #.

Needs tests. See https://github.com/spotify/luigi/blob/master/test/contrib/bigquery_gcloud_test.py#L272

luigi/contrib/bigquery_avro.py
Mandalorian007: Cleaned up the confusion considerably here.  Realistically grabbing the data from "avro_field.doc" only works in 3 scenarios:  PrimitiveSchema, MapSchema, and PrimitiveSchema nested in a UnionSchema.  To make the code more explicit I moved this call directly to where it was needed.  And just for clarity all other times the call returns 'None' which is automatically dropped so the field would populate no data to the Big Query description

luigi/contrib/bigquery_avro.py
Mandalorian007: Effectively what this does is Big Query generates a "key" attribute that doesn't have any direct interaction possible with the avro schema.  In this use case the code "skips" the key and directly starts work with the value attribute.  Cleaned up the comments to try and express that better.

luigi/contrib/bigquery_avro.py
Mandalorian007: Changed the code to have a description above about what is currently supported and changed the comments to reflect that it simply is not yet implemented

luigi/contrib/bigquery_avro.py
Mandalorian007: added the details above and removed all excess logging

Mandalorian007:
@ulzha Testing suite has been updated and has been successfully run locally 👍   Let me know if you see anything else :)

Mandalorian007:
@ulzha Boom Flake8 issues all resolved 👍

Mandalorian007:
@ulzha any chance on you can take a look at this today?

PR #2227, opened 2017-09-05
Fix ListParameter Hashability

dlstadther:
## Description

With the merge of #2115, `ListParameter` normalization uses the `_recursively_freeze()` method to transform lists of dictionaries into tuples of `_FrozenOrderedDict`. However, `ListParameter`'s `parse()` and `serialize()` methods were not updated to utilize the `_DictParamEncoder`, allowing JSON serialization.

This PR resolves the serialization of lists of dictionaries provided to `ListParameter`.

## Motivation and Context

Fixes unhashability of `ListParameter` when provided a list of dictionaries caused by #2115.

Added hashability test for list of dictionaries.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -821,7 +821,7 @@ def _recursively_freeze(value):
     """
     if isinstance(value, Mapping):
         return _FrozenOrderedDict(((k, _recursively_freeze(v)) for k, v in value.items()))
-    elif isinstance(value, list):
+    elif isinstance(value, list) or isinstance(value, tuple):
         return tuple(_recursively_freeze(v) for v in value)
     return value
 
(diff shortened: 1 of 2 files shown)


dlstadther:
@Tarrasch @j-ostrich Could y'all review? Thanks!

j-ostrich:
I think the same changes should be made for `TupleParameter` since its members can also be objects. Other than that, LGTM.

dlstadther:
`TupleParameter` and `ListParameter` only differ in their `parse()` methods. However, i'm not sure if it makes sense to subclass one of them as the other. Tuples and Lists are similar of use, but not in design or purpose.

Should I subclass `TupleParameter` as a class of `ListParameter`, only overriding the `parse()` method?

j-ostrich:
I don't see the benefit of subclassing besides saving a couple of lines of code, but I'm also not against it.  LGTM, but I don't have write access so you'll have to wait for @Tarrasch to take a look.

dlstadther:
@joemeszaros I see you added `DictParameter`. Do you think you could review this? It would be much appreciated! Then i'll merge.

dlstadther:
I'll merge this tomorrow (2017-09-12) unless I hear otherwise.

joemeszaros:
@dlstadther Sorry for the late response, but  I was on holiday and got back just now.

dlstadther:
@joemeszaros No worries. Feel free to still review (although merged). If you have problems or suggestions, i'm happy to submit another PR with those changes.

PR #2228, opened 2017-09-07
Fix "IF NOT EXISTS" Postgres 8 incompatibility

angelinemm:
Fix Postgres 8 incompatibility

## Description
Commit https://github.com/spotify/luigi/pull/2064/commits/a41e4c4384845c21f5854cf697b3f83535dedb5f introduced a postgres 8 incompatibility by using IF NOT EXISTS in the sql query that creates the marker table. It was done to simplify code and getting rid of exception handling.

IF NOT EXISTS is supported only since postgres 9.1

## Motivation and Context
Postgres 8 is arguably quite an old version but it's the version that amazon's redshift is based on.
It makes sense for companies that use redshift in production to use postgres 8 in development environment.  

I ran the tests suggested here: https://github.com/spotify/luigi/blob/master/CONTRIBUTING.rst
Also this commits is a revert, so it reintroduces code that was already used before.

### DIFF
--- a/luigi/contrib/postgres.py
+++ b/luigi/contrib/postgres.py
@@ -204,19 +204,25 @@ class PostgresTarget(luigi.Target):
         connection.autocommit = True
         cursor = connection.cursor()
         if self.use_db_timestamps:
-            sql = """ CREATE TABLE IF NOT EXISTS {marker_table} (
+            sql = """ CREATE TABLE {marker_table} (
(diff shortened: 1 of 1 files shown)


dlstadther:
@ant1j Any strong opinions against reverting your PR #2064 ?

ant1j:
If it helps compatibility, no problem.

dlstadther:
Thanks @angelinemm for contributing!

PR #2229, opened 2017-09-07
Fix unreachable variable status. Replaced by statuses

waltercool:
Original status variable caused a crash by logging a null
variable.

Signed-off-by: Pablo Cholaky <waltercool@slash.cl>


dlstadther:
@waltercool Could you please resolve conflicts? Thanks!

waltercool:
The same fix was applied one day ago under a45d0f3106314af0133384e1509d84d3eb85acf3, so this is actually on master and no need to fulfill this PR.

Thanks.

PR #2230, opened 2017-09-11
Add HA functionality to WebHdfsClient

adaitche:
## Motivation and Description

HDFS supports High Availability (HA) through the use of multiple namenodes. This is currently not supported by WebHdfsClient as it allows to specify only one namenode. Since version 2.1.0, the hdfs package allows to specify multiple namenodes, by passing an url containing multiple namenodes separated by ';'. This functionality is forwarded through to WebHdfsClient in this commit.

I tested the HA functionality manually on a cluster with two namenodes. I did not add unit tests, as for this one would need to extend WebHdfsMiniCluster to support multiple namenodes, which is non-trivial and probably overkill.


adaitche:
@tarrasch As the main author of the webhdfs client, could you take a quick look at this patch?

Tarrasch:
Thanks @adaitche :)

PR #2231, opened 2017-09-11
Silence Parameter type warning in WebHdfsClient

adaitche:
## Motivation and Description
In the webhdfs configuration, the parameter `user` has the default value None, which triggers the annoying warning `parameter.py:259: UserWarning: Parameter None is not of type string.` every time a WebHdfsClient is created. This commit prevents this by setting the default value to `''`.

Manual testing that the warning is gone.


dlstadther:
@adaitche Thanks!

PR #2233, opened 2017-09-14
 FileSystemExternalProgramTask, redirects stdout from externalProgram…

dnul:
…Task to task Target

This PR adds a new type of ExternalProgramTask which redirects output of the task the task target. This allows , for example, to run shell commands and use the output in the workflow. 

Simply modified ExternalProgramTask which creates a tmp file to capture stdout. Instead, we open the task target and use that fd to capture the ExternalProgram output. 

## Motivation and Context
This allows better integration of external shell commands into Luigi, by directly capturing its output. In particular I was using some s3 CLI commands which produced a custom output. By putting all those commands into an .sh file and capturing its output I was able to easy integrate those scripts instead of rewriting them using boto bindings.

I've included some basic tests and I'am currently using the modifications.

### DIFF
--- a/luigi/contrib/external_program.py
+++ b/luigi/contrib/external_program.py
@@ -37,6 +37,7 @@ import sys
 import tempfile
 
 import luigi
+from ..target import FileSystemTarget
 
 logger = logging.getLogger('luigi-interface')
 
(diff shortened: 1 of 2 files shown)


luigi/contrib/external_program.py
```diff
@@ -203,3 +204,54 @@ def program_environment(self):
...
+                    logger.info('Program stderr:\n{}'.format(stderr))
+
+            if not success:
+                raise ExternalProgramRunError(
+                    'Program failed with return code={}:'.format(proc.returncode),
+                    args, env=env, stdout=stdout, stderr=stderr)
+
+        finally:
+            tmp_stderr.close()
+            tmp_stdout.close()
```
dlstadther: Flake8 complains about lack of newline at EOF

luigi/contrib/external_program.py
```diff
@@ -37,6 +37,7 @@
 import tempfile
 
 import luigi
+from ..target import FileSystemTarget
```
dlstadther: Is this the best way to import `FileSystemTarget`?

PR #2239, opened 2017-09-26
Speed up task_list when beyond limit

daveFNbuck:
## Description
In task_list, just return the count for the status when that count is over the max_shown_tasks limit without iterating to check upstream status. This means we can't show counts for upstream statuses, so those are replaced with "unknown" in the visualizer when this happens.

## Motivation and Context
When the number of tasks get into the millions, even refreshing the visualizer can take a minute or more, causing havoc in the pipeline. Since all we really want in these situations is the counts, we can skip the more expensive bits of computation and just return the sizes. This prevents doing upstream checks, but saves a lot of time.

We may want to institute a higher threshold so we can get upstream numbers if you're only a little above the limit for returning all tasks.

Added unit tests, have been running this for about a day and it works at scale to reduce visualizer refreshes from minutes to seconds.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -474,6 +474,12 @@ class SimpleTaskState(object):
     def get_active_tasks_by_status(self, *statuses):
         return itertools.chain.from_iterable(six.itervalues(self._status_tasks[status]) for status in statuses)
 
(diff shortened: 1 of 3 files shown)


dlstadther:
Thanks @daveFNbuck !

PR #2241, opened 2017-09-27
update task_parameter and set value to TEXT type

### LINKED ISSUE #2240: luigi parameter value column is set to limit size of 256
But we need values longer than that. Can the limit be raised? In postgres this can just be `TEXT` type.

Jimexist:
fixes #2240


dlstadther:
I don't have any experience with this portion of Luigi. @erikbern @stephenpascoe could y'all take a look given your past contributions to this module? Thanks!

Jimexist:
thanks for bringing up the conversation. travis failed because i used postgres syntax but the sqlite didn't support.

another thing that i did also want to confirm is that `task_id` index was not a *unique index* but it was used one, i.e. fetching by task_id is assumed to return one row. this was not enforced in database level - but i think it should be.

Jimexist:
thanks for fixing this

Issue #2047, opened 2017-02-28
Set transaction flag for database queries


raymondlewallen:
Would like a way to set enable transaction true/false for Postgres.  This would enable to run some commands that require to be run outside of a transaction block, such as redshift "VACUUM" statement.  I've started working on overriding the query code to allow me to do this.

dlstadther:
@raymondlewallen Feel free to contribute! 

It looks like you'd need to update [`PostgresQuery`](https://github.com/spotify/luigi/blob/master/luigi/contrib/postgres.py#L348) and then [`RedshiftQuery`](https://github.com/spotify/luigi/blob/master/luigi/contrib/redshift.py#L577) (as a courtesy :) ).

I've had this same need (to execute vacuum). My solution was to just manually implement the execution and output myself (below).

```python
def run(self):
       connection = psycopg2.connect('dbname={db} host={host} port={port} user={user} password={pass}'.format(args_dict))
        cursor = connection.cursor()

        # change isolation level temporarily to allow vacuum
        old_isolation_level = connection.isolation_level
        connection.set_isolation_level(0)
        cursor.execute(self.query)   # self.query is just 'vacuum;'
        connection.commit()
        connection.set_isolation_level(old_isolation_level)

        # output - update market table
        connection_target = self.output().connect()
        self.output().touch(connection_target)
        connection_target.commit()
        connection_target.close()
```

see [here](http://stackoverflow.com/questions/1017463/postgresql-how-to-run-vacuum-from-code-outside-transaction-block) and [here](http://initd.org/psycopg/docs/connection.html#connection.autocommit) for two psycopg2 solutions to execute a vacuum.

dmohns:
One way I can think of is to specify some sort of dictionary in the `PostgresQuery` tasks.  Then it could get passed to `run` and called from `connection.set_session(<dict>)`. That way we get everything from here http://initd.org/psycopg/docs/connection.html#set_session , in particular `autocommit=True` allows to run `VACUUM` queries. 

I am uncertain though: Is this the luigi way of doing things?

dlstadther:
>  I've started working on overriding the query code to allow me to do this.
@raymondlewallen Any update here? Feel free to open a WIP PR if necessary.

dckc:
[...]

PR #2242, opened 2017-09-27
Adding autocommit property for Redshift queries for Vacuums, etc.

jamesmcm:
This solves  #2047 

I've added the ability to override the autocommit property in Postgres Queries (used by Redshift Queries) - this will allow the user to make queries in autocommit mode as some operations require this.

The user only needs to set:
`autocommit = True `
in their RedshiftQuery subclass to make use of it.

Specifically this was done so that we could run VACUUMs in Redshift from Luigi.

I've tested it locally against our database, but I couldn't get the tox tests to run on my machine.

(Sorry for the 2 separate commits, it seems my attempt at squashing them failed)

### DIFF
--- a/luigi/contrib/postgres.py
+++ b/luigi/contrib/postgres.py
@@ -348,7 +348,8 @@ class PostgresQuery(rdbms.Query):
     Template task for querying a Postgres compatible database
 
     Usage:
-    Subclass and override the required `host`, `database`, `user`, `password`, `table`, and `query` attributes. Optionally one can override the `autocommit` attribute to put the connection for the query in autocommit mode.
+    Subclass and override the required `host`, `database`, `user`, `password`, `table`, and `query` attributes. 
+    Optionally one can override the `autocommit` attribute to put the connection for the query in autocommit mode.
 
     Override the `run` method if your use case requires some action with the query result.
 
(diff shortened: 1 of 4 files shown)


luigi/contrib/postgres.py
```diff
@@ -359,6 +360,8 @@ class PostgresQuery(rdbms.Query):
 
     def run(self):
         connection = self.output().connect()
+        if self.autocommit:
```
dlstadther: You'll want to initialize this in `rdbms.Query`. Set its default to `False`?

jamesmcm:
Okay, please check the new commit. I couldn't find an easy way of getting the autocommit status from the postgres server, so for now the test just uses the attribute on the connection.

luigi/contrib/rdbms.py
```diff
@@ -178,6 +182,10 @@ def table(self):
     def query(self):
         return None
 
+    @abc.abstractproperty
```
dlstadther: I don't think this needs to be an `abstractproperty`. You shouldn't require the user to override it. Just a plain `@property` should suffice.

jamesmcm: Oops, I fixed this now. Thanks.

jamesmcm:
Thanks, it looks good now :)

dlstadther:
Thanks @jamesmcm for contributing!

Issue #2244, opened 2017-10-02
Redshift COPY should use the columns defined


jamesmcm:
Currently the Redshift COPY is not using the defined columns (if present) - these are only used in table creation. This means that the COPY job relies upon the order of the columns being the same as in the CSV file.

This is a big problem, as it can cause data to be inserted into the wrong columns. 

## Example

For example, imagine we are importing two fields from a pull from another DB via a CSV in S3:

`order_id, us_dollars_float`

The DB maintainers then decide they will drop the column of us_dollars_float and instead send it as an integer in cents, so we now receive:

`order_id, us_dollars_cents_int`

However we don't want to drop the old column in our Redshift table yet as we still have reports using it that we can't change right now.

So our RS table looks like:

`order_id, us_dollars_float, us_dollars_cents_int`

This means if we now do the COPY without specifying the columns we will insert in to the old column with the new data (now in cents!), as it only uses the column ordering.

I.e. Luigi will do:
`COPY our_table FROM 'ourfile.csv' ...` 

Whereas we want:

`COPY our_table (order_id, us_dollars_cents_int) FROM 'ourfile.csv'...`

This is a big problem for us, so I'll submit my Pull Request immediately.

jamesmcm:
See PR #2245

jamesmcm:
Merged now :) If using the columns variable be sure to set them to match your Redshift table.

PR #2245, opened 2017-10-02
Use columns defined for COPY to Redshift so we never rely on order

jamesmcm:
## Description

See issue #2244 

## Motivation and Context
Right now COPYs to Redshift are dependent upon the column ordering which can break imports.

I've tested it locally with our columns defined.

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -334,7 +334,7 @@ class S3CopyToTable(rdbms.CopyToTable, _CredentialsMixin):
         If both key-based and role-based credentials are provided, role-based will be used.
         """
         logger.info("Inserting file: %s", f)
-        if len(self.columns)>0:
+        if len(self.columns) > 0:
             colnames = ",".join([x[0] for x in self.columns])
             cursor.execute("""
              COPY {table} ({colnames}) from '{source}'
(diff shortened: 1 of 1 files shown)


luigi/contrib/redshift.py
```diff
@@ -334,16 +334,30 @@ def copy(self, cursor, f):
...
+            cursor.execute("""
+             COPY {table} from '{source}'
+             CREDENTIALS '{creds}'
+             {options}
+             ;""".format(
+                table=self.table,
+                source=f,
+                creds=self._credentials(),
+                options=self.copy_options)
+            )
```
dlstadther: Perhaps you could do something more like below to reduce duplicate code

```python
colnames = ''
if len(self.columns) > 0:
    colnames = ",".join(x[0] for x in self.columns])
cursor.execute("""\
    copy {table} {cols} from '{source}'
    credentials '{creds}'
    {options}
    ;""".format(
        table=self.table,
        colnames='({})'.format(colnames),
        source=f,
        creds=self._credentials(),
        options=self.copy_options
    )
```

jamesmcm: Fixed :)

dlstadther:
I'm going to pull this branch down and verify its compatibility with my ETLs (as i'm not familiar with the data mapping options with COPY). Will report back soon

jamesmcm:
Thanks, I don't think it should break anything, unless you have defined the columns there but are reliant on it importing them out of order, and I can't really think of any use case where that is desirable.

dlstadther:
Thanks @jamesmcm !

PR #2249, opened 2017-10-09
Lazy completion check

TMaYaD:
## Description
This tries to improvise luigi scheduler by addressing bottlenecks

## Motivation and Context

When luigi worker starts it builds a dependency graph and checks for completion of all tasks.
If we start **w** workers with **t** tasks each, we are instantly checking **(w * t)** tasks for completion which generally are `output().exists()` calls. In a typical use case **w** is expected to increase linearly with **t** so that we could parallelise efficiently. Instead, we end up with a bottleneck with O(n^2) calls.

Changing completion check to lazy addresses this.
- First worker builds the entire graph with with `PENDING` status and submits it to the scheduler
- Scheduler merges it with it's existing graph and returns with a list of tasks to complete
- Worker checks each task it receives for completion and proceeds to try and run it.
  - If task is already completed, worker informs scheduler about the same and gets next task
  - If task is not completed, worker proceeds to run it and informs the scheduler about the result(`DONE`/`FAIL`)
- Scheduler as usual makes sure only one worker is working on a task so there are no collisions even while we mark `DONE` tasks as `PENDING`

However, starting with the PENDING status for tasks results in already DONE/FAIL tasks to be (wrongly) transitioned to PENDING when a new worker starts and submits it's graph. Hence a `NEW` task state is introduced. Which is a pseudo state and replaced by scheduler to existing task state or `PENDING` as appropriate. This change makes the workers run tasks correctly. 

One final issue remaining is that now the task status is not synchronised between worker and scheduler and as a result the execution summary is off. We remedy this by returning the modified task status(and other info) from scheduler to worker and updating the worker's copy of the task.

This is tested manually using `examples.touch_n.NTask` 
The automated test suite could not be run.


TMaYaD:
Should we make this optional behaviour with a flag similar to `parallel-scheduling`?

luigi/worker.py
```diff
@@ -180,16 +180,12 @@ def run(self):
             t0 = time.time()
             status = None
 
-            if _is_external(self.task):
-                # External task
-                # TODO(erikbern): We should check for task completeness after non-external tasks too!
-                # This will resolve #814 and make things a lot more consistent
```
Tarrasch: So does this fix this TODO?

luigi/scheduler.py
```diff
@@ -691,11 +698,11 @@ def dump(self):
 
     @rpc_method()
     def prune(self):
-        logger.info("Starting pruning of task graph")
+        logger.debug("Starting pruning of task graph")
```
Tarrasch: Ideally these changes would be in a separate commit

examples/touch_n.py
```diff
@@ -0,0 +1,37 @@
+import luigi
```
Tarrasch: * Can you add a how-to-run-header for this file?
 * Is `touch_n ` a good name fir this fie? I'm thinking maybe naming it something to clarify this illustrates lazy completion checks.

luigi/scheduler.py
```diff
@@ -885,6 +895,25 @@ def add_task(self, task_id=None, status=PENDING, runnable=True,
             self._state.get_worker(worker_id).tasks.add(task)
             task.runnable = runnable
 
+        return {
```
Tarrasch: So now this function returns something?

luigi/worker.py
TMaYaD: Yes if I got the intent right. @erikbern may need to take a look and see if I'm missing anything.

luigi/scheduler.py
TMaYaD: Will refactor out.

examples/touch_n.py
TMaYaD: - I'll add a header. 
- This is not just for lazy completion checks. I use this to simulate long running tasks with (simple) outputs. Nevertheless I agree it could use a better name.

luigi/scheduler.py
TMaYaD: Yeah, It returns almost the same variables as it takes (in slightly different format).
I needed only the `status` but I thought it makes better sense to include as much serialisable task as possible for completeness's sake.

TMaYaD:
This does not change the order of phases. I personally feel it's better to have separate scheduling phase and running phase. This patch doesn't change that. It moves the task.complete() check from scheduling phase to running phase.
So to begin with the scheduler treat everything as todo and it marks finished tasks as **done** vs **_already_ done** both of which seems to be same implementation wise.

The bug you are talking about could be that if you delete an output before the task is pruned out of scheduler and try to reschedule it, the scheduler might treat it as already done and won't run it. I'll have to test this. It occurred to me as a possibility as I'm typing this.

I'll spend some time this week to get tests running

PR #2251, opened 2017-10-10
Adding voyages-sncf.com as a company using luigi

Lucas-C:
We presented our use of luigi at voyages-sncf.com last Tuesday at a meetup of the French Python society:
http://nantes.afpy.org


dlstadther:
Thanks @Lucas-C !

PR #2253, opened 2017-10-12
Add bigquery extract task alias

yudppp:
## Description
Add BigqueryExtractTask alias

## Motivation and Context

i like Big**q**uery syntax, but this func has not alias


yudppp:
Oh sorry. I see.For backwards compatibility. 

I use BigQuery syntax in future.
I close this PR

Thank you.

PR #2262, opened 2017-10-18
add Leipzig University Library as user

miku:
## Description

Update README, add Leipzig University Library as user.


dlstadther:
Thanks @miku !

PR #2267, opened 2017-10-24
Added worker-id command line parameter

TMaYaD:
## Description
This change adds a `--worker-id` which will be used for worker to identify itself. The user is expected to make sure these are unique.

## Motivation and Context
We needed to start and (safely) shutdown workers based on resource availability. In order to use `disable-worker` rpc call we needed the worker id. But there is no way to know the same. We can't rely on the included host name since we are using ECS to run the workers inside containers
This change allows us to set the worker-id using a known uuid and other details and use the same to disable it later.

This is already available is a function parameter (which doesn't seem to be used anywhere). We tested it in our own setup without any issues.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -59,7 +59,7 @@ from luigi.target import Target
 from luigi.task import Task, flatten, getpaths, Config
 from luigi.task_register import TaskClassException
 from luigi.task_status import RUNNING
-from luigi.parameter import FloatParameter, IntParameter, BoolParameter
+from luigi.parameter import Parameter, FloatParameter, IntParameter, BoolParameter
 
 try:
     import simplejson as json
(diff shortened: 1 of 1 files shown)


dlstadther:
@TMaYaD I've emailed Travis CI's support about our past few PR builds not going through. They seem to think it's due the fact that no one from your `lexentbio` org has ever logged into Travis before. Is there any way you or someone else from lexentbio (maybe @abhik) can do so and restart your PR builds? Thanks!

TMaYaD:
@dlstadther Done. Builds are running now. But all of the PRs are failing `flake8` tests. I could use some help to fix those.

dlstadther:
Interesting... I'm inclined to ignore some of these pep8 rules in the flake8 testing as they've gotten more strict and a little excessive IMO. I'm just not yet sure how to modify `.travis.yml` to add exclusions to the flake8 matrix

Issue #2270, opened 2017-10-26
Configuration - Postgres section - explicit schema definition 


Gr1f0n6x:
Hello!
In current version of configuration file it's impossible to define explicitly table schema only if i add schema as a part of future table-name like this:
`[postgres]
marker-table=schema.table-name`

What if we add additional parameter for postgres (or maybe for others DB too) like 
`marker-schema=schema`

It will be very useful for many cases, because of now it depends on postgres' **search_path** value which we can't change sometimes.

And it's also concerning tables which luigi creates - tasks, task_parameters and task_events. I want to define schema for them too, but there is not any parameters which can be changed to configure them.

PR #2271, opened 2017-10-26
marker-table space [issue_#2270]

Gr1f0n6x:
## Description

I've changed few lines in postgres.py file and add additional parameter: **marker-space** with default value: **public**
## Motivation and Context

https://github.com/spotify/luigi/issues/2270
 

I've tested this changes with new parameters:
```
[postgres]
marker-space=space_name

```
And without this line and all my workers have finished their work successfully. Of course, marker table was created in configured space or, by default, in public space

Also i want to add this functionality to schedulers tables (tasks, task_parameters and task_events) because now it's create tables in spaces depends on external DB configs

### DIFF
--- a/luigi/contrib/postgres.py
+++ b/luigi/contrib/postgres.py
@@ -105,6 +105,7 @@ class PostgresTarget(luigi.Target):
     This will rarely have to be directly instantiated by the user.
     """
     marker_table = luigi.configuration.get_config().get('postgres', 'marker-table', 'table_updates')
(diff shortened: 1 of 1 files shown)


Gr1f0n6x:
The origin of this problem is that i want to store all luigi's records in particular schema (so, marker table is the least problem. The main - tables created by luigi - tasks, task_events and task_parameters). By default (in postgres) this tables created in schema which configured in search_path and i can't change it. 

And after that, i want to add logic to restore the history not from the file but from DB. Maybe it is possible now, but i didn't find the solution.

PR #2273, opened 2017-10-30
Preserve filter on server on reload

daveFNbuck:
## Description
Run processHashChange before updateTasks on page load so the server-side filtering box gets properly checked.

## Motivation and Context
processHashChange only gets called after updateTasks, so filter on the checkbox only gets checked after we make the server request without filtering. This results in a checked box but no server-side filtering.

Ran the server locally with a small max_shown tasks and tried refreshing with a server-side filter. Without this change, no tasks are shown after the refresh. With this change, the filtered tasks are shown.

### DIFF
--- a/luigi/static/visualiser/js/visualiserApp.js
+++ b/luigi/static/visualiser/js/visualiserApp.js
@@ -1182,6 +1182,7 @@ function visualiserApp(luigi) {
             }
         });
 
+        processHashChange();
         updateTasks();
         bindListEvents();
 


dlstadther:
@daveFNbuck I'm trying to fix these excessive new flake8 tests over on #2274 but somehow managed to get hit with many more annoying flake8 failures. Do you have experience with the flake8 linter to know why this happened or how to resolve?

PR #2274, opened 2017-10-31
Fix new flake8 tests

dlstadther:
## Description

Update flake8 compliance

## Motivation and Context

Flake8 started linting against stricter standards. This patch fixes the new ones.

* E722 - bare except
* E741 - ambiguous variable name
* E742 - ambiguous class definition

Travis will run.

### DIFF
--- a/luigi/contrib/docker_runner.py
+++ b/luigi/contrib/docker_runner.py
@@ -227,7 +227,7 @@ class DockerTask(luigi.Task):
                 container_name = self.name
             try:
                 message = e.message
-            except:
+            except AttributeError:
                 message = str(e)
             self.__logger.error("Container " + container_name +
                                 " exited with non zero code: " + message)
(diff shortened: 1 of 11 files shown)


dlstadther:
Added test exclusions to wrong tox line :/

daveFNbuck:
Why are these ones excessive? E741 and E742 only catch three test cases where we use l and I as class and variable names. For the bare except, I see a couple we can fix but the rest can just get around it by catching BaseException. There are only 13 instances of bare excepts in the repo, so it's still a quick fix. Note that if we had this enabled, it would have caught the latest docker commit, which has a bare except where it should be catching an `AttributeError`.

The reason your commit fails is that you modified the wrong line. The line you modified only runs flake8 on the docs. The line above is the one that runs it on everything else.

dlstadther:
@daveFNbuck I've taken your guidance and fixed the current flake8 failures. I didn't spend the time to really determine what the `except` clauses should be catching except for the docker `AttributeError`.

daveFNbuck:
I think most of them were meant to catch arbitrary exceptions, so that's fine. The redshift one should probably just be removed. No point in catching an exception just to re-raise it.

dlstadther:
@daveFNbuck Good for me to merge?

daveFNbuck:
LGTM

PR #2275, opened 2017-11-02
Cleans up dangerous function default variables ([] and {})

sklarsa:
## Description
Changes function default variables `{}` -> `dict()` and `[]` -> `list()`

## Motivation and Context
Just cleaning up some of the errors from https://landscape.io/github/spotify/luigi/1347/messages/error

Ran `tox -e py27-nonhdfs` and everything passes

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -174,7 +174,7 @@ class BigQueryClient(object):
 
         return True
 
-    def make_dataset(self, dataset, raise_if_exists=False, body={}):
+    def make_dataset(self, dataset, raise_if_exists=False, body=dict()):
         """Creates a new dataset with the default permissions.
 
            :param dataset:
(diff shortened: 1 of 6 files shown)


dlstadther:
Thanks @sklarsa ! The errors here are waiting on #2274 for resolution

sklarsa:
Sounds good.  Let me know if you need anything else from my end

dlstadther:
Thanks @sklarsa !

jhorman:
Is this still dangerous. The default args are still mutable. Isn't the right approach to set these to None, and default in code.

kwilcox:
Correct @jhorman, this did not fix what was intended. However, I took a quick look and none of the mutable arguments are changed within the functions. It would be better to default these to `None` and set the defaults accordingly inside of the function as you mentioned but this shouldn't be causing any issues as is.

Tarrasch:
@sklarsa, can you please send a new PR fixing this as @jhorman pointed out?

sklarsa:
Yep I'll have one out this week

sklarsa:
In `make_dataset` in `luigi/contrib/bigquery.py:177`, it appears that `body` is actually required since the first line of code assumes its existence and the `'id'` key... Not sure what to do here @mikekap?

PR #2276, opened 2017-11-02
Added JupyterNotebookTask to run Jupyter notebooks as luigi tasks

mattiaciollaro:

### DIFF
--- a/luigi/contrib/jupyter_notebook.py
+++ b/luigi/contrib/jupyter_notebook.py
@@ -1,0 +1,194 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2012-2015 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""
+Template tasks for running external programs as luigi tasks.
+
+This module is intended for when you need to execute a Jupyter notebook as a 
+task within a luigi pipeline.
+
+`JupyterNobebookTask` allows you to pass parameters to the Jupyter notebook 
+through a dictionary (`pars`). When the `run` method is executed, the `pars`
+dictionary is written to a `notebook_title.ipynbpars` temporary JSON file
+in the same directory where the notebook that you want to execute is saved
+(here `notebook_title` is the title of the notebook).
+Inside the notebook, you can then retrieve the values of the parameters in 
+`pars` by reading the temporary `notebook_title.ipynbpars` JSON file.
+
+By default, the `path` attribute of `self.input()` and `self.output()` 
+are added to the temporary JSON file via the `pars` attribute with keys
+`input` and `output` respectively.
+
+`JupyterNotebookTask` inherits from the standard `luigi.Task` class.
+As usual, you should override the default `requires` and `output` methods.
+
+The template task `JupyterNotebookTask` implements a run method that wraps the 
+nbformat/nbconvert approach to executing Jupyter notebooks as scripts.
+See the `Executing notebooks from the command line` section of the `nbconvert`
+module documentation for more information.
+"""
+
+import os
+import json
+
+import luigi
+import nbformat
+import nbconvert
+import logging
+
+
+logger = logging.getLogger('luigi-interface')
+
+
+def get_file_name_from_path(input_path):
+    """
+    Utility to extract the name of a file without the file extension from a 
+    given path.
+
+    This function extracts the file name without the file extension from a given
+    path.
+    For example, if path='~/some_dir/some_file.ext', then
+    get_filename_from_path(path) will return 'some_file'.
+
+    Args:
+        input_path: a path to a file
+
+    Returns:
+        a string containing the name of the file without file extension
+    """
+
+    base_name = path.basename(path.normpath(input_path))
+    file_name = base_name.split('.')[0]
+    return(file_name)
+
+
+class JupyterNotebookTask(luigi.Task):
+    """
+    This is a template task to execute a Jupyter notebook as a luigi task.
+
+    This class has the following key attributes:
+
+    - nb_path: the full path to the Jupyter notebook (required)
+
+    - kernel_name: the name of the kernel to be used in the notebook execution
+                   (required)
+
+    - timeout: maximum time (in seconds) allocated to run each cell. If -1 (the
+               default), no timeout limit is imposed.
+
+    - pars: a dictionary of parameters to be passed to the Jupyter notebook.
+            These parameters can be retrieved inside the notebook by reading
+            the `notebook_title.ipynbpars` temporary JSON file created during the
+            execution of the task (where `notebook_title` is the title
+            of the Jupyter notebook you want to execute).
+            By default, paths associated with `self.input()` and `self.output()` 
+            are included with keys `input` and `output` in `pars`, and therfore
+            can be read from inside the notebook as well.
+            The `pars` dictionary is empty by default. 
+    
+    Example
+    #######
+    You would add something like the following block inside a notebook titled 
+    `my_notebook`:
+    
+    ```
+    # reading the temporary JSON file
+    with open('./my_notebook.ipynbpars') as pars:
+        parameters = json.load(pars)
+
+    # extracting the task's self.input() paths 
+    requires_path = parameters.get('input')
+
+    # extracting the task's self.output() paths
+    output_path = parameters.get('output')
+
+    # extracting a custom parameter named `my_par`
+    my_own_par = parameters.get('my_par')
+
+    ```
+
+    `requires_path` is itself a dictionary if the task's `requires` method 
+    returns a dictionary; otherwise, `requires_path` is a list.
+    """
+
+    nb_path = luigi.Parameter(
+        default=None
+    )
+
+    kernel_name = luigi.Parameter(
+        default=None
+    )
+
+    timeout = luigi.IntParameter(
+        default=-1
+    )
+
+    pars = {}
+
+    def run(self):
+
+        # get notebook name
+        notebook_name = get_file_name_from_path(self.nb_path)
+
+        # set requires pars
+        if type(self.requires()) is dict:
+            self.pars['input'] = {
+                tag: req.path for tag, req in self.input().items()
+            }
+        else:
+            self.pars['input'] = [req.path for req in self.input()]
+
+        # set output pars
+        self.pars['output'] = self.output().path
+
+        # write pars to temporary file
+        tmp_file_path = os.path.join(
+            os.path.dirname(self.nb_path), '%s.ipynbpars' % notebook_name
+        )
+
+        with open(tmp_file_path, 'w') as parameters:
+            json.dump(self.pars, parameters)
+
+        # run notebook
+        logger.info('=== Running notebook: %s ===' % notebook_name)
+
+        # get notebook format version
+        nb_version = int(nbformat.__version__.split('.')[0])
+
+        # run notebook
+        try:
+            with open(self.nb_path, 'r') as nb:
+                nb = nbformat.read(nb, as_version=nb_version)
+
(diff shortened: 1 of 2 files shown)


erikbern:
cool!

mattiaciollaro:
Still need to fix a couple of things in the docs and then it should be in good shape

dlstadther:
Thanks @mattiaciollaro ! All the flake8 errors here will be resolved by #2274

dlstadther:
@mattiaciollaro The remaining flake8 errors are due to this PR. Please address.

Also, do you have a coworker of whom has ipy and luigi experience to review?

Thanks!

mattiaciollaro:
Yes, I think @prncc would be perfect to review this!

mattiaciollaro:
@dlstadther Looks good now on the travis side, but I'm not sure on how to deal with the codecov errors. Do you have any suggestions?

dlstadther:
@mattiaciollaro Glad to see a green build on Travis. I'm not concerned about the codecov errors. Just started seeing these a few weeks ago and it's because of the [inability to compare this PR to the base commit](https://docs.codecov.io/docs/error-reference#section-missing-base-commit). IMO, nothing to worry about.

mattiaciollaro:
Ok, perfect. And thanks! I am still doing some more testing with the help of @prncc. I will make a final commit when fully ready.

Two important points that @prncc raised that I didn't think about are:

1. this class, as is, will not support concurrent execution of different instances of the same task; this is because in order to pass parameters from outside of the notebook to the notebook a JSON file is created at a given location and multiple instances could potentially try to write to the same file at the same time.
The original use case of this class that I had in mind, however, is for simple pipelines of notebooks where concurrent execution is not a requirement. If concurrent execution is a requirement for a project, perhaps one should translate the notebook to a script and make it into a task using some other class (e.g. luigi.contrib.external_program.ExternalProgramTask).

2. the path attributes from self.input() and self.output() can be passed to the notebook only if the targets associated with the task's self.input() and self.output() have a path attribute. I am changing the code to catch the AttributeError and pass None if self.input()[...].path or self.output[...].path does not exist.

dlstadther:
To point number 1, I don't have an immediate problem with your class not supporting concurrent execution from the start. However, if it would be trivial to add concurrency support that would be awesome! Is there a way you could use the task_id as part of the filename to attempt for uniqueness?

mattiaciollaro:
The previous points (concurrent execution and catching the AttributeError for targets that don't have a path attributes) are now both addressed and resolved in the last commit.

In particular, concurrent execution is now supported (by combining your idea of using the task id with passing the path of the temp JSON file as an environment variable).
For targets that don't have a path attribute, the corresponding entry is set to None and therefore empty in the JSON file.

luigi/contrib/jupyter_notebook.py
```diff
@@ -0,0 +1,310 @@
...
+When the task is executed, the ``pars`` dictionary is written to a
+temporary JSON file in the same directory that contains the notebook.
+
+Inside the notebook, you can retrieve the values of the parameters in
+``pars`` by reading the temporary JSON file. The path to the temporary file is
+accessible via the environment variable ``PARS``.
+For example, in a Python notebook, you can read the ``pars`` dictionary as
+follows:
+
+.. code-block:: python
```
prncc: These example code blocks are very nice to have. I would add another block with an example of how to run the task. This could be along the lines you sent me previously. In particular, 

1. Look up the available kernels with `jupyter kernelspec list` (which usually correspond to the Python versions).
2. After starting `luigid` run:

```
luigi --module jupyter_notebook JupyterNotebookTask --nb-path=./my_notebook.ipynb --kernel-name=python3
```

luigi/contrib/jupyter_notebook.py
```diff
@@ -0,0 +1,310 @@
...
+    )
+
+    json_action = luigi.Parameter(
+        default='delete'
+    )
+
+    pars = {}
+
+    def _form_input(self):
+
```
prncc: The `_form_input` and `_form_output` methods are a bit hard to read due to the lambdas, so I would consider adding a short docstring to each.

luigi/contrib/jupyter_notebook.py
```diff
@@ -0,0 +1,310 @@
...
+            }
+        else:
+            out = [req.path for req in self.output()]
+
+        return out
+
+    def run(self):
+
+        # check arguments
+        if not self.nb_path:
```
prncc: I already mentioned this to you, but my personal preference is to make the parameters required (by omitting `default=None`) and letting the task fail from Luigi's level. If you include a `description` parameter, the user will have a natural place to look for instructions. This *is* more explicit, as you said.

luigi/contrib/jupyter_notebook.py
```diff
@@ -0,0 +1,310 @@
...
+                kernel_name=self.kernel_name
+            )
+
+            ep.preprocess(nb, {
+                'metadata': {
+                    'path': os.path.dirname(self.nb_path)
+                }
+            })
+
+            logger.info('=== Done! ===')
```
prncc: Just pointing out that this is only displayed when the notebook executes correctly. You could add a line like `logger.info('=== Done with errors! ===')` to the `except` block.

luigi/contrib/jupyter_notebook.py
```diff
@@ -0,0 +1,310 @@
...
+
+    kernel_name = luigi.Parameter(
+        default=None
+    )
+
+    timeout = luigi.IntParameter(
+        default=-1
+    )
+
+    json_action = luigi.Parameter(
```
prncc: I think that `luigi.parameter.ChoiceParameter` is appropriate here. It also may happen that the kept file is wiped between runs, so if this concerns you in addition to the `run_id` you can add the a timestamp to the filename when saving, or have an option to log the params.

mattiaciollaro:
Thanks for the input @prncc! :) Will address soon and reply item by item.

luigi/contrib/jupyter_notebook.py
```diff
@@ -0,0 +1,310 @@
...
+        return out
+
+    def _form_output(self):
+
+        if isinstance(self.output(), dict):
+            out = {
+                tag: req.path for tag, req in self.output().items()
+            }
+        else:
+            out = [req.path for req in self.output()]
```
prncc: I forgot to mention (and based on our conversation you noticed and fixed): inputs and outputs may not be iterable.

luigi/contrib/jupyter_notebook.py
mattiaciollaro: `json_action` is now a ChoiceParameter. Thanks for the advice. Also, a time stamp is now added to the name of the temp JSON file.

luigi/contrib/jupyter_notebook.py
mattiaciollaro: I documented `_form_input` and `_form_output` with a doctoring that explains the logic behind them.

luigi/contrib/jupyter_notebook.py
mattiaciollaro: Removed defaults. Luigi will now throw its own errors if any required parameter is missing.
I also added a `description` to each parameter.

luigi/contrib/jupyter_notebook.py
mattiaciollaro: Thanks for pointing this out. Added logging in the `except` block.

luigi/contrib/jupyter_notebook.py
mattiaciollaro: Yes, this was definitely a bug. It is fixed now by getting values safely by leveraging the `flatten` utility in `luigi.task` and treating the `dict` case separately from the `list`/iterable case.
Thanks for pointing this out!

luigi/contrib/jupyter_notebook.py
mattiaciollaro: I will shortly add some more documentation containing an example of how to run the task. You're right, that's a good idea :)

mattiaciollaro:
@prncc:
- Is it possible to use nbconvert save the executed notebook as HTML or PDF? That would be a useful feature addition for me personally.

This can be done by using `nbconvert`. However, I am not sure whether one can *simultaneously* execute and convert the notebook. This may require running the notebook twice (once to execute it and once to export it to HTML or PDF) inside of the task. I'll take a look into it.
Any ideas on how to mark the task as complete in this case? Would the exported notebook be part of the items returned by the task's `output()` method?

- Minor: I think it's clearer to use full words for parameter and variable names. Notably pars and nb_path could be parameters and notebook_path.

Parameters name have been updated: `pars` -> `parameters`, `nb_path` -> `notebook_path`.

prncc:
I actually meant the scenario where you execute the notebook and save the outputs (in addition to the code cells) via something like:

```
jupyter nbconvert --to html --execute notebook.ipynb
```

In any case, just a side-remark! @ mention me when you want me to look over the changes again. please.

mattiaciollaro:
@prncc I made the suggested changes!

I am a little hesitant about the conversion to pdf or html for now. I will definitely add it as a feature in the future, but I need to figure out exactly how to avoid running the notebook twice inside the task and how to include in the task output the html/pdf if the user wants the converted notebook saved too.
This is a feature that requires more code and more testing. I see the value in it thought, and it will definitely become an option in JupyterNotebookTask.
Note that this will also add more requirements (pandoc, latex, etc.) to JupyterNotebookTask.

Another thing I decided to drop (instead of adding) is any R examples in the docs for now. I am considering writing a blog post about how to use this class and reference it in the Luigi docs. Instead of having a little example in little space, I'd rather have a longer and more detailed discussion with a runnable example. To this end I am writing a dedicated repository that will be the blog post companion.

Let me know your thoughts, and thanks!

prncc:
Great addition on parameter parsing, Mattia!

Two comments here:

1. It's not inconceivable that some people would be using the `PARS` environment variable, so that would ideally be customizable.

2. It would be useful to have a utility for loading parameters from the notebooks, so that the user is exposed to implementation details only minimally. So, from the notebook it could looks something like the following...

```
from luigi.contrib.jupyter_notebook_task import load_parameters

parameters = load_parameters(environment_variable='MY_VAR')
```

where 

```
import os
import json

def load_parameters(environment_variable='PARS'):
    with open(os.environ[environment_variable]) as parameters_file:
        return json.load(parameters_file)
```

mattiaciollaro:
Thanks @prncc !
I incorporated your feedback in the last commit:

1. `load_parameters` is implemented exactly as you described
(of course, this utility only works with Python notebooks... for R notebooks or notebooks written in other languages, the user has to read the JSON file manually, or write a similar utility)

2. the task has now an additional `luigi.Parameter` named `environment_variable` which defaults to 'PARS'

 The documentation is up to date with these last two changes.

Thanks once more for your valuable feedback!

prncc:
LGTM

mattiaciollaro:
@dlstadther Please feel free to let me know how this looks to you. Some more features may be added in the future, but for the time being this is the final version :)

prncc:
@dlstadther Anything else you need from me (the reviewer) on this PR?

dlstadther:
@mattiaciollaro Is there any chance you could add any tests? (Just realized a test file wasn't included in this PR)

mattiaciollaro:
@dlstadther Added some test! Most of the focus is on reading the parameters, which is the only tricky part I think.

A test that performs a full run of a small jupyter notebook is also included, but likely won't work on Travis (requires nbformat, nbconvert, jupyter_client). However, this test can be run locally.

test/contrib/jupyter_notebook_test.py
```diff
@@ -0,0 +1,405 @@
...
+    #     pass
+
+    # def test_json_action_keep(self):
+    #     pass
+
+    # def test_json_action_delete(self):
+    #     pass
+
+    # def test_invalid_key(self):
+    #     pass
```
dlstadther: Is there a reason for these being present and commented out?

test/contrib/jupyter_notebook_test.py
```diff
@@ -0,0 +1,405 @@
...
+
+logger = logging.getLogger('luigi-interface')
+
+try:
+    from jupyter_client import KernelManager
+    import nbformat as nbf
+    from nbformat.v4 import (
+        new_notebook,
+        new_code_cell
+    )
```
dlstadther: For the final test in this file, you can add these dependencies to `tox.ini` so that the actual test for running a JupyterNotebookTask is confirmed.

theotheo:
Please could you tell me why it wasn't merged after all?

dlstadther:
@theotheo it was abandoned by the author and exceeded the stale duration.

PR #2285, opened 2017-11-13
Convert dates to datetimes for DateHourParameter

mkennely:
## Description

Base class for DateHourParameter throws an error when a date instead of a datetime is passed as an argument.  Code was added to convert dates to datetimes when passed to DateHourParameters in order to alleviate aforementioned issue.

## Motivation and Context

The current Luigi set up I work with utilizes base tasks to avoid repeating code.  Some processes use DateParameters and others use DateHourParameters and errors arise when trying to pass dates to DateHourParameters

Unit test is included

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -479,6 +479,12 @@ class _DatetimeParameterBase(Parameter):
             return str(dt)
         return dt.strftime(self.date_format)
 
+    @staticmethod
+    def _convert_to_dt(dt):
+        if not isinstance(dt, datetime.datetime):
+            dt = datetime.datetime.combine(dt, datetime.time.min)
+        return dt
+
     def normalize(self, dt):
         """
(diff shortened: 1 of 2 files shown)


mkennely:
Not real sure who to tag. So @Tarrasch @mikekap , could you review?

Tarrasch:
I consider it to be a type error, hence a crash makes sense to me. Not sure if I want a behavior change. 

-----

The code of the PR is very clean though. :)

dlstadther:
It should be noted that `DateParameter` already does the opposite [here](https://github.com/spotify/luigi/blob/master/luigi/parameter.py#L380-L381) - converts datetime objects to dates. Makes sense to me that datetime should be able to take a date and assume midnight. However, I can also see the type error desire too

Tarrasch:
Hmm... well I guess it kinda makes sense that you always round down to the lowest of granularity time.  LGTM. :)

PR #2287, opened 2017-11-14
Add Synetiq to `Who uses Luigi?`.

danthelion:
We have been using Luigi for our main data pipeline for more than 3 years now, I think we fit on the list!


dlstadther:
Thanks @danthelion !

Tarrasch:
> I think we fit on the list!

You definitely do! And so great that you had a presentation about it too. :)

Issue #2209, opened 2017-08-14
RPCError: Received null response from remote scheduler


pavel-popov:
Hi Luigi community,

Often I get Workers failing with error: 

```
File "/usr/lib/python2.7/site-packages/luigi/scheduler.py", line 115, in rpc_func
  return self._request('/api/{}'.format(fn_name), actual_args, **request_args)
File "/usr/lib/python2.7/site-packages/luigi/rpc.py", line 152, in _request
    raise RPCError("Received null response from remote scheduler %r" % self._url)
PMRPCError: Received null response from remote scheduler 'http://path-to-central-scheduler.example.com:8082'
```

The reason for this is unclear, but it definetely takes place when scheduler is under load. Examples when I observe these failures:
- lone Worker started with 10 worker processes, overall graph contains several hundred tasks to be processed. Worker processes start working. After increasing number of worker processes from GUI from 10 to 20 the Worker fails
- 12 Workers starting, each covers it's month via RangeHourly, about 600 tasks are due on each Worker, 5 worker-processes per Worker. First four Workers are running without issues, others are receiving null response and die.

The workload is to run Apache Spark apps (via spark-submit) on Yarn cluster (deploy-mode=cluster), workers are running in specific Docker containers where nothing except `spark-submit` host applications are executed.

I checked Scheduler logs and found nothing strange - bunch of responses to ping and various api endpoints. 

What are the directions I should be looking into / how to debug this behaviour?

Env: luigi 2.6.2 on Python 2.7 on CentOS 7

pavel-popov:
I made a test with gradually increasing number of workers-processes from 10 up to 30 by 1 or 2. After 32 processes Worker died and in GUI there is a message "2 unread message(s)" near "Disable" worker button.

On the server where Worker is started I have 40 CPU and 252 Gb RAM - server utilization is near 0 during all time Worker was running.

Tarrasch:
Perhaps check the code-paths related to "changing number of workers" if there's anything that potentially could be a bug.

PR #2290, opened 2017-11-18
Handle multiprocessing with request sessions

jhorman:
When luigi forks with `--workers` it uses the same requests `Session` object leading to bugs.

## Description

`Session` is not multiproc safe. It uses http keep alives, and re-using the session across procs can lead to weird multiplexing. This was the cause of my RPCErrors, from what I can tell. 

```
RPCError: Received null response from remote scheduler
```

## Motivation and Context

Relates to #2209

I ran under load, 50 workers with and without the fix. With the fix I no longer get the `RPCErrors`

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -81,6 +81,7 @@ class RequestsFetcher(object):
         from requests import exceptions as requests_exceptions
         self.raises = requests_exceptions.RequestException
         self.session = session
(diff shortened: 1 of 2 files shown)


jhorman:
Added a basic test @dlstadther

Tarrasch:
I've wanted to hunt down this bug for a long time! Thanks!

PR #2291, opened 2017-11-21
Pass kwargs to `discovery.build()` when instantiating `GSCClient`.

danthelion:
## Description
https://github.com/google/google-api-python-client/blob/89906ac33b37c6017c893841743aa4f45729c91f/googleapiclient/discovery.py#L175

https://github.com/google/google-api-python-client/blob/89906ac33b37c6017c893841743aa4f45729c91f/googleapiclient/discovery.py#L297

The `build` and `build_from_document` methods has some useful keyword arguments that are currently not useable from the `GSCClient` interface.

## Motivation and Context
My use case for these changes is getting rid of warning log spam because of a default argument (`cache_discovery=True`)

https://github.com/google/google-api-python-client/issues/299
https://github.com/google/oauth2client/issues/673

Locally, running my pipeline with the `cache_discovery=False` argument passed on instantiation.

### DIFF
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -108,14 +108,18 @@ class GCSClient(luigi.target.FileSystem):
       as the ``descriptor`` argument.
     """
     def __init__(self, oauth_credentials=None, descriptor='', http_=None,
-                 chunksize=CHUNKSIZE):
+                 chunksize=CHUNKSIZE, **discovery_build_kwargs):
         self.chunksize = chunksize
         authenticate_kwargs = gcp.get_authenticate_kwargs(oauth_credentials, http_)
 
+        build_kwargs = authenticate_kwargs.copy()
+        build_kwargs.update(discovery_build_kwargs)
+
         if descriptor:
-            self.client = discovery.build_from_document(descriptor, **authenticate_kwargs)
+            self.client = discovery.build_from_document(descriptor, **build_kwargs)
         else:
-            self.client = discovery.build('storage', 'v1', cache_discovery=False, **authenticate_kwargs)
+            build_kwargs.setdefault('cache_discovery', False)
+            self.client = discovery.build('storage', 'v1', **build_kwargs)
 
     def _path_to_bucket_and_key(self, path):
         (scheme, netloc, path, _, _) = urlsplit(path)


ADiegoCAlonso:
@dlstadther These changes are quite simple and would help a lot to debug using the logs. Any possibility of an admin having a look?

dlstadther:
@danthelion could you rebase this branch with the new version of gcs?

danthelion:
@dlstadther Done.

luigi/contrib/gcs.py
```diff
@@ -108,14 +108,17 @@ class GCSClient(luigi.target.FileSystem):
...
 
+        build_kwargs = authenticate_kwargs.copy()
+        build_kwargs.update(discovery_build_kwargs)
+
         if descriptor:
-            self.client = discovery.build_from_document(descriptor, **authenticate_kwargs)
+            self.client = discovery.build_from_document(descriptor, **build_kwargs)
         else:
-            self.client = discovery.build('storage', 'v1', cache_discovery=False, **authenticate_kwargs)
+            self.client = discovery.build('storage', 'v1', **build_kwargs)
```
dlstadther: You've undone the default `cache_discovery` value supplied by #2361 . This addition was recommended by @himikof . Did you intentionally remove?

danthelion: Yes, the original intention of my PR was regarding this parameter a well.
I think it's better to modify it on the client side (by propagating it with `build_kwargs`) and `True` is always a sane default for caching functionalities.

dlstadther: @himikof Would you like to voice your opinion on moving the default back to `True`?

himikof: Well, though I in general agree that turning caching on is a sane default in general, I do disagree in this particular case. The discovery caching in `googleapiclient` is mostly broken, as indicated in the linked issues, and nobody tried to fix this upstream for several years. The chances of file caching working are even more reduced by #2361 dropping `oauth2client` dependency.
The best solution would be to have some kind of simple in-memory cache without external dependencies by default in the upstream `googleapiclient`, but I do not believe this is a Luigi issue.
What Luigi can do, though, is to ensure GCP-related functionality works out of the box, without getting scary `ImportError` log spam or learning about API discovery caching (which is not that important to know about).
So I would say that getting a way to pass custom discovery kwargs is certainly an improvement, but I feel that Luigi should change the default of the `cache_discovery` value, at least until the discovery caching works out of the box in the upstream library. It is a simple matter of doing `kwargs.setdefault('cache_discovery', False)`.

danthelion:
@himikof I have added `'cache_discovery' = False` as a default parameter for the discovery build call, I think you are right about it, produces way too many logs when caching is enabled and it should be fixed in upstream.

PR #2294, opened 2017-11-25
Removed message 'No Instance(s) Available.' in Windows when starting …

hirolau:
## Description

In Window when starting luigi the message `No Instance(s) Available.` is shown in the terminal. This is due to the fact that command `wmic` throws this error if no process with the given pid is found. We really do not care about this error, so this change redirects the error message to `nul`

The method of redirecting standard error was found here:

https://stackoverflow.com/questions/4507312/how-to-redirect-stderr-to-null-in-cmd-exe

## Motivation and Context
The message was confusing to new users and gave no important information.

See for example here: https://stackoverflow.com/questions/40593709/luigi-no-instances-available

Updated the unit tests to work with Windows, and have run my pipeline without problems.

### DIFF
--- a/luigi/lock.py
+++ b/luigi/lock.py
@@ -37,7 +37,7 @@ def getpcmd(pid):
     """
     if os.name == "nt":
         # Use wmic command instead of ps on Windows.
-        cmd = 'wmic path win32_process where ProcessID=%s get Commandline' % (pid, )
+        cmd = 'wmic path win32_process where ProcessID=%s get Commandline 2> nul' % (pid, )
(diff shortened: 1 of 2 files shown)


hirolau:
Sorry, happened to miss that os was already imported. Is this ok or should I make a totally new branch and create a new request?

hirolau:
I am a bit lost of why my pull-request would fail the `codecov/patch`. Any idea what I can do to fix it?

hirolau:
So, what is needed to get this one merged?

Tarrasch:
Thanks for this!

PR #2295, opened 2017-11-29
Optional parameter

daveFNbuck:
## Description
Create an OptionalParameter class that serializes None as the empty string and use this class whenever we have a parameter with default None. Empty string gets parsed to None.

## Motivation and Context
I get a long series of type warnings from default None values in the Luigi codebase whenever I run my pipeline. This gets rid of those, and should be safe as long as no one needs to differentiate between the empty string and None.

I added some unit tests for the OptionalParameter class. I haven't tried running a pipeline with it yet, but I will soon.


daveFNbuck:
Forgot to update this, but I've been using this in production about 2 weeks now and it works fine.

Issue #2202, opened 2017-08-04
Feature: Integration with Databricks API


colemanja91:
Databricks (Spark-as-a-service) recently added native support to Airflow for their Jobs API (https://docs.databricks.com/api/latest/jobs.html, https://databricks.com/blog/2017/07/19/integrating-apache-airflow-with-databricks.html). 

I'd like to work on adding a similar functionality to Luigi as a new contrib submodule. Still thinking through what targets/outputs would be, will add thoughts here as I start working through it. 
Any thoughts or comments?

a1xt06:
I think a DatabricksSparkSubmitTask using Databricks' Jobs API and another module that could kick-off a cluster using Databricks' Cluster API would be very useful.

colemanja91:
@a1xt06 I hadn't thought about kicking off a cluster! Might take a pass at that too. Was also wondering if having a set of DBFS targets would be helpful.

a1xt06:
DBFS targets might be useful, but as a first pass, using Luigi's S3Target for targets would probably be enough, especially considering most DBFS endpoints will be associated with mounted s3 buckets. I'm guessing doing DBFS endpoints would require using Databricks' DBFS API, and I haven't looked into that API.

At the same time, it might be nice to have DBFS endpoints as a part of a roadmap, considering Databricks' is trying to add Azure support (maybe?) later this year, so the S3Target might not be enough for Luigi on Microsoft's cloud.

colemanja91:
@a1xt06 I just added a pull request (took me long enough, eh); if you have the ability to do some light review or testing would appreciate your input. 

And since now Databricks does have Azure support (with DBFS pointing to Azure Blob Storage) it may be more helpful to have DBFS targets defined. I might be able to get started on this over the holiday break.

kabirkhan:
See https://github.com/spotify/luigi/pull/2312 for an Azure Data Lake file storage option. I may write a module to support Azure Blob Storage as well.

wkeithvan:
Are there still any plans to add this functionallity? I see a stale PR [here](https://github.com/spotify/luigi/pull/2296) that started to get it to work...

PR #2296, opened 2017-12-03
WIP: Databricks submit

colemanja91:
## Description

Added new submodule `luigi.contrib.databricks` with base functionality for submitting jobs and creating/deleting clusters via the Databricks API (https://docs.databricks.com/api/latest/index.html)
## Motivation and Context

#2202 
Similar to how `luigi.contrib.dataproc` adds support for remote Spark job execution via GCP; and 'luigi.contrib.kubernetes` supports remote jobs configured over API calls.  

I would still like to add unit tests and an example before merge, but would like an initial round of feedback before going much further. 

In particular, I am wondering if there is a better way to handle tracking of clusters when running Create/Delete. For GCP, it is easy to track as the cluster name is expected to be a unique identifier; in Databricks, the cluster name is not required to be unique - it's not even required at all (i.e., you can create a cluster with "" as the name). Instead, a new `cluster_id` parameter is returned from the initial create cluster API call, and that is used as reference. 

Right now I'm essentially requiring a user-defined `signal_complete()` method which writes out the `cluster_id` to a target; the target must be loaded in to whatever upstream task requires it. 

One alternative: treat `cluster_name` as globally-unique as far as Luigi is concerned. Then it's just a GET call to `/clusters/list`, and grab the one with a matching name and `state=='RUNNING'`. The risk is that if someone manually creates a cluster with that name in the same workspace, the task would use that cluster, which may be an ill-fit for the job. 

I ran basic jobs with this code; unit tests are coming next


colemanja91:
Hmmmm.... curious about the code coverage here. @Tarrasch is this just a matter of needing more tests?

Tarrasch:
I don't know why code coverage is running. I thought I disabled it at some point...

PR #2301, opened 2017-12-06
Documented way for starting Luigi tasks inside Python code

Gr1f0n6x:
## Description

Information about another existing way for starting Luigi tasks
## Motivation and Context
This way can be useful if user wants to get dynamic params from anywhere before start the task and pass them automatically


Tarrasch:
Looks good:

 * Can you change the filename from "comman_line" to "running_luigi" or something?
 * Perhaps remove the class file code?
 * If you mention `run()` you should also mention `build` I think to be fair, and also link to the whole run interface code module.

erikbern:
I don't think this is the right way to run tasks from inside Python. Why deal with the command line interface at all

erikbern:
Look at `interface.build`: https://github.com/spotify/luigi/blob/master/luigi/interface.py#L251

We should probably document that somewhere

Tarrasch:
Actually yes, looking at the code it seems it heeds quite strong warnings against using it (written by myself a long time ago).

Gr1f0n6x:
Oh, yes, i had to document build, not run (as i see now). It's all because i use run method in my code despite on warnings about it

dlstadther:
Any update here @Gr1f0n6x ?

Gr1f0n6x:
@dlstadther yes, sorry for delay (a lot of work), i'll continue to work on this issue and try to finish another one (parameter visibility)

Tarrasch:
Wow, what an improvement! Thanks!

PR #2302, opened 2017-12-06
Fixes dangerous default parameter values to make them immutable

sklarsa:
Per comments in #2275, fixes dangerous default values by setting them to None

## Description
See above

## Motivation and Context
#2275


Tarrasch:
Thanks!

Issue #2298, opened 2017-12-05
Redshift COPY fails in luigi 2.7.1 when columns are not provided


dmohns:
Running Redshift COPY jobs with `columns = None` to prohibit table creation fails in luigi 2.7.1 with 
```
TypeError: object of type 'NoneType' has no len()
```

The root cause seems to be https://github.com/spotify/luigi/pull/2245/files#diff-778ea3db4cccaf4de6564889c5eb670fR338

A possible solution would be to change the line to
```python
if self.columns and len(self.columns) > 0:
```
unless I am missing some reason to explicitly ask only for `len(self.columns)`.

dlstadther:
This seems like a reasonable solution. The motivation here was to handle the [default `column`](https://github.com/spotify/luigi/blob/master/luigi/contrib/rdbms.py#L71) definition or some structure with length > 0. Really, we should ensure that it is not `None` and an iterable with length greater than zero.

Feel free to PR!

PR #2303, opened 2017-12-07
Allow NoneType columns in Redshift COPY job

dmohns:
## Description
Preceding booleanness check lowers requirements for the columns attribute and ensures backward-compatibility.

## Motivation and Context
It fixes issue #2298.

### DIFF
--- a/test/contrib/redshift_test.py
+++ b/test/contrib/redshift_test.py
@@ -40,9 +40,11 @@ class DummyS3CopyToTableBase(luigi.contrib.redshift.S3CopyToTable):
     user = 'dummy_user'
     password = 'dummy_password'
     table = luigi.Parameter(default='dummy_table')
-    columns = (
-        ('some_text', 'varchar(255)'),
-        ('some_int', 'int'),
+    columns = luigi.TupleParameter(
+        default=(
+            ('some_text', 'varchar(255)'),
+            ('some_int', 'int'),
+        )
     )
 
     copy_options = ''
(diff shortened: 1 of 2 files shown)


dmohns:
@dlstadther I added some basic tests for the cases you suggested. I am not too familiar with writing tests. Happy to receive further comments.

test/contrib/redshift_test.py
```diff
@@ -214,6 +216,123 @@ def test_s3_copy_to_temp_table(self, mock_redshift_target, mock_copy):
             (task.table,),
         )
 
+    @mock.patch("luigi.contrib.redshift.RedshiftTarget")
+    def test_s3_copy_with_valid_columns(self, mock_redshift_target):
+        task = DummyS3CopyToTableKey(columns=(
+            ('some_text', 'varchar(255)'),
+            ('some_int', 'int'),
```
dlstadther: Since this is what is defined as the default, you should be able to just `task = DummyS3CopyToTableKey()`. Right?

dmohns: The idea was to emphasize on the valid columns types. But re-thinking it: I agree that it should be clear enough what the test is doing even when simply using the default `columns` attribute from `DummyS3CopyToTableKey()`.

Side note: For changes like this is there any preference whether to amend preceding commits or add them to a new commit?

dlstadther: A new commit is fine. We squash/merge PRs and amend the message where necessary. So no need to rebase on your end (unless you just want to).

dlstadther:
Sorry for the delay here @dmohns . Thanks for your contribution!

PR #2306, opened 2017-12-08
Fix a warning in googleapiclient usage in contrib.gcs

himikof:
The constructor of `googleapiclient.http.MediaFileUpload` is marked with `@positional(2)`, which means it only accepts 2 positional arguments (`self` and `filename`). This change should get rid of mysterious `__init__() takes at most 2 positional arguments (3 given)` message in the task output.


himikof:
@Tarrasch , ping

This should be OK to merge, right?

Tarrasch:
Thanks for the ping and sorry for the delay.

PR #2307, opened 2017-12-11
 Fix MRO on tasks using util.requires #2054

### LINKED ISSUE #2054: util.inherits and requires use inheritence, breaking the MRO
As discussed in on [the mailing list](https://groups.google.com/forum/#!topic/luigi-user/7m0uZkyyYow), using @inherits messes up the method resolution order and breaks `super()`

```
>>> class Foo(luigi.Task): pass
...
>>> @inherits(Foo)
... class Bar(luigi.Task): pass
...
>>> Bar.__mro__
(<class '__main__.Bar'>, <class '__main__.Bar'>, <class 'luigi.task.Task'>, <type 'object'>)
```

Since Bar is now twice on the list, any calls to super will result in infinite loops.

This problem comes from @inherits using inheritance to modify the class. We should look into whether the class can be monkey-patched instead of subclassed, so that inheritance works as expected.

DavidS3141:
## Description
Add a test for luigi.util.requires which fails, as the inherits and requires decorators mess with the MRO and a super call results in infinite loops (#2054).
Document a bit of the source code for clarity.
Instead of subclassing in the decorators add the methods clone_parent and requires to the task that inherits by monkey patching them as proposed in #2054.
This is my first contribution to a bigger GitHub project, so please watch out for newbie mistakes :)

## Motivation and Context
This fixes the issue #2054!

I added a test specifically looking at the MRO of a requiring task, which fails on the current master and returns ok on my fix. Running all tests using `tox -e py27-nonhdfs` returned the same 30 skips and 3 errors both on the master and my fix (excluding my new test for the master). Those tests have no apparent connection to luigi.util.requires and inherits (see [test-output.txt](https://github.com/spotify/luigi/files/1548207/test-output.txt)).

## Possible issue with this PR
I am not sure what the exact reason was for using a wrapped class to add the methods clone_parent and requires in the first place, at least I could not come up with a scenario, that would require subclassing instead of monkey patching. If there is such a reason/scenario one might want to check if this PR breaks it!

### DIFF
--- a/luigi/util.py
+++ b/luigi/util.py
@@ -274,8 +274,11 @@ class inherits(object):
         self.task_to_inherit = task_to_inherit
 
     def __call__(self, task_that_inherits):
+        # Get all parameter objects from the underlying task
         for param_name, param_obj in self.task_to_inherit.get_params():
+            # Check if the parameter exists in the inheriting task
             if not hasattr(task_that_inherits, param_name):
+                # If not, add it to the inheriting task
                 setattr(task_that_inherits, param_name, param_obj)
 
         # Modify task_that_inherits by subclassing it and adding methods
--- a/luigi/util.py
+++ b/luigi/util.py
@@ -281,14 +281,12 @@ class inherits(object):
                 # If not, add it to the inheriting task
                 setattr(task_that_inherits, param_name, param_obj)
 
-        # Modify task_that_inherits by subclassing it and adding methods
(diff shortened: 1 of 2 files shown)


DavidS3141:
Yes, it does make sense. I just copied all tests for the `requires` decorator and adapted them for `inherits`. They are all running fine!

Tarrasch:
Thanks! Great job!

PR #2309, opened 2017-12-11
Code format

Gr1f0n6x:
## Description

Added some semicolons + spaces in `if` blocks between arguments

## Motivation and Context
The major part of JS code has semicolons at the end of statement, so i think that it would be good to keep this code-style in other parts too.
Also it concerns spaces between arguments in `if` statements

I ran my jobs with this code and it works for me (as i see - nothing bad happened)


Tarrasch:
Did you test it? :)

Gr1f0n6x:
Yep, i ran some tasks =)
Did i break something?

Tarrasch:
Thanks! Merged!

PR #2310, opened 2017-12-11
Added cleanup of tempfiles in case ftp-get command failure.

hirolau:
## Description

The `get` command of `RemoteTarget` ftp class leaves an empty tmp files if not successful.

This change catches the error, removes the tempfile and then raises the original errors again.

## Motivation and Context

Luigi is meant to run often, a repeating failure could lead to many temp-files being created on disk for no good reason. This has happened for me from time to time due to configurations errors or missing ftp-files.

I have edited the ftp test case to run with Python 3, then added a test case for this. I have tested to "fail" a download on both FTP and SFTP.

* Note that I was unable to run the original test as the test expected a file to exist "/tmp/luigi-test-ftp" when the instructions ask us to add this folder. I have to change this to use a file inside that folder instead. Then the test-case ran successfully.


himikof:
Could a `tempfile.NamedTemporaryFile` be used instead of re-implementing it?

kwilcox:
The tests are hardcoding `/tmp` and should probably use `tempfile.gettempdir()`.

PR #2316, opened 2017-12-19
Introduced process_requires

TMaYaD:
## Description
Adds a `process_requires` method to `Task` class. Similar to `_requires` and `process_resources`, this allows a Template task to inspect and modify dependencies before passing them to Luigi worker.

This is different from `_requires` or `deps` in that this works with dynamic dependencies yielded from `run` as well.

## Motivation and Context
We have a template task that needed to pass some parameters along the dependency tree. This change could be used to obviate the need for using `clone` in every Task:
```
def process_requires(self, requires):
    for req in luigi.task.flatten(requires):
        req.job_id = self.job_id
    return requires
```

"We ran our jobs with this code and it works for us."
I could use some help in adding unit tests!

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -575,6 +575,18 @@ class Task(object):
         """
         return []  # default impl
 
+    def process_requires(self, requires):
+        """
+        Override in "template" tasks which themselves are supposed to be
+        subclassed and thus have their requires() overridden (name preserved to
+        provide consistent end-user experience), yet need to introduce
+        custom behaviour into dependencies.
+
+        This defers from _requires in that this will not directly call task.requires
+        and this is being used by dependencies yielded by run
+        """
+        return requires
+
     def requires(self):
         """
         The Tasks that this Task depends on.
@@ -627,7 +639,7 @@ class Task(object):
         Returns the flattened list of requires.
         """
         # used by scheduler
-        return flatten(self._requires())
+        return flatten(self.process_requires(self._requires()))
 
     def run(self):
         """
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -147,7 +147,7 @@ class TaskProcess(multiprocessing.Process):
             except StopIteration:
                 return None
 
-            new_req = flatten(requires)
+            new_req = flatten(self.task.process_requires(requires))
             if all(t.complete() for t in new_req):
                 next_send = getpaths(requires)
             else:


Tarrasch:
This looks overall sound:

 * Of course some sort of tests are required
 * Maybe I'm tired but I felt reading the docstring was really confusing. Can you phrase it with simpler sentences?
 * Perhaps have an example, it serves as documentation and maybe more users will adopt it.

ramyala:
@Tarrasch could you merge this upstream?

luigi/task.py
```diff
@@ -575,6 +575,40 @@ def output(self):
...
         return []  # default impl
 
+    def process_requires(self, requires):
+        """
+        Override in "template" tasks which themselves are supposed to be
+        subclassed and thus have their requires() overridden (name preserved to
+        provide consistent end-user experience), yet need to introduce
+        custom behaviour into dependencies.
+
+        Differences with :py:meth:`_requires`:
```
Tarrasch: We shouldn't mention the internal function should we?

luigi/task.py
```diff
@@ -575,6 +575,40 @@ def output(self):
...
+            requirements where as this method works with dynamic dependencies yielded from
+            :py:meth:`run` also.
+
+        ex:
+            class ParentTask(luigi.Task):
+                foo = `bar`
+
+                def process_requires(self, requirements):
+                    for(req in flatten(requirements)):
+                        req.foo = self.foo
```
Tarrasch: Seem scary to modify input arguments like that. Not sure if the docs should encourage it.

luigi/task.py
```diff
@@ -575,6 +575,40 @@ def output(self):
...
+
+                    return requirements
```
Tarrasch: what about just 1 line `return flatten(requirements) + self._paent_task_extra_requirements()`?

luigi/task.py
```diff
@@ -575,6 +575,40 @@ def output(self):
...
+
+            class B(ParentTask):
+                def requires(self):
+                    return luigi.LocalPathTask('my/local/path')
+
+            B().deps()[0].foo # => 'bar'
```
Tarrasch: I know `deps()` isn't prefixed with `_` but I think we would like to make it private, so better not referencing it either.

luigi/task.py
TMaYaD: This won't do. `process_requires` is for modifying the `requires` returned by the task, not add more requirements, which is already covered by the use-case described in `_requires`. Also using such an override with `process_requires` would result in same dependency being attempted multiple times in case of dynamic dependencies yielded from `run`.

luigi/task.py
TMaYaD: In our use case `foo` is actually used for log tags used to identify logs of a single pipeline throughout the run.
It's a powerful ability and I agree it could be dangerous to modify the input arguments and that it should be used with caution.

luigi/task.py
TMaYaD: Documentation for `_requires` demonstrates similar use case(Advanced use with template tasks). So I think it's okay.

luigi/task.py
TMaYaD: We could remove this. I'm open to ideas on how we can demonstrate the results of the override?

luigi/task.py
Tarrasch: Can you add to the doc-string for `_requires()` (and also for `deps()`) and just mention "DEPRECATED don't override". For the `_requires()`-use-cases I would it if `process_requires()` could be the new go-to-replacement for it.

luigi/task.py
Tarrasch: Sure. Feel free to use it like that, but let's not encourage it in the docs?

luigi/task.py
Tarrasch: Well, you would also "modify" your dependency multiple times in case of dynamic dependencies right?

Can you aim to make this new function the "recommended replacement" for `_requires()`? It's probably just a very old mistake that we encourage users to override a private function like that.

luigi/task.py
Tarrasch: I think it's fine to just remove this line for now. The behavior is already described in text above right?

TMaYaD:
Fair enough, I'll work on it and update this PR.

PR #2318, opened 2017-12-21
Deprecate param_args

TMaYaD:
## Description
This change deprecates `param_args` on `Task` and changes the equality check to use `param_kwargs` instead. This is the only place that uses the property.

The property itself duplicates the information already available in `param_kwargs` and `get_params()`  and that's one more thing to maintain and possibly go out of sync.

This aims to not change behaviour so tests are left intact.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -532,7 +532,7 @@ class Task(object):
         return task_str
 
     def __eq__(self, other):
-        return self.__class__ == other.__class__ and self.param_args == other.param_args
+        return self.__class__ == other.__class__ and self.param_kwargs == other.param_kwargs
 
     def complete(self):
         """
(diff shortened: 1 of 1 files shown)


Tarrasch:
I'm totally fine with this, but why did you change this? Honestly I have forgotten that they are stored and public like that. Any reason we don't also deprecate the `param_kwargs`?

TMaYaD:
A bit of background might help:
We are using `process_requires` from #2316 to add some dynamic attributes to tasks and faced an issue with the attributes not showing up on scheduler dashboard. In the process of debugging it I came across these two variables. 
```
```

Tarrasch:
LGTM, does anyone else have any opinion?

Issue #1012, opened 2015-06-19
TypeError: 'str' does not support the buffer interface in luigi.contrib.hive


red1ynx:
Hi,

I'm running luigi task with python3.4 and trying call print(HiveTableTarget(table = "tbl", database = "db").exists()).
I'm getting error the following stacktrace:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/luigi/worker.py", line 137, in run
    new_deps = self._run_get_new_deps()
  File "/usr/local/lib/python3.4/dist-packages/luigi/worker.py", line 88, in _run_get_new_deps
    task_gen = self.task.run()
  File "hive.py", line 10, in run
    print(str(target.exists()))
  File "/usr/local/lib/python3.4/dist-packages/luigi/contrib/hive.py", line 376, in exists
    return self.client.table_exists(self.table, self.database)
  File "/usr/local/lib/python3.4/dist-packages/luigi/contrib/hive.py", line 141, in table_exists
    return stdout and table.lower() in stdout
TypeError: 'str' does not support the buffer interface
```

I changed the last line in the stacktrace into 

```
return stdout and table.lower() in str(stdout)
```

and it works now.

Is this a bug or do I use wrong python version or something else?

Thanks

Tarrasch:
It's probably just that our python 3 users haven't tried out Hive yet. You was the first one. Please add a test case and fix the bug in a PR. :)

pmerienne:
Hi,
This issue still exists. @red1ynx fix seems to be ok. @Tarrasch what do you need to include this fix in the next release?

mfcabrera:
ping @Tarrasch @red1ynx  can we create a PR for this. Fix seems trivial here.

Tarrasch:
@pmerienne You can create at PR with this fix included. See our CONTRIBUTING.md file. Good luck! :)

PR #2323, opened 2018-01-04
Fix Python 3 TypeError in contrib.hive.HiveTableTarget.exists()

sd2k:
## Description
Stdout is `bytes` which is fine in Python 2 since the rest of the strings are too, but in Python 3 we need it to be `unicode` so we can do string comparisons. This fixes a long-standing bug, it seems!

## Motivation and Context
See #1012.


Tarrasch:
Thanks!

PR #2325, opened 2018-01-05
Added SlurmJobTask to run as luigi tasks

jcftang:
This Task is derived from the existing SGE Task class. This SlurmJobTask can execute jobs on a slurm based cluster. It is known to work with python 2.7 and 2.4 (on centos7).

It is setup so that it allows users to allow for nodes, memory, partition, time and gres resources to be requested.

### DIFF
--- a/luigi/contrib/slurm.py
+++ b/luigi/contrib/slurm.py
@@ -54,7 +54,7 @@ The following is an example usage (and can also be found in ``slurm_tests.py``)
         def work(self):
             logger.info('Running test job...')
             with open(self.output().path, 'w') as f:
-                f.write('this is a test')
+                f.write('this is a test\n')
 
         def output(self):
             return luigi.LocalTarget(os.path.join('/home', 'testfile_' + str(self.i)))
@@ -99,6 +99,7 @@ try:
     import cPickle as pickle
 except ImportError:
     import pickle
+import itertools
 
 import luigi
 from luigi.contrib.hadoop import create_packages_archive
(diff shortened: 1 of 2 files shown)


Tarrasch:
I think you need reviewers that know Slurm here. :)

pietromarchesi:
I'd be interested in seeing this working. Is it moving forward / do you need help?

jcftang:
@pietromarchesi I got some feedback from @nathantsoi and there's been a few changes, I think the plugin requires more rigourous use by others before this can really progress any further.

nathantsoi:
thx @jcftang for the great work, i've been using this in production for a few months now and it's working great.

the final thing i've been tweaking is allowing dynamic increases to memory allocation on task failures via the scheduler. i'm not sure if I'm doing this the right way, if anyone has input: https://github.com/nathantsoi/luigi/commit/7c07e020b778860de51c9f9f28b33bd0f821d87a

the idea is to double the allowed memory up to a given limit on every failure due to out of memory errors

the reason for this final feature is to allow tasks with wide-ranging memory requirements to complete successfully (perhaps not efficiently, but at least finish) with minimal user configuration and interaction

jcftang:
AFAIK this is failing on some pep8 checks, I can clean this up and see what people think, I'm still interested in this getting merged.

luigi/task.py
```diff
@@ -276,7 +276,7 @@ def trigger_event(self, event, *args, **kwargs):
                 except KeyboardInterrupt:
                     return
                 except BaseException:
-                    logger.exception("Error in event callback for %r", event)
+                    logger.exception("Error in event callback (%s) for %r", callback, event)
```
Tarrasch: Always put changes like this in separate PRs please

nathantsoi: ahh, sry. i believe it was in another pr somewhere, thought this got merged into master... but maybe not...

Issue #2324, opened 2018-01-05
Inconsistent task_id if using DictParameter


BinrOp:
If we pass dict parameters, the task_ids will not be consistent across multiple executions.

# Reproduce issue
**Dependencies**
For `luigi==2.7.1` and `Python 3.5.2`.

**Run this script multiple times:**
```import luigi
import luigi.parameter
from luigi import DictParameter
import json

class Task(luigi.Task):
    my_dict = DictParameter()


params_str_values = []
task_ids = []
for i in range(100):
    t = Task(my_dict={'a': 1, 'b': 2})

    task_ids.append(t.task_id)
    str_params = t.to_str_params()
    params_str_values.append(str_params['my_dict'])

print("Using Original DictParameter, for 100 runs:")
print("unique dict params: {}".format(set(params_str_values)))
print("unique task ids: {}".format(set(task_ids)))


class TmpFixedDictParameter(DictParameter):

    def serialize(self, x):
        return json.dumps(x,
                          cls=luigi.parameter._DictParamEncoder,
                          # NOTE: change this to the original version
                          sort_keys=True)

class Task(luigi.Task):
    my_dict = TmpFixedDictParameter()

params_str_values = []
task_ids = []
for i in range(100):
    t = Task(my_dict={'a': 1, 'b': 2})

    task_ids.append(t.task_id)
    str_params = t.to_str_params()
    params_str_values.append(str_params['my_dict'])

print("Using Fixed DictParameter, for 100 runs:")
print("unique dict params: {}".format(set(params_str_values)))
print("unique task ids: {}".format(set(task_ids)))
```
**Output:**
```
# run 1 
Using Original DictParameter, for 100 runs:
unique dict params: {'{"b": 2, "a": 1}'}
unique task ids: {'Task___b___2___a___1__f39c2636f7'}
Using Fixed DictParameter, for 100 runs:
unique dict params: {'{"a": 1, "b": 2}'}
unique task ids: {'Task___a___1___b___2__4949d6be98'}

# run 2
Using Original DictParameter, for 100 runs:
unique dict params: {'{"a": 1, "b": 2}'}
unique task ids: {'Task___a___1___b___2__4949d6be98'}
Using Fixed DictParameter, for 100 runs:
unique dict params: {'{"a": 1, "b": 2}'}
unique task ids: {'Task___a___1___b___2__4949d6be98'}
```

The consistency in one execution seems related to python dictionary handling.

[...]

PR #2330, opened 2018-01-15
Make task_id deterministic with DictParameter

BinrOp:
## Description
Sort DictParameter keys when serializing,

## Motivation and Context
The DictParameter serialization is not deterministic and so is not the task_id of a task containing a DictParameter. Note that this behaviour might vary among versions of python.

The changes are minimal and should not break anything (not worsening existing behaviour).
  
This PR fixes this issue https://github.com/spotify/luigi/issues/2324

Tested with  `luigi==2.7.1` and `Python 3.5.2.` in different processes.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -916,7 +916,7 @@ class DictParameter(Parameter):
         return json.loads(s, object_pairs_hook=_FrozenOrderedDict)
 
     def serialize(self, x):
-        return json.dumps(x, cls=_DictParamEncoder)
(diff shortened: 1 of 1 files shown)


Tarrasch:
Hmmm... Not sure how this should work. Given that the name is "DictParameter" we should be able to handle normal (unordered) dicts as input. And I suppose "demanding" that input is sorted will break more code than this will break.

Does anyone else who uses this want tho chime in?

Issue #2334, opened 2018-01-18
[contrib.kubernetes] Hangs on 'Pod errors: Unschedulable'


yonromai:
When asking for more resources (e.g. CPU) than the Kubernetes cluster can allocate, Kubernetes fails the job (rightfully) with a status `Pod errors: Unschedulable`.

The `KubernetesJobTask` logic, however, fails to catch this error and hangs:
```
DEBUG: Kubernetes job example-luigi-b8ffbfbbeb7b429bb8cc1e81c6a879c9 is still running
DEBUG: Kubernetes job example-luigi-b8ffbfbbeb7b429bb8cc1e81c6a879c9 is still running
DEBUG: Kubernetes job example-luigi-b8ffbfbbeb7b429bb8cc1e81c6a879c9 is still running
...
```
(FWIW, I'm using GKE here)

yonromai:
@mcapuccini ^

mcapuccini:
Hi @yonromai. Thanks for reporting the issue. The problem should on this block of code: https://github.com/spotify/luigi/blob/6ce970846c0ca8c2850bc168b686340e15982b2b/luigi/contrib/kubernetes.py#L160-L171, which only checks for pod status "running", "succeeded" and throws error for everything else. I think that `pykube`, the library that we are currently using for the Kubernetes Task, returns running when the pod is unschedulable, and that you need to run an additional command to query the unschedulable status. Unfortunately I don't have time to work on it at the moment, but since you are getting the error in one of your settings, it could be a nice contribution if you could sort it out. I'd be happy to review and merge your patch 🙂 .

yonromai:
@mcapuccini thanks for coming back to me. 

I took a look at it yesterday, I think the problem might rather be in the `self.__get_job_status()`. 
I played with it and printed the contents of `pod.obj['spec']`, `node.obj['spec']` and `job.obj`. All of these objects had the same data whether the job was actually scheduled or not (because too much resources was requested). 
Also,  `node.unschedulable` was `False` in both cases.

I'm wondering if the problem could be with pykube/GKE compatibility.

PR #2335, opened 2018-01-19
Various improvements for luigi.contrib.kubernetes

yonromai:
This PR contains a bugfix (#2334) and various small improvements for the Kubernetes module, each one in its own commit.

### DIFF
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -43,7 +43,7 @@ logger = logging.getLogger('luigi-interface')
 try:
     from pykube.config import KubeConfig
     from pykube.http import HTTPClient
-    from pykube.objects import Job
+    from pykube.objects import Job, Pod
 except ImportError:
     logger.warning('pykube is not installed. KubernetesJobTask requires pykube.')
 
@@ -159,6 +159,7 @@ class KubernetesJobTask(luigi.Task):
 
     def __track_job(self):
         """Poll job status while active"""
+        self.__verify_job_has_started()
         while (self.__get_job_status() == "running"):
             self.__logger.debug("Kubernetes job " + self.uu_name
                                 + " is still running")
@@ -181,16 +182,33 @@ class KubernetesJobTask(luigi.Task):
         """
         pass
 
+    def __verify_job_has_started(self):
+        """Asserts that the job has successfully started"""
+        # Verify that the job started
+        jobs = Job.objects(self.__kube_api) \
+            .filter(selector="luigi_task_id="+ self.job_uuid) \
+            .response['items']
+
+        assert len(jobs) > 0, "Kubernetes job " + self.uu_name + " not found"
+        self.__job = jobs[0]
+
+        # Verify that the pod started
+        pods = Pod.objects(self.__kube_api) \
+            .filter(selector="job-name=" + self.uu_name) \
+            .response['items']
+
+        assert len(pods) > 0, "No pod scheduled by " + self.uu_name
+
+        for pod in pods:
+            for cond in pod['status']['conditions']:
+                if 'message' in cond:
+                    assert cond['status'] != 'False', \
+                        "[ERROR] %s - %s" % (cond['reason'], cond['message'])
+
     def __get_job_status(self):
         """Return the Kubernetes job status"""
-        # Look for the required job
-        jobs = Job.objects(self.__kube_api).filter(selector="luigi_task_id="
-                                                            + self.job_uuid)
-        # Raise an exception if no such job found
-        if len(jobs.response["items"]) == 0:
-            raise RuntimeError("Kubernetes job " + self.uu_name + " not found")
         # Figure out status and return it
-        job = Job(self.__kube_api, jobs.response["items"][0])
+        job = Job(self.__kube_api, self.__job)
         if ("succeeded" in job.obj["status"] and job.obj["status"]["succeeded"] > 0):
             job.scale(replicas=0)  # Downscale the job, but keep it for logging
             return "succeeded"
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -32,6 +32,7 @@ Requires:
 
 Written and maintained by Marco Capuccini (@mcapuccini).
 """
+from datetime import datetime
 
 import luigi
 import logging
@@ -48,7 +49,7 @@ except ImportError:
     logger.warning('pykube is not installed. KubernetesJobTask requires pykube.')
 
 
-class kubernetes(luigi.Config):
+class Kubernetes(luigi.Config):
     auth_method = luigi.Parameter(
         default="kubeconfig",
         description="Authorization method to access the cluster")
@@ -63,19 +64,20 @@ class kubernetes(luigi.Config):
 class KubernetesJobTask(luigi.Task):
 
     __POLL_TIME = 5  # see __track_job
-    kubernetes_config = kubernetes()
+    kubernetes_config = Kubernetes()
 
     def _init_kubernetes(self):
         self.__logger = logger
         self.__logger.debug("Kubernetes auth method: " + self.auth_method)
-        if(self.auth_method == "kubeconfig"):
+        if self.auth_method == "kubeconfig":
             self.__kube_api = HTTPClient(KubeConfig.from_file(self.kubeconfig_path))
-        elif(self.auth_method == "service-account"):
+        elif self.auth_method == "service-account":
             self.__kube_api = HTTPClient(KubeConfig.from_service_account())
         else:
             raise ValueError("Illegal auth_method")
         self.job_uuid = str(uuid.uuid4().hex)
-        self.uu_name = self.name + "-luigi-" + self.job_uuid
+        now = datetime.utcnow()
+        self.uu_name = "%s-%s-%s" % (self.name, now.strftime('%Y%m%d%H%M%S'), self.job_uuid[:16])
 
     @property
     def auth_method(self):
@@ -184,8 +186,7 @@ class KubernetesJobTask(luigi.Task):
             time.sleep(self.__POLL_TIME)
             status = self.__get_job_status()
 
-        if status == "FAILED":
-            raise RuntimeError("Kubernetes job " + self.uu_name + " failed")
+        assert status != "FAILED", "Kubernetes job " + self.uu_name + " failed"
 
         # status == "SUCCEEDED"
         self.__logger.info("Kubernetes job " + self.uu_name + " succeeded")
@@ -239,7 +240,15 @@ class KubernetesJobTask(luigi.Task):
         assert len(pods) > 0, "No pod scheduled by " + self.uu_name
 
         for pod in pods:
-            for cond in pod.obj['status']['conditions']:
+            status = pod.obj['status']
+            for cont_stats in status['containerStatuses']:
+                if 'terminated' in cont_stats['state']:
+                    t = cont_stats['state']['terminated']
+                    err_msg = "Pod %s %s (exit code %d). Logs: `kubectl logs pod/%s`" % (
+                        pod.name, t['reason'], t['exitCode'], pod.name)
+                    assert t['exitCode'] == 0, err_msg
+
+            for cond in status['conditions']:
                 if 'message' in cond:
                     if cond['reason'] == 'ContainersNotReady':
                         return False
@@ -300,7 +309,7 @@ class KubernetesJobTask(luigi.Task):
         # Update user labels
         job_json['metadata']['labels'].update(self.labels)
         # Add default restartPolicy if not specified
-        if ("restartPolicy" not in self.spec_schema):
+        if "restartPolicy" not in self.spec_schema:
             job_json["spec"]["template"]["spec"]["restartPolicy"] = "Never"
         # Submit job
         self.__logger.info("Submitting Kubernetes Job: " + self.uu_name)
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -38,6 +38,8 @@ import logging
 import uuid
 import time
 
+from enum import Enum
+
 logger = logging.getLogger('luigi-interface')
 
 try:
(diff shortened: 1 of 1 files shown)


yonromai:
@mcapuccini @andrewsmartin PTAL

luigi/contrib/kubernetes.py
```diff
@@ -157,18 +157,35 @@ def max_retrials(self):
         """
         return self.kubernetes_config.max_retrials
 
+    @property
+    def delete_on_success(self):
+        """
+        Delete the Kubernetes workload if the job has ended successfully.
+        """
+        return True
```
mcapuccini: This should be False IMO. In many cases it is important to keep track of what was ran on the infrastructure along with the logs. If the Pod is deleted, then there is no way to do auditing. I am open for discussion though.

luigi/contrib/kubernetes.py
```diff
@@ -157,18 +157,35 @@ def max_retrials(self):
...
+
+    @property
+    def print_pod_logs_on_exit(self):
+        """
+        Fetch and print the pod logs once the job is completed.
+        """
+        return True
```
mcapuccini: I would have this to False as well. When you run a parallel workload the output would be difficult to read.

luigi/contrib/kubernetes.py
```diff
@@ -157,18 +157,35 @@ def max_retrials(self):
...
+
     def __track_job(self):
         """Poll job status while active"""
-        while (self.__get_job_status() == "running"):
-            self.__logger.debug("Kubernetes job " + self.uu_name
-                                + " is still running")
+        while not self.__verify_job_has_started():
+            time.sleep(1)
+        self.__print_kubectl_hints()
+        while True:
```
mcapuccini: Having while(true) kind of loops is a bad programming practice. I suggest to move the return conditions in the while statement.

luigi/contrib/kubernetes.py
```diff
@@ -157,18 +157,35 @@ def max_retrials(self):
...
+        """
+        return True
+
     def __track_job(self):
         """Poll job status while active"""
-        while (self.__get_job_status() == "running"):
-            self.__logger.debug("Kubernetes job " + self.uu_name
-                                + " is still running")
+        while not self.__verify_job_has_started():
+            time.sleep(1)
```
mcapuccini: Please put something like `self.__logger.debug("Kubernetes job " + self.uu_name + " not running yet")` here

mcapuccini: Maybe we shoud use `__POLL_TIME ` for the sleep

yonromai: I've never had to wait more than 1s for the job to start, so waiting more seems like wasting some time but I don't have too strong feelings about this. Will change.

luigi/contrib/kubernetes.py
yonromai: I don't believe that's how Kubernetes Jobs are supposed to be used - logs should be used for this purpose. Additionally, when Luigi DAGs are recurrently scheduled, not removing the successful jobs creates clutter and forces an extra cleanup step. Also, please note that if a job is not successful, the pod will not be removed - maintaining troubleshooting capabilities when needed.

yonromai:
@mcapuccini no worries, thank you for the review. Will make the changes and loop you in.

yonromai:
@mcapuccini done with changes. PTAL

andrewsmartin:
LGTM

luigi/contrib/kubernetes.py
mcapuccini: Yep, you are totally right!

luigi/contrib/kubernetes.py
```diff
@@ -63,19 +64,20 @@ class kubernetes(luigi.Config):
...
             self.__kube_api = HTTPClient(KubeConfig.from_file(self.kubeconfig_path))
-        elif(self.auth_method == "service-account"):
+        elif self.auth_method == "service-account":
             self.__kube_api = HTTPClient(KubeConfig.from_service_account())
         else:
             raise ValueError("Illegal auth_method")
         self.job_uuid = str(uuid.uuid4().hex)
-        self.uu_name = self.name + "-luigi-" + self.job_uuid
+        now = datetime.utcnow()
+        self.uu_name = "%s-%s-%s" % (self.name, now.strftime('%Y%m%d%H%M%S'), self.job_uuid[:16])
```
mcapuccini: I like to have the date in the job name, but lsn't it nice to have `luigi` in it as well? Any argument against it? Also why are you taking only 16 chars from the hex ID?

luigi/contrib/kubernetes.py
```diff
@@ -181,27 +203,90 @@ def signal_complete(self):
...
+        assert len(jobs) == 1, "Kubernetes job " + self.uu_name + " not found"
+        return Job(self.__kube_api, jobs[0])
+
+    def __print_pod_logs(self):
+        for pod in self.__get_pods():
+            logs = pod.logs(timestamps=True).strip()
+            self.__logger.info("Fetching logs from " + pod.name)
+            if len(logs) > 0:
+                for l in logs.split('\n'):
+                    self.__logger.info(l)
```
mcapuccini: This is very nice to have, but I am a bit concerned about messy mixed output from parallel jobs. We have some use cases in which we run 50 jobs in parallel, and it works nicely in terms of performance (with no need of Spark/Hadoop). What do you think?

luigi/contrib/kubernetes.py
```diff
@@ -181,27 +203,90 @@ def signal_complete(self):
...
+                assert 'waiting' not in cont_stats['state'],\
+                    "Pod %s %s. Logs: `kubectl logs pod/%s`" % (pod.name,
+                                                                cont_stats['state']['waiting']['reason'],
+                                                                pod.name)
+
+            for cond in status['conditions']:
+                if 'message' in cond:
+                    if cond['reason'] == 'ContainersNotReady':
+                        return False
+                    assert cond['status'] != 'False', \
```
mcapuccini: `cond['status'] == 'True'` may be easier to read, but maybe this is not a boolean?

luigi/contrib/kubernetes.py
```diff
@@ -181,27 +203,90 @@ def signal_complete(self):
...
+        for pod in pods:
+            status = pod.obj['status']
+            for cont_stats in status.get('containerStatuses', []):
+                if 'terminated' in cont_stats['state']:
+                    t = cont_stats['state']['terminated']
+                    err_msg = "Pod %s %s (exit code %d). Logs: `kubectl logs pod/%s`" % (
+                        pod.name, t['reason'], t['exitCode'], pod.name)
+                    assert t['exitCode'] == 0, err_msg
+
+                assert 'waiting' not in cont_stats['state'],\
```
mcapuccini: Does this mean that a waiting pod is going to fail? Sometimes we want to have more pods than workers, so we can have a sort of queuing system.

luigi/contrib/kubernetes.py
```diff
@@ -217,6 +302,7 @@ def run(self):
                 }
             },
             "spec": {
+                "activeDeadlineSeconds": 100,
```
mcapuccini: If I get this correctly, this should be passed as parameter and be way higher IMO, I have jobs that take way longer to complete.

luigi/contrib/kubernetes.py
yonromai: Right, so the reason is that Kubernetes constrains job names to be <64 characters. I ran into issues when I added the date, so I tried to reduce the amount of characters that are added to the name. 
After removing `-luigi` and halving the uuid, there are 32 characters available for the name (otherwise there would be 9 characters max for the jobname).

luigi/contrib/kubernetes.py
yonromai: Interesting point - I guess we can change the logger to output the worker id so interleaved logs and be used easily. 
The best solution to me seems to be to do less logging within luigi (this code is turned off by default now) and more within the docker container, and then collect the container logs.

luigi/contrib/kubernetes.py
yonromai: So I think this name (& the [k8s doc](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#job-termination-and-cleanup)) is confusing: It seems to me that it's the deadline for the Job to be scheduled. Once the Job is running, this timeout doesn't apply anymore.

luigi/contrib/kubernetes.py
yonromai: Not 100% sure what are all the values `status` can take - I just noticed that `False` means something went wrong.

luigi/contrib/kubernetes.py
mcapuccini: Sounds good.

luigi/contrib/kubernetes.py
mcapuccini: Yep, good to have it optional for now.

luigi/contrib/kubernetes.py
mcapuccini: I see. Yes, then I agree is more robust in this way.

luigi/contrib/kubernetes.py
mcapuccini: So does this apply to `ContainerCreating`? If so I would remove this line just to be safe, or at least I would have it parametrized.

luigi/contrib/kubernetes.py
```diff
@@ -247,10 +247,10 @@ def __verify_job_has_started(self):
                         pod.name, t['reason'], t['exitCode'], pod.name)
                     assert t['exitCode'] == 0, err_msg
 
-                assert 'waiting' not in cont_stats['state'],\
-                    "Pod %s %s. Logs: `kubectl logs pod/%s`" % (pod.name,
-                                                                cont_stats['state']['waiting']['reason'],
-                                                                pod.name)
+                if 'waiting' in cont_stats['state']:
+                    wr = cont_stats['state']['waiting']['reason']
+                    assert wr == 'ContainerCreating', "Pod %s %s. Logs: `kubectl logs pod/%s`" % (
```
mcapuccini: I think you should add the status `Pending` as well to your assert.

luigi/contrib/kubernetes.py
yonromai: I don't understand the safety concern with this parameter: my last commit `don't fail on ContainerCreating` means that `ContainerCreating` (which happens to me if I push a new docker image and kickoff luigi right after) is not considered as an error and allows retries. 
`activeDeadlineSeconds ` mean that if the job retries for 100s and still doesn't get scheduled, the luigi task should fail (instead of potentially hang for ever).
Agreed about the parameterization.

luigi/contrib/kubernetes.py
yonromai: I have never encountered this case, are you sure that `pod.obj['status'][0]['containerStatuses']['state']['waiting']['reason']` takes `Pending` as value?

mcapuccini: Nope, I am not sure 🙂 . Gonna trust you there.

mcapuccini:
I will ask a collaborator of mine to test this branch. We have a production workflow that uses the k8s, runner (https://www.biorxiv.org/content/early/2017/11/24/213603). If you don't mind waiting a couple of days, I think that would be a good test. Otherwise you can go ahead and merge it, and we'll test it later. Thank you so much @yonromai for sorting this out.

yonromai:
@mcapuccini No worries, thanks for the review.
The last release of luigi happened yesterday so I guess there is no time pressure

yonromai:
@mcapuccini please keep me posted on the results of your tests

carmat88:
@mcapuccini Unfortunately I couldn't find a good time slot for testing it. Will be able to do it beginning of next week (likely on Monday). Would that be fine?

mcapuccini:
@carmat88 no worries on my side 🙂

yonromai:
I guess that'll do

carmat88:
@yonromai @mcapuccini I've just finished to test this branch with this already long tested [workflow](https://github.com/phnmnl/MTBLS233-Jupyter) of ours.

I would claim that everything went fine as no errors were returns and thus the pre-processing analysis completed successfully. However I was wondering whether it would be better to have a unit test that could generate an error for which this patch was meant to resolve, thus officially identifying/reporting such kind of error.

What do you think about this idea?

mcapuccini:
Yep, it would be good if possible

yonromai:
That's a good point - I was thinking about the same thing actually. 
I wish there was some testing in place for the Kubernetes contrib module when I wrote this PR.
Such testing would most likely involve running minikube on CI, so that's probably gonna be some work - also reproducing runtime kubernetes problem might be hard to do in a local environment. 

I still think we should eventually do it but I'd prefer to not increase the scope of this PR to add some new testing capabilities.

Thought?

carmat88:
@yonromai They both sound fair and reasonable suggestions I'd say (i.e. potential unit-test and not increasing scope of this PR). Cheers.

mcapuccini:
There is some unit testing in place, however it is not run in Travis. Minikube uses virtualization, and it is hard within an AWS instance. Yep, let's leave it as future improvement.

luigi/contrib/kubernetes.py
```diff
@@ -181,27 +211,90 @@ def signal_complete(self):
...
+        if "succeeded" in job.obj["status"] and job.obj["status"]["succeeded"] > 0:
+            job.scale(replicas=0)
+            if self.print_pod_logs_on_exit:
+                self.__print_pod_logs()
+            if self.delete_on_success:
+                for pod in self.__get_pods():
+                    self.__logger.info("Deleting Pod " + pod.name)
+                    pod.delete()
+                self.__logger.info("Deleting Job " + job.name)
+                job.delete()
```
himikof: This cleanup loop seems racy/unreliable in case of API errors. The apiserver can handle cleanup of the dependent objects by itself, if told to do so (documented at https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/). The way to do it here would be to pass `{"kind":"DeleteOptions","apiVersion":"v1","propagationPolicy":"Foreground"}` to the job deletion request, like this:
```py
delete_options_cascade = {"kind": "DeleteOptions", "apiVersion": "v1", "propagationPolicy": "Foreground"}
r = self.__kube_api.delete(json=delete_options_cascade, **job.api_kwargs())
if r.status_code != 404:
    self.__kube_api.raise_for_status(r)
```

luigi/contrib/kubernetes.py
```diff
@@ -43,12 +44,12 @@
...
     from pykube.config import KubeConfig
     from pykube.http import HTTPClient
-    from pykube.objects import Job
+    from pykube.objects import Job, Pod
 except ImportError:
     logger.warning('pykube is not installed. KubernetesJobTask requires pykube.')
 
 
-class kubernetes(luigi.Config):
+class Kubernetes(luigi.Config):
```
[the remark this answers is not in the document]
yonromai: nice catch, that's right configs are lower case by convention - will fix

luigi/contrib/kubernetes.py
yonromai: Fair point. Why did you pick `"propagationPolicy":"Foreground"`? Was thinking of using background cascading deletion. Will fix.

luigi/contrib/kubernetes.py
```diff
@@ -63,19 +64,20 @@ class kubernetes(luigi.Config):
 class KubernetesJobTask(luigi.Task):
 
     __POLL_TIME = 5  # see __track_job
-    kubernetes_config = kubernetes()
+    kubernetes_config = Kubernetes()
```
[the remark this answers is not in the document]
yonromai: indeed - will fix

yonromai:
@himikof thanks for the review, PTAL at the last commit.
Looks like someone broke the build (unrelated to this changes) - will investigate.

yonromai:
@himikof PTAL (I fixed the master build and rebased this changes on master)

yonromai:
@Tarrasch did someone change the codecov settings: these checks seem new, right?

Tarrasch:
@yonromai, Feel free to ignore it. I've seen it before but I've never liked it as it's completely unreliable. We've probably misconfigured it, I think it needs tricky configuration because some of our coverage runs run in separate processes. Also see https://github.com/spotify/luigi/blob/master/codecov.yml

PR #2339, opened 2018-01-25
Fix unicode formatting.

pkaleta:
On Python 2.x, `scheduler.Task.pretty_id` methods fails with the following error when a parameter's value contains a unicode code point:

```
  File "/Users/piotr/counsyl/luigi/test/scheduler_test.py", line 34, in test_pretty_id_unicode
    task.pretty_id
  File "/Users/piotr/counsyl/luigi/luigi/scheduler.py", line 338, in pretty_id
    param_str = ', '.join('{}={}'.format(key, value) for key, value in sorted(self.params.items()))
  File "/Users/piotr/counsyl/luigi/luigi/scheduler.py", line 338, in <genexpr>
    param_str = ', '.join('{}={}'.format(key, value) for key, value in sorted(self.params.items()))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2192' in position 0: ordinal not in range(128)
```

## Description
Do not attempt to convert unicode to str, but return unicode object instead.

I have included unit tests. I've also ran my jobs with this code and it works for me.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -335,8 +335,8 @@ class Task(object):
 
     @property
     def pretty_id(self):
-        param_str = ', '.join('{}={}'.format(key, value) for key, value in sorted(self.params.items()))
-        return '{}({})'.format(self.family, param_str)
+        param_str = ', '.join(u'{}={}'.format(key, value) for key, value in sorted(self.params.items()))
+        return u'{}({})'.format(self.family, param_str)
 
 
 class Worker(object):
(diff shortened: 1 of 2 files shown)


pkaleta:
Thanks @dlstadther

pkaleta:
friendly ping on this one

pkaleta:
Thanks @dlstadther!

Tarrasch:
Thanks @pkaleta!

PR #2340, opened 2018-01-27
Fix status message button in worker tab when only progress is set.

riga:
## Description

Hi there,

it appears that the status message button is missing in the worker tab when there are no messages, but a task progress percentage is set. In the task list tab, the button is shown for that case, so I assume this is just a tiny bug.

The first commit in this PR should fix it. The second commit just adds some quotes, so feel to squash =)


Tarrasch:
It seems like the email that you're commiting from is not associated with your GitHub account. Can you please add your email to your GitHub account? By doing this you increase the traceability for this project. Thank you.

You can run `git show --format='%ae'` to see what email you commited with. Make sure it's present among your [emails registered at GitHub](https://github.com/settings/emails)

riga:
Sorry, looks like a messed up my git config.
Closed and reopened as #2344 for simplicity.

PR #2343, opened 2018-01-31
Add possibility to specify Redshift column compression.

jamesmcm:
## Description

I have added the possibility to specify Redshift column compression encodings in the coldefs tuple. At the moment we never specify them so if the table is created via Luigi, Redshift will estimate the best compression given the data imported in _only_ that first COPY. 

## Motivation and Context

Sometimes we know a priori what compression we want to use, and so this allows us to specify it in the coldefs directly with an additional element in the tuple.

Redshift Compression Encodings: https://docs.aws.amazon.com/redshift/latest/dg/c_Compression_encodings.html

We might also consider adding similar default value support in the future (i.e. to set some columns to have a default of 0 rather than NULL) - but in this case we would likely need to change the current tuple structure we use.

I have tested it on our development cluster and checked that the tables are created with the corresponding compression encodings.

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -299,6 +299,24 @@ class S3CopyToTable(rdbms.CopyToTable, _CredentialsMixin):
                 table_attributes=self.table_attributes)
 
             connection.cursor().execute(query)
+        elif len(self.columns[0]) == 3:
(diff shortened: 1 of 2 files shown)


dlstadther:
Sorry for the delay here! Merging!

dlstadther:
Thanks @jamesmcm !

PR #2344, opened 2018-01-31
Show status message button in worker tab when only progress is set.

riga:
## Description

(This is a clone of PR #2340 which lacked the committer email address.)

Hi there,

it appears that the status message button is missing in the worker tab when there are no messages, but a task progress percentage is set. In the task list tab, the button is shown for that case, so I assume this is just a tiny bug.

Here is a screenshot:

[missing button](https://dl.dropboxusercontent.com/s/ce0h2ob7oudtob7/missing_button.png?dl=0)

The message button is visible in the task list, but hidden in the worker list.

The single commit in this PR should fix it.


Tarrasch:
Thanks @riga! :)

Issue #2299, opened 2017-12-05
memory leak in luigi scheduler task history database code


ozzieba:
We are seeing memory issues very similar to #430 , with Luigi 2.6.2. In particular, over time memory usage grows to >6GB, and only goes back down after a restart.
Looking at the [code](https://github.com/spotify/luigi/blob/56fbbcb9be4c97a88c25fa527881f5bbade9cd41/luigi/db_task_history.py#L109) , it seems like self.tasks is storing all the tasks in history, and never gets cleaned. Could we add some kind of process to periodically clean old tasks? Alternatively, could we use some kind of priority queue that removes the oldest entry (past a certain size) on insertion?
@Tarrasch

ZhangXingyaoTHU:
Hi there, we found a similar problem about memory leaking. After executing about 1000 tasks, the memory usage of luigi will rise to about 50GB. Just like this issue, the only going back is a restart.
the current version we used is 2.6.1, will an update work?
@Tarrasch

PR #2345, opened 2018-02-02
Fix memory leak in db_task_history

ozzieba:
This fixes https://github.com/spotify/luigi/issues/2299, where memory usage grows linearly over time when using db_task_history. With this change memory usage essentially plateaus after a "retention" period.


Tarrasch:
I don't use the feature. Can anyone who uses db_task_history please review?

ozzieba:
Sorry, need to make a couple small changes to make this pass on python 3, will do soon

PR #2346, opened 2018-02-03
Allow release of resources during task running.

riga:
## Description

At scheduling time, the luigi scheduler needs to be aware of the maximum resource consumption a task might have once it runs. For some tasks, however, it can be beneficial to release or reduce a resource between two steps within their run method (e.g. after some heavy computation). In this case, a different task waiting for that particular resource can already be scheduled.

I simply added another method to the `TaskStatusReporter` which is forwarded to the task in `TaskProcess._run_get_new_deps`. The scheduler also got two new rpc methods `set_running_task_resources` and `get_running_task_resources`.

(Maybe the `TaskStatusReporter` should be renamed since it's doing more than just reporting the status now.)

Within a task, the resources can be updated via `set_running_resource()` which gives you complete access to the resource dict held by the scheduler for that task. Another approach could be something like a method `release_resource()` which is used to just *reduce* resource values. Right now, one can also increase resources which is certainly dangerous. Which way do you prefer?

## Motivation and Context

We have some long running tasks which control remote jobs on some computing grids. Although those grids have batch systems / queues, we want to use luigi's resource system to ensure we don't exceed a certain number (few k) of running jobs.

I also added test cases and some docs =)

**edit**: In the third commit, I propagated the running resources to the visualizer (didn't think about that when I opened the PR).


riga:
The nonhdfs tests fail, the others are fine. I guess it‘s because a remote scheduler is used in one of the test cases. Any advise?

Tarrasch:
> The nonhdfs tests fail, the others are fine. I guess it‘s because a remote scheduler is used in one of the test cases. Any advise?

I would simply disable the test case for the remote scheduler if it's to hard to get it working, as long as the in-memory scheduler test works. I think there's already a couple of tests like that.

> Which way do you prefer?

I think you're the best person to form this API, since you are the one using/needing it. Some things to consider:

 * You can still make the  luigi Task not do the invalid changes that the code says. For example if the start resource is 5, make changes to >5 no
 * What would make your code easier? To have `set_resources()`, `release_resources()` or `decrease_resources()`. I slightly lean towards the decrease version, so I could just specify the one resource I want to decrease and not respecify the others, like `decrease_resources({'my_resource': 3})`. But I let you pick. :)

luigi/scheduler.py
```diff
@@ -970,8 +970,9 @@ def _used_resources(self):
         used_resources = collections.defaultdict(int)
         if self._resources is not None:
             for task in self._state.get_active_tasks_by_status(RUNNING):
-                if getattr(task, 'resources_running', task.resources):
-                    for resource, amount in six.iteritems(getattr(task, 'resources_running', task.resources)):
+                resources_running = getattr(task, "resources_running", task.resources)
+                if resources_running:
+                    for resource, amount in six.iteritems(resources_running):
```
Tarrasch: Nice code improvement :)

test/task_running_resources_test.py
```diff
@@ -0,0 +1,118 @@
...
+    reduce_foo = luigi.BoolParameter()
+
+    def process_resources(self):
+        return {"foo": 2}
+
+    def run(self):
+        if self.reduce_foo:
+            self.set_running_resources({"foo": 1})
+
+        time.sleep(2)
```
Tarrasch: I suppose it was too hard to make it work without using `time.sleep` right?

daveFNbuck:
The API needs to be designed so that you can't increase any of the resources, as that can push you over the limit. We want to be able to guarantee that you never exceed a resource limit.

test/task_running_resources_test.py
riga: Yeah, this time also interferes with the workers `waint_interval`. I couldn't come up yet with a better approach to check that the scheduler really allows to run the two tasks in parallel...

riga:
I disabled the remote scheduler tests on travis. Some other tests fail, I'm not entirely sure if my changes cause the problems...

riga:
Could you maybe start the tests again? I'm curious whether this was just a glitch. Thanks! =)

dlstadther:
Travis build has been restarted

riga:
Mh, again the nonhdfs tests fail. In particular:

- contrib.docker_runner_test.TestDockerTask
- contrib.spark_test.PySparkTaskTest

I don't see how these tests are related to the changes in this PR :/

dlstadther:
@riga I think this is a Travis issue. We didn't merge anything with failing tests, but these two tests have been failing for all new PRs.

riga:
Hi! Is there any progress on this matter?

dlstadther:
@riga These tests were fixed in #2356 

Mind pulling those changes in here? Thanks!

riga:
[One test](https://travis-ci.org/spotify/luigi/jobs/348363902#L2149) still fails, but it is not clear to me what is actually causing this ...

test/task_running_resources_test.py
```diff
@@ -0,0 +1,120 @@
...
+        # as the total foo resource (3) is smaller than the requirement of two tasks (4),
+        # the scheduler is forced to run them serially which takes longer than 4 seconds
+        task_a = ResourceTestTask(param="a")
+        task_b = ResourceTestTask(param="b")
+
+        with self.assert_duration(min_duration=4):
+            luigi.build([task_a, task_b], self.factory, workers=2,
+                        scheduler_port=self.get_http_port())
+
+    @skipOnTravis("https://travis-ci.org/spotify/luigi/jobs/338398994")
```
Tarrasch: There was really no way to get these tests to run on Travis? I'm not sure having them helps at all then. This is a luigi core feature you are adding, such important functionalities should be properly tested even on Travis.

riga: Agreed. Maybe the travis errors I've seen at the beginning were also related to #2356. I'll check again.

riga:
I noticed that the same task resource dicts are used in multiple places, esp.:

- when setting `task.resources_running` to `task.resources` before a task starts running
- when setting the `resources_running` of batch tasks

As this PR introduces dynamic `resources_running`, the initial resource dict needs to be copied at those places → 036b813.

riga:
I fixed one of the two failing tests in this PR.

I wasn't aware that there were hardcoded attributes in `Task.no_unpicklable_properties`. Those attributes consist only of temporary task callbacks added in `TaskProcess`, right before the task's run method is called. As a new callback is added in this PR (`decrease_running_resources`), the spark task tests failed.

What do you think about making this dynamic (depending on a mapping defined on class level)? In addition, `Task.no_unpicklable_properties` can use this mapping to obtain the list of unpicklable props at runtime. I implemented this in dc2a74a and c319bf1.

I'm about the solve the remaining failure.

Tarrasch:
Awesome to see progress. I commented on the commits you linked too.

riga:
> Sounds good to me. If you're tired on being stuck with #2346, you can submit this as a separate PR.

I *think* I'm close to fixing the issue, just getting a tornado error

```
RuntimeError: Cannot share PollIOLoops across processes
```

This is closely related to what you guys experienced as well ;)
https://github.com/spotify/luigi/blob/master/test/server_test.py#L45-L53

test/task_running_resources_test.py
```diff
@@ -0,0 +1,118 @@
...
+from helpers import RunOnceTask
+
+import luigi
+import luigi.scheduler
+import luigi.worker
+import luigi.rpc
+import server_test
+
+
+luigi.notifications.DEBUG = True
```
Tarrasch: Is this copied from anywhere? Is it necessary? You just want to delete this line

test/task_running_resources_test.py
```diff
@@ -0,0 +1,118 @@
...
+        # the total "foo" resource (3) is sufficient to run both tasks in parallel shortly after
+        # the first task started, so the entire process should not exceed 4 seconds
+        task_c = ResourceTestTask(param="c", reduce_foo=True)
+        task_d = ResourceTestTask(param="d")
+
+        with self.assert_duration(max_duration=4):
+            luigi.build([task_c, task_d], self.factory, workers=2,
+                        scheduler_port=self.get_http_port())
+
+    def test_local_scheduler(self):
```
Tarrasch: btw, Maybe this test can be outside of  `server_test.ServerTestBase`?

test/task_running_resources_test.py
```diff
@@ -0,0 +1,118 @@
...
+            luigi.build([task_a, task_b], self.factory, workers=2,
+                        scheduler_port=self.get_http_port())
+
+    def test_remote_scheduler_parallel(self):
+        # parallel test
+        # run two tasks and the first one lowers its requirement on the "foo" resource, so that
+        # the total "foo" resource (3) is sufficient to run both tasks in parallel shortly after
+        # the first task started, so the entire process should not exceed 4 seconds
+        task_c = ResourceTestTask(param="c", reduce_foo=True)
+        task_d = ResourceTestTask(param="d")
```
Tarrasch: You shouldn't need to use new task names right? might as well call them "a" and "b"

test/task_running_resources_test.py
```diff
@@ -0,0 +1,118 @@
...
+
+        with self.assert_duration(max_duration=4):
+            luigi.build([task_c, task_d], self.factory, workers=2,
```
Tarrasch: I wonder if you get strange behavior because of the simplistic and unintuitive way `luigi.build` is implemented. I think it just runs those two guys serially (it won't attempt `task_d` before `task_c` is complete. But you might just want to test to build with a single WrapperTask that depends on both of those tasks.

riga: Yep, you're right, will be simplified in the next commit.

test/task_running_resources_test.py
riga: done

test/task_running_resources_test.py
riga: done

test/task_running_resources_test.py
riga: done

riga:
The test that demonstrates how two tasks with concurrent resources can be accelerated required remote scheduler. I adapted some of the `server_test._ServerTest` features to setup the server, remote scheduler, etc.

riga:
Anything else I can/should do? =)

test/task_running_resources_test.py
```diff
@@ -0,0 +1,138 @@
...
+
+            w.add(task)
+            w.run()
+
+            self.assertEqual(sch.get_running_task_resources(task.task_id)["resources"]["foo"], 1)
+
+
+class ConcurrentRunningResourcesTest(unittest.TestCase):
+
+    def get_app(self):
```
Tarrasch: Can you remove this in a follow-up? You don't need it anymore right?

Tarrasch:
Awesome! Well done!

test/task_running_resources_test.py
riga: Yup!

tiamot:
Where is the documentation on how this is used?
I am attempting to reduce resources for long running tasks to allow more tasks to run in parallel.
Based on the current documentation https://luigi.readthedocs.io/en/stable/luigi_patterns.html?highlight=decrease#decreasing-resources-of-running-tasks

I attempted to do the following with the expectation that after 10 seconds the first task would release the resource and allow the second task to run in parallel: 
```from time import sleep
import luigi


class Y(luigi.Task):
    wait = luigi.IntParameter(default=10)
    n = luigi.IntParameter()

    resources = ({'A': 1})

    def output(self):
        return luigi.LocalTarget('out{}.txt'.format(self.n))

    def run(self):
        sleep(self.wait)
        self.decrease_running_resources({'A': 0})
        sleep(self.wait)
        f = self.output().open('w')
        f.close()


class Z(luigi.WrapperTask):
    tasks = [Y(n=x) for x in range(20)]

    def requires(self):
        yield self.tasks
```
`$ luigid`
`$ luigi --module x Z --workers 20`

Tarrasch:
See discussion in #2576 which you created.

PR #2351, opened 2018-02-12
Add support for luigi email attachments

dlstadther:
## Description

Adds support for 0 or more attachments per email notification.

## Motivation and Context

Given Luigi already has notification code for emails, particularly SMTP, the addition of attachment support would reduce extraneous code for Luigi users who utilize email for report dispersal via Luigi.

Goal is to not affect current usage, but to add to a feature to Luigi notification support.

Not yet. Tests will be forthcoming.

### DIFF
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -28,6 +28,7 @@ In particular using the config `receiver` should set up Luigi so that it will se
 '''
 
 import logging
+import os
 import socket
 import sys
 import textwrap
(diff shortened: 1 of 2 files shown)


dlstadther:
During this patch investigation, I noticed the lack of apparent usage of `image_png` noted in the various email notifications. From what I can tell, this parameter existed in the original code from @erikbern and @freider [here](https://github.com/spotify/luigi/commit/20add9da9861870074deea29f899e37f55b45d00#diff-b438c8dc80f50172445c72006dc7c8f7). However, I wanted to confirm with someone else that it's kosher to remove before I go and do that.

Tarrasch:
I think it's safe to remove that. :)

dlstadther:
I'll see what I can do in the next week or so to improve tests. Otherwise, this is good to go.

dlstadther:
Had no luck on adding more tests around email attachment

PR #2353, opened 2018-02-13
Verbose worker error logging

nathantsoi:
## Description

Add some details to the type checking exception

## Motivation and Context

When debugging, I've found this helpful to quickly pinpoint the source of the exception

Yes


Tarrasch:
Thanks!!

Issue #2355, opened 2018-02-16
Master build broken


yonromai:
Hey, the master build has been broken for ~2 weeks. Looks like a docker issue:
`ContainerError: Command 'test -d  /tmp/luigi-test' in image 'busybox:latest' returned non-zero exit status` 
from: https://travis-ci.org/spotify/luigi/jobs/336516969
@ulzha wanna take a look? If you don't have time, no big deal - I can take a look at it sometimes next week.

Tarrasch:
Easiest is to just disable it (send a PR) similar to how many other tests are disabled.. Sorry for the inconvenience!

yonromai:
fixed by: https://github.com/spotify/luigi/pull/2356

Tarrasch:
Thank you!

PR #2356, opened 2018-02-19
fix contrib.docker_runner exit code check #2355

yonromai:
Prior to `3.0.0`, `docker-py` used to return the status code as an `int` - it now returns a `dict` (cf. [2.7.0 `Container.wait` doc](http://docker-py.readthedocs.io/en/2.7.0/containers.html#docker.models.containers.Container.wait) vs [3.0.1 `Container.wait` doc](http://docker-py.readthedocs.io/en/3.0.1/containers.html#docker.models.containers.Container.wait))

The fix is to check if the return type is a dict, to attempt to maintain backward compatibility.


ravwojdyla:
+1

Issue #1936, opened 2016-11-28
Add LSF Support


boombard:
Make a production-ready version of https://github.com/dattalab/luigi/blob/lsf/luigi/lsf.py based on the existing SGE implementation. I will make a pull request

dlstadther:
@boombard Status update on PR?

ypriverol:
Any progress on this. Would be great to have support for LSF clusters in Luigi.

markmcdowall:
Hi @Tarrasch,

I have been in contact with @alexbw about the Luigi LSF update. We now have a working update on the LSF branch of a fork.

Would you like the changes merged to the master branch of the fork and then for a pull request to be made to Luigi or would you like it to remain on the branch?

Cheers,

Mark

dlstadther:
@markmcdowall Feature branch PR will be fine!

PR #2373, opened 2018-03-13
Additions to provide support for the Load Sharing Facility (LSF) job scheduler

markmcdowall:
## Description
Provides support for the use of the Load Sharing Facility (LSF) job scheduler for the distributed computation of tasks within a high performance cluster environment. The work is based off the SGE wrapper and work that had been done by @alexbw.

## Motivation and Context
Needed this so that we could use the Luigi infrastructure on our local LSF cluster

https://github.com/spotify/luigi/issues/1936

I have run this code as part of a local pipeline and everything ran smoothly. There is also a small unittest script

## Notes
Updates were based on the work by Alex with changes so that jobs could run within the latest version of Luigi.
Pulled in the latest changes from master
Added new parameters for naming of jobs, listing queues and memory sizes.
Added in a print out to provide a rough timing for how long a bsub job ran for. This is dependent on the polling time.
The LSF wrapper has been moved into the contrib folder
Some code tidying based on pylint output
Added doc sections and copied over the Apache 2.0 license from Spotify. You'll need to add your institute into the list, I have left a gap for this already


dlstadther:
@markmcdowall Please address travis errors and add tests were applicable, if possible.

markmcdowall:
Hi @dlstadther,

I am still getting a failure with the docs not building correctly. Would you be able to advise on why it might be failing?

Cheers,

Mark

dlstadther:
`/home/travis/build/spotify/luigi/luigi/contrib/lsf.py:docstring of luigi.contrib.lsf.LSFJobTask.run:5: ERROR: Unexpected indentation.`

You also have some other Py3 errors

Tarrasch:
@markmcdowall, Can you please update the PR title and description to be more useful for casual reaters? What's LSF? Who should care etc. Thanks. :)

markmcdowall:
Hi @dlstadther 

There is now a green light from the builds with the docs building correctly.

Cheers,

Mark

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+    memory_flag = luigi.Parameter(default='8192', significant=False)
+    queue_flag = luigi.Parameter(default='queue_name', significant=False)
+    runtime_flag = luigi.IntParameter(default=60)
+    job_name_flag = luigi.Parameter(default='')
+    poll_time = luigi.FloatParameter(
+        significant=False, default=5,
+        description="specify the wait time to poll bjobs for the job status")
+    save_job_info = luigi.BoolParameter(default=False)
+    output = luigi.Parameter(default="")
+    extra_bsub_args = luigi.Parameter("")
```
dlstadther: I think you meant for `extra_bsub_args` to have a default empty string

Have you considered using `OptionalParameter`s with defaults of None?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+        self.tmp_dir = os.path.join(base_tmp_dir, task_name)
+        # Max filename length
+        max_filename_length = os.fstatvfs(0).f_namemax
+        self.tmp_dir = self.tmp_dir[:max_filename_length]
+
+        LOGGER.info("Tmp dir: %s", self.tmp_dir)
+        os.makedirs(self.tmp_dir)
+
+        # Dump the code to be run into a pickle file
+        logging.debug("Dumping pickled class")
```
dlstadther: Why have we switched from using `LOGGER` to using `logging`?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+        self._dump(self.tmp_dir)
+
+        # Make sure that all the class's dependencies are tarred and available
+        logging.debug("Tarballing dependencies")
```
dlstadther: Here too

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+        # Grab luigi and the module containing the code to be run
+        packages = [luigi] + [__import__(self.__module__, None, None, 'dummy')]
```
dlstadther: Why are we adding two lists here?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+
+    def _run_job(self):
+        """
+        Build a bsub argument that will run lsf_runner.py on the directory we've specified.
+        """
+
+        args = []
+
+        if isinstance(self.output(), list):
+            log_output = self.output()[0].path.split("/")
```
dlstadther: Why would `output()` be a list?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+            args += ["-J", str(self.job_name_flag)]
+        args += ["-o", "/".join(log_output[0:-1]) + "/job.out"]
+        args += ["-e", "/".join(log_output[0:-1]) + "/job.err"]
+        args += self.extra_bsub_args.split()
+
+        # Find where our file is
+        runner_path = lsf_runner.__file__
+        # assume source is next to compiled
+        if runner_path.endswith("pyc"):
+            runner_path = runner_path[:-3] + "py"
```
dlstadther: Why are you renaming pyc to py?

If the intention is to rename `.pyc` to `.py`, why don't you just remove the `c` or something?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+
+        args += [runner_path]
+        args += [self.tmp_dir]
+
+        # That should do it. Let the world know what we're doing.
+        LOGGER.info(" ".join([str(a) for a in args]))
```
dlstadther: Do you want to provide some context to this log record? i.e. `The following lsf args have been specified: {}`

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+            [str(a) for a in args],
+            stdin=subprocess.PIPE, stdout=subprocess.PIPE, cwd=self.tmp_dir)
+        output = run_job_proc.communicate()[0]
+
+        # ASSUMPTION
+        # The result will be of the format
+        # Job <123> is submitted ot queue <myqueue>
+        # So get the number in those first brackets.
+        # I cannot think of a better workaround that leaves logic on the Task side of things.
+        LOGGER.info("### JOB SUBMISSION OUTPUT:" + str(output))
```
dlstadther: Since you're going string concatenation, your output will not include a space between `:` and your output var.

General critique here... You using both string concatenation and `%` interpolation throughout. Pick one (preferably `%` or `.format` (I'm personally partial to format)) and stick with it please.

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+        self.job_id = int(output.split("<")[1].split(">")[0])
+        LOGGER.info(
+            "Job %ssubmitted as job %s",
```
dlstadther: Do you mean to not have a space between job_id and submitted?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+                LOGGER.info("Job is running...")
+                if time0 == 0:
+                    time0 = int(round(time.time()))
+            elif lsf_status == "PEND":
+                self.job_status = PENDING
+                LOGGER.info("Job is pending...")
+            elif lsf_status == "DONE" or lsf_status == "EXIT":
+                # Then the job could either be failed or done.
+                errors = self.fetch_task_failures()
+                if errors == '':
```
dlstadther: Alternatively, `if not errors:`. Especially if `errors` could be `None`

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+                    self.job_status = DONE
+                    LOGGER.info("Job is done")
+                    time1 = int(round(time.time()))
+
+                    # Return a near estimate of the run time to with +/- the
+                    # self.poll_time
+                    LOGGER.info(
+                        "### JOB COMPLETED: %s%s in %s seconds",
+                        self.job_name_flag + ' ',
```
dlstadther: Why would you concat a space to a interpolated var instead of just including the space in the string to be interpolated?

luigi/contrib/lsf_runner.py
```diff
@@ -0,0 +1,85 @@
...
+try:
+    import dill as pickle
+except ImportError:
+    import pickle
+import logging
+import tarfile
+
+
+def do_work_on_compute_node(work_dir):
+    print("LSF RUNNER: LOAD THE DATA AND DO THE THING")
```
dlstadther: Why are we printing?

luigi/contrib/lsf_runner.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Extract the necessary dependencies
+    extract_packages_archive(work_dir)
+
+    # Open up the pickle file with the work to be done
+    os.chdir(work_dir)
+    with open("job-instance.pickle", "r") as f:
+        job = pickle.load(f)
+
+    # job = pickle.loads("job-instance.pickle")
```
dlstadther: Is this supposed to be commented out?

luigi/contrib/lsf_runner.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def main(args=sys.argv):
+    """Run the work() method from the class instance in the file "job-instance.pickle".
+    """
+    try:
+        # Set up logging.
+        logging.basicConfig(level=logging.WARN)
+        work_dir = args[1]
+        print("PATHS:" + " | ".join(args))
```
dlstadther: Why printing?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,362 @@
...
+
+import os
+import subprocess
+import time
+import sys
+import logging
+import random
+import shutil
+try:
+    import dill as pickle
```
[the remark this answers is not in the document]
markmcdowall: I implemented using the dill function as the head node for submitting to the LSF cluster is not identical to the nodes in the cluster. This was causing issues with the pickling/unpickling of the objects. Dill was more robust to this scenario.

I'll add this as a comment in the code.

luigi/contrib/lsf.py
markmcdowall: This was to allow for the return of a single output file location. Or if there are multiple output files then provide those as a list of string locations.

I can either comment this in the code, or if there is a better way or more conformant to way to perform this action that matches code used else where in Luigi that is fine.

luigi/contrib/lsf.py
markmcdowall: Looks like this might have been dev code to ensure that the latest version of the py script was called. Python should check that the pyc timestamp is newer than the py timestamp anyway so I have removed this from the code. This could always be reintroduced if there is a problem along with a debug flag.

luigi/contrib/lsf.py
markmcdowall: This is so that if a job_name_flag is not submitted there is not a double space

luigi/contrib/lsf.py
markmcdowall: By mistake when I was tidying up. Will correct this on the one on line 172.

luigi/contrib/lsf.py
markmcdowall: Will merge into a list of 2 elements

luigi/contrib/lsf.py
markmcdowall: Sorry, but I went for the `%` version as the majority of instances already used that form. Happy to change if it matches in with the rest of the code base though.

luigi/contrib/lsf.py
markmcdowall: Moved this logic out to pre compose the required string. Adding the space was to avoid having a double space if there was no name specified.

markmcdowall:
Hi @dlstadther,

I have addressed, or commented on, each of he issues that you raised and Travis is giving the green light.

Cheers,

Mark

markmcdowall:
Hi @dlstadther ,

Thanks for merging. As for the second reviewer would this need to be someone else that has already contributed to the Luigi repo? If so is there a list of people that are happy to be approached.

dlstadther:
@lairdm, could you oblige @markmcdowall 's review request? Thanks!

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,354 @@
...
+    """
+    Tracking is done by requesting each job and then searching for whether the job
+    has one of the following states:
+    - "RUN",
+    - "PEND",
+    - "SSUSP",
+    - "EXIT"
+    based on the LSF documentation
+    """
+    cmd = "bjobs -noheader -o stat %d" % job_id
```
lairdm: Should you use new style .format()? Old style hasn't been deprecated yet, but the threat is still hanging out there.

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,354 @@
...
+    track_job_proc = subprocess.Popen(
```
lairdm: Do you need to put this in a try block, or is it alright if things die loudly if the sub-process can't be made?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,354 @@
...
+        cmd, stdout=subprocess.PIPE, shell=True)
+    status = track_job_proc.communicate()[0].strip('\n')
+    return status
+
+
+def kill_job(job_id):
+    """
+    Kill a running LSF job
+    """
+    subprocess.call(['bkill', job_id])
```
lairdm: Does Luigi support returning if a kill was successful or not?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,354 @@
...
+        else:
+            errors = ''
+        return errors
+
+    def fetch_task_output(self):
+        """
+        Read in the output file
+        """
+        # Read in the output file
+        with open(os.path.join(self.tmp_dir, "job.out"), "r") as f_out:
```
lairdm: Nitpicky, but with errors you check if the file exists, is there a reason the same logic isn't mirrored here?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,354 @@
...
+        random_id = '%016x' % random.getrandbits(64)
+        task_name = random_id + self.task_id
+        # If any parameters are directories, if we don't
+        # replace the separators on *nix, it'll create a weird nested directory
+        task_name = task_name.replace("/", "::")
+
+        self.tmp_dir = os.path.join(base_tmp_dir, task_name)
+        # Max filename length
+        max_filename_length = os.fstatvfs(0).f_namemax
+        self.tmp_dir = self.tmp_dir[:max_filename_length]
```
lairdm: I'm not sure I follow what you're trying to do. I'm unsure if f_namemax is the max length of a filename component or the entire path. Either way, you're truncating the base_tmp_dir + task_name concatenated path to f__namemax then adding packages.tar a bit later. If f_namemax is the max total path length, wouldn't making it longer be an issue? If it's the max length of a component of the path, shouldn't it be truncated before prepending the base path?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,354 @@
...
+            log_output = self.output().path.split("/")
+
+        args += ["bsub", "-q", self.queue_flag]
+        args += ["-n", str(self.n_cpu_flag)]
+        args += ["-M", str(self.memory_flag)]
+        args += ["-R", "rusage[%s]" % self.resource_flag]
+        args += ["-W", str(self.runtime_flag)]
+        if self.job_name_flag:
+            args += ["-J", str(self.job_name_flag)]
+        args += ["-o", "/".join(log_output[0:-1]) + "/job.out"]
```
lairdm: If you want to drop the last component of the path, why this way rather than rsplit as with the path earlier, then us the OS safe os.path module?

luigi/contrib/lsf.py
```diff
@@ -0,0 +1,354 @@
...
+
+            # dest_dir = self.output().path
+            shutil.move(self.tmp_dir, "/".join(log_output[0:-1]))
+
+        # Now delete the temporaries, if they're there.
+        self._finish()
+
+    def _track_job(self):
+        time0 = 0
+        while True:
```
lairdm: Is it ok if we get stuck in this call forever if the job never releases or is suspended?

luigi/contrib/lsf.py
markmcdowall: Not really as blank will get returned and handled later in the script

luigi/contrib/lsf.py
markmcdowall: The only response would be from LSF saying that it is terminating the job with the given id. There is nothing from LSF to say if it was successful or not.

luigi/contrib/lsf.py
markmcdowall: This should only happen if jobs are hanging while in the queue to get onto LSF. If a job has been held back this could be due to the LSF manager prioritising other jobs over yours or you could have held your jobs from completing to clear a backlog of other jobs that are waiting.

markmcdowall:
Hi @dlstadther ,

Sorry for the delay. I have gone through the comments made by @lairdm about the code and made relevant changes and also run a few pipelines to check that the changes did not break anything.

Are you happy with the changes that have been made to the code?

Cheers,

Mark

luigi/contrib/lsf.py
```diff
@@ -236,8 +235,8 @@ def _run_job(self):
         args += ["-W", str(self.runtime_flag)]
         if self.job_name_flag:
             args += ["-J", str(self.job_name_flag)]
-        args += ["-o", os.path.join(log_output[0:-1], "job.out")]
-        args += ["-e", os.path.join(log_output[0:-1], "job.err")]
+        args += ["-o", os.path.join("/".join(log_output[0:-1]), "job.out")]
```
lairdm: Sorry, why are you flipping back to "/".join()? That's not OS safe. Unless I'm missing something.

lairdm:
Other than the "/".join() question, I'm happy. Thanks.

luigi/contrib/lsf.py
markmcdowall: I've modified the code to use the os.path.split and os.path.join functions.

The reason for using the "/".join() is that the os.path.join could only take in strings. Using the os.path.split should get around this issue.

markmcdowall:
Hi @dlstadther ,

Thank you for accepting the changes. Do I need to provide any further information or is it now over to you guys?

Cheers,

Mark

PR #2374, opened 2018-03-13
Version 2.7.3

honnix:
## Description
Prepare for the next release.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ if os.environ.get('READTHEDOCS', None) == 'True':
 
 setup(
     name='luigi',
-    version='2.7.2',
+    version='2.7.3',
     description='Workflow mgmgt + task scheduling + dependency resolution',
     long_description=long_description,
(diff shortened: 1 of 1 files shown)


dlstadther:
Is there a new release schedule? Or event which causes you guys to initiate a release?

honnix:
2.7.3 has been uploaded to pypi, and there is a draft being prepared right now. You should be able to read it I think. https://github.com/spotify/luigi/releases/tag/untagged-7b7143d52c5be8f8af21

Tarrasch:
> Is there a new release schedule? Or event which causes you guys to initiate a release?

I'm curious too. But even if you just do it sporadically I think it's really great. Thanks!

honnix:
Technically our squad still own Luigi although almost no contribution. We are back to the business of occasional release which largely depends on internal ask. :)

We will for sure be careful with major release and will make sure the main contributors and maintainers are involved in that case.

PR #2375, opened 2018-03-13
Add parameterized backoff limit.

andrewsmartin:

### DIFF
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -159,6 +159,13 @@ class KubernetesJobTask(luigi.Task):
         return self.kubernetes_config.max_retrials
 
     @property
+    def backoff_limit(self):
+        """
+        Maximum number of retries before considering the job as failed.
+        """
+        return 6
+
+    @property
     def delete_on_success(self):
         """
         Delete the Kubernetes workload if the job has ended successfully.
@@ -322,6 +329,7 @@ class KubernetesJobTask(luigi.Task):
             },
             "spec": {
                 "activeDeadlineSeconds": self.active_deadline_seconds,
+                "backoffLimit": self.backoff_limit,
                 "template": {
                     "metadata": {
                         "name": self.uu_name
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -162,6 +162,7 @@ class KubernetesJobTask(luigi.Task):
     def backoff_limit(self):
         """
         Maximum number of retries before considering the job as failed.
+        See: https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#pod-backoff-failure-policy
         """
         return 6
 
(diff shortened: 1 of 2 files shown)


andrewsmartin:
@mcapuccini PTAL

luigi/contrib/kubernetes.py
```diff
@@ -158,6 +158,13 @@ def max_retrials(self):
         """
         return self.kubernetes_config.max_retrials
 
+    @property
+    def backoff_limit(self):
+        """
+        Maximum number of retries before considering the job as failed.
+        """
+        return 6
```
mcapuccini: Is there any particular reason to have 6? 🙂

andrewsmartin: 6 is the default when not specified, so it seems like a sensible default here as well.

flylo: feel like we should default to one here

andrewsmartin: @flylo hm I guess we could, but I thought it would be best to stay consistent with the default behavior here, and we can override this in our internal wrapper to be 1.

flylo: sounds reasonable!

mcapuccini:
LGTM!

luigi/contrib/kubernetes.py
honnix: Maybe add a link to k8s doc where it says `6` as default value?

honnix:
It would be better we can make `codecov` happy.

andrewsmartin:
@mcapuccini any advise on fixing this coverage failure? Not sure I'm totally understanding the cause, and I'm not sure what could be added to the test that would address it.

andrewsmartin:
cc @Tarrasch ^

Tarrasch:
@andrewsmartin It's about test coverage. Is it possible to add a test case testing the code you've added?

andrewsmartin:
@Tarrasch the coverage failures seem pretty flaky, for example changes in the coverage of luigi.contrib.gcs, unrelated to this PR

honnix:
Not quite sure that test file is actually running properly. It says requireing `minikube`.

Tarrasch:
@honnix is right. Please ignore codecov this time. Next time you open a PR (or rebase this one I think) you should not run into problems anymore since I commited this https://github.com/spotify/luigi/commit/14eea1bc126aafd36ebd151b39c4eda8966e7087

mcapuccini:
It's not going to be easy to run minikube on travis, since minikube needs virtualization and I am not sure it can be run in the VMs that CI leaverages.

mcapuccini:
The tests are skipped as a matter of fact in CI.

andrewsmartin:
Thanks @honnix and @Tarrasch - i rebased to include that commit.

honnix:
LGTM
I don't think testing against minikube is a good idea anyway as part of CI. :)

PR #2376, opened 2018-03-13
Add method to push files using multiprocessing in GCS

tmattio:
This PR adds multiprocessing capabilities to the GCS module. The implementation can probably be improved, but I wanted to start the discussion on these questions:

- Should this be implemented on the library side (IMO yes)
- What would be your idea of a design for implementing this?

## Description

This adds a method called `put_multiple` that allows the user of the module to put multiple files at a time. The filenames on the bucket will be the base names of the provided filepaths.
The method provides a parameter `num_process`. If the parameter value is more than 1, the upload will be executed in multiprocessing.

## Motivation and Context

The upload to GCS is really slow. The API backend does not support the upload of several MediaFile, but standard tools such as `gsutil` use multiprocessing to speed up the upload. This PR merely tries to mimic the behavior of other standard tools.

Here is a benchmark for the upload of 10 files of 1MB (using 10 processes):
```
mutliproc: 4163.86 ms
single proc: 23692.74 ms
```

I added a sample unit test for the method.


dlstadther:
Please fix flake8 tests :)

tmattio:
@dlstadther I fixed the tests and a previous commit from @Tarrasch fixes the code coverage

luigi/contrib/gcs.py
```diff
@@ -249,6 +249,28 @@ def put(self, filename, dest_path, mimetype=None, chunksize=None):
...
+
+    def put_multiple(self, filenames, directory_path, mimetype=None, chunksize=None, num_process=1):
+        args = [
+            {
+                'filename': filename,
+                'dest_path': os.path.join(directory_path, os.path.basename(filename)),
+                'mimetype': mimetype,
+                'chunksize': chunksize,
+            }
+            for filename in filenames
```
dlstadther: There is no guard against submitting `filenames` as a string. The iterable won't fail until L245

luigi/contrib/gcs.py
```diff
@@ -249,6 +249,28 @@ def put(self, filename, dest_path, mimetype=None, chunksize=None):
 
         self._do_put(media, dest_path)
 
+    def _put_star(self, kwargs):
```
dlstadther: I'm not sure this hidden method name makes intuitive sense. Why not just call it `_put(self, kwargs)`?

tmattio: The only purpose is to expand the kwargs because the multiprocessing `map` method does not allow to pass multiple parameters. 

I would think the `_put(self, kwargs)` method is the hidden implementation of the `put` method, which it's not. A more explicit name name could be `_forward_args_to_put` maybe?

tmattio:
@dlstadther Is there any change that you would like me to make to merge this?
I've been using this in production for a couple of months now and everything is working properly.

dlstadther:
@tmattio I restarted the tests and green build all the way!

dlstadther:
Thanks for the contribution @tmattio !

PR #2378, opened 2018-03-18
Add option for tasks to run on main process

naor2013:
## Description

When using multiple workers, the main process is used for the scheduler and the tasks run on the rest of the cores. If you're using a function that utilizes multiple cores, running that function not from the main core can cause some issues. So in some cases, you want your workflow to run on multiple cores except from one task that you want to run on the main core. I added an open to add a property to your task "run_on_main_process" which you can set to true (default is false) and using that will force that task to run on the main process.

## Motivation and Context

Using functions from libraries that use multiple cores (like Scikit-Learn) will have unexpected outcomes when running from a core that's not the main core. Adding that property will solve that issue.

I tried it and it worked. Plus, I only changed 1 line of core code with that so I'm not sure how to add unit tests for it (Would love for some help with that if needed).

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -284,6 +284,10 @@ class Task(object):
         # TODO(erikbern): we should think about a language-agnostic mechanism
         return self.__class__.__module__
 
+    @property
+    def run_on_main_process(self):
+        return False
+
     _visible_in_registry = True  # TODO: Consider using in luigi.util as well
 
     __not_user_specified = '__not_user_specified'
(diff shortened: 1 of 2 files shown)


luigi/task.py
```diff
@@ -284,6 +284,10 @@ def task_module(self):
         # TODO(erikbern): we should think about a language-agnostic mechanism
         return self.__class__.__module__
 
+    @property
+    def run_on_main_process(self):
```
Tarrasch: Please add docs here. And also warn users that this mode should be avoided whenever possible, because any blocking IO will make the keep-alive-thread not run.

Issue #2259, opened 2017-10-16
Adjusting the worker rank function to use last modified time rather than submitted time


nugend:
Currently it accounts for the assigned rank and the submitted time. But in large graphs, this can lead to the most recently added tasks *only* executing (since their retry timeout expires before the other tasks are attempted). We're trying this out locally to determine if it works for us.

Does anyone foresee any unanticipated side-effects?

nugend:
We've been using this for a while and it seems to resolve the issue we were seeing with unfair scheduling of tasks. I would strongly suggest said change being put into mainline.

Tarrasch:
I'm happy to merge such optimization, please share more insights of how things are better in the PR description. :)

nugend:
@Tarrasch Have you had a chance to look at the PR?

PR #2379, opened 2018-03-22
use task's last updated time for scheduler rank

nugend:
## Description

In very large task graphs with task retry enabled (awaiting updates), the tasks which are scheduled first may never execute because by the time the newer tasks inserted into the graphs expire their re-attempt timeouts, the scheduler will decide to run them again. This results in the earlier scheduled tasks not executing until the later scheduled tasks have completed.

## Motivation and Context
It makes running very large task graphs with task reattempts enabled work as expected. It doesn't seem to affect scheduling otherwise (in a task graph without re-attempts, the tasks which are updated won't be considered in the first place)

See #2259

We have been running with this change in production for six months and it has been working as expected. If this breaks or otherwise justifies a test, I haven't tested it though.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -982,7 +982,7 @@ class Scheduler(object):
         :return:
         """
 
-        return task.priority, -task.time
+        return task.priority, -task.updated
 
     def _schedulable(self, task):
         if task.status != PENDING:


Tarrasch:
I'm fine with this I think. But can you check and understand why so many tests fail? Are the tests overasserting or is something essential breaking with this change?

nugend:
Is there something significant about it only failing under the "nonhdfs" build profiles?

I believe the `test_resend_batch_on_get_work_retry` is a natural consequence of the proposed change resulting in the work ordering constraints being loosened. So if you agree, I would run `sorted` on the array contents.

WRT `test_tasks_parallel` and `test_lock_resources_for_second_worker` I'm not sure how these tests are being affected by the proposed change since it doesn't seem like the priority system should interact with resource constraints. Does work *have* to be returned in the same order for resource constraints to function?

Tarrasch:
> Is there something significant about it only failing under the "nonhdfs" build profiles?

Well, "nonhdfs" was a poorly choosen name by me. But it's the "main" or "core" tests, including the scheduler tests.

> I believe the test_resend_batch_on_get_work_retry is a natural consequence of the proposed change resulting in the work ordering constraints being loosened. So if you agree, I would run sorted on the array contents.

Sure. Actually you probably want [`assertCountEqual`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual).

> WRT test_tasks_parallel and test_lock_resources_for_second_worker I'm not sure how these tests are being affected by the proposed change since it doesn't seem like the priority system should interact with resource constraints. Does work have to be returned in the same order for resource constraints to function?

Can you investigate this further on your own? `test_lock_resources_for_second_worker` was very recently submitted by @riga in #2346 tough.

PR #2382, opened 2018-03-29
WIP: Add support for running MRJob jobs as luigi Tasks (rebase)

bnewbold:
This is a rebase of @Jyrsa's PR #1731, which never got merged. The only changes were to accommodate the move from oauth2 to the `google.auth` package. I did not re-write any git commit metadata (eg, change titles).

I tried running `tox` to test, but it looks like there are a whole bunch of test environment needs, so i'm submitting a PR early to have this run on travis.

# Original PR Description

I added a luigi.contrib.mrjob package to support running MRJob steps (particularly on EMR) as part of Luigi workflows.

It's based on waluigi by @danielhfrank with a couple of bells and whistles added that I thought would make sense when testing workflows with small loads on local machines so that moving to EMR or Hadoop would be as effortless as possible.

### Description

A class MRJobTask that parses it's input into the correct format for a MRJob JobRunner and initializes and runs the runner. Support for adding configuration parameters to the MRJob, either everything or an external config file.

The user is assumed to have some previous experience with MRJob.

Also a class LocalMRJobTarget for testing workflows using LocalRunner on local machine and a LocalMRJobTask to boot that I believe follows the naming convention.

### Motivation and Context

I liked the idea of luigi and wanted to use it an envisioned automated data pathway that consisted of both local and EMR steps.

Also,
#294

### Have you tested this? If so, how?

I have included tests that are more integration than unit tests for the basic functionality.

I'll admit that there are a few cryptic functions that could use more testing.

### Notes

I'm not sure about the copyright notices. I put in my company's name as I did this on their time (and got permission to publish). If you want to change that to Spotify like everywhere else let me know.

Also this is my first meaningful pull request to an open source project. Be gentle :)

### DIFF
--- a/luigi/contrib/dataproc.py
+++ b/luigi/contrib/dataproc.py
@@ -15,9 +15,14 @@ try:
     from googleapiclient import discovery
     from googleapiclient.errors import HttpError
 
-    DEFAULT_CREDENTIALS, _ = google.auth.default()
-    authenticate_kwargs = gcp.get_authenticate_kwargs(DEFAULT_CREDENTIALS)
-    _dataproc_client = discovery.build('dataproc', 'v1', cache_discovery=False, **authenticate_kwargs)
+    try:
+        DEFAULT_CREDENTIALS, _ = google.auth.default()
+        authenticate_kwargs = gcp.get_authenticate_kwargs(DEFAULT_CREDENTIALS)
+        _dataproc_client = discovery.build('dataproc', 'v1', cache_discovery=False, **authenticate_kwargs)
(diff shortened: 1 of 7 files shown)


bnewbold:
Tests seem to be failing now because `boto3` is installed by default (mrjob depends on it), which results in a bunch of `NoRegionError` failures in all test targets. These tests normally don't run usually (eg, "SKIP: boto3 is not installed. BatchTasks require boto3"), and presumably depend on local configuration to run.

bnewbold:
This wasn't as easy as I hoped, closing.

PR #2384, opened 2018-04-04
Add google-auth-httplib2 as dependency

honnix:
## Description

Adding a new dependency.
## Motivation and Context

As discussed here https://github.com/spotify/luigi/pull/2361#issuecomment-378661375

This is used by https://github.com/google/google-api-python-client/blob/v1.6.6/googleapiclient/discovery.py#L371

Another option is to fix `luigi.contrib.bigquery` somehow.
 

Manually installing this package worked for me.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,6 +24,7 @@ deps=
   py27-gcloud: avro
   py33-gcloud,py34-gcloud,py35-gcloud,py36-gcloud: avro-python3
   gcloud: google-auth==1.4.1
+  gcloud: google-auth-httplib2==0.0.3
   google-compute-engine
   coverage>=4.1,<4.2
   codecov>=1.4.0


ulzha:
tox.ini is just adding it as a dependency for tests; didn't you mean to add it as a dependency for the real thing?

It's a bit unfortunate that contrib modules aren't separate packages with their own `setup.py`s; that would be the place to declare dependencies...

honnix:
I did mean to put it as real thing. Sorry for the mistake. Another PR coming.

honnix:
Now the problem is, I don't know where to put it. `setup.py` doesn't seem to be a good place. @ulzha as you mentioned, how do we handle contrib deps?

honnix:
That looks like a big change in code structure. Not so sure we want to spend that much effort atm.

PR #2390, opened 2018-04-11
having a PR for bumping version seems useful

honnix:
## Description

Release process documentation change.

## Motivation and Context

It would be easier to understand what PRs have been merged since last release.

N/A


honnix:
Never mind, I think I misinterpreted how Github PR list works. And it seems we missed some PRs in the latest release note. I will fix.

Tarrasch:
When I look for PR's that's been going in I usually just look at the commit list and let GitHub figure out which PR they are associated with.

PR #2391, opened 2018-04-11
append is vulnerable of xss

honnix:
## Description

Fix cross-site-scripting issue.

## Motivation and Context

http://api.jquery.com/append/ points out that `append` can introduce
cross-site-scripting (XSS) vulnerabilities.

In this particular case, we have identified an XSS issue as following:

Spin up a Luigi instance and navigate to /static/visualiser/index.html#tab=graph&visType=svg&taskId=%3Cscript%3Ealert(1)%3C%2Fscript%3E.
You should see an alert box pop up displaying 1.

Use `text` instead of `append` should fix it.

I tested it locally manually.

### DIFF
--- a/luigi/static/visualiser/js/visualiserApp.js
+++ b/luigi/static/visualiser/js/visualiserApp.js
@@ -382,7 +382,7 @@ function visualiserApp(luigi) {
                 }
             } else {
                 $("#searchError").addClass("alert alert-error");
-                $("#searchError").append("Couldn't find task " + taskId);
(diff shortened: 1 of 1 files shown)


Tarrasch:
Great job!

honnix:
As this is a security patch, it would better to get it out asap and somehow notify users.

Tarrasch:
@honnix, you could make a new release and send an email to the user group?

honnix:
Yes we will do that.

PR #2396, opened 2018-04-12
Revert "Add google-auth-httplib2 as dependency"

honnix:
Reverts spotify/luigi#2384

This is not intended because it is supposed to be a installation dependency.


honnix:
I guess it still makes sense to be part of test deps.

Issue #2397, opened 2018-04-12
Specify encoding of the configuration file


mpoussevin:
Hello,

I use Luigi on Windows with UTF8 configuration files.
It seems impossible to set the encoding to use. Luigi uses default CP1252 to open the configuration and incorrectly parses some characters. 

I will also open a PR with a proposed solution and update in the documentation.
Thank you for this great project :smile: 

Mickaël.

PR #2398, opened 2018-04-12
Environment variable to set the encoding of the configuration file.

mpoussevin:
Possible solution for the issue I opened : #2397 

## Description

I added a `_config_encoding`  field in `LuigiConfigParser` (in configuration.py).
I check the LUIGI_CONFIG_ENCODING environment variable to see if a value is given. 
If a value is given, I check in `codecs` whether the encoding exists or not and set the field to the encoding or None when the encoding is unknown (a warning message is sent).
If no value is given, I keep the field to None.

The documentation is updated accordingly in the configuration.rst file.
I took the liberty to change the level of configurable options [core], [worker], ... to simplify the navigation menu.

## Motivation and Context

I run Luigi on Windows. It opens configuration files with CP1252 encoding while they are in UTF8.

I created a small script in the example folder (hello_encoding_world.py) to test it.


mpoussevin:
The first version was not considering Python2. 
Passing the `_config_encoding` to the `read()` method raised a TypeError.
I catch the error and fallback to default behaviour for Python2.

Is there a better way to do it ?

Tarrasch:
Is the `-` to `*` conversion related to this?

mpoussevin:
> Is the `-` to `*` conversion related to this?

Hello, I don't get which conversion you're talking about. Could you please tell me more ?

dlstadther:
Also, please resolve merge conflict. Update with `master` branch

dlstadther:
Hey @mpoussevin  I'm not sure we're communicating well related to the .rst changes. @Tarrasch and I have not been questioning the use of `*` or `-` as unordered list items, but rather the use of a sequence of them (`-------` or `*******`) for header denotation.

mpoussevin:
Hello @dlstadther  

Thanks, I get it now. 
I propose using `---` for page title, `~~~` for level 1 sections (as in `doc/tasks.rst`) and `***` for level 2 sections (I couldn't find any .rst with level 2 sections though).

Tell me if that's OK !

mpoussevin:
Wow. I don't get how changing only the `configuration.rst` file between commit 880a711 and 7aa39af breaks the `flake8` test.
Any idea ?

mpoussevin:
OK, investigating further, I found out that setting UTF8 with the TOML parser yields `unicode` objects instead of  `str` objects : the `serialize` method from `Parameter` fails `str(u'hé')` fails in python 2. 
I don't want to change the behavior of `serialize`. I'll think of a way out.

mpoussevin:
I don't think it's worth it, too many modifications in the core code to use Unicode for small benefits. 
Closing the PR.
Thanks guys.

PR #2399, opened 2018-04-13
Add configuration to enable/disable the pause button

kevinwang:
## Description
This adds a new config value which, when set to false, will turn disable the functionality of the pause/unpause RPC methods and also hide the pause toggle from the visualiser.

## Motivation and Context
We wish to have the ability to disable this functionality because curious engineers may click the button without knowing what it does and accidentally pause all tasks.

I ensured that when the config is set to false, the button is hidden from the visualiser.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -716,6 +716,10 @@ worker_disconnect_delay
   scheduler before removing it and marking all of its running tasks as
   failed. Defaults to 60.
 
+pause_enabled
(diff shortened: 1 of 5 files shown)


Tarrasch:
Looks great! Would you mind at least make an API-unittest even if you don't have a unittest for the UI part?

Tarrasch:
Thanks. Hopefully one day (in far distant future) we'll have proper ACLs and stuff. :)

PR #2400, opened 2018-04-14
Remove unnecessary method in ConcurrentRunningResourcesTest.

riga:
## Description

This (trivial) PR removes an unnecessary method in `test.task_running_resources_test.ConcurrentRunningResourcesTest`. The test itself might be a good example to show how to use a remote scheduler in unit tests so it's worth cleaning it up.

The change was proposed by @Tarrasch in https://github.com/spotify/luigi/pull/2346#pullrequestreview-110284998.


Tarrasch:
Thank you! Indeed I think you test case can be an example to follow!

PR #2401, opened 2018-04-14
Add worker config to force multiprocessing

riga:
## Description

At the moment, luigi uses multiprocessing for tasks only when the number of requested workers is > 1. In some situations, it might be helpful if this was also the case for exactly 1 worker (e.g. when the executed code does something nasty with the main process, TensorFlow in our case). I added a worker config `force_multiprocessing` which, when `True`, forces multiprocessing to be used independent of the number of workers. I also added a few lines to the docs.

Yup, there are two additional tests now in `test/worker_multiprocess_test.py` that check if the created `TaskProcess` instance received the correct multiprocessing flag.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -270,6 +270,12 @@ check_unfulfilled_deps
   resource-intensive.
   Defaults to true.
 
+force_multiprocessing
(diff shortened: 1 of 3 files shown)


Tarrasch:
Awesome! Thanks!

PR #2404, opened 2018-04-24
Change task result parameter value to TEXT

### LINKED ISSUE #2240: luigi parameter value column is set to limit size of 256

kwilcox:
This is a manual rebase of #2241 onto master with one additional commit.

## Description

Parameter values longer than 256 characters currently fail to be written to the SQL result store. This changes the task value column to TEXT to avoid the issue.

## Motivation and Context

Fixes #2240
Fixes #1885

I've tested this against a PostgreSQL setup, both a new database and existing database, and it works.  I have not tested any other dialects (help plz). The existing tests **should** cover MySQL and SQLite? Marking WIP until I can confirm this works with the required dialects.

### DIFF
--- a/luigi/db_task_history.py
+++ b/luigi/db_task_history.py
@@ -247,16 +247,35 @@ def _upgrade_schema(engine):
     :param engine: SQLAlchemy engine of the underlying database.
     """
     inspector = reflection.Inspector.from_engine(engine)
-    conn = engine.connect()
+    with engine.connect() as conn:
 
-    # Upgrade 1.  Add task_id column and index to tasks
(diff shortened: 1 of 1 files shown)


kwilcox:
I tested this extensively against PostgreSQL and SQLite and it migrates things correctly if moving from a previous table on PostgreSQL. SQLite doesn't support changing column types without changing the table name and copying the data over so a warning message is logged if the migration can't be run.

kwilcox:
@Jimexist @oivoodoo would you like to confirm this is working for you before merge?

oivoodoo:
@kwilcox it looks great. I believe that it would work. thank you!

PR #2406, opened 2018-04-24
this feels more like a warning than error

honnix:
## Description

Log exception as warning when retrying communicating with remote scheduler.

## Motivation and Context

Since this happens during retrying, a warning seems to be more suitable.

Tested it manually.
```
import logging
from logging import getLogger

logging.basicConfig(level=logging.INFO)
logger = getLogger('test')

try:
    10 / 0
except Exception as exp:
    logger.exception("an error with exception")
    logger.warning("a warning with exception" , exc_info=True)
```
```
python test.py
ERROR:test:an error with exception
  File "test.py", line 8, in <module>
    10 / 0
ZeroDivisionError: integer division or modulo by zero
WARNING:test:a warning with exception
  File "test.py", line 8, in <module>
    10 / 0
ZeroDivisionError: integer division or modulo by zero
```

@dlstadther PTAL Does this make sense?

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -141,7 +141,8 @@ class RemoteScheduler(object):
             except self._fetcher.raises as e:
                 last_exception = e
                 if log_exceptions:
-                    logger.exception("Failed connecting to remote scheduler %r", self._url)
+                    logger.warning("Failed connecting to remote scheduler %r", self._url,
+                                   exc_info=True)
                 continue
         else:
             raise RPCError(


dlstadther:
The only time I've received this is when `luigid` isn't running. Since it isn't running when i expect it to, an exception feels expected. Then again, since the code allows the run to continue, it also makes sense to have the logging be a warning...

honnix:
We have experienced that when luigid is somehow overloaded, we got `ConnectionReset`, `Timeout`, etc. And after a couple of retries, it could recover.

dlstadther:
@honnix Then sounds like a valid update. You've got my approval!

PR #2417, opened 2018-05-09
Pass on 'port' argument to PostgresTarget (fixes #2416)

### LINKED ISSUE #2416: PostgresQuery does not honour `port`
`PostgresQuery` fails when running a query against a PostgreSQL database on a non-standard port (other than 5432).

This is because the `PostgresTarget` which is the output of this task (and through which the DB connection is [established](https://github.com/spotify/luigi/blob/master/luigi/contrib/postgres.py#L362)) does not pass on the `port` argument (see [here](https://github.com/spotify/luigi/blob/master/luigi/contrib/postgres.py#L383)).

maxalbert:
Should I add a regression test for this, or is it not worth it? If it is, could someone point me to relevant files (e.g. where the postgres test server is set up)? Thanks!

### DIFF
--- a/luigi/contrib/postgres.py
+++ b/luigi/contrib/postgres.py
@@ -386,5 +386,6 @@ class PostgresQuery(rdbms.Query):
             user=self.user,
             password=self.password,
             table=self.table,
-            update_id=self.update_id
+            update_id=self.update_id,
+            port=self.port
         )
--- a/test/contrib/postgres_test.py
+++ b/test/contrib/postgres_test.py
@@ -87,6 +87,7 @@ class DummyPostgresQuery(luigi.contrib.postgres.PostgresQuery):
     date = luigi.DateParameter()
 
     host = 'dummy_host'
+    port = 'dummy_port'
     database = 'dummy_database'
     user = 'dummy_user'
     password = 'dummy_password'


luigi/contrib/postgres.py
```diff
@@ -386,5 +386,6 @@ def output(self):
             user=self.user,
             password=self.password,
             table=self.table,
-            update_id=self.update_id
+            update_id=self.update_id,
+            port=self.port
```
dlstadther: There is no default value for `self.port` here. Aka: port is not inherited from rdbms.Query

It looks like it might also be an option to provide your port via `:1234` at [the end of your host string ](https://github.com/maxalbert/luigi/blob/89399079f46144d54959ed7703d83aac5c7d993e/luigi/contrib/postgres.py#L125-L126)

maxalbert: Whoops, good catch about the missing default! (I'm a bit surprised this didn't make the tests fail.)

Is there a reason that `rdbms.Query` does not have a `port` attribute? Sounds like it would be worth adding? I'm happy to include it in this PR, unless there is an argument against it?

Regarding your second comment, it's true that specifying the port at the end of the host string is a workaround, but from a user's perspective this is not obvious (without reading the source code), so it would be nice if the `port` argument was supported - especially because it's such a standard argument in database connection strings.

dlstadther: I'm not intimately aware why `rdbms.Query` doesn't have a port specified. I'll leave it up to your discretion of it makes sense to include it. If you do chose to include it, require it to be overwritten and make the appropriate updates everywhere `rdbms.Query` is inherited (i think just PostgresQuery and RedshiftQuery).

That's fair. I'm not against an explicit `port` specification.

PR #2422, opened 2018-05-14
Add Glossier as part of the companies that use Luigi

cabouffard:
## Description
Let's spread the love!


dlstadther:
Thanks for sharing, @cabouffard ! Looking forward to the next blog in this series!

PR #2423, opened 2018-05-14
Reenables true multithreaded s3 copy

dlstadther:
Boto3 update removed multithreaded file copy in favor of multithreaded multipart file copy.

## Description

Re-enables true multithreaded file copy. (Logic taken directly from pre-boto3 s3.py)

## Motivation and Context

Boto3's `TransferConfig` appears to enable multithreaded copy capabilities for multipart files, but not for the copying of many small files.

I've run with test code. Other than that, i'm letting travis test it.

### DIFF
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -30,6 +30,8 @@ import os
 import os.path
 import warnings
 
+from multiprocessing.pool import ThreadPool
+
 import botocore
 
 try:
@@ -323,6 +325,9 @@ class S3Client(FileSystem):
         total_keys = 0
 
         if self.isdir(source_path):
+            copy_jobs = []
+            management_pool = ThreadPool(processes=threads)
+
             (bucket, key) = self._path_to_bucket_and_key(source_path)
             key_path = self._add_path_delimiter(key)
             key_path_len = len(key_path)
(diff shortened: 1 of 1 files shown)


luigi/contrib/s3.py
```diff
@@ -340,8 +345,19 @@ def copy(self, source_path, destination_path, threads=100, start_time=None, end_
...
-                        copy_source, dst_bucket, dst_prefix + path, Config=transfer_config, ExtraArgs=kwargs)
+                    the_kwargs = {'Config': transfer_config, 'ExtraArgs': kwargs}
+                    job = management_pool.apply_async(self.s3.meta.client.copy,
+                                                      args=(copy_source, dst_bucket, dst_prefix + path),
+                                                      kwds=the_kwargs)
+                    copy_jobs.append(job)
+
+            # Wait for the pools to finish scheduling all the copies
+            management_pool.close()
+            management_pool.join()
```
Tarrasch: Please just check that this doesn't have a ContextManager interface maybe?

dlstadther: A few minutes of Google searching doesn't return valuable resources towards this desire.

ouanixi:
@dlstadther tests that are skipped on travis are tests that have shown issues with moto (mocking library for boto), it appears when testing multipart functionality, there is a race condition expressed in same tests passing and failing intermittently.

dlstadther:
I've spent some time trying to test the true multithreaded s3 copy, but cannot come up with a reasonable test to identify the behavior.

* Confirmed that `TestS3Client.test_copy_dir` continues to run successfully with this code
* Tried to add test to show that as threads increase, runtime reduces, but for this to be a consistent speed improvement the number and size of files has to increase such to be unreasonable (runtime)

@Tarrasch or @ouanixi , do either of you have any suggestions?

Thanks,

dlstadther:
bump

Tarrasch:
I wouldn't bother to much about writing unit-tests for multi-threaded code if it's already working in production for you. I wouldn't let that block me from merging.

dlstadther:
In this case, I'll merge when I return from vacation. Thanks @Tarrasch !

dlstadther:
Unable to provide unit-tests for multi-threading within method. Code has been running successfully copying millions of files daily in production for ~2 months.

Merging...

PR #2424, opened 2018-05-18
Clarify wording in execution summary

ulzha:
## Motivation and Context

Some users were confused by the existing wording.

Updated unittests.


Tarrasch:
Great! Thanks. :)

PR #2426, opened 2018-05-18
Send messages from scheduler to tasks

riga:
## Description

This PR enables the scheduler to send messages to particular tasks. More precisely, messages are sent to workers via the RPC message system which dispatches them to a specified task. Running tasks 
can access a `multiprocessing.Queue` object (named `scheduler_messages`, set via the `StatusReporter`) which stores incoming messages. Users can implement custom behavior to react to certain messages (see code example below).

The new config `scheduler.send_messages` controls whether a scheduler is capable of sending messages in the first place. By defining `accepted_messages` a task can define whether it accepts messages, independent of the scheduler the worker is talking to. The message button is only visible in the UI when the scheduler is able to send messages, and the task can receive them (see screenshots).

There is also a simple "respond" mechanism so tasks can immediately respond to incoming messages which is shown in the scheduler.

"Send message" action button:

Action button

Message prompt:

Message prompt

Awaiting response:

Awaiting response

Response arrived:

Response arrived

When `Await response` is not checked, the prompt closes upon `Send`. Otherwise, the prompt remains open and a response container is shown. This triggers a simple polling to fetch and display the actual response.

## Motivation and Context

There are many use cases when the messaging could be useful. One particular example is machine learning. Our training tasks publish their current results to the scheduler. Although they contain predefined termination criteria, it could be nice to manually trigger *graceful* training termination if the training stagnates:

```python
class Training(luigi.Task):

    def requires(self):
        ...

    def output(self):
        return luigi.LocalTarget("my_ml_model.pb")

    def run(self):
        # setup the training
        model = ...

        for _ in training_loop():
            # training stuff with internal termination criterion
            model.train(...)

            # check messages
            if not self.scheduler_messages.empty():
                msg = self.scheduler_messages.get()
                if msg.content == "terminate":
                    break
                else:
                    msg.respond("unknown message")

        # save the model
        model.save(self.output().path)
```

Yep, I added unit tests and docs.

### DIFF
--- a/doc/luigi_patterns.rst
+++ b/doc/luigi_patterns.rst
@@ -334,7 +334,7 @@ messages:
         ...
 
         # configure the task to accept all incoming messages
-        accepted_messages = True
+        accepts_messages = True
 
         def run(self):
             # this example runs some loop and listens for the
--- a/doc/luigi_patterns.rst
+++ b/doc/luigi_patterns.rst
@@ -317,3 +317,39 @@ built-in solutions. In the case of you're dealing with a file system
 :meth:`~luigi.target.FileSystemTarget.temporary_path`. For other targets, you
 should ensure that the way you're writing your final output directory is
 atomic.
+
+Sending messages to tasks
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The central scheduler is able to send messages to particular tasks. When a running task accepts 
+messages, it can access a `multiprocessing.Queue <https://docs.python.org/3/library/multiprocessing.html#pipes-and-queues>`__
+object storing incoming messages. You can implement custom behavior to react and respond to
+messages:
+
+.. code-block:: python
+
+    class Example(luigi.Task):
+
+        # common task setup
+        ...
+
+        # configure the task to accept all incoming messages
+        accepted_messages = True
+
+        def run(self):
+            # this example runs some loop and listens for the
+            # "terminate" message, and responds to all other messages
+            for _ in some_loop():
+                # check incomming messages
+                if not self.scheduler_messages.empty():
+                    msg = self.scheduler_messages.get()
+                    if msg.content == "terminate":
+                        break
+                    else:
+                        msg.respond("unknown message")
+
+            # finalize
+            ...
+
+Messages can be sent right from the scheduler UI which also displays responses (if any). Note that
+this feature is only available when the scheduler is configured to send messages (see the :ref:`scheduler-config` config), and the task is configured to accept them.
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -1264,12 +1264,15 @@ class Scheduler(object):
             'tracking_url': getattr(task, "tracking_url", None),
             'status_message': getattr(task, "status_message", None),
             'progress_percentage': getattr(task, "progress_percentage", None),
-            'enabled_scheduler_messages': self._config.send_messages,
+            'enabled_scheduler_messages': False,
         }
         if task.status == DISABLED:
             ret['re_enable_able'] = task.scheduler_disable_time is not None
         if include_deps:
             ret['deps'] = list(task.deps if deps is None else deps)
+        if self._config.send_messages and task.status == RUNNING and task.worker_running:
+            worker = self._state.get_worker(task.worker_running)
+            ret['enabled_scheduler_messages'] = worker.info.get('receive_messages', False)
         return ret
 
     @rpc_method()
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -149,6 +149,8 @@ class scheduler(Config):
 
     pause_enabled = parameter.BoolParameter(default=True)
 
+    send_messages = parameter.BoolParameter(default=True)
+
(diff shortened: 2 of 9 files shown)


luigi/worker.py
```diff
@@ -1148,3 +1155,14 @@ def set_worker_processes(self, n):
 
         # tell the scheduler
         self._scheduler.add_worker(self._id, {'workers': self.worker_processes})
+
+    @rpc_message_callback
+    def dispatch_scheduler_message(self, task_id, message):
```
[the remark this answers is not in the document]
riga: Implemented a draft if this reply/response mechanism.

luigi/scheduler.py
```diff
@@ -932,6 +936,31 @@ def disable_worker(self, worker):
...
+    @rpc_method()
+    def send_scheduler_message(self, worker, task, message):
+        if not self._config.send_messages:
+            return {"messageId": None}
+
+        message_id = str(uuid.uuid4())
+        self._state.get_worker(worker).add_rpc_message('dispatch_scheduler_message', task_id=task,
+                                                       message_id=message_id, message=message)
+
+        return {"messageId": message_id}
```
Tarrasch: yuck, I just realize that we mix camelCase and snake_case in the key-names. But it seems snake_case is more used. Maybe use that?

luigi/worker.py
```diff
@@ -392,6 +420,10 @@ class worker(Config):
                                           description='If true, use multiprocessing also when '
                                           'running with 1 worker')
 
+    receive_messages = BoolParameter(default=True,
```
Tarrasch: Maybe `can_receive_messages`?

Also I'm thinking of having the default to `False`, so the UI button doesn't appear by default.

luigi/worker.py
```diff
@@ -447,17 +479,17 @@ def __init__(self, scheduler=None, worker_id=None, worker_processes=1, assistant
         if scheduler is None:
             scheduler = Scheduler()
 
+        self._config = worker(**kwargs)
+
+        assert self._config.wait_interval >= _WAIT_INTERVAL_EPS, "[worker] wait_interval must be positive"
+        assert self._config.wait_jitter >= 0.0, "[worker] wait_jitter must be equal or greater than zero"
+
```
Tarrasch: Why the need to move this code?

Tarrasch: (feel free to if you have a reason, I'm mostly curious)

luigi/worker.py
```diff
@@ -540,7 +572,8 @@ def _generate_worker_info(self):
         # Generate as much info as possible about the worker
         # Some of these calls might not be available on all OS's
         args = [('salt', '%09d' % random.randrange(0, 999999999)),
-                ('workers', self.worker_processes)]
+                ('workers', self.worker_processes),
+                ('receive_messages', self._config.receive_messages)]
```
Tarrasch: Hmmm... wait, why is this a property of the worker...

Thinking about it I'm sure it makes sense. But I really would like to see some user docs.

luigi/scheduler.py
```diff
@@ -932,6 +936,31 @@ def disable_worker(self, worker):
     def set_worker_processes(self, worker, n):
         self._state.get_worker(worker).add_rpc_message('set_worker_processes', n=n)
 
+    @rpc_method()
+    def send_scheduler_message(self, worker, task, message):
```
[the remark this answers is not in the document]
riga: ✔︎

luigi/scheduler.py
riga: ✔︎

luigi/worker.py
riga: This was required as `_generate_worker_info()` needs to access `_config`.

luigi/worker.py
riga: Mh, but you have a point. Another alternative would be to have sth like a "can_receive_messages()" hook (defaulting to false) defined on a task. Actually this makes more sense as the worker should just dispatch and let tasks decide what to do with messages. Are you fine with that change?

Tarrasch: Well. I think it would make sense to both have message for tasks and workers.

Worker messages could be "shut down asap", "shut down task x", "decrease num workers" (we already have that, don't we)? Tasks messages I imagine are mostly task-specific.

----

I'm not sure how to implement it. Anyway, I'm fine with any changes, I'm sure if you think it becomes better so will I. :)

doc/luigi_patterns.rst
```diff
@@ -317,3 +317,39 @@ built-in solutions. In the case of you're dealing with a file system
...
+
+.. code-block:: python
+
+    class Example(luigi.Task):
+
+        # common task setup
+        ...
+
+        # configure the task to accept all incoming messages
+        accepted_messages = True
```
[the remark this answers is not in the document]
riga: ✔︎

Tarrasch:
Are you happy with this now? Is it ready to merge? Is it reasonably tested?

riga:
Yep, I'm happy with the PR. Docs and tests should be fine.

Tarrasch:
Thanks!! This is so cool!

PR #2427, opened 2018-05-24
Improve handling of bool parameters.

riga:

The current BoolParameter implementation does not allow to negate parameters on the command line. Example:

```python
class MyTask(luigi.Task):
    param_a = luigi.BoolParameter(default=False)
    param_b = luigi.BoolParameter(default=True) # cannot be set to False on the CLI
    ...
```

So right now, there is no way to set `param_b` to False from the cli. Of course, users can redefine `param_b` to mean the opposite e.g. `no_param_b`, but this might become cumbersome in the long run (at least for us it kind of does). Also, there are some luigi-internal bool parameters (e.g. `send_failure_email`, `check_unfulfilled_deps`) which are true by default, so one has to toggle the value manually in the config file for running individual tasks.

The behavior I have in mind is:

```bash
luigi MyTask
# -> param_a = False (as before)

luigi MyTask --param_a
# -> param_a = True (as before)

luigi MyTask --param_a true
# -> param_a = True (new)

luigi MyTask --param_a false
# -> param_a = False (new)

# same behavior for param_b except for the default as before
luigi MyTask
# -> param_b = True (as before)
```

To achieve this, one needs to pass different values of `nargs` and `const` to the ArgumentParser,  which requires only a minor change. I also updated `BoolParameter.normalize()` which was originally introduced in #1461. Until now it returned `bool(value)` or `None`. I think it is safe to return the result of `parse()` or `None` for back-compat in case of an error.

Does this sound reasonable to you?

I extended all existing tests that deal with BoolParamter parsing to cover the update behavior.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -612,8 +612,31 @@ class FloatParameter(Parameter):
 
 class BoolParameter(Parameter):
     """
-    A Parameter whose value is a ``bool``. This parameter have an implicit
-    default value of ``False``.
+    A Parameter whose value is a ``bool``. This parameter has an implicit default value of
+    ``False``. For the command line interface this means that the value is ``False`` unless you
+    add ``"--the-bool-parameter"`` to your command without giving a parameter value. This is
+    considered *implicit* parsing (the default). However, in some situations one might want to give
+    the explicit bool value (``"--the-bool-parameter true|false"``), e.g. when you configure the
+    default value to be ``True``. This is called *explicit* parsing. When omitting the parameter
+    value, it is still considered ``True`` but to avoid ambiguities during argument parsing, make
+    sure to always place bool parameters behind the task family on the command line when using
+    explicit parsing.
+
+    You can toggle between the two parsing modes on a per-parameter base via
+
+    .. code-block:: python
+
+        class MyTask(luigi.Task):
+            implicit_bool = luigi.BoolParameter(parsing=luigi.BoolParameter.IMPLICIT_PARSING)
+            explicit_bool = luigi.BoolParameter(parsing=luigi.BoolParameter.EXPLICIT_PARSING)
+
+    or globally by
+
+    .. code-block:: python
+
+        luigi.BoolParameter.parsing = luigi.BoolParameter.EXPLICIT_PARSING
+
+    for all bool parameters instantiated after this line.
     """
 
     IMPLICIT_PARSING = "implicit"
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -297,9 +297,12 @@ class Parameter(object):
     def _parser_global_dest(param_name, task_name):
         return task_name + '_' + param_name
 
-    @staticmethod
-    def _parser_action():
-        return "store"
+    @classmethod
+    def _parser_kwargs(cls, param_name, task_name=None):
+        return {
+            "action": "store",
+            "dest": cls._parser_global_dest(param_name, task_name) if task_name else param_name,
+        }
 
 
 class OptionalParameter(Parameter):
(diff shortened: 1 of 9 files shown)


test/helpers.py
```diff
@@ -160,7 +160,7 @@ def run_locally(self, args):
         temp = CmdlineParser._instance
         try:
             CmdlineParser._instance = None
-            run_exit_status = luigi.run(['--local-scheduler', '--no-lock'] + args)
+            run_exit_status = luigi.run(args + ['--local-scheduler', '--no-lock'])
```
dlstadther: Is there a reason for the order swap here?

luigi/parameter.py
```diff
@@ -618,19 +621,32 @@ def __init__(self, *args, **kwargs):
...
 
-    @staticmethod
-    def _parser_action():
-        return 'store_true'
+    @classmethod
+    def _parser_kwargs(cls, *args, **kwargs):
+        parser_kwargs = super(BoolParameter, cls)._parser_kwargs(*args, **kwargs)
+        parser_kwargs.update({
+            "nargs": "?",
+            "const": True,
```
dlstadther: What is the importance of `nargs` and `const`?

test/helpers.py
riga: So you're fine with that change? If yes, I'll also add a note to the docs on that.

dlstadther: Yeah, i'm fine with it - so long as tests pass (they aren't currently)

Tarrasch:
Did you consider adding a different version of BoolParameter or a parameter to it's constructor? I'm so afraid of changing the existing behavior will cause breakage (even if we think it's backward compatible)

riga:
@tarrasch This is actually what I fear as well. I added a parameter `improved_parsing` to the BoolParameter init which steers the parsing behavior. It defaults to the class attribute `improved_parsing` which itself defaults to False, so the current behavior is still standard. However, one can do

```python
luigi.BoolParameter.improved_parsing = True
```

at an early stage to trigger the improved parsing for all parameters, including internal ones.

Tarrasch:
Sounds good! Could we rename it to something like this?

    luigi.BoolParameter.parsing = luigi.BoolParameter.(IMPLICIT_PARSING/EXPLICIT_PARSING)

test/cmdline_test.py
```diff
@@ -125,12 +125,12 @@ def test_cmdline_local_scheduler(self, logger):
 
     @mock.patch("logging.getLogger")
     def test_cmdline_other_task(self, logger):
-        luigi.run(['--local-scheduler', '--no-lock', 'SomeTask', '--n', '1000'])
+        luigi.run(['SomeTask', '--local-scheduler', '--no-lock', '--n', '1000'])
```
Tarrasch: This shouldn't be needed to change right?

test/cmdline_test.py
```diff
@@ -125,12 +125,12 @@ def test_cmdline_local_scheduler(self, logger):
...
         self.assertEqual(dict(MockTarget.fs.get_all_data()), {'/tmp/test_1000': b'done'})
 
     @mock.patch("logging.getLogger")
     def test_cmdline_ambiguous_class(self, logger):
-        self.assertRaises(Exception, luigi.run, ['--local-scheduler', '--no-lock', 'AmbiguousClass'])
+        self.assertRaises(Exception, luigi.run, ['AmbiguousClass', '--local-scheduler', '--no-lock'])
```
Tarrasch: Same here, we don't want code who pass `--local-scheduler` first to break.

riga:
Yes, you're right, they don't require any changes.

test/cmdline_test.py
riga: ✔︎

test/cmdline_test.py
riga: ✔︎

riga:
The requested changes are included now, so feel free to review again :)

Tarrasch:
Sorry for the late review. It looks all good but I would like some docs too. Simple class-documentation for `BoolParameter` would be easiest I think.

PR #2428, opened 2018-05-24
Add `target_table` to `PostgresTarget.exists()`

grauscher:
## Description
Adding `target_table` to the `exists` method of PostgresTarget checks if a specific `update_id` exists for a specific table, rather than any table.

## Motivation and Context
This way, one can use a "general" `update_id` (e.g. update_id = '2018-05-01', if a Pipeline has to be executed on a daily basis) for every task involved in this Pipeline.
Since a `PostgresTarget` can only point to one specific table anyway, it makes sense to constraint its `exists` method to check for an `update_id` of this specific table.

In the current implementation, every task in a Pipeline has to have a specific `update_id`. 
This often results in the table name being integrated to the `update_id` (e.g.: 'table-a-2018-05-01', 'table-b-2018-05-01', ...), which brings redundancy, since the table name is in the column `target_table` anyway.

### DIFF
--- a/luigi/contrib/postgres.py
+++ b/luigi/contrib/postgres.py
@@ -169,10 +169,9 @@ class PostgresTarget(luigi.Target):
         cursor = connection.cursor()
         try:
             cursor.execute("""SELECT 1 FROM {marker_table}
-                WHERE update_id = %s
+                WHERE (update_id, target_table) = (%s, %s)
                 LIMIT 1""".format(marker_table=self.marker_table),
-                           (self.update_id,)
-                           )
+                           (self.update_id, self.table))
             row = cursor.fetchone()
         except psycopg2.ProgrammingError as e:
             if e.pgcode == psycopg2.errorcodes.UNDEFINED_TABLE:


Tarrasch:

dlstadther:
What is your use case to need to use a "general" `update_id`? (primarily curious)
The primary purpose of the update table is to store persistent record of "unique" luigi-based database operations. Off hand, i'm not thinking of a broad use-case to need a general update_id here

grauscher:
The "general" `update_id` I was thinking, was something like: '2018-05', for a table that only has to be calculated once a month.

What I usually do is to append the table name to the update_id, like '<table_name>-2018-05'.
But I actually find this redundant, since the name of the column is right there, on the next column of the same table...

grauscher:
I like using the `marker_table`, where the "luigi-based database operations" are stored, as a _log_.
I can see, very clearly, which pipelines already ran.

Being able to aggregate all update_id's of tasks of a Pipeline into one, would help me control the execution flow, filtering this table, or applying group by's.

And it also seems more _elegant_ to me.

dlstadther:
Alright, cool. You've sold me on _why_ this update should be included in Luigi.

Could you add a short test case to be sure that something like update_id/target_table ('2018-05-23', foobar) and ('2018-05-24', foobar) are unique entries in table_updates? (or something like that)

Lastly, I wonder where a good place may be to include some useful documentation about the various ways table_updates can be utilized (such as your use-cases) to help new Luigi users.

@Tarrasch What are your thoughts on some form of documentation about helpful Luigi style/use of various Luigi modules?

jamesmcm:
This seems okay, as you say, it is backward compatible since we don't change the actual update_id itself.

I should add that we solved this in-house by adding the following to the Redshift target:

```python
target.marker_table = 'updates.{}_{}'.format(self.table.replace('.', '_'), camel_to_snake(self.__class__.__name__))
```

That way, each task has its own updates table in the updates schema (this was for Redshift).

grauscher:
@jamesmcm that is indeed an interesting approach!

Just a comment: I think for users who understand the back engine, it works fine.
However for new users, who are not acquainted to the infrastructure, it introduces an extra step.

dlstadther:
@grauscher so sorry for the delay getting back to you here!!

I've thought about this some more and agree with @jamesmcm . So long as this is backwards compatible and defaults to the current behavior, I'm find with this patch.

Yes, I do mean adding a test to `test/contrib/postgres_test.py`

Issue #2342, opened 2018-01-30
Make capturing of stdout / stderr optional in ExternalProgramTask


jytug:
Currently, during execution of an `ExternalProgramTask` the standard output is captured and redirected to the logger after the execution. It would be useful, during debugging of long-running tasks (for instance a `SparkSubmitTask`), to disable this mechanism, and see the standard output on the fly.

PR #2430, opened 2018-05-25
Add a "capture_output" parameter to ExternalProgramTask

jytug:
## Description
A `capture_output` is added to `ExternalProgramTask`, so that
the output is captured only optionally (which is still the default
behaviour).

## Motivation and Context
This change can make debugging long-running tasks easier
by allowing the user to see the output on the fly. The current
behaviour remains unchanged by having the parameter set
to True by default

I have been running jobs this way by hot-fixing luigi
in `dist-packages`

Simple usage:

```
In [1]: from luigi.contrib.external_program import ExternalProgramTask

In [2]: class LsTask(ExternalProgramTask):
   ...:     def program_args(self):
   ...:         return ['ls']
   ...:     

In [3]: task = LsTask(capture_output=False)

In [4]: task.run()
bin  codecov.yml  CONTRIBUTING.rst  doc  examples  ISSUE_TEMPLATE.md  LICENSE  luigi  MANIFEST.in  PULL_REQUEST_TEMPLATE.md  README.rst  RELEASE-PROCESS.rst  scripts  setup.py  test  tox.in
```
*Edit*:
in reference to issue #2342

### DIFF
--- a/luigi/contrib/external_program.py
+++ b/luigi/contrib/external_program.py
@@ -53,7 +53,13 @@ class ExternalProgramTask(luigi.Task):
     and you can optionally override :py:meth:`program_environment` if you want to
     control the environment variables (see :py:class:`ExternalPythonProgramTask`
     for an example).
+
+    By default, the output (stdout and stderr) of the run external program
+    is being captured and displayed after the execution has ended. This
+    behaviour can be overriden by passing ``--capture-output False``
     """
+
+    capture_output = luigi.BoolParameter(default=True, significant=False)
 
     def program_args(self):
         """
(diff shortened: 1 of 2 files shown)


Tarrasch:
Good idea. I would like to see some docs and tests though. :)

jytug:
I would like to add a unit test in `tests/contrib/external_program_test.py`. However, I can't seem to find a way to run those tests. And invocation like:

```
python -m unittest test/contrib/external_program.py
```

results in the following error:

```
ImportError: Failed to import test module: test/contrib/external_program
  File "/Users/filip/miniconda2/envs/p3/lib/python3.6/unittest/loader.py", line 153, in loadTestsFromName
    module = __import__(module_name)
ModuleNotFoundError: No module named 'test/contrib/external_program'
```

dlstadther:
@jytug take a look at [CONTRIBUTING.rst](https://github.com/spotify/luigi/blob/master/CONTRIBUTING.rst)

PR #2435, opened 2018-05-30
Add table constraints support to Redshift

cabouffard:
## Description
Add the ability to add table constraint to Redshift tables.

## Motivation and Context
While the table keys constraints (UNIQUE, PRIMARY, FOREIGN) aren't
enforced by Redshift, it's still useful to be able to set that data.
Given the situation that there is a monitoring task that manually checks
if the table constraints are respected, it's useful to be able to set
multiple primary keys on a given table.

Based on the syntax of the CREATE commands of Redshift (see link down
below) we think it's useful to add support to it.

https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_NEW.html

We've been using this feature for a few months now without problems.


luigi/contrib/redshift.py
```diff
@@ -308,12 +324,18 @@ def create_table(self, connection):
...
+                table_constraints = ',' + self.table_constraints
+
             query = ("CREATE {type} TABLE "
-                     "{table} ({coldefs}) "
+                     "{table} ({coldefs} {table_constraints}) "
                      "{table_attributes}").format(
                 type=self.table_type,
                 table=self.table,
                 coldefs=coldefs,
+                table_constraints=table_constraints,
```
dlstadther: Out of the scope of this PR, but if you'd like it would be awesome to reduce the duplicated code here for where len == 2 and len == 3 for the creation of `query`. The only difference is the definition of `coldefs`

cabouffard: I would agree with you that the whole Redshift contrib should be refactored. There would a lot of value to make it a little bit more DRY.

If I have more time in the upcoming weeks I'll take a jab at refactoring it so that it's a little bit saner! :)

cabouffard:
@dlstadther I've added specs like you've suggested. It seems though that there is a spec that randomly fails that is not related to my edit. What would you suggest the next step be?

cabouffard:
Now that I think more about this, maybe the constraint should be a tuple instead of a string. I'll think about it overnight and see where I stand on this!

cabouffard:
I've decided to keep going on with the string instead of the tuple to follow the same behavior as `table_attributes`. So you've got my green light on this! :)

PR #2436, opened 2018-05-30
Check against goal_task_family's family attribute

elainexmas:
## Description

`find_deps` in `luigi.tools.deps` constructs the dependency path between task and upstream_task_family by recursively calling `dfs_paths`.  The base condition in `dfs_paths` performs a comparison between `start_task.task_family` and `goal_task_family`.  However, `start_task` is a fully-parameterized Task object and `goal_task_family` becomes a fully-parameterized Task object beyond the initial level of recursion.  Thus, the equality check between `start_task.task_family` and `goal_task_family` will return False and no path is returned.

## Motivation and Context

Without this change, valid calls to `find_deps(task, upstream_task_family)` always return `set([])`, even though a path exists between `task` and `upstream_task_family`

### DIFF
--- a/luigi/tools/deps.py
+++ b/luigi/tools/deps.py
@@ -57,7 +57,7 @@ def get_task_requires(task):
 def dfs_paths(start_task, goal_task_family, path=None):
     if path is None:
         path = [start_task]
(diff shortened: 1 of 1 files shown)


Tarrasch:
Can you add tests?

Issue #2431, opened 2018-05-29
Demos in docs are still using print XXX, how about using print(XXX) for better python3 support?


licuiwei:
I saw that Luigi says to be supporting both python2 and python3, but in the documents of [Running from the Command Line](https://luigi.readthedocs.io/en/latest/running_luigi.html), the demos are still using print in a way like "print self.x + self.y", which python3 doesn't support.
I'm using python3 and when I tried the demos, it gave me errors like this:
```
File "C:\python\my_module.py", line 9
print self.x + self.y
^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(self.x + self.y)?
```
So I think using "print(self.x+self.y)" instead may be better, because print() function is supported by both python2 and python3.
By fixing this, the docs and demos could be more friendly to beginners :).
PS. This "print" may also exist in other documents, I think these should all be fixed.

dlstadther:
Feel free to submit a PR to update Demos for Python3 support

PR #2438, opened 2018-06-04
Use print(XXX) in documents for better python3 support 

licuiwei:
## Description

Changed all print XXX to print(XXX) in docs.
## Motivation and Context

Issue https://github.com/spotify/luigi/issues/2431
 

Not tested because only document changed.


yoichi:
There is `print d.date` in doc/parameters.rst.
How about changing it together?

dlstadther:
Thanks for the doc cleanup, @licuiwei !

PR #2439, opened 2018-06-06
Add schema to DB if not exists for Redshift

cabouffard:
## Description
This adds the functionality to allow for the schema to be created if we
find a schema as part of the table name.

## Motivation and Context
Based on the current implementation, code exists to create the table if
it doesn't exist. It's a fair assumption that we would like to have the
same behavior for schemas.

Without this feature, the task could potentially crash and throw an error
saying that it cannot create the table under the related schema if a
schema is specified in the table name and that it doesn't exist.

Tested this feature manually in a test database + added a few specs to make sure that the core behavior works.

### DIFF
--- a/luigi/contrib/redshift.py
+++ b/luigi/contrib/redshift.py
@@ -267,6 +267,16 @@ class S3CopyToTable(rdbms.CopyToTable, _CredentialsMixin):
         finally:
             cursor.close()
 
+    def create_schema(self, connection):
+        """
+        Will create the schema in the database
+        """
+        if '.' not in self.table:
+            return
+
+        query = 'CREATE SCHEMA IF NOT EXISTS {schema_name};'.format(schema_name=self.table.split('.')[0])
+        connection.cursor().execute(query)
+
     def create_table(self, connection):
         """
         Override to provide code for creating the target table.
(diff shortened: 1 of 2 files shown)


dlstadther:
Hey @cabouffard , Thanks for all your Luigi activity recently - it's refreshing to see devs actively pushing for improvements :)

Questions regarding this PR:
1) Under what real-world situations would you expect your luigi Tasks to be creating new schema in your Redshift DW?
2) Do we need to consider the increase in runtime of `S3CopyToTable` as we continue to increase the existence checks? (check if update_id exists in marker table, check if table exists, check if schema exists, etc.)

To be clear, I'm not saying this shouldn't be added. I'm just seeking some additional use-case explanation and raising awareness to runtime speed considerations.

Thanks!

cabouffard:
@dlstadther 

1) In our day-to-day, we have source tables that we read from and depending on some data in those source table, we output the result in a new or an existing schema. That said, if source A requires schema B to be created, we would initially have a task that runs named `CreateSchema` that would create the schema. Then we've reflected upon that and we thought to ourselves that maybe `S3CopyToTable` should own that behavior. You literally can't create a table if the schema doesn't exist, so it made sense in our eyes to create the schema under that situation.

So in summary; depending on our source tables data, we have to create different schemas to reflect the type of data it contains so that we can differentiate between multiple sources of data dynamically.

Another idea that we've used that in the past, but no more, would be that each developer would have their own set of schemas and the output of the tasks would get saved in their schema. Since this can yield to multiple schema creations, we felt like the pipeline should take care of that step instead of having to us, manually create the required schemas for each developer.

2) I can concur that there is a lot of checks happening under that query but I think that it kinds of make sense. Mostly because the task itself may not work as expected if it doesn't those checks. I think it's a necessary evil. Also, I don't think that those checks necessary deteriorate the performance by a sufficient amount that we have to re-think it, but that call is over this contribution. Maybe we could think about adding some flags to some checks so that it doesn't do them automatically for the people who want better performance?

On another subject, we've been using Luigi for more than a year now and we've made a lot of tweaks that felt like improvements to us. I took it upon myself to merge back with the community, so you'll probably see a few more PRs coming from us! :)

PR #2440, opened 2018-06-07
Add metadata columns to the RDBMS contrib

cabouffard:
The goal of this feature is to allow metadata column to exists for
specific tables created for Redshift.

Given the scenario where we would always have to have a `created_tz`
column at the end of every table generated by that contrib we could do
the following:

```python
class UpdateTableTask(redshift.S3CopyToTable):

    def metadata_columns(self):
        return [('created_tz', 'TIMESTAMP')]

    def metadata_queries(self):
        query =  'UPDATE {0} ' \ 
                 'SET created_tz = CURRENT_TIMESTAMP ' \
                 'WHERE created_tz IS NULL'.format(self.table)
        return [query]
```

Adding a layer of abstraction over this feature, you could easily
add many default behavior for specific tables for versioning the table
and more.

This feature is opt-in by default since we don't want this break other
people's pipeline after integrating this.

Do we think that this would be a feature that you guys would be interested in merging? I'm unsure if this becomes too specific for our use-case, but it should help some other people to have that type of behavior if need be.

If that's something that you guys are interested in merging, then I'll spend some time in writing a proper testing suite!

Let me know!

### DIFF
--- a/luigi/contrib/rdbms.py
+++ b/luigi/contrib/rdbms.py
@@ -28,7 +28,8 @@ logger = logging.getLogger('luigi-interface')
 
 
 class _MetadataColumnsMixin(object):
-    """
+    """Provice an additinal behavior that adds columns and values to tables
+
     This mixin is used to provide an additional behavior that allow a task to
     add generic metadata columns to every table created for both PSQL and
     Redshift.
@@ -49,7 +50,9 @@ class _MetadataColumnsMixin(object):
     def enable_metadata_columns(self):
         return False
 
-    def _add_metadata_columns(self, cursor):
+    def _add_metadata_columns(self, connection):
+        cursor = connection.cursor()
+
         for column in self.metadata_columns:
             if len(column) == 0:
                 logger.info('Unable to infer column information from column {column} for {table}'.format(column=column, table=self.table))
@@ -65,11 +68,11 @@ class _MetadataColumnsMixin(object):
             schema, table = self.table.split('.')
             query = "SELECT 1 AS column_exists " \
                     "FROM information_schema.columns " \
-                    "WHERE table_schema = LOWER('{0}') AND table_name = LOWER('{1}') AND column_name = LOWER('{2}') LIMIT 1".format(schema, table, column_name)
+                    "WHERE table_schema = LOWER('{0}') AND table_name = LOWER('{1}') AND column_name = LOWER('{2}') LIMIT 1;".format(schema, table, column_name)
         else:
             query = "SELECT 1 AS column_exists " \
                     "FROM information_schema.columns " \
-                    "WHERE table_name = LOWER('{0}') AND column_name = LOWER('{1}') LIMIT 1".format(table, column_name)
+                    "WHERE table_name = LOWER('{0}') AND column_name = LOWER('{1}') LIMIT 1;".format(self.table, column_name)
 
         cursor.execute(query)
         result = cursor.fetchone()
@@ -79,13 +82,18 @@ class _MetadataColumnsMixin(object):
         if len(column) == 1:
             raise ValueError("_add_column_to_table() column type not specified for {column}".format(column=column[0]))
         elif len(column) == 2:
-            query = "ALTER TABLE {table} ADD COLUMN {column}".format(table=self.table, column=' '.join(column))
+            query = "ALTER TABLE {table} ADD COLUMN {column};".format(table=self.table, column=' '.join(column))
         elif len(column) == 3:
-            query = "ALTER TABLE {table} ADD COLUMN {column} ENCODE {encoding}".format(table=self.table, column=' '.join(column[0:2]), encoding=column[3])
+            query = "ALTER TABLE {table} ADD COLUMN {column} ENCODE {encoding};".format(table=self.table, column=' '.join(column[0:2]), encoding=column[2])
         else:
             raise ValueError("_add_column_to_table() found no matching behavior for {column}".format(column=column))
 
         cursor.execute(query)
+
+    def post_copy_metacolumns(self, cursor):
+        logger.info('Executing post copy metadata queries')
+        for query in self.metadata_queries:
+            cursor.execute(query)
 
 
 class CopyToTable(luigi.task.MixinNaiveBulkComplete, _MetadataColumnsMixin, luigi.Task):
--- a/luigi/contrib/rdbms.py
+++ b/luigi/contrib/rdbms.py
@@ -27,7 +27,68 @@ import luigi.task
 logger = logging.getLogger('luigi-interface')
 
 
-class CopyToTable(luigi.task.MixinNaiveBulkComplete, luigi.Task):
+class _MetadataColumnsMixin(object):
+    """
+    This mixin is used to provide an additional behavior that allow a task to
+    add generic metadata columns to every table created for both PSQL and
+    Redshift.
+    """
+    @property
+    def metadata_columns(self):
+        """Returns the default metadata columns.
+
+        Those columns are columns that we want each tables to have by default.
+        """
+        return []
+
+    @property
+    def metadata_queries(self):
+        return []
+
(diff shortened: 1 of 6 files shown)


luigi/contrib/redshift.py
```diff
@@ -37,6 +37,66 @@
                    "Will crash at runtime if postgres functionality is used.")
 
 
+class _MetadataColumnsMixin(object):
```
dlstadther: This isn't distinct to Redshift as i believe this could also function in Postgres. Possibly move this mixin to `rdbms.py`?

luigi/contrib/redshift.py
```diff
@@ -321,6 +381,9 @@ def create_table(self, connection):
             raise ValueError("create_table() found no columns for %r"
                              % self.table)
 
+        if self.do_enable_metadata_columns:
```
dlstadther: rename to `enable_metadata_columns`

luigi/contrib/redshift.py
```diff
@@ -338,6 +401,9 @@ def run(self):
         self.copy(cursor, path)
         self.post_copy(cursor)
 
+        if self.do_enable_metadata_columns:
+            self.post_copy_metacolumns(cursor)
```
dlstadther: I'm a bit confused on how this would interact with `post_copy`. Seems like you would never run both together.

cabouffard:
>This seems like a cool addition; however, i'm not completely sold on it yet. It really feels like a specific wrapper around post_copy, but it's executed along side it.

You would be right to say that it's a wrapper around `post_copy` because in our original implementation, we were using `post_copy` instead of `post_copy_metacolumns`. 

I would agree that they are pretty much the same but we use `post_copy` as task-specifics post copy queries. For instance, we happen to use `post_copy` to deduplicate possible duplicated output. 

Alongside this, we also want to be able to run generic updates. For instance, `created_tz` timestamp. It made sense to decouple them, in our mind, to differentiate task-specific `post_copy` and generic updates to table with `post_copy_metacolumns`. 

I've created a gist that may better demonstrate how we use it: https://gist.github.com/cabouffard/73cfc798bef9ec09fc91807b99a6c5a9

That said, having this separated in two different methods, allow us to not have to override the content of `post_copy`  if we want to add generic metadata_columns behavior.

>The end goal you're trying to achieve is to easily modify a set series of metadata columns after performing a Redshift Copy. This feels like a finer-grained version of post_copy.

Totally agreed with this statement!

luigi/contrib/redshift.py
cabouffard: I've highly an example at the PR level displaying a use-case where we have both cases that run.

cabouffard:
I feel like I haven't sold on your on this one, is that the case @dlstadther ? :)

dlstadther:
Hey @cabouffard 

After thinking on this all morning, I think my hesitations here are more-so `S3CopyToTable` structural issues than feature addition problems.

So, let's continue with getting your metadata feature pushed through. I believe I would like like to see the Mixin moved to rdbms (assuming it's also applicable to Postgres (which i believe is the case)).

Can you also mesh some of your gist together in a docstring for the Mixin? I believe another hesitation I had (prior to your gist) was lack of implementation clarity.

I very much appreciate all your recent PRs! Sorry for the delay addressing them to completion!

cabouffard:
@dlstadther No worries mate! 

I've added a whole bunch of text, some extra bit of documentation and changed some code-part with the suggestions you've provided. 

Is there anything else that you would like for me to do? I'm happy to hear any suggestions!
Thanks! :)

cabouffard:
@dlstadther Hey, what's the latest on this one? :)

dlstadther:
Sorry for my delayed response here @cabouffard . I pretty sure i'm all good here, but let me review real quick once more this afternoon :)

luigi/contrib/rdbms.py
```diff
@@ -27,7 +27,125 @@
 logger = logging.getLogger('luigi-interface')
 
 
-class CopyToTable(luigi.task.MixinNaiveBulkComplete, luigi.Task):
+class _MetadataColumnsMixin(object):
+    """Provide an additinal behavior that adds columns and values to tables
```
dlstadther: typo in `additional`

luigi/contrib/rdbms.py
```diff
@@ -27,7 +27,125 @@
...
+
+    This mixin is used to provide an additional behavior that allow a task to
+    add generic metadata columns to every table created for both PSQL and
+    Redshift.
+
+    Example:
+
+        This is a use example or how this mixin could come handy and how to use
```
dlstadther: Are you trying to say `This is an example use-case of how this mixin could come in handy.`?

luigi/contrib/rdbms.py
```diff
@@ -27,7 +27,125 @@
...
+        it.
+
+        .. code:: python
+
+            class CommonMetaColumnsBehavior(object):
+                def update_report_execution_date_query(self):
+                    query = 'UPDATE {0} ' \
+                            "SET date_param = DATE '{1}' " \
+                            'WHERE date_param IS NULL'.format(self.table, self.date)
```
dlstadther: Can you use the same quote style for these lines? Rather than going back and forth between single and double quote.

luigi/contrib/rdbms.py
```diff
@@ -27,7 +27,125 @@
...
+                def metadata_queries(self):
+                    queries = [self.update_created_tz_query()]
+                    if self.date:
+                        queries.append(self.update_report_execution_date_query())
+
+                    return queries
+
+
+            class RedshiftCopyCSVToTableFromS3(CommonMetaColumnsBehavior, redshift.S3CopyToTable):
+                "We have some business override heres that would only add noise to the
```
dlstadther: `here` (not `heres`)

luigi/contrib/rdbms.py
```diff
@@ -27,7 +27,125 @@
...
+    def enable_metadata_columns(self):
+        return False
+
+    def _add_metadata_columns(self, connection):
+        cursor = connection.cursor()
+
+        for column in self.metadata_columns:
+            if len(column) == 0:
+                logger.info('Unable to infer column information from column {column} for {table}'.format(column=column, table=self.table))
+                break
```
dlstadther: Should this raise an exception?

test/contrib/rdbms_test.py
```diff
@@ -0,0 +1,251 @@
...
+        return 's3://%s/%s' % (BUCKET, KEY)
+
+
+class DummyS3CopyToTableKey(DummyS3CopyToTableBase):
+    aws_access_key_id = AWS_ACCESS_KEY
+    aws_secret_access_key = AWS_SECRET_KEY
+
+
+# Using Redshift as the test bed since Redshift implements RDBMS. Could have
+# opted for PSQL but I'm less familar with that contrib.
```
dlstadther: Perhaps make this comment a docstring to add emphasis. I almost asked why redshift was being used, but then i saw this comment

luigi/contrib/rdbms.py
cabouffard: I was thorned with that decision. Raising an exception would halt the pipeline and other tasks from running while breaking out of the function would simply not add the meta_columns like it would be expected.

Breaking would proceed with the execution of the pipeline but may yield potentially unexpected results, while raising would halt everything from running. 

I think that you bring a fair point and I'll implement an exception as this is how it seems that we're managing that type of cases throughout that contrib.

cabouffard:
@dlstadther Hey ya, we're all green and I've added your suggestions / fixed typos! :)

luigi/contrib/rdbms.py
dlstadther: I agree that neither solution is the most ideal. But i'm more comfortable being explicit in what must be done to achieve happy path. Anything straying from that to raise an exception.

PR #2441, opened 2018-06-10
bugfix bigquery bulk complete failing when argument is a generator

drj42:
## Description

This change fixes a bug that prevents BigQuery tasks from working as ranged tasks (RangeDaily, et. al.) in python 3.

## Motivation and Context

RangeDaily tasks are broken for big query tasks in python3, because the big query bulk_complete method is invoked, with the return value of a call to map, as an argument.  In python 3, map returns a generator, while in python 2 it returns a list.

The MixinBigQueryBulkComplete.bulk_complete method invokes len() on this argument, and so fails if that argument is a generator.

Current unit tests pass for me.

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -420,11 +420,10 @@ class MixinBigQueryBulkComplete(object):
 
     @classmethod
     def bulk_complete(cls, parameter_tuples):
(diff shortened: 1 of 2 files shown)


Tarrasch:
Nice. Thanks for the fix!

PR #2446, opened 2018-06-18
Build task.param_kwargs dynamically.

riga:

In the [current task constructor](https://github.com/spotify/luigi/blob/bda236d0cf41b849c25a82bda02d5ecbebd2bf6a/luigi/task.py#L441), all task parameters at initialization time are stored in `param_kwargs`:

```python
        params = self.get_params()
        param_values = self.get_param_values(params, args, kwargs)

        # Set all values on class instance
        for key, value in param_values:
            setattr(self, key, value)

        # Register kwargs as an attribute on the class. Might be useful
        self.param_kwargs = dict(param_values)  # <--- set once
```

Dynamic changes of parameter values are not reflected in this dict, which is somewhat dangerous. Consider this example using [util.common_params](https://github.com/spotify/luigi/blob/bda236d0cf41b849c25a82bda02d5ecbebd2bf6a/luigi/util.py#L233) (which uses `param_kwargs`):

```python
class MyTask(luigi.Task):

    weight = luigi.FloatParameter(default=0.)
    unit = luigi.ChoiceParameter(default="g", choices=("g", "kg"))

        super(MyTask, self).__init__(*args, **kwargs)

        # store weight internally as grams
        if self.unit == "kg":
            self.weight *= 1000

    def requires(self):
        required_cls = ... # class of the required task
        common_params = luigi.util.common_params(self, required_class)
        #
        # -> common_params["weight"] will have the wrong value when unit is kg!
        #
        return required_cls(**common_params)
```
(ok, this might not be the best example...imagine a situation where a parameter value changes after `Task.__init__`)

`util.common_params` uses the `param_kwargs` attribute which is not synchronized to the actual task parameters at the time it is called.

This PR makes `param_kwargs` a property so its content becomes dynamic. There is a small implication on `task.__eq__` which currently compares the `param_kwargs` dict between two instances. This might be too inefficient with dynamic properties, but I think it should be safe to compare `task.task_id` instead.

## Have you tested this?

No new feature was added, so the existing tests should completely cover changed code.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -437,9 +437,6 @@ class Task(object):
         for key, value in param_values:
             setattr(self, key, value)
 
-        # Register kwargs as an attribute on the class. Might be useful
-        self.param_kwargs = dict(param_values)
-
         self._warn_on_wrong_param_types()
         self.task_id = task_id_str(self.get_task_family(), self.to_str_params(only_significant=True))
         self.__hash = hash(self.task_id)
@@ -451,7 +448,11 @@ class Task(object):
     @property
     def param_args(self):
         warnings.warn("Use of param_args has been deprecated.", DeprecationWarning)
-        return tuple(self.param_kwargs[k] for k, v in self.get_params())
+        return tuple(self.param_kwargs.values())
+
+    @property
+    def param_kwargs(self):
+        return {param: getattr(self, param) for param, _ in self.get_params()}
 
     def initialized(self):
         """
@@ -526,13 +527,10 @@ class Task(object):
         """
         Build a task representation like `MyTask(param1=1.5, param2='5')`
         """
-        params = self.get_params()
-        param_values = self.get_param_values(params, [], self.param_kwargs)
-
         # Build up task id
         repr_parts = []
-        param_objs = dict(params)
-        for param_name, param_value in param_values:
+        param_objs = dict(self.get_params())
+        for param_name, param_value in self.param_kwargs.items():
             if param_objs[param_name].significant:
                 repr_parts.append('%s=%s' % (param_name, param_objs[param_name].serialize(param_value)))
 
@@ -541,7 +539,7 @@ class Task(object):
         return task_str
 
     def __eq__(self, other):
-        return self.__class__ == other.__class__ and self.param_kwargs == other.param_kwargs
+        return self.__class__ == other.__class__ and self.task_id == other.task_id
 
     def complete(self):
         """
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -448,7 +448,8 @@ class Task(object):
     @property
     def param_args(self):
         warnings.warn("Use of param_args has been deprecated.", DeprecationWarning)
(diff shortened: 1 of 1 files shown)


luigi/task.py
```diff
@@ -541,7 +540,7 @@ def __repr__(self):
         return task_str
 
     def __eq__(self, other):
-        return self.__class__ == other.__class__ and self.param_kwargs == other.param_kwargs
+        return self.__class__ == other.__class__ and self.task_id == other.task_id
```
Tarrasch: We probably could do this anyway. :)

riga: Yeah, we even experienced a noticeable speed-up during tree building when dealing with > O(100) tasks as the dict comparison (== length + keys + values comparison) is avoided (depends on the use-case of course).

riga:
> I'm quite sceptic to this. Is it ever necessary to mutate the parameters?

I guess this strongly depends on the use case / personal task setup. In our analysis code, we definitely need to change some parameters (but only on rare occasions) and avoiding this would have some inconvenient implications. But this is just how we use it, so this might not really apply to all users ;)

Independently, I think it's somewhat confusing if a simple `setattr` on a task parameter value such as

```python
task = MyTask(param=1)
task.param = 2
# task.param_kwargs["param"] => 1
```

can lead to sync issues between the `param_kwargs` field (which is used in various places) and the actual parameter value. Instead of making `param_kwargs` dynamic, one could also disable the setter on task parameters.

Tarrasch:
> one could also disable the setter on task parameters.

I would prefer this. What's wrong with simply having computed fields?

riga:
> I would prefer this.

Me too. How about this addition to the base task:

```python
    def __getattribute__(self, attr):
        if attr != "__class__" and isinstance(getattr(self.__class__, attr, None), luigi.Parameter):
            return self.param_kwargs[attr]
        else:
            return super(Task, self).__getattribute__(attr)

    def __setattr__(self, attr, value):
        if isinstance(getattr(self.__class__, attr, None), luigi.Parameter):
            self.param_kwargs[attr] = value
        else:
            super(Task, self).__setattr__(attr, value)
```

Unfortunately this is not possible with `__getattr__` (which I would prefer), since the parameters are registered on the class with the same name, so `__getattr__` would never be invoked for them. But the above implementation is safe nevertheless. One could also drop https://github.com/spotify/luigi/blob/1192003388c7e661962a76102e7668256a67022d/luigi/task.py#L436-L439 then (as well as the current changes in this PR).

Tarrasch:
I think we're over-thinking this. What about a seperate PR with only the `.task_id ==` change? I would love to make the parameter flags unsettable but I don't want to break compatibility now. Also I'm not sure the `__get/setattribute__`` magic is worth it. Maybe just encourage our users to not set the instantiated parameters?

riga:
I think you're right. Especially the `getattribute` is maybe just too much for this case. Will open a new PR with the `task_id` comparison.

PR #2449, opened 2018-06-28
Allow to inject a context manager around TaskProcess.run

ulzha:
## Description
Adds a configurable context manager around TaskProcess.

## Motivation and Context
The intent with this is to enable customizable collection of logs per task (stdout and stderr output from any Task's run() method, including any subprocesses spawned by it), as opposed to having to read through the worker's logs where outputs from all the tasks are interspersed.

Inspired by CaptureOutput from the opensource [capturer](https://github.com/xolox/python-capturer) package.

Such configurable context managers could be useful extension points for "plugins" made for other purposes too. (Related: https://github.com/spotify/luigi/issues/1897)

Has unittests.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -30,6 +30,7 @@ ways between versions. The exception is the exception types and the
 
 import collections
 import getpass
+import importlib
 import logging
 import multiprocessing
 import os
@@ -59,7 +60,7 @@ from luigi.target import Target
 from luigi.task import Task, flatten, getpaths, Config
 from luigi.task_register import TaskClassException
 from luigi.task_status import RUNNING
-from luigi.parameter import FloatParameter, IntParameter, BoolParameter
+from luigi.parameter import BoolParameter, FloatParameter, IntParameter, Parameter
 
 try:
     import simplejson as json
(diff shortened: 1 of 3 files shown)


luigi/worker.py
```diff
@@ -419,6 +441,12 @@ class worker(Config):
     force_multiprocessing = BoolParameter(default=False,
                                           description='If true, use multiprocessing also when '
                                           'running with 1 worker')
+    task_process_context = Parameter(default=None,
```
Tarrasch: Did you consider making this a "list" by making a comma-separate name of classnames?

But I guess users can make their single context-manager itself have multiple ones.

luigi/worker.py
```diff
@@ -258,6 +259,27 @@ def terminate(self):
...
+# Introduce a convention shared for extension points other than TaskProcess?
+# Use https://docs.openstack.org/stevedore?
+class ContextManagedTaskProcess(TaskProcess):
+    def __init__(self, context, *args, **kwargs):
+        super(ContextManagedTaskProcess, self).__init__(*args, **kwargs)
+        self.context = context
+
+    def run(self):
+        if self.context:
+            logger.debug('Instantiating ' + self.context)
```
Tarrasch: Maybe something you can grep easier like `"Importing module for "`?

luigi/worker.py
ulzha: Yes, exactly, that would be the next step I could think of; about to add in a separate PR.

Tarrasch:
Feel free to merge this. Perhaps someone else from Spotify should review too?

NatashaL:
> So cool! Have you tested this in real code yet?

Yes, tested  👍

luigi/worker.py
lgeiger: This will throw a warning because `None` is not allowed in string parameter:
```python-traceback
UserWarning: Parameter "task_process_context" with value "None" is not of type string.
```

bar: > This will throw a warning because `None` is not allowed in string parameter:
> 
> ```
> UserWarning: Parameter "task_process_context" with value "None" is not of type string.
> ```

I wonder why can't it be defaulted to `""` instead, like in
```
task_process_context = Parameter(default="",
```
@ulzha any thoughts?

dlstadther: Alternatively, you can leave the None default and make it an `OptionalParameter`

PR #2454, opened 2018-07-10
ChoiceParameter should raise an error when passed invalid choice

jotinha:
## Description
Validates `ChoiceParameter` when initializing a task by moving the logic from `.parse` to `.normalize`. Currently this validation only runs when calling from the command line, when the method `.parse` is called. Moving it to `.normalize` ensure it always runs

## Motivation and Context
I was having an issue where I was requiring a task with an invalid ChoiceParameter but no ValueError was raising at task init time. 

```
class Foo(luigi.Task):
     args = luigi.ChoiceParameter(var_type=str, choices=["1", "2", "3"])

Foo(args="bar") #should raise an Error
```

Added tests

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -1201,8 +1201,11 @@ class ChoiceParameter(Parameter):
 
     def parse(self, s):
         var = self._var_type(s)
+        return self.normalize(var)
+
+    def normalize(self, var):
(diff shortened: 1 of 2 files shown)


Tarrasch:
Nice fix. Thanks!

Issue #2455, opened 2018-07-13
[contrib.s3] Can't use S3Target outside luigi Task


michcio1234:
Using `luigi.contrib.s3.S3Target` outside of `luigi.Task` results in `AttributeError: module 'boto3' has no attribute 's3'`.

Consider file `good.py`:
```python
import luigi
from luigi.contrib.s3 import S3Target

class Task(luigi.Task):
    def output(self):
        return S3Target('s3://bucket/file')

    def run(self):
        self.output().open('w').close()

if __name__ == '__main__':
    luigi.build([Task()], local_scheduler=True)
```

and a file `bad.py`:
```python
from luigi.contrib.s3 import S3Target
# import boto3.s3.transfer  # <-- uncomment this and it starts working

S3Target('s3://bucket/file').open('w').close()
```

If you don't have AWS credentials configured, running `good.py` results in an expected error (`Forbidden`). 

I would expect that running `bad.py` will have the same result, but instead that's what I get:
```
Traceback (most recent call last):
  File "bad.py", line 4, in <module>
    S3Target('s3://bucket/file').open('w').close()
  File "/home/mdank/anaconda3/envs/s3-bug/lib/python3.7/site-packages/luigi/target.py", line 330, in close
    self.move_to_final_destination()
  File "/home/mdank/anaconda3/envs/s3-bug/lib/python3.7/site-packages/luigi/contrib/s3.py", line 558, in move_to_final_destination
    self.tmp_path, self.path, **self.s3_options)
  File "/home/mdank/anaconda3/envs/s3-bug/lib/python3.7/site-packages/luigi/contrib/s3.py", line 289, in put_multipart
    transfer_config = boto3.s3.transfer.TransferConfig(
AttributeError: module 'boto3' has no attribute 's3'
```
However, if you include `import boto3.s3.transfer` in `bad.py`, it starts working as expected. 

Relevant output of `pip freeze`:
```
boto3==1.7.57
botocore==1.10.57
luigi==2.7.6
s3transfer==0.1.13
tornado==4.5.3
```

michcio1234:
Probably should mention @ouanixi.

ouanixi:
Thanks for flagging this @michcio1234 

It looks like this is what's happening here
https://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-submodules-are-added-to-the-package-namespace-trap

PR #2458, opened 2018-07-13
Fix transfer config import

ouanixi:
As per https://github.com/spotify/luigi/issues/2455
It looks as though I fell in this trap.
I tried this locally and things are working fine


ouanixi:
@dlstadther this should be a simple one. 
Sorry about this

dlstadther:
What happened with all the commits?

dlstadther:
I'll just squash them. Thanks!

PR #2461, opened 2018-07-15
Fix scheduler_visualisation_test

Gr1f0n6x:
## Description

Replace luigi.Parameter by luigi.IntParmeter to avoid messages like:
UserWarning: Parameter "product" with value "1" is not of type string.

Add dynamic complete method for some tasks to avoid messages like:
Task scheduler_visualisation_test.FailingTask(task_id=8) without outputs has no custom complete() method

## Motivation and Context

This test provides a lot of redundant messages in test log which can be removed 
 

Yes, i ran this test locally and result was similar to result before my changes


test/scheduler_visualisation_test.py
```diff
@@ -585,7 +638,11 @@ class X(luigi.Task):
 
     def test_worker_list_disabled_worker(self):
         class X(luigi.Task):
-            pass
+            def run(self):
```
Tarrasch: Can you use the RunOnceTask instead of doing this charade everywhere? https://github.com/spotify/luigi/search?q=RunOnceTask&unscoped_q=RunOnceTask

Gr1f0n6x: Why i didn't notice this class before =\

test/scheduler_visualisation_test.py
```diff
@@ -535,13 +521,10 @@ def complete(self):
         self.assertEqual(1, worker['workers'])
 
     def test_worker_list_pending_uniques(self):
-        class X(luigi.Task):
-
-            def complete(self):
-                return False
+        class X(RunOnceTask):
```
Tarrasch: This isn't the same isn't it?

test/task_history_test.py
```diff
@@ -46,7 +46,7 @@ def test_run(self):
         th = SimpleTaskHistory()
         sch = luigi.scheduler.Scheduler(task_history_impl=th)
         with luigi.worker.Worker(scheduler=sch) as w:
-            class MyTask(luigi.Task):
+            class MyTask(RunOnceTask):
```
Tarrasch: You don't need to change these things...

test/task_test.py
```diff
@@ -211,12 +209,9 @@ def run(self):
 
     def test_externalize_same_id_with_task_namespace(self):
         # Dependent on the new behavior from spotify/luigi#1953
-        class MyTask(luigi.Task):
+        class MyTask(RunOnceTask):
             task_namespace = "something.domething"
 
-            def run(self):
```
Tarrasch: Again this isn't equivalent and you didn't need to change this...

However in this case I guess we have a more defined task so this is okay too.

PR #2466, opened 2018-07-20
Add Python 3.7 compatibility

dlstadther:
## Description

Python 3.7.0 has been released. Time to check for Luigi compatibility.

## Motivation and Context

Let's try to keep up with most recent Python versions!

Letting Travis do this :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ services:
   - elasticsearch
   - mysql
   - docker
+  - postgresql
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,14 @@ env:
       # - TOXENV=visualiser
       # Disabling this test because of intermittent failures :-/
 
+sudo: false
+
 # Python 3.5 has to go here until Travis adds it to the default build images.
 # https://github.com/travis-ci/travis-ci/issues/4794#issuecomment-143758799
+#
+# Python 3.7 needs to have its dist/sudo flags set in matrix so they don't
+# break the other python builds.
+# https://github.com/travis-ci/travis-ci/issues/9815
 matrix:
   include:
     - python: 3.5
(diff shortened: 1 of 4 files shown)


dlstadther:
If successful, will need to update README

dlstadther:
At least one error is caused by: https://github.com/spulec/moto/issues/1706

dlstadther:
With regard to the moto error, we'll need to update our moto version with the Python 3.7 compatible version.

We're currently testing against `moto<1.0`. Moto 1.0.0 [denotes breaking changes for decorators](https://github.com/spulec/moto/blob/master/CHANGELOG.md#100).

So an additional PR may need to be submitted to update our moto test dependency. I'll submit a naive WIP one to see how badly things break.

dlstadther:
Thanks so much @dalejung ! I hadn't found the time to resolve the postgres issue. I've copied your build changes here.

dalejung:
:+1: Had to wrangle with travis updating some personal libs. Happy the info was useful elsewhere.

dlstadther:
@ulzha  How's this look to you?

At some point, we should probably do something about our test speed. We're now exceeding 30m to find out if everything passes. What are your (and @Tarrasch 's) thoughts on cutting down to Py2.7, 3.7 explicit support? I understand many people may be using 34/35/36, but testing `nonhdfs` for each gets resource hungry.

honnix:
Hard decision, but 30m is way too much. I wish I could find some stats of which Python versions are more popular than the other. :/

honnix:
I guess my above answer is not helping at all. :D How about we drop 34/35 first? Give 36 one more year?

dlstadther:
I'd support dropping tests for 34/35. If we go that route, i'd suggest updating our README to denote previous support for those versions, but that they are no longer included in the the automated test suite.

honnix:
Sound like a plan. Let's go for it. BTW, in readme we say 3.3 still, lol.

dlstadther:
Cool. I push those changes here shortly. Then this will be good to merge. Thanks @honnix !

honnix:
👍

honnix:
Is this worth a minor version bump? I'm drafting and releasing `2.7.9` and this PR can go to `2.8.0` perhaps.

dlstadther:
If we're dropping explicit (tested) support for Python versions, i'd say this deserves `2.8.0`.

Thanks for taking care of `2.7.9`!

honnix:
:shipit:

PR #2468, opened 2018-07-23
Check for default before warning of "task_process_context" as None

HyunTruth:
## Description

Add a check for default in `Parameter._warn_on_wrong_param_type` before giving out warning by adding an additional if clause to skip the cases of where the parameter is default
```
if param_value == self._default:
    return
```

## Motivation and Context
Currently, the parameter `task_process_context` is being passed as None for default.
However, in the `Parameter._warn_on_wrong_param_type` method, check for string is made and 
if the `param_value` is not a string, the warning is raised.
Therefore, as I was not giving any parameter input for `task_process_context`, every job that I ran displayed this warning, and I thought it would be a good idea to check for default only for the `Parameter` class, to resolve this issue.

I ran my jobs with this code and it works fine for me - without the warning.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -258,7 +258,7 @@ class Parameter(object):
         if self.__class__ != Parameter:
             return
         if param_value == self._default:
-            return
+            return        
         if not isinstance(param_value, six.string_types):
             warnings.warn('Parameter "{}" with value "{}" is not of type string.'.format(param_name, param_value))
 
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -257,6 +257,8 @@ class Parameter(object):
     def _warn_on_wrong_param_type(self, param_name, param_value):
         if self.__class__ != Parameter:
             return
+        if param_value == self._default:
+            return
         if not isinstance(param_value, six.string_types):
             warnings.warn('Parameter "{}" with value "{}" is not of type string.'.format(param_name, param_value))
 


dlstadther:
Is it easy enough to convert `task_process_context` to an [`OptionalParameter`](https://github.com/spotify/luigi/pull/2295) such that `None` is a valid default?

Actually, i believe @ulzha 's mentioned the intention of including [a followup PR changing the behavior slightly](https://github.com/spotify/luigi/pull/2449#discussion_r201366043).

HyunTruth:
Yeah, that sounds like a better option - I wasn't aware of the addition of an `OptionalParameter ` class.
If @ulzha intends to have a follow up which is more consistent with the master luigi codebase, then it is fine. Thanks.

gardenunez:
Is there any issue or PR in progress to handle the OptionalParameter approach?

dlstadther:
@gardenunez seems like this PR has been abandoned. I believe we were waiting on @ulzha to submit a follow-up PR (per a mention in his earlier PR), but nothing has been heard.

Regardless of a possible future PR, i'd like to see this Parameter warning be put to rest in the near term.

evictor:
@dlstadther Do you need a new PR w/ the change? Also wondering when we can get this warning to go away... I can send a PR if you like

dlstadther:
Given this issue has not already been resolved, yes.

evictor:
Ugh I'm confused now actually. I see it has already been changed to `OptionalParameter` in master. Sorry for added confusion

Issue #2473, opened 2018-07-24
RedshiftQuery doesn't set port 


gregroberts:
by default, when tasks from luigi.contrib.redshift have `RedshiftTarget` as an output, they do not set the port. e.g. in RedshiftQuery:

`    

    def output(self):
        """
        Returns a RedshiftTarget representing the executed query.
        Normally you don't override this.
        """
        return RedshiftTarget(
            host=self.host,
            database=self.database,
            user=self.user,
            password=self.password,
            table=self.table,
            update_id=self.update_id
        )
`

RedshiftTarget is based on PostgresTarget.
So if your redshift host doesn't include the port, these Tasks will fail, because they will default back to 5432. 
so if your Redshift instance is running on the default port, your tasks will fail until you explicitly override the output method, passing the port to the RedshiftTarget constructor.

Not a massive issue, for sure, but for consistencies sake, this ought to be changed.

dlstadther:
Can you not just specific the port with `<host>:<port>` and it be extracted within [PostgresTarget's __init__()](https://github.com/spotify/luigi/blob/master/luigi/contrib/postgres.py#L125-L129)?

gregroberts:
Thats true. I guess I'm just being pedantic because it means you have to
format the arguments differently for different databases.

gregroberts:
By default surely the preferable behaviour is for the task to behave
equivalent to PostgresQuery, and just using the default Redshift port?

dlstadther:
If you have a non-breaking code recommendation to enable the optional ability to provide explicit port, i'll review!

gregroberts:
Will get on that tomorrow :)

PR #2474, opened 2018-07-25
Added default port behaviour for Redshift

gregroberts:
## Description
RedshiftTarget is a subclass of PostgresTarget, for obvious reasons. 

In PostgresTarget, if we do not provide a port, we fallback to 5432, the default Postgres port.
In RedshiftTarget, if we do not provide a port, we still fallback to 5432, even though the default Redshift port is 5439.

This PR fixes this by defining the default port as a static class variable on PostgresTarget, as opposed to a default argument on __init__. This allows RedshiftTarget to override this default, without changing any other behaviour.

## Motivation and Context
From https://github.com/spotify/luigi/issues/2473
I only noticed this behaviour when I started using Redshift. I had been supplying the port in PostgresQuery Tasks, even though this never actually got passed to PostgresTarget! So, when I switched to doing the same with Redshift, nothing worked, because RedshiftTarget wasn't being supplied the port by RedshiftQuery, so RedshiftTarget was defaulting to 5432.

I looked 'very hard' at any code which is affected by the change, and can't see any reason it would fail.

### DIFF
--- a/luigi/contrib/postgres.py
+++ b/luigi/contrib/postgres.py
@@ -106,7 +106,7 @@ class PostgresTarget(luigi.Target):
     """
     marker_table = luigi.configuration.get_config().get('postgres', 'marker-table', 'table_updates')
 
-    #if not supplied, fall back to default Postgres port 
+    # if not supplied, fall back to default Postgres port
     DEFAULT_DB_PORT = 5432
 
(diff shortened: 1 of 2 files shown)


dlstadther:
Thanks @gregroberts !

PR #2475, opened 2018-07-29
Multiple requires

### LINKED ISSUE #1934: Feature request: list of tasks for @requires
Hi all,

Would it be possible to pass a list of tasks to `@requires` please?

If you know they're going to have individual param signatures but need to share the majority of the parameters, I'd like to be able to do

```
from __future__ import print_function
import luigi
from luigi.utils import inherits, requires


class A(luigi.ExternalTask):
    """First ext class"""
    par_a = luigi.Parameter(default="A")

    def run(self):
        print(self.par_1, self.par_2, self.par_a)

    def output(self):
        return luigi.LocalTarget('/tmp/log-{t.param_a}'.format(t=self))


class B(luigi.ExternalTask):
    """Second ext class"""
    par_b = luigi.Parameter(default="B")

    def run(self):
            print(self.par_1, self.par_2, self.par_b)

    def output(self):
        return luigi.LocalTarget('/tmp/log-{t.param_b}'.format(t=self))


@requires([A, B])
class C(luigi.Task):
    param_c = luigi.Parameter(default="c")

    def run(self):
        print(self.par_1, self.par_2, self.c)

    def output(self):
        return luigi.LocalTarget('/tmp/log-{t.param_c}'.format(t=self))

```

what do you think?

A

StasDeep:
## Description
Now `requires` decorator can take multiple tasks to require, as well as `inherits` decorator (no need to stack them now). But `clone_parent` method becomes ambiguous in cases when there are 2+ tasks to inherit. I've kept it as is to not break anything and added `clone_parents` method, that is used inside `requires` decorator.

## Motivation and Context
Resolves #1934.

I have included new unit tests.

### DIFF
--- a/luigi/util.py
+++ b/luigi/util.py
@@ -250,7 +250,9 @@ def common_params(task_instance, task_cls):
 
 class inherits(object):
     """
-    Task inheritance. Supports multiple tasks inheritance.
+    Task inheritance.
+
+    *New after Luigi 2.7.6:* multiple arguments support.
 
     Usage:
 
@@ -315,6 +317,9 @@ class inherits(object):
 class requires(object):
     """
     Same as :class:`~luigi.util.inherits`, but also auto-defines the requires method.
+
+    *New after Luigi 2.7.6:* multiple arguments support.
+
     """
 
     def __init__(self, *tasks_to_require):
@@ -331,10 +336,7 @@ class requires(object):
         # If only one task is required, this single task is returned.
         # Otherwise, list of tasks is returned
         def requires(_self):
-            if len(self.tasks_to_require) == 1:
-                return _self.clone_parent()
-
-            return _self.clone_parents()
+            return _self.clone_parent() if len(self.tasks_to_require) == 1 else _self.clone_parents()
         task_that_requires.requires = requires
 
         return task_that_requires
--- a/luigi/util.py
+++ b/luigi/util.py
@@ -52,7 +52,7 @@ parameter handling can spiral out of control.  Each downstream task becomes
 more burdensome than the last.  Refactoring becomes more difficult.  There
 are several ways one might try and avoid the problem.
 
-**Approach 1**:  Parameters via command line or config instead of ``requires``.
+**Approach 1**:  Parameters via command line or config instead of :func:`~luigi.task.Task.requires`.
 
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


dlstadther:
I am not familiar with these features. I'll try to look into them this week and provide you a review afterward. Thanks for the submission!

luigi/util.py
```diff
@@ -249,61 +250,91 @@ def common_params(task_instance, task_cls):
...
-        # Modify task_that_requres by adding methods
+        # Modify task_that_requires by adding requires method.
+        # If only one task is required, this single task is returned.
+        # Otherwise, list of tasks is returned
         def requires(_self):
-            return _self.clone_parent()
+            if len(self.tasks_to_require) == 1:
+                return _self.clone_parent()
+
+            return _self.clone_parents()
```
dlstadther: Possibly make this a one-liner (equally clear, but prettier)

```python
return _self.clone_parent() if len(self.tasks_to_require) == 1 else _self.clone_parents()
```

luigi/util.py
```diff
@@ -249,61 +250,88 @@ def common_params(task_instance, task_cls):
...
 
+            def run(self):
+               print self.m # this will be defined
+               # ...
+
+        @inherits(AnotherTask, YetAnotherTask):
+        class MySecondTask(luigi.Task):
+            def requires(self):
+               return self.clone_parents()
+
```
Tarrasch: Can you add a 'note' that this was added after version 2.7.6?

test/decorator_test.py
```diff
@@ -91,11 +97,24 @@ def test_overwriting_defaults(self):
...
+        self.assertEqual(self.e_stacked.param1, self.b.param1)
+        self.assertEqual(self.e_stacked.param2, self.b.param2)
+
+    def test_empty_inheritance(self):
+        def create_task_with_empty_inheritance():
+            @inherits()
+            class shouldfail(luigi.Task):
+                pass
+
+        self.assertRaises(TypeError, create_task_with_empty_inheritance)
```
Tarrasch: Can you replace this with the with decorator for assertRaises?

test/decorator_test.py
```diff
@@ -226,53 +245,79 @@ def test_wrong_common_params_order(self):
...
+        self.assertEqual(y.m, 56)
+        self.assertEqual(y.n, 42)
+
+    def test_empty_requires(self):
+        def create_task_with_empty_requires():
+            @requires()
+            class shouldfail(luigi.Task):
+                pass
+
+        self.assertRaises(TypeError, create_task_with_empty_requires)
```
Tarrasch: Same here.

Tarrasch:
Huge thanks for updating the documentation as well!

dlstadther:
Interestingly, I just came across an attempt at this same intention almost 2 years ago in #1845

StasDeep:
@Tarrasch could you please explain how I should test it in production?

dlstadther:
I believe he just means use your own code for your own pipelines and verify it functions the way you expect.

StasDeep:
@Tarrasch @dlstadther I believe this is ready to be merged.

dlstadther:
Did production tests of this code function as expected in your "live" pipelines @StasDeep ?

StasDeep:
@dlstadther yes, everything works as expected.

PR #2476, opened 2018-07-29
Replace luigi.Task by RunOnceTask in scheduler_visualisation_test

Gr1f0n6x:
## Description

Replace luigi.Parameter by luigi.IntParmeter to avoid messages like:
UserWarning: Parameter "product" with value "1" is not of type string.

Add dynamic complete method for some tasks to avoid messages like:
Task scheduler_visualisation_test.FailingTask(task_id=8) without outputs has no custom complete() method
## Motivation and Context

This test provides a lot of redundant messages in test log which can be removed

Ran tests locally

p.s. probably, in one day, i will be able to do this with 1st attempt ^_^

### DIFF
--- a/test/scheduler_visualisation_test.py
+++ b/test/scheduler_visualisation_test.py
@@ -19,7 +19,7 @@ from __future__ import division
 import os
 import tempfile
 import time
-from helpers import unittest
+from helpers import unittest, RunOnceTask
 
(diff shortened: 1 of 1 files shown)


Tarrasch:
Thanks! I changed the commit message to emphasize this is about tests only.

PR #2477, opened 2018-07-30
Fix race condition

### LINKED ISSUE #2357: race condition on lock :(
I think there is a race condition on 
**luigi/lock.py", line 93, in acquire_for**

```python
# Check if there is a pid file corresponding to this name
if not os.path.exists(pid_dir):
  os.mkdir(pid_dir)
```
between `if not [...]` and `mkdir` another process could have created the pid_dir resulting  in an unhandled OSError exception.

I think that ignore the exception if the folder exists will be enough, or perhaps just create the folder and say sorry if it was just there...

StasDeep:
## Description
Fix race condition in `luigi.lock.acquire_for`.

## Motivation and Context
Fixes #2357.

Not tested, as it's pretty hard to reproduce.

### DIFF
--- a/luigi/lock.py
+++ b/luigi/lock.py
@@ -21,6 +21,7 @@ This prevents multiple identical workflows to be launched simultaneously.
 """
 from __future__ import print_function
 
+import errno
 import hashlib
 import os
 import sys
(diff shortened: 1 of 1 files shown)


luigi/lock.py
```diff
@@ -104,8 +105,14 @@ def acquire_for(pid_dir, num_available=1, kill_signal=None):
...
-        os.mkdir(pid_dir)
-        os.chmod(pid_dir, 0o777)
+        # Try/except for handling race condition
+        try:
+            os.mkdir(pid_dir)
+            os.chmod(pid_dir, 0o777)
+        except OSError as exc:
+            if exc.errno != errno.EEXIST:
+                raise
+            pass
```
dlstadther: Why have the `if not os.path.exists(pid_dir)` if this try/except will handle the case when it already exists? While still correct to have both, I believe the Pythonic way is just to try and allow to fail. The try/except should handle both the normal case when the `pid_dir` already exists AND when this race condition occurs.

StasDeep: @dlstadther done. It also resolved codecov checks.

Tarrasch:
Is it worth while to add a unit test?

StasDeep:
@Tarrasch I'm afraid [there's no good general way to trigger a race condition in testing](https://softwareengineering.stackexchange.com/questions/196105/testing-multi-threaded-race-conditions).

Tarrasch:
Fair enough. Thanks for this fix!

PR #2478, opened 2018-07-30
Fix attribute forwarding for tasks with dynamic dependencies

riga:
## Description

This PR fixes the forwarding of attributes (mostly callbacks such as `set_status_message`) from `TaskProcess`es to running `Task`s that yield dynamic dependencies within their `run()` method.

## Motivation and Context

Currently, the attributes are forwarded and reset like this:

https://github.com/spotify/luigi/blob/7d2c5574cb53106044a25afdddf151165471a741/luigi/worker.py#L137-L146

However, when the `run` method returns a generator, attributes are reset again before the yielded dependencies are handled (right below the referenced code). Actually, this should happen right before `_run_get_new_deps` returns. I added a context-managed method that handles the attributes so I could place the (only) call to `_run_get_new_deps` in that context.

## Have you tested this?

I added a new test case that checks if attributes are forwarded for tasks that don't yield dependencies and for those who do. The `TaskForwardedAttributesTest.test_yielding_task` test would actually fail in the current implementation, but passes with the proposed change.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -257,10 +257,8 @@ class TaskProcess(multiprocessing.Process):
         # forward configured attributes to the task
         for reporter_attr, task_attr in six.iteritems(self.forward_reporter_attributes):
             setattr(self.task, task_attr, getattr(self.status_reporter, reporter_attr))
-
         try:
             yield self
-
         finally:
             # reset attributes again
             for reporter_attr, task_attr in six.iteritems(self.forward_reporter_attributes):
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -37,6 +37,7 @@ import os
 import signal
 import subprocess
 import sys
+import contextlib
 
 try:
     import Queue
@@ -135,15 +136,7 @@ class TaskProcess(multiprocessing.Process):
         self.check_unfulfilled_deps = check_unfulfilled_deps
 
     def _run_get_new_deps(self):
-        # forward some attributes before running
-        for reporter_attr, task_attr in six.iteritems(self.forward_reporter_attributes):
-            setattr(self.task, task_attr, getattr(self.status_reporter, reporter_attr))
-
         task_gen = self.task.run()
-
-        # reset attributes again
-        for reporter_attr, task_attr in six.iteritems(self.forward_reporter_attributes):
-            setattr(self.task, task_attr, None)
 
         if not isinstance(task_gen, types.GeneratorType):
             return None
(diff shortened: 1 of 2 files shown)


test/task_forwarded_attributes_test.py
```diff
@@ -0,0 +1,76 @@
...
+        sch = luigi.scheduler.Scheduler()
+        with luigi.worker.Worker(scheduler=sch) as w:
+            w.add(task)
+            w.run()
+        return task
+
+    def test_non_yielding_task(self):
+        task = self.run_task(NonYieldingTask())
+
+        self.assertEqual(task.attributes_while_running, forwarded_attributes)
```
dlstadther: Am i correct that this test passes before AND after the code changes?

test/task_forwarded_attributes_test.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def test_yielding_task(self):
+        task = self.run_task(YieldingTask())
+
+        self.assertEqual(task.attributes_before_yield, forwarded_attributes)
+        self.assertEqual(task.attributes_after_yield, forwarded_attributes)
```
dlstadther: But this one would fail before this code change?

test/task_forwarded_attributes_test.py
riga: Yes, that's correct.

test/task_forwarded_attributes_test.py
riga: Exactly.

test/task_forwarded_attributes_test.py
```diff
@@ -0,0 +1,76 @@
...
+class YieldingTask(NonYieldingTask):
+
+    def run(self):
+        self.attributes_before_yield = self.gather_forwarded_attributes()
+
+        yield RunOnceTask()
+
+        self.attributes_after_yield = self.gather_forwarded_attributes()
+
+        RunOnceTask.run(self)
```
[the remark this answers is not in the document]
riga: ✔︎

test/task_forwarded_attributes_test.py
```diff
@@ -0,0 +1,76 @@
...
+#
+
+from helpers import LuigiTestCase, RunOnceTask
+
+import luigi
+import luigi.scheduler
+import luigi.worker
+
+
+forwarded_attributes = set(luigi.worker.TaskProcess.forward_reporter_attributes.values())
```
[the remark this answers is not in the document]
riga: ✔︎

luigi/worker.py
```diff
@@ -258,6 +252,20 @@ def terminate(self):
...
 
+    @contextlib.contextmanager
+    def _forward_attributes(self):
+        # forward configured attributes to the task
+        for reporter_attr, task_attr in six.iteritems(self.forward_reporter_attributes):
+            setattr(self.task, task_attr, getattr(self.status_reporter, reporter_attr))
+
+        try:
+            yield self
+
```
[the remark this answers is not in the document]
riga: ✔︎

riga:
Anything else I can do here? =)

dlstadther:
Nah, LGTM. All reviewer comments have been addressed.

Issue #2215, opened 2017-08-23
TypeError: a bytes-like object is required, not 'str' with SGEJobTask


sdhutchins:
I'm trying to run the following code as a test.

```python
import logging
import os
import luigi
from luigi.contrib.sge import SGEJobTask

logger = logging.getLogger('luigi-interface')


class TestJobTask(SGEJobTask):

    SGEJobTask.parallel_env = 'pbs'
    SGEJobTask.shared_tmp_dir = '/work5/r2295/tmp'
    SGEJobTask.job_name = 'sdhtest'

    i = luigi.Parameter()

    def work(self):
        logger.info('Running test job...')
        with open(self.output().path, 'wb') as outfile:
            outfile.write(b'This is a test!!!')

    def output(self):
        return luigi.LocalTarget(os.path.join('/work5/r2295/tests/sge_tests',
                                              'testfile_' + str(self.i)))


if __name__ == '__main__':
    tasks = [TestJobTask(i=str(i), n_cpu=i+1) for i in range(3)]
    luigi.build(tasks, local_scheduler=True, workers=3)
```

I continue to get the following error:

```
DEBUG: Checking if TestJobTask(i=0) is complete
INFO: Informed scheduler that task   TestJobTask_0_d010c9c495   has status   PENDING
DEBUG: Checking if TestJobTask(i=1) is complete
INFO: Informed scheduler that task   TestJobTask_1_64a8d99e05   has status   PENDING
DEBUG: Checking if TestJobTask(i=2) is complete
INFO: Informed scheduler that task   TestJobTask_2_a9facda7c5   has status   PENDING
INFO: Done scheduling tasks
INFO: Running Worker with 3 processes
DEBUG: Asking scheduler for work...
DEBUG: Pending tasks: 3
DEBUG: Asking scheduler for work...
DEBUG: Pending tasks: 2
INFO: [pid 9313] Worker Worker(salt=162029851, workers=3, host=sequoia, username=r2295, pid=9310) running   TestJobTask(i=0)
INFO: Tmp dir: /work5/r2295/tmp/TestJobTask_0_d010c9c495-3c3a12a51392d284
DEBUG: Asking scheduler for work...
DEBUG: Pending tasks: 1
INFO: [pid 9314] Worker Worker(salt=162029851, workers=3, host=sequoia, username=r2295, pid=9310) running   TestJobTask(i=1)
[...]

PR #2480, opened 2018-08-04
Fix bytes/str confusion in luigi.contrib.sge

mivade:
This PR addresses #2215.

### DIFF
--- a/luigi/contrib/sge.py
+++ b/luigi/contrib/sge.py
@@ -276,11 +276,13 @@ class SGEJobTask(luigi.Task):
             self.job_file = os.path.join(out_dir, 'job-instance.pickle')
             if self.__module__ == '__main__':
                 d = pickle.dumps(self)
-                module_name = os.path.basename(sys.argv[0]).rsplit('.', 1)[0]
-                d = d.replace('(c__main__', "(c" + module_name)
-                open(self.job_file, "w").write(d)
+                module_name = os.path.basename(sys.argv[0]).rsplit('.', 1)[0].encode()
+                d = d.replace(b'(c__main__', b"(c" + module_name)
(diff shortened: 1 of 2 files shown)


sdhutchins:
Awesome. @mivade. I had been lazy and neglected all of the changes I made in that regard (I had to do this for a version I have working in a different project). If you want any help with the failed checks, I can help.

mivade:
I'd be happy to add a test if someone has suggestions on how exactly to test this. Looking through the existing tests, it doesn't look like the affected code was ever actually tested in the first place. At any rate, there still appears to be something wrong because when I try to run a simple test task, I just get

```pytb
AttributeError: Can't get attribute 'SimpleTask' on <module '__main__' from '/home1/depalati/src/luigi/luigi/contrib/sge_runner.py'
```

dlstadther:
I have no experience with sge so i'm just going to tag the various contributors to this package :)

@brcopeland @ivannotes and @jfeala

leonqli:
> I'd be happy to add a test if someone has suggestions on how exactly to test this. Looking through the existing tests, it doesn't look like the affected code was ever actually tested in the first place. At any rate, there still appears to be something wrong because when I try to run a simple test task, I just get
> 
> ```
> AttributeError: Can't get attribute 'SimpleTask' on <module '__main__' from '/home1/depalati/src/luigi/luigi/contrib/sge_runner.py'
> ```
> 
> and despite this, luigi thinks that everything ran correctly.

How is the latest status? I still have the same issue...

mivade:
I'm no longer working with SGE and never got feedback on what kinds of tests should be added. I'll leave this open until someone takes it over or submits a new PR.

PR #680, opened 2015-01-27
Fixed various things that Landscape was complaining about

erikbern:

PR #2482, opened 2018-08-07
S3 client refactor

gardenunez:
## Description
Small refactor over S3Client and s3_test module.

## Motivation and Context
The motivation came because we are using Luigi actively in my organization and we recently use this module and I saw some small things I can update as a starting point to contribute to the project and get more deeply in near future. 

Since I haven't change any behaviour I didn't add new tests but I checked that all jobs passed in my custom TravisCI configuration.

### DIFF
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -249,7 +249,7 @@ class S3Client(FileSystem):
         :param destination_s3_path: URL for target S3 location
         :param kwargs: Keyword arguments are passed to the boto function `put_object`
         """
-        self._check_deprecated_argument(kwargs)
+        self._check_deprecated_argument(**kwargs)
 
         # put the file
         self.put_multipart(local_path, destination_s3_path, **kwargs)
@@ -257,11 +257,11 @@ class S3Client(FileSystem):
     def put_string(self, content, destination_s3_path, **kwargs):
         """
         Put a string to an S3 path.
-        :param content: Object data
+        :param content: Data str
         :param destination_s3_path: URL for target S3 location
         :param kwargs: Keyword arguments are passed to the boto3 function `put_object`
         """
-        self._check_deprecated_argument(kwargs)
+        self._check_deprecated_argument(**kwargs)
         (bucket, key) = self._path_to_bucket_and_key(destination_s3_path)
 
         # validate the bucket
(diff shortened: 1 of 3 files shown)


luigi/contrib/s3.py
```diff
@@ -500,17 +501,31 @@ def _get_s3_config(self, key=None):
...
+    def _is_root(key):
         return (len(key) == 0) or (key == '/')
 
-    def _add_path_delimiter(self, key):
+    @staticmethod
+    def _add_path_delimiter(key):
         return key if key[-1:] == '/' or key == '' else key + '/'
 
+    @staticmethod
+    def _check_deprecated_argument(arguments):
```
dlstadther: Wouldn't it be best to make this `_check_deprecated_argument(**kwargs)`? to keep with Python practice

luigi/contrib/s3.py
```diff
@@ -243,24 +245,23 @@ def get_key(self, path):
...
-                'encrypt_key deprecated in boto3. Please refer to boto3 documentation for encryption details.')
+        self._check_deprecated_argument(kwargs)
 
         # put the file
         self.put_multipart(local_path, destination_s3_path, **kwargs)
 
     def put_string(self, content, destination_s3_path, **kwargs):
         """
         Put a string to an S3 path.
+        :param content: Object data
```
dlstadther: it should already be clear by the method name, but could we `:param content: Data str` or something. `Object data` just makes me hesitate.

luigi/contrib/s3.py
```diff
@@ -188,7 +188,7 @@ def exists(self, path):
         logger.debug('Path %s does not exist', path)
         return False
 
-    def remove(self, path, recursive=True):
+    def remove(self, path, recursive=True, skip_trash=True):
```
dlstadther: What is this?

dlstadther:
Restarting build due to Travis connection error

luigi/contrib/s3.py
gardenunez: the parent function in FileSystem class has this attribute. Thought will be good to keep compatibility.

dlstadther: This looks to be something added to FileSystem for hadoop/hive/webhdfs [long ago](https://github.com/spotify/luigi/pull/680). It serves no purpose for S3 given that there is no "trash" for s3.

It's also worth noting that [LocalFileSystem did not include this arg](https://github.com/spotify/luigi/blob/master/luigi/local_target.py#L94) either.

For that, i'm going to say to exclude `skip_trash` for S3Client's `remove()`.

But nice eye for checking it's parent!

gardenunez:
Agree on giving it a couple of days. Thank you for the review.

dlstadther:
Thanks @gardenunez

PR #2486, opened 2018-08-10
Make logging of RPC retries configurable

ulzha:
## Motivation and Context
Logging of intermittent RPC exceptions and retries as they happen makes logs spammy and obscures other errors, so we want a way to disable it. If the exceptions persist, the final RPCError does log the last exception anyway.

Especially when long running (hanging, even) tasks are considered, with periods of silence in the logs, any intermittent "Failed connecting to remote scheduler" logging from KeepAliveThread looks very flashy and misleads troubleshooting often, in our experience.

The health of communication channels (presence of intermittent failures) probably should be easy to  monitor through dedicated tooling, as opposed to logging it interspersed with task logs.

Ought to be covered with unit tests.

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -116,6 +116,7 @@ class RemoteScheduler(object):
 
         self._rpc_retry_attempts = config.getint('core', 'rpc-retry-attempts', 3)
         self._rpc_retry_wait = config.getint('core', 'rpc-retry-wait', 30)
+        self._log_exceptions = config.getboolean('core', 'log-exceptions', True)
 
         if HAS_REQUESTS:
             self._fetcher = RequestsFetcher(requests.Session())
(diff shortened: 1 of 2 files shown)


luigi/rpc.py
```diff
@@ -116,31 +116,34 @@ def __init__(self, url='http://localhost:8082/', connect_timeout=None):
 
         self._rpc_retry_attempts = config.getint('core', 'rpc-retry-attempts', 3)
         self._rpc_retry_wait = config.getint('core', 'rpc-retry-wait', 30)
+        self._rpc_log_retries = config.getboolean('core', 'rpc-log-retries', True)
```
Tarrasch: Oh, this part of the code still doesn't use the [config class](http://luigi.readthedocs.io/en/stable/configuration.html#configuration-classes) syntax? Oh well maybe better done in a separate PR.

test/rpc_test.py
```diff
@@ -72,6 +72,36 @@ def test_retry_rpc_limited(self):
...
+    def test_log_rpc_retries_disabled(self, mock_logger):
+        """
+        Tests that retries of an RPC method are not logged
+        """
+
+        fetch_results = [socket.timeout, socket.timeout, socket.gaierror]
+        try:
+            self.get_work(fetch_results)
+            self.fail("get_work should have thrown RPCError")
+        except luigi.rpc.RPCError as e:
```
Tarrasch: Can't you use self.assertRaises?

ulzha: This was done to assert about e.sub_exception.

PR #2487, opened 2018-08-10
Added optional choice for hdfs clients 

### LINKED ISSUE #2481: Use of KerberosClient with webhdfs
Currently webhdfs uses InsecureClient by default. However, the hdfs package also has an KerberosClient which can be used to do authenticated requests. 

Our hdfs system uses Kerberos and I would like it if KerberosClient is added as an option. 

The KerberosClient does not need any additional settings (as far as I can test). And thus only an additional parameter is needed: `use_kerberos_client`. 

Are you willing to accept a pull request for this?

pietermarsman:
## Description
Added optional choice for hdfs clients. Previously only InsecureClient was supported. This PR adds KerberosClient and TokenClient. 

## Motivation and Context
Some hdfs work with kerberos authentication. Mine does... This adds support for those clusters. 

This fixes #2481.

I have tested the KerberosClient in practice and it works! (Thousand times faster than hadoopcli :) ) 

Not tested the TokenClient though, because I do not have access to a Hadoop cluster with token authentication.

### DIFF
--- a/luigi/contrib/hdfs/webhdfs_client.py
+++ b/luigi/contrib/hdfs/webhdfs_client.py
@@ -43,9 +43,7 @@ class webhdfs(luigi.Config):
                               description='Port for webhdfs')
     user = luigi.Parameter(default='', description='Defaults to $USER envvar',
                            config_path=dict(section='hdfs', name='user'))
-    client_type = luigi.Parameter(default='insecure',
-                                  description='Type of client to use. One of insecure, kerberos or token')
-    token = luigi.Parameter(default='', description='Hadoop delegation token, only used when client_type="token"')
+    client_type = luigi.Parameter(default='insecure', description='Type of client to use. Can be insecure or kerberos')
 
 
 class WebHdfsClient(hdfs_abstract_client.HdfsFileSystem):
@@ -56,12 +54,11 @@ class WebHdfsClient(hdfs_abstract_client.HdfsFileSystem):
     <https://hdfscli.readthedocs.io/en/latest/api.html>`__.
     """
 
-    def __init__(self, host=None, port=None, user=None, client_type=None, token=None):
+    def __init__(self, host=None, port=None, user=None, client_type=None):
         self.host = host or hdfs_config.hdfs().namenode_host
         self.port = port or webhdfs().port
         self.user = user or webhdfs().user or os.environ['USER']
         self.client_type = client_type or webhdfs().client_type
-        self.token = token or webhdfs().token
 
     @property
     def url(self):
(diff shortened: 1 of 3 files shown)


luigi/contrib/hdfs/webhdfs_client.py
```diff
@@ -43,6 +43,9 @@ class webhdfs(luigi.Config):
                               description='Port for webhdfs')
     user = luigi.Parameter(default='', description='Defaults to $USER envvar',
                            config_path=dict(section='hdfs', name='user'))
+    client_type = luigi.Parameter(default='insecure',
```
Tarrasch: Maybe use ChoiceParameter?

luigi/contrib/hdfs/webhdfs_client.py
```diff
@@ -73,8 +78,18 @@ def client(self):
...
+            import hdfs
+            return hdfs.InsecureClient(url=self.url, user=self.user)
+        elif self.client_type == 'kerberos':
+            from hdfs.ext.kerberos import KerberosClient
+            return KerberosClient(url=self.url)
+        elif self.client_type == 'token':
+            import hdfs
+            return hdfs.TokenClient(url=self.url, token=self.token)
+        else:
+            raise ValueError("Error: Unknown client type specified in webhdfs client_type"
```
Tarrasch: You could skip this if you use the ChoiceParameter.

luigi/contrib/hdfs/webhdfs_client.py
```diff
@@ -43,6 +43,9 @@ class webhdfs(luigi.Config):
                               description='Port for webhdfs')
     user = luigi.Parameter(default='', description='Defaults to $USER envvar',
                            config_path=dict(section='hdfs', name='user'))
+    client_type = luigi.Parameter(default='insecure',
+                                  description='Type of client to use. One of insecure, kerberos or token')
+    token = luigi.Parameter(default='', description='Hadoop delegation token, only used when client_type="token"')
```
dlstadther: @daveFNbuck Added an `OptionalParameter`. Would that be applicable here?

pietermarsman:
Only adding KerberosClient now.

And I've added some really basic test. Don't know what your standards are, so if you have any specific test in mind let me know. 

Also changed Parameter to ChoiceParameter. This removed the need for the ValueError is the else clause, so now I made InsecureClient the default in the else clause.

In the next couple of days I will test this code in a couple of projects.

pietermarsman:
(Y)(Y)

pietermarsman:
👍 👍

dlstadther:
Thanks for reporting back @pietermarsman ! Glad to hear things functioning as expected

PR #2488, opened 2018-08-13
Fix S3Client.copy return value consistency

dlstadther:
## Description

Adds support for num/size copy response for single-file copies.

## Motivation and Context

S3Client.copy doesn't return a tuple if not copying a directory. This patch adds support for num/size tuple copy response for non-directories.

Adds assertions for minimum values to existing copy tests.
All s3_test.py tests pass locally.

### DIFF
--- a/test/contrib/s3_test.py
+++ b/test/contrib/s3_test.py
@@ -425,7 +425,6 @@ class TestS3Client(unittest.TestCase):
         self._run_copy_test(self.test_put_multipart_empty_file)
 
     @mock_s3
-    @skipOnTravis('https://travis-ci.org/spotify/luigi/jobs/145895385')
     def test_copy_empty_dir(self):
         """
         Test copying an empty dir
--- a/test/contrib/s3_test.py
+++ b/test/contrib/s3_test.py
@@ -453,7 +453,9 @@ class TestS3Client(unittest.TestCase):
             self.assertTrue(s3_client.exists(file_path))
 
         s3_dest = 's3://mybucket/copydir_new/'
-        s3_client.copy(s3_dir, s3_dest, threads=10, part_size=copy_part_size)
+        response = s3_client.copy(s3_dir, s3_dest, threads=10, part_size=copy_part_size)
+
+        self._run_copy_response_test(response)
 
         for i in range(n):
             original_size = s3_client.get_key(s3_dir + str(i)).size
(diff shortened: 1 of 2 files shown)


dlstadther:
Just tagging a couple other recent `luigi/contrib/s3.py` contributors to quicken the review process. Your review would be appreciated!

@gardenunez @ouanixi

test/contrib/s3_test.py
```diff
@@ -527,3 +511,11 @@ def _run_multipart_test(self, part_size, file_size, **kwargs):
         key_size = s3_client.get_key(s3_path).size
         self.assertEqual(file_size, key_size)
         tmp_file.close()
+
+    def _run_copy_response_test(self, response):
```
gardenunez: I'd find the way to check that the size is strictly equal to the sum instead of using GreaterEqual. Same thing to the number of items.

test/contrib/s3_test.py
```diff
@@ -454,15 +453,17 @@ def test_copy_dir(self):
             self.assertTrue(s3_client.exists(file_path))
 
         s3_dest = 's3://mybucket/copydir_new/'
-        s3_client.copy(s3_dir, s3_dest, threads=10, part_size=copy_part_size)
+        response = s3_client.copy(s3_dir, s3_dest, threads=10, part_size=copy_part_size)
```
gardenunez: I'd have a test for empty directory and empty file cases.

dlstadther: There already exists a `test_copy_empty_file`, but I'll add a test for empty directory

dlstadther:
@gardenunez Thanks for the review and suggestions!

> Another suggestion would be to extract copy-dir and copy-file logic inside copy to a separate function. Maybe even to have copy_dir and copy_file functions instead of copy, this one I still not sure.

I have no opposition to this refactor, but it is outside of the scope of this PR. Feel free to contribute it separately though :)

test/contrib/s3_test.py
```diff
@@ -425,6 +424,26 @@ def test_copy_empty_file(self):
         """
         self._run_copy_test(self.test_put_multipart_empty_file)
 
+    @mock_s3
+    @skipOnTravis('https://travis-ci.org/spotify/luigi/jobs/145895385')
```
gardenunez: can you elaborate?

test/contrib/s3_test.py
```diff
@@ -527,3 +531,13 @@ def _run_multipart_test(self, part_size, file_size, **kwargs):
         key_size = s3_client.get_key(s3_path).size
         self.assertEqual(file_size, key_size)
         tmp_file.close()
+
+        return 1, key_size
+
+    def _run_copy_response_test(self, response, expected_num=None, expected_size=None):
```
gardenunez: :+1:

test/contrib/s3_test.py
dlstadther: Sorry. Copy/paste. Will remove to try to green build without it.

test/contrib/s3_test.py
```diff
@@ -235,7 +235,7 @@ def test_put_multipart_less_than_split_size(self):
         # 5MB is minimum part size
         part_size = 8388608
         file_size = 5000
-        self._run_multipart_test(part_size, file_size)
+        return self._run_multipart_test(part_size, file_size)
```
Tarrasch: Why do you return the last thing you do in a test case??? I thought you only used self.assertBlah()

dlstadther: The `test_put_multipart...` tests are used below as args to `self._run_copy_test(...)` and I need the output of `self._run_multipart_test` to verify if the correct num and size of files were copied

PR #2489, opened 2018-08-13
Use passed password when create a redis connection

Gr1f0n6x:
## Description

Use passed password when create a redis connection
## Motivation and Context

Currently, password is ignored when creating a RedisTarget, but in some cases there is a needance in secured connection
 

There is no new logic, so, existing tests should cover it

### DIFF
--- a/luigi/contrib/redis_store.py
+++ b/luigi/contrib/redis_store.py
@@ -68,6 +68,7 @@ class RedisTarget(Target):
         self.redis_client = redis.StrictRedis(
             host=self.host,
             port=self.port,
+            password=self.password,
             db=self.db,
             socket_timeout=self.socket_timeout,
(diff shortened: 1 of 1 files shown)


dlstadther:
Odd that password would have never been passed to StrictTarget. I haven't used this module before, so here's to hoping that @pbamba (the only other author of `redis_store.py`) can provide some aid.

If not, then LGTM.

Gr1f0n6x:
Probably, it's a typo, because in tests there is also password param (assigned to None, but anyway) which is passed to RedisStore

pbamba:
Good catch @Gr1f0n6x! Indeed it was probably a typo back then, from what I remember I didn't use it with authentication. LGTM as well.

PR #2491, opened 2018-08-14
Remove Scheduler config backwards compatibility

dlstadther:
## Description

Removes backwards compatibility of long-deprecated scheduler configuration variables

## Motivation and Context

TODO note to remove backwards compatibility from [>3 years ago](https://github.com/spotify/luigi/pull/727).

Given this is a deprecation of something from Luigi 1.x (and we're already in 2.x), I don't think this requires a milestone version increase.

Travis :)

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -123,8 +123,6 @@ def rpc_method(**request_args):
 
 
 class scheduler(Config):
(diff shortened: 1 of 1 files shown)


Tarrasch:
Totally, nice cleanup! :)

PR #2495, opened 2018-08-18
Support 'b' flag in open() by setting the correct default format lazily

dalejung:
## Description
When `format` is `None`, do not set it to the default format in the Constructor, but wait until `open()` is called. If we find the 'b' flag, then use `Nop` so format doesn't muck with binary data.

## Motivation and Context

#1647

Included tests and on personal Tasks.

So this might not admittedly be *the* way. The format/mode juggling is tricky since they were designed to be independent of each other and `mode` isn't really the builtin `open`'s mode since the format really decides whether we're writing bytes or str.

### DIFF
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -49,7 +49,7 @@ from luigi import six
 from luigi import configuration
 from luigi.format import get_default_format
 from luigi.parameter import OptionalParameter, Parameter
-from luigi.target import FileAlreadyExists, FileSystem, FileSystemException, FileSystemTarget, AtomicLocalFile, MissingParentDirectory
+from luigi.target import FileAlreadyExists, FileSystem, FileSystemException, FileSystemTarget, AtomicLocalFile, MissingParentDirectory, LazyFormatMixin
 from luigi.task import ExternalTask
 
 logger = logging.getLogger('luigi-interface')
@@ -639,7 +639,7 @@ class ReadableS3File(object):
         self.close()
 
 
-class S3Target(FileSystemTarget):
+class S3Target(FileSystemTarget, LazyFormatMixin):
     """
     Target S3 file object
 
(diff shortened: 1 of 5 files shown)


luigi/target.py
```diff
@@ -348,3 +349,20 @@ def __exit__(self, exc_type, exc, traceback):
...
+
+
+class LazyFormatMixin:
+    @property
+    def format(self):
+        if self._format is None:
+            self._set_default_format()
+        return self._format
+
+    def _set_default_format(self, mode=None):
```
dlstadther: It's odd to me that the method `_set_default_format` take an optional arg of `mode`...

luigi/contrib/s3.py
```diff
@@ -639,7 +639,7 @@ def __iter__(self):
         self.close()
 
 
-class S3Target(FileSystemTarget):
+class S3Target(FileSystemTarget, LazyFormatMixin):
```
dlstadther: Any way to explicitly test this within s3_test's S3Target tests?

luigi/local_target.py
```diff
@@ -128,19 +128,16 @@ def rename_dont_move(self, path, dest):
         self.move(path, dest, raise_if_exists=True)
 
 
-class LocalTarget(FileSystemTarget):
+class LocalTarget(FileSystemTarget, LazyFormatMixin):
```
dlstadther: Same here?

luigi/target.py
```diff
@@ -348,3 +349,23 @@ def __exit__(self, exc_type, exc, traceback):
         if exc_type:
             return
         return super(AtomicLocalFile, self).__exit__(exc_type, exc, traceback)
+
+
+class LazyFormatMixin:
+    _format = None
```
dlstadther: Be vary of class variables vs instance variables when it comes to modifying them

luigi/contrib/s3.py
dalejung: Anything outside of the FileSystemTarget tests? Originally I didn't have the S3 target updated but the tests failed (correctly) so I had to update S3Target to fix. So it is testing S3

luigi/target.py
dalejung: You want this changed to 1. set _format lazily or 2. require subclasses to populate `_format = None`? Normally only do this when the default is None since you can't modify the class instance via self.

dlstadther: I guess i was thinking

```python
class MyClass:
    _var = None

# vs

class YourClass:
        super(YourClass, self).__init__(*args, **kwargs)
        self._var = None
```

dalejung:
@dlstadther So I added a warning in the case where 'wb' might've been requested with a TextFormat. In reality that would only occur if someone accessed `.format` before calling `open('wb')`. So
```
t = LocalTarget(path, 'wb')
t.format
t.open('wb) # warns
```
Note that previously writing a str to `LocalTarget(path, 'wb')` worked since we ignored 'b'. This will now error.

luigi/local_target.py
dalejung: err, I think maybe the context for this was on another comment? not sure what you were referrring to :/

dlstadther:
@ulzha , could you or @Tarrasch give this a final sign off?

Tarrasch:
I think we can merge if the PR creator have tested it in production. it's a bit too time consuming for me to get into this kind of python code.

@dalejung If you can commit to fixing any potential breakages, can we merge?

PR #2498, opened 2018-08-22
added a new event 'progress'

honnix:
@ulzha PTAL

## Description

Added a new event named `event.core.progress`.

## Motivation and Context

This event can be fired by the task itself while running. The purpose is
for the task to report progress, metadata or any generic info so that
event handler listening for this can keep track of the progress of
running task.

I have included unit tests.

### DIFF
--- a/luigi/event.py
+++ b/luigi/event.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-''' Definitions needed for events. See :ref:`Events` for info on how to use it.'''
+""" Definitions needed for events. See :ref:`Events` for info on how to use it."""
 
 
 class Event(object):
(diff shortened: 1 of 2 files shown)


ulzha:
LGTM

ulzha:
One urging side thought - there is some existing code already, catering for similar communication:
* [`TaskStatusReporter`](https://github.com/spotify/luigi/blob/de4629a4c7c562b53fee2045aabd84189f897593/luigi/worker.py#L288), that has ` update_progress_percentage` method, among others, for sending some communication from the task
* [`SchedulerMessage`](https://github.com/spotify/luigi/blob/de4629a4c7c562b53fee2045aabd84189f897593/luigi/worker.py#L319), that allows receiving communication from the scheduler by the task

I guess anyway the use of events would be preferable, could allow us to eventually replace the ad-hoc code?

honnix:
Are those for communication from worker to scheduler? If that is the case, it doesn't seem to fit very well to our use case.

honnix:
@Tarrasch 👍 for that, will fix.

luigi/event.py
```diff
@@ -25,6 +25,10 @@ class Event(object):
     DEPENDENCY_PRESENT = "event.core.dependency.present"
     BROKEN_TASK = "event.core.task.broken"
     START = "event.core.start"
+    #: This event can be fired by the task itself while running. The purpose is
```
honnix: Tried following existing inline doc. How does it look? @Tarrasch

ulzha:
I didn't mean to include communication with the scheduler in our solution. It was just an aside comment, that we'll now be "reporting progress" in 2 different ways, and that I prefer event approach among those.

honnix:
OK. Got it.

PR #2499, opened 2018-08-23
Added launchType and networkConfiguration support for ECS task.

### LINKED ISSUE #2497: contrib.ecs with launchType and networkConfiguration parameters
I am a user of contrib.ecs but I find that I cannot launch my tasks without defining the two following parameters:

```
launchType='EC2'|'FARGATE',
networkConfiguration={
'awsvpcConfiguration': {
    'subnets': [
        'string',
    ],
    'securityGroups': [
        'string',
    ],
    'assignPublicIp': 'ENABLED'|'DISABLED'
}
}
```
(the above is taken from boto3 ECS.Client.run_task [documentation](https://boto3.readthedocs.io/en/latest/reference/services/ecs.html#ECS.Client.run_task))

Would it be possible to expose those from the contrib.ecs interface? Or did I miss something?

akddv:
## Description
Add optional parameters to support launchType and networkConfiguration arguments of boto3 ECS run_task function.

## Motivation and Context
Resolves #2497 

### DIFF
--- a/luigi/contrib/ecs.py
+++ b/luigi/contrib/ecs.py
@@ -131,11 +131,17 @@ class ECSTask(luigi.Task):
     :param cluster: str defining the ECS cluster to use.
         When this is not defined it will use the default one.
 
+    :param launch_type: str defining the launch type on which to run your task.
+
+    :param network_conf: dict describing the network configuration for the task.
+
     """
 
     task_def_arn = luigi.OptionalParameter(default=None)
     task_def = luigi.OptionalParameter(default=None)
     cluster = luigi.Parameter(default='default')
+    launch_type = luigi.OptionalParameter(default=None)
+    network_conf = luigi.OptionalParameter(default=None)
 
     @property
     def ecs_task_ids(self):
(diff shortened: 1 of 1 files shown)


akddv:
After checking out the HEAD on master, I tried running the test with the following command:
> PYTHONPATH=. nosetests test/contrib/ecs_test.py

and got the error message:
> Exception: default is not a cluster

So I created a cluster and had to register an instance to it with a code mostly taken from [moto test suite](https://github.com/spulec/moto/blob/master/tests/test_ecs/test_ecs_boto3.py#L524).

Even with that, the two tests 'test_registered_task' and 'test_override_command' will fail with the error message:

> 'Exception: hello-world:1 is not a task_definition'.

the test 'test_unregistered_task' will go through without exceptions but the 'run_task' function will return a list of thousands of tasks which will make the '_track_tasks' function dump a lot of logs on screen and it will keep looping as the tasks statuses seems to stay 'RUNNING'.

I realized that it is hard for me to simply run the test without yet making any changes to the code.
Could you provide some guidance to run the test?

akddv:
I could not run this test successfully, and it does not seems that moto supports those fields anyway.
I found out that the KubernetesJobTask fits better my needs (it does even check the container exit code unlike ECSTask), so I am closing this PR.
Thanks a lot for your help.

PR #2490, opened 2018-08-14
Update tornado version

dlstadther:
## Description

Check for compatibility of tornado's new milestone release

## Motivation and Context

Tornado 5.1.0 has been released. Also, a [community member asked if Tornado >=5 was compatible](https://groups.google.com/forum/#!topic/luigi-user/ABJ3SKnkNHU)

## Have you tested this? If so, how?
 

Travis :)

PR #2503, opened 2018-08-24
Revert "Update tornado version"

honnix:
Reverts spotify/luigi#2490


honnix:
That looks weird.

Tarrasch:
Thanks for being on top of these things!

PR #2504, opened 2018-08-24
revert tornado upgrade

honnix:
## Description

Revert the change of tornado version requirement.

## Motivation and Context

Since 5.0 tornado requires Python 2.7.9+ and 3.4+.

Travis.


honnix:
@dlstadther PTAL

honnix:
Yeah I will.

Issue #2171, opened 2017-07-05
Seperate out AWS into a seperate repo


dlstadther:
In this issue I propose:

* We copy out the code for aws-related stuff into it's own repository. I created a an empty repository, see [luigi-orchestrator/luigi-aws#1](https://github.com/luigi-orchestrator/luigi-aws/issues/1). Add to pypi
* We remove the implementation from here and replace it with a package dependency. We keep the tests as a way to check backward compatibility
* Celebrate. We have accomplished:
    * First (or second) attempt of splitting the kitchen-sink contrib/ folder.
    * It becomes so much easier to work with the repos if they are separate.

This is also a bit in the direction of #1897, but much more trivial, it requires no extra software. Just a build dependency.

Note that this design consciously suggests different repos, as we want different travis builds, different people being pinged on changes, different set of owners. etc. etc.

dckc:
> It becomes so much easier to work with the repos if they are separate.

Easier in what way? What pain point would this be addressing?

From the perspective of a user, I can only see costs.

Tarrasch:
@dckc, I think the main motivations for me would be:

 * Travis would run tests much faster - This is obviously a solved problem among Google and Facebook Inc, where they use monorepos. But I'm not sure if Travis handles it as well. Currently the `.travis.yml` file is cluttured and I suspect it scares away contributors.
 * The dependencies in the pip-package could finally start to make sense, and we can remove [these hacks](https://github.com/spotify/luigi/blob/b47b2679d99980d486b1527229a9c8b4caa5a314/luigi/contrib/bigquery.py#L26-L31) which exist in every other contrib repo.
 * Code coverage for the "core" repo would go up. :)
 * We can have different maintainers for different repos, one idea:
   * Core - @Tarrasch
   * AWS - @dlstadther 
   * Hadoop - @daveFNbuck 
   * GCP - @Spotify (@uldis and @fabriziodemaria with others)

These are some thoughts for now. But thanks for bringing this question up, it's definitely not to late to change our minds yet. :)

What would be the painpoints for you as a user, except for the first transition?

dckc:
Thanks for filling me in on the background. You sold me at "Travis would run tests much faster".

The cost for users like me would be installing the extra bits -- AWS, say. It's most likely a small cost and so far, for me, hypothetical. I don't use AWS or the others yet.

PR #2506, opened 2018-08-27
Fix exception when toml lib is not installed

orsinium:
## Description

Fix issue from comments to #2457 and some related stuff:
1. Add informative message, when `LUIGI_CONFIG_PATH` with `*.toml` file is presented in env vars, but `toml` lib is not installed yet.
1. Raise `ImportError` instead of simple logging, because this is critical issue, that doesn't allow lib works fine.
1. Add warning when `LUIGI_CONFIG_PATH` and `LUIGI_CONFIG_PARSER` point to different parsers.
1. Add info about toml support in `installation` section.

Some minimal tests included and works fine.

### DIFF
--- a/luigi/configuration/core.py
+++ b/luigi/configuration/core.py
@@ -46,12 +46,12 @@ def get_config(parser=PARSER):
 
     parser_class = PARSERS[parser]
     if not parser_class.enabled:
-        logger.error((
-                "Parser not installed yet. "
-                "Please, install luigi with required parser:\n"
-                "pip install luigi[{parser}]"
-            ).format(parser)
+        msg = (
+            "Parser not installed yet. "
+            "Please, install luigi with required parser:\n"
+            "pip install luigi[{parser}]"
         )
+        raise ImportError(msg.format(parser))
 
     return parser_class.instance()
 
@@ -72,12 +72,20 @@ def add_config_path(path):
     parser_class = PARSERS[parser]
 
     if not parser_class.enabled:
-        logger.error((
+        msg = (
                 "Parser not installed yet. "
                 "Please, install luigi with required parser:\n"
                 "pip install luigi[{parser}]"
-            ).format(parser)
+            )
+        raise ImportError(msg.format(parser))
+
+    if parser != PARSER:
+        msg = (
+            "Config for {added} parser added, but used {used} parser. "
+            "Set up right parser via env var: "
+            "export LUIGI_CONFIG_PARSER={added}"
         )
+        warnings.warn(msg.format(added=parser, used=PARSER))
 
     # add config path to parser
     parser_class.add_config_path(path)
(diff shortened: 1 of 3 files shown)


luigi/configuration/core.py
```diff
@@ -66,9 +66,26 @@ def add_config_path(path):
...
+        parser = PARSER
+    parser_class = PARSERS[parser]
+
+    if not parser_class.enabled:
+        msg = (
+                "Parser not installed yet. "
+                "Please, install luigi with required parser:\n"
+                "pip install luigi[{parser}]"
+            )
+        raise ImportError(msg.format(parser=parser))
```
dlstadther: This `if not parser_class.enabled` is checked at least twice here. Can that be extracted into a method or something for DRY programming purposes?

luigi/configuration/core.py
```diff
@@ -66,9 +66,26 @@ def add_config_path(path):
...
+
+    if parser != PARSER:
+        msg = (
+            "Config for {added} parser added, but used {used} parser. "
+            "Set up right parser via env var: "
+            "export LUIGI_CONFIG_PARSER={added}"
+        )
+        warnings.warn(msg.format(added=parser, used=PARSER))
```
dlstadther: Does `warnings.warn()` also write to logs?

orsinium: I use it because `warnings.warn` used [in many other places in project](https://github.com/spotify/luigi/search?q=warnings.warn&unscoped_q=warnings.warn). Yes, [warnings can be captured](https://docs.python.org/3/library/logging.html#logging.captureWarnings).

luigi/configuration/core.py
orsinium: Fixed

orsinium:
Can someone fix the TravisCI? It have troubles with AWS keys. It's something related to #2171?

Tarrasch:
> Can someone fix the TravisCI? It has troubles with AWS keys. It's something related to #2171?

@dlstadther do you know why there's so many crashes (does the tests rely on external services?).

luigi/configuration/core.py
```diff
@@ -40,19 +40,21 @@
     PARSER = DEFAULT_PARSER
 
 
+def _check_parser(parser_class, parser):
+    if not parser_class.enabled:
+        msg = (
+            "Parser not installed yet. "
+            "Please, install luigi with required parser:\n"
+            "pip install luigi[{parser}]"
```
Tarrasch: This can only be toml right? It's fine to hard code it too.

dlstadther:
moto version between last passing build and current is the same. However, boto3 increased from 1.7.84 to 1.8.2 and botocore from 1.10.84 to 1.11.2

[Last passing build deps](https://travis-ci.org/spotify/luigi/jobs/420230608#L528)
[Current failing build deps](https://travis-ci.org/spotify/luigi/jobs/421663295#L528)

dlstadther:
Builds are green again. I suspect that the boto issue we were experiencing was a Travis issue.

orsinium:
Thank you ^^

PR #2508, opened 2018-08-28
Refactor/s3 copy

gardenunez:
## Description
Extract copy directory and copy file logic out of copy function in S3 client.

## Motivation and Context
Is just a small refactor over copy function for having more readability and reuse copy file logic also in copy directory. 
The idea came by reviewing this PR: https://github.com/spotify/luigi/pull/2488

I didn't change any behaviour so tests remain the same.

### DIFF
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -295,7 +295,8 @@ class S3Client(FileSystem):
         self.s3.meta.client.upload_fileobj(
             Fileobj=open(local_path, 'rb'), Bucket=bucket, Key=key, Config=transfer_config, ExtraArgs=kwargs)
 
-    def copy(self, source_path, destination_path, threads=100, start_time=None, end_time=None, part_size=8388608, **kwargs):
+    def copy(self, source_path, destination_path, threads=100, start_time=None, end_time=None, part_size=8388608,
+             **kwargs):
         """
         Copy object(s) from one S3 location to another. Works for individual keys or entire directories.
         When files are larger than `part_size`, multipart uploading will be used.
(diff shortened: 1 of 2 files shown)


luigi/contrib/s3.py
```diff
@@ -322,55 +322,52 @@ def copy(self, source_path, destination_path, threads=100, start_time=None, end_
...
-                result.get()
-
-            end = datetime.datetime.now()
-            duration = end - start
-            logger.info('%s : Complete : %s total keys copied in %s' %
-                        (datetime.datetime.now(), total_keys, duration))
-
-            return total_keys, total_size_bytes
+            return self._copy_dir(src_bucket, src_key, dst_bucket, dst_key, end_time, source_path, start,
+                                  start_time, threads, total_keys, transfer_config, **kwargs)
```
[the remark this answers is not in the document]
gardenunez: @dlstadther I noticed that tests are failing in master branch. Some of them related to `s3` tests. Should we create an issue for that? I will take a closer look at that before finishing this PR.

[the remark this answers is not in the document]
dlstadther: @gardenunez Yes, there is currently an issue with Travis S3 tests. I haven't been able to look into it too far, but suspect the issue is with moto given that it's our fake s3client credential that may be raising the issue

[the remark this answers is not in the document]
dlstadther: moto version between last passing build and current is the same. However, boto3 increased from 1.7.84 to 1.8.2 and botocore from 1.10.84 to 1.11.2

dlstadther:
Build errors appear to have been a Travis issue as they have since resolved without change on our end.

gardenunez:
I have made the changes suggested by @dlstadther. Please take a look.

luigi/contrib/s3.py
```diff
@@ -319,61 +317,59 @@ def copy(self, source_path, destination_path, threads=100, start_time=None, end_
...
+            self._copy_file(src_bucket, src_key, dst_bucket, dst_key, transfer_config, **kwargs)
+            return 1, item.size
+
+    def _copy_file(self, src_bucket, src_key, dst_bucket, dst_key, transfer_config, **kwargs):
+        copy_source = {
+            'Bucket': src_bucket,
+            'Key': src_key
+        }
+
+        self.s3.meta.client.copy(copy_source, dst_bucket, dst_key, Config=transfer_config, ExtraArgs=kwargs)
```
dlstadther: It is still weird to be that `_copy_dir()` and `_copy_file()` don't return the same thing. `_copy_dir()` returns a tuple and `_copy_file()` returns `None`

dlstadther: I wonder if it's really more efficient to use boto's `meta.client.copy()` for a single file rather than treating it with the same logic (minus the iteration over a dir) as `_copy_dir()`

gardenunez: Regarding returning values I agree, it should be consistent.
Regarding efficiency, I think calling `meta.client.copy` could be more efficient and at the same time less consistent regarding Luigi S3 `_copy_file` logic. But since now we don't do much there, I agree on using the boto3 function. Good catch.  
What do you think?

gardenunez:
Will I was doing the previous changes I that we are sending to `_copy_file` and `_copy_dir` the `source_path`, `src_bucket` and `src_key` args, this is redundant. Maybe sending `copy_source` ( AKA `source_path` would be enough and more close to meta.client.copy definition. What do you think?

gardenunez:
Another thing is that I think copy nested directories are not supported, would that be desirable?

dlstadther:
I agree on supplying `source-path` to method as the other parts can be created from that.

`_copy_dir()` already supports nesteddir copying

gardenunez:
Changes committed. @dlstadther Let me know what do you think?

luigi/contrib/s3.py
```diff
@@ -309,71 +309,70 @@ def copy(self, source_path, destination_path, threads=100, start_time=None, end_
...
-                'Key': src_key
-            }
-            item = self.get_key(source_path)
-            total_size_bytes += item.size
-            self.s3.meta.client.copy(
-                copy_source, dst_bucket, dst_key, Config=transfer_config, ExtraArgs=kwargs)
+            return self._copy_file(source_path, dst_bucket, dst_key, transfer_config, **kwargs)
+
+    def _copy_file(self, source_path, dst_bucket, dst_key, transfer_config, **kwargs):
+        src_bucket, src_key = self._path_to_bucket_and_key(source_path)
```
dlstadther: If we're going to move the split of `source_path` into src_bucket/src_key here, why not also do the same with `destination_path`?

That way, the only two required args for `_copy_*()` are `source_path` and `destination_path`. That also mirrors the requirements of `copy()`

dlstadther: That would also mean moving transfer_config

gardenunez: make sense

gardenunez: Done

luigi/contrib/s3.py
```diff
@@ -309,71 +312,69 @@ def copy(self, source_path, destination_path, threads=100, start_time=None, end_
         :returns tuple (number_of_files_copied, total_size_copied_in_bytes)
         """
 
-        start = datetime.datetime.now()
-
-        (src_bucket, src_key) = self._path_to_bucket_and_key(source_path)
-        (dst_bucket, dst_key) = self._path_to_bucket_and_key(destination_path)
-
         # don't allow threads to be less than 3
         threads = 3 if threads < 3 else threads
```
dlstadther: I wrote this line long ago and have always wanted to update it to `threads = max(threads, 3)`. This PR seems appropriate since you've already modified other code in this method

Issue #2507, opened 2018-08-28
502 Bad Gateway error tied to 


rayrrr:
We use Luigi as an unpinned requirement that rebuilds whenever we deploy, in a cloud-hosted environment.

We deployed some changes today that did not involve our Python code; after this deploy, our service went offline with 502 Bad Gateway responses from the scheduler.

Our team dug in and found the following error:

```
AttributeError: 'Task' object has no attribute 'param_visibilities'
 File "/usr/local/lib/python3.6/site-packages/luigi/scheduler.py", line 825, in add_task
```

It looks like the Scheduler class needs to be updated to work with the latest changes to how the Task class handles param visibility included in release 2.7.7.

Gr1f0n6x:
@rayrrr hm, it's strange =\ Probably, i missed something
I will look at this.
If it is possible, could you provide some code to reproduce it?

rayrrr:
@Gr1f0n6x can't provide our company code, sorry, but it looks as though it'll happen with any Task run against a scheduler.

Gr1f0n6x:
Yes, as i thought. The problem was in saving old versioned tasks.
```
    result = getattr(self._scheduler, method)(**arguments)
  File "/Users/grifon/anaconda/lib/python3.6/site-packages/luigi/scheduler.py", line 825, in add_task
    if not task.param_visibilities:
AttributeError: 'Task' object has no attribute 'param_visibilities'
```

rayrrr:
@Gr1f0n6x we do stave state in a pickle file as per the [built-in `state_path` attribute](https://luigi.readthedocs.io/en/stable/api/luigi.scheduler.html?highlight=state_path#luigi.scheduler.scheduler.state_path) of the scheduler but that's it. We currently run our jobs as "nightly batch jobs," so there were no old tasks waiting to be executed when this happened to us AFAIK. 

The bug has something to do with https://github.com/spotify/luigi/blob/master/luigi/scheduler.py#L825 which is a conditional statement dealing with an attribute called `task.param_visibilities`, which doesn't exist in the case of the bug. It appears that this portion of the code is intending to set this attribute if it doesn't already exist...but it also seems this attribute is added on to the instance of the class at some point after it is created...so if something didn't happen as expected in that whole flow, this bug could result, because

```
>>> class Task(object):
...     builtin = 'hello'
... 
>>> task = Task()
>>> task.builtin
'hello'
>>> task.param_visibility
Traceback (most recent call last):
[...]

PR #2509, opened 2018-08-28
Fix error #2507

Gr1f0n6x:
## Description

Replace direct attribute accessing by using built-n function `getattr`
## Motivation and Context

Fixes https://github.com/spotify/luigi/issues/2507
 

I tried to reproduce this issue without and with this patch. As local test showed, this patch fix this issue

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -822,7 +822,7 @@ class Scheduler(object):
             task.family = family
         if not getattr(task, 'module', None):
             task.module = module
-        if not task.param_visibilities:
+        if not getattr(task, 'param_visibilities', None):
             task.param_visibilities = _get_default(param_visibilities, {})
         if not task.params:
             task.set_params(params)


luigi/scheduler.py
```diff
@@ -822,7 +822,7 @@ def add_task(self, task_id=None, status=PENDING, runnable=True,
             task.family = family
         if not getattr(task, 'module', None):
             task.module = module
-        if not task.param_visibilities:
+        if not getattr(task, 'param_visibilities', None):
```
rayrrr: `hasattr(task, 'param_visibilities')` is more concise with the same effect, but either one works.

rayrrr: Good point. Stick with `getattr`. Thanks!

rayrrr:
Approved, though you will have to figure out why the Travis build is failing and take appropriate action.

Gr1f0n6x:
There is a lot of failed S3 tests. I'm not sure that this change has broken it, but i'll re-check it anyway

dlstadther:
No specific luigi PR appears to be the cause of the s3 test failures. I suspect it has to do with a dependency update. I've take a couple minutes and found that boto3 and botocore versions increased between passing and failing builds. But don't have more time to dedicate.

Any and all efforts provided to identify and resolve this issue will be most appreciated!

PR #2517, opened 2018-09-06
Expose NoSectionError at package level

honnix:
## Description
Expose `NoSectionError` at package level so that `from luigi.configuration import NoSectionError` would work.

## Motivation and Context

It used to work like this until https://github.com/spotify/luigi/pull/2457, so we should probably bring it back for backward compatibility.

TBD

### DIFF
--- a/luigi/configuration/__init__.py
+++ b/luigi/configuration/__init__.py
@@ -24,4 +24,5 @@ __all__ = [
     'get_config',
     'LuigiConfigParser',
     'LuigiTomlParser',
+    'NoSectionError'
 ]
--- a/luigi/configuration/__init__.py
+++ b/luigi/configuration/__init__.py
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from .cfg_parser import LuigiConfigParser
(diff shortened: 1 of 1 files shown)


ulzha:
+1

I guess the reason we "should probably" is just so that we keep compatibility?

The whole idea of using NoSectionError exception as a way for checking if a section exists, seems like abuse of exceptions.

honnix:
Yeah the reason is to keep backward compatibility. It is indeed a bad idea to abuse exception. Maybe we should instead let it break in this case.

honnix:
@Tarrasch @dlstadther wdyt?

dlstadther:
Am i correct that `NoSectionError` is just coming from `configparser`/`ConfigParser`? Other places in Luigi which need that just reimport from config parser. So i'd say no need to keep in package __init__

honnix:
You are right it is from `ConfigParse`. The problem is not Luigi itself but user code because they could access from `init`, and now that breaks. The risk is low and very easy to fix if we ever get complaints.

dlstadther:
Yeah, i'm fine letting that break. At some point, we need to drop support for certain things. This is a small case for that.

honnix:
👍 Let's go for it.

PR #2519, opened 2018-09-12
MockTarget mode fixed to accept read and write like r* or w*

PedroRossi:
## Description
I changed the mock target class to accept open modes such as rb, rt, wb and others

## Motivation and Context
I did it because I am testing some luigi tasks using pickle and the wb wasn't working because of the mode == 'w' instead of mode[0] == 'w' comparation

I have included a test for bytearray at mock_test.py

### DIFF
--- a/luigi/mock.py
+++ b/luigi/mock.py
@@ -131,7 +131,7 @@ class MockTarget(target.FileSystemTarget):
         """
         self.move(*args, **kwargs)
 
-    def open(self, mode):
+    def open(self, mode='r'):
         fn = self.path
         mock_target = self
 
@@ -158,7 +158,7 @@ class MockTarget(target.FileSystemTarget):
                 super(Buffer, self).write(data)
 
             def close(self):
-                if mode == 'w':
+                if mode[0] == 'w':
                     try:
                         mock_target.wrapper.flush()
                     except AttributeError:
@@ -174,15 +174,15 @@ class MockTarget(target.FileSystemTarget):
                 return self
 
             def readable(self):
-                return mode == 'r'
+                return mode[0] == 'r'
 
             def writeable(self):
-                return mode == 'w'
+                return mode[0] == 'w'
 
             def seekable(self):
                 return False
 
-        if mode == 'w':
+        if mode[0] == 'w':
             wrapper = self.format.pipe_writer(Buffer())
             wrapper.set_wrapper(wrapper)
             return wrapper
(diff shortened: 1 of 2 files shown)


luigi/mock.py
```diff
@@ -158,7 +158,7 @@ def write(self, data):
                 super(Buffer, self).write(data)
 
             def close(self):
-                if mode == 'w':
+                if mode[0] == 'w':
```
dlstadther: is it possible for `mode` to be `None`? If so, this will throw an exception

PedroRossi: From python2 and python3 when we open a file we use it like `f = open("file.extension", mode)` where mode should be a string, if not it should throw an exception, should I add a test for this also?

dlstadther: Python allows you to `with open("file.extension")` where the mode is defaulted to `r`. However, i'm just not clear on the behavior here given we're overriding `open`. It wouldn't hurt to have  a test for sanity :)

PedroRossi: Yes, I am going to write more tests about it and notify then. Thank you for the comments :)

PedroRossi: @dlstadther I refactored the byte test and added and default open action as you mentioned in the discussion and added an error check test if type is None 👍

test/mock_test.py
```diff
@@ -42,6 +43,31 @@ def test_with(self):
...
+            self.assertEqual(list(b), ['bar'])
+
+    def test_mode_none_error(self):
+        t = MockTarget("foo")
+        try:
+            with t.open(None) as b:
+                b.write("bar")
+        except TypeError as error:
+            self.assertIsNotNone(error)
+            self.assertIn("'NoneType' object", str(error))
```
dlstadther: can this test use [`assertRaises`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises)?

PedroRossi: Yes

test/mock_test.py
```diff
@@ -42,6 +43,30 @@ def test_with(self):
...
+        with t.open() as b:
+            self.assertEqual(list(b), ['bar'])
+
+    def test_mode_none_error(self):
+        t = MockTarget("foo")
+        try:
+            with t.open(None) as b:
+                b.write("bar")
+        except TypeError:
+            self.assertRaises(TypeError)
```
dlstadther: with this being in an `except`, we can't guarantee that the assertion is risen

i was expecting

```python
t = MockTarget("foo")
with self.assertRaises(TypeError):
    with t.open(None) as b:
        b.write("bar")
```

PedroRossi: Ok, going to refactor it then

PedroRossi:
No problem, it was my first time doing a PR to an open source project and I was able to learn more of python :)

PR #2526, opened 2018-09-21
Add option to avoid s3 bucket validation

dlstadther:
## Description

Add option to not validate s3 bucket during PUT* operations

## Motivation and Context

S3 permissions can be very granular. It is possible for them to be such that an IAM Role only has
permissions on a specific subdir within a S3 bucket. In this event, _validate_bucket() may fail due
to a 403, when in fact the permissions exist to put* within the specified path.

Added tests, which passed locally

### DIFF
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -247,6 +247,7 @@ class S3Client(FileSystem):
         Put an object stored locally to an S3 path.
         :param local_path: Path to source local file
         :param destination_s3_path: URL for target S3 location
+        :param validate_bucket: bool indicator of validating existence of target s3 bucket root (default: True)
         :param kwargs: Keyword arguments are passed to the boto function `put_object`
         """
         self._check_deprecated_argument(**kwargs)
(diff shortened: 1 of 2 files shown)


dlstadther:
Tagging a couple others for reviews (who recently contributed to s3):
@ouanixi @gardenunez

dlstadther:
Not sure why only an unrelated Python 3.5 test is failing...

dlstadther:
And now the build passes :)

dlstadther:
Thanks for taking a look, @ouanixi  :)

ouanixi:
Not sure why we need to validate the bucket tbh (at least not this way)! That method either returns a bool or raises an exception. It's never used in a way that its return value is needed/used! Making it only used to raise exceptions (clumsily since when the bucket doesn't exist it doesn't raise)

I'd argue we should just get rid of it!! What do you think? (Unless I'm missing something of course)

dlstadther:
@ouanixi I agree that it doesn't make too much sense why it's there. I would be happy to remove that code path all together. My initial stab was to impact existing code as much as possible.

Individuals can error handle as they wish without the bucket validation. It really isn't overly useful anyway. My use case for removing it shows that it can provide false errors too.

PR #2528, opened 2018-09-24
Remove s3 bucket validation prior to file upload

dlstadther:
Replaces #2526 

## Description

Removes bucket validation from s3 PUT operations

## Motivation and Context

S3 permissions can be very granular. It is possible for them to be such that an IAM Role only has
permissions on a specific subdir within a S3 bucket. In this event, _validate_bucket() may fail due
to a 403, when in fact the permissions exist to put* within the specified path.

It was then determined that the `_validate_bucket()` check provided very little benefit to Luigi. The removal of this staticmethod should not break any code.

Added tests

### DIFF
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -264,9 +264,6 @@ class S3Client(FileSystem):
         self._check_deprecated_argument(**kwargs)
         (bucket, key) = self._path_to_bucket_and_key(destination_s3_path)
 
-        # validate the bucket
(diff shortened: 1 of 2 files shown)


dlstadther:
@ouanixi

gardenunez:
LGTM.
It happened to me in the past that have the permissions to put but not to head makes Luigi fails.

ouanixi:
Looks good to me :)

Issue #2533, opened 2018-09-26
Using S3Client to download files fails when the path contains a question mark


aminevg:
As per the title, using the S3Client in `luigi.contrib.s3` to download files breaks when the S3 path contains a question mark.

This works:
```python
import boto3
client = boto3.client('s3')
client.download_file('bucket', 'hello?world', 'local_path') # works
```
While this fails:
```python
from luigi.contrib.s3 import S3Client
S3Client().get('s3://bucket/hello?world', 'local_path') # 403/404 error
```

I've narrowed it down to the `_path_to_bucket_and_key` function outputting the wrong result:
```python
S3Client()._path_to_bucket_and_key('s3://bucket/hello?world') # --> ('bucket', 'hello')
```
This is likely due to `urlsplit` interpreting a question mark as a delimiter between the path and the query in normal URL paths. Will submit a PR if welcome/when I have time.

PR #2534, opened 2018-09-26
Fix S3Client's _path_to_bucket_and_key to support keys with question marks

aminevg:
## Description
Fix S3Client's _path_to_bucket_and_key to support keys with question marks.

## Motivation and Context
Fixes https://github.com/spotify/luigi/issues/2533.

Before:
```python
S3Client()._path_to_bucket_and_key('s3://bucket/hello?world') # --> ('bucket', 'hello')
```

After:

```python
S3Client()._path_to_bucket_and_key('s3://bucket/hello?world') # --> ('bucket', 'hello?world')
```

### DIFF
--- a/test/contrib/s3_test.py
+++ b/test/contrib/s3_test.py
@@ -503,6 +503,11 @@ class TestS3Client(unittest.TestCase):
             copy_size = s3_client.get_key(s3_dest + str(i)).size
             self.assertEqual(original_size, copy_size)
 
+    def test__path_to_bucket_and_key(self):
+        self.assertEqual([('bucket', 'noquestionmarkhere'), ('bucket', 'questionmark?here')],
+                         [S3Client._path_to_bucket_and_key('s3://bucket/noquestionmarkhere'),
+                          S3Client._path_to_bucket_and_key('s3://bucket/questionmark?here')])
+
     @mock_s3
     def _run_copy_test(self, put_method, is_multipart=False):
         create_bucket()
--- a/test/contrib/s3_test.py
+++ b/test/contrib/s3_test.py
@@ -504,9 +504,10 @@ class TestS3Client(unittest.TestCase):
             self.assertEqual(original_size, copy_size)
 
     def test__path_to_bucket_and_key(self):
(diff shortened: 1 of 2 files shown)


aminevg:
@dlstadther Thanks for the review! 
Added a test, could you confirm it's written correctly? :)

test/contrib/s3_test.py
```diff
@@ -503,6 +503,11 @@ def test_copy_dir(self):
             copy_size = s3_client.get_key(s3_dest + str(i)).size
             self.assertEqual(original_size, copy_size)
 
+    def test__path_to_bucket_and_key(self):
+        self.assertEqual([('bucket', 'noquestionmarkhere'), ('bucket', 'questionmark?here')],
+                         [S3Client._path_to_bucket_and_key('s3://bucket/noquestionmarkhere'),
+                          S3Client._path_to_bucket_and_key('s3://bucket/questionmark?here')])
```
dlstadther: While this does what you want it to, i believe it's generally better to split unittest tests into very small tests or just having a couple quick/easy assertions. Such as

```python
def test__path_to_bucket_and_key(self):
    self.assertEqual(('bucket', 'dir'), S3Client._path_to_bucket_and_key('s3://bucket/dir))

def test__path_to_bucket_and_key_with_question_mark(self):
    self.assertEqual(('bucket', 'dir?arg'), S3Client._path_to_bucket_and_key('s3://bucket/dir?arg))
```

or even within 1 unittest test, with multiple assertions:

```python
def test__path_to_bucket_and_key(self):
    self.assertEqual(('bucket', 'dir'), S3Client._path_to_bucket_and_key('s3://bucket/dir))
    self.assertEqual(('bucket', 'dir?arg'), S3Client._path_to_bucket_and_key('s3://bucket/dir?arg))
```

The first example above would present easier identification during a test failure (bc the test named `test__path_to_bucket_and_key_with_question_mark` would fail and that's very clear what the problem is). The latter may be easier to read, but test failure wouldn't be as explicit.

aminevg: @dlstadther Seems fair! I went with the first approach, what do you think? :)

Issue #2535, opened 2018-09-26
502 Bad Gateway error tied to public_params


kockockockoc:
Update from luigi-2.7.5 to 2.7.8 and we do save state in a pickle file.

@Gr1f0n6x , does it look like https://github.com/spotify/luigi/issues/2507 ?

```

2018-09-26 14:43:21,542 tornado.access[40044] ERROR: 500 GET /api/task_list?data=...
    Traceback (most recent call last):
  File "/usr/lib64/python2.7/site-packages/tornado/web.py", line 1467, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/usr/lib/python2.7/site-packages/luigi/server.py", line 77, in get
    result = getattr(self._scheduler, method)(**arguments)
  File "/usr/lib/python2.7/site-packages/luigi/scheduler.py", line 1430, in task_list
    serialized = self._serialize_task(task.id, include_deps=False)
  File "/usr/lib/python2.7/site-packages/luigi/scheduler.py", line 1284, in _serialize_task
    'display_name': task.pretty_id,
  File "/usr/lib/python2.7/site-packages/luigi/scheduler.py", line 351, in pretty_id
    param_str = ', '.join(u'{}={}'.format(key, value) for key, value in sorted(self.public_params.items()))
AttributeError: 'Task' object has no attribute 'public_params'
```

Gr1f0n6x:
@kockockockoc Yes, it is exactly the same probem :(
At this moment you can fix it manually or make a git cherry-pick this commit: ` 2c98b5667c8d66c6975687f9f78b414a0709c275`
https://github.com/spotify/luigi/pull/2509/commits/2c98b5667c8d66c6975687f9f78b414a0709c275

kockockockoc:
Hmm, this issue similar but not quite the same, IMHO: "param_visibilities" vs "public_params"

Gr1f0n6x:
@kockockockoc yes, you're right. 
Probably, for backward compatibility we need to add something like that:
```
if hasattr(self, public_params):
    param_str = ', '.join(u'{}={}'.format(key, value) for key, value in sorted(self.public_params.items()))
else:
    param_str = ', '.join(u'{}={}'.format(key, value) for key, value in sorted(self.params.items()))
```

And also in this part: `'params': task.public_params,` for reliability.
Shame on me, but i didn't expect errors like that :(

I'll take a look in this issue more precisely tomorrow

PR #2537, opened 2018-09-27
Fix error: AttributeError: 'Task' object has no attribute

Gr1f0n6x:
## Description

Add additional checks for attribute 'public_params' and method 'set_params'
## Motivation and Context

https://github.com/spotify/luigi/issues/2535
 

I have included unit tests for cases when scheduler tries to load old versioned tasks and make an attempt to accept new attribute or call new method.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -348,7 +348,8 @@ class Task(object):
 
     @property
     def pretty_id(self):
-        param_str = ', '.join(u'{}={}'.format(key, value) for key, value in sorted(self.public_params.items()))
+        params = self.public_params if hasattr(self, 'public_params') else self.params
+        param_str = ', '.join(u'{}={}'.format(key, value) for key, value in sorted(params.items()))
         return u'{}({})'.format(self.family, param_str)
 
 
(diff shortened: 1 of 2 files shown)


Tarrasch:
Is this just to resolve an upgrade issue if you use pickled files? I'd rather not add this much complexity for that.

Gr1f0n6x:
Yes, it's exactly for this purpose.  

When i've tried to resolve this I found a comment about pickling/unpickling old versioned files and that It can lead to issues like this. 
So, if this is ok than this PR can be closed.

Tarrasch:
Backward compatibility would be nice if could be done in a much simpler and general way, and not unique to every kind of change we do to our representation of tasks.

Perhaps simply that the pickle file from now on has a version number, and we keep track of when we have to drop the file because the server can't support the file being loaded.

So I vote for closing this. Any other opinions?

kwilcox:
I'd also vote on closing this (but thanks for the work!). Putting task representation changes in the CHANELOG and mentioning that pickle users will need to remove their old files will be more maintainable.

PR #2541, opened 2018-09-28
Reduce TravisCI Test Runtime

dlstadther:
Travis tests are taking way too long (and we're being too greedy with their free resources). Trying to reduce the overhead here.

## Description

Brute force nose attributes of (`apache`, `aws`, `contrib`) excluded from `core` (previously `nonhdfs`). (`apache`, `aws`) will be added for most recent Python 2.X and 3.X versions.

## Motivation and Context

CI needs to be quicker - seeing what can be done to segment differently.

TravisCI :)


dlstadther:
py-27-aws only adds 2:30m

dlstadther:
Not immediately sure why apache tests ran some postgres stuff. Will look into this more at a later time.

dlstadther:
py27-apache <2m

dlstadther:
< 10m build

dlstadther:
@Tarrasch @honnix @ulzha 
Looking to get one more :+1: before merging.

The only contribution negative is that new TestCases in `/tests/contrib/` will need to add a Nose Attr label, otherwise will be included in `core`

honnix:
Looks awesome. Good job!

Just one minor comment. `apache` refers to `Apache Hadoop Ecosystem`, right? Wonder whether this name would be too generic. On the other hand, I think people will get it.

dlstadther:
Correct. `apache` -> apache hadoop ecosystem

Thanks for taking a look!

Tarrasch:
Awesome! :dancer: :tada:  :balloon: :100:  :trophy:

PR #2547, opened 2018-10-08
Make capture_output non-positional in ExternalProgramTask

adaitche:
## Description

Make the `capture_output` parameter of `ExternalProgamTask` non-positional, so that it doesn't catch parameters from sub-classes.

## Motivation and Context

Imagine the following task

```python
class MyTask(ExternalProgamTask):
    foo = luigi.Parameter()
    ...
```
When this task is instantiated via `MyTask("bar")` the supplied argument will be catched by `capture_output` and won't be assigned to `foo`. This behaviour is surprising. 

In my case it led to a failing pipeline: we were using `PySparkTask`, which is derived from `ExternalProgramTask`. After the introduction of the `capture_output` parameter in the recent release (2.7.9), `capture_output` catched a positional parameter of a deriving task and the pipeline failed with `MissingParameterException`.

PS: We had the luigi version pinned in our `requirements.txt` in the following way: `luigi==2.7.*`, with the intention to get only bugfixes through new patch versions. The introduction of `capture_output` seems more like a new feature to me and hence I would have expected the minor version to be incremented. Does luigi follow semantic versioning?

### DIFF
--- a/luigi/contrib/external_program.py
+++ b/luigi/contrib/external_program.py
@@ -59,7 +59,7 @@ class ExternalProgramTask(luigi.Task):
     behaviour can be overriden by passing ``--capture-output False``
     """
 
-    capture_output = luigi.BoolParameter(default=True, significant=False)
+    capture_output = luigi.BoolParameter(default=True, significant=False, positional=False)
 
     def program_args(self):
         """


adaitche:
@ekbelova

Tarrasch:
The short story about semantic versioning is that we all agree that it's "good". But [we don't](https://github.com/spotify/luigi/blob/master/RELEASE-PROCESS.rst) because it's quite hard to track for the maintainers. There's no good tooling. There's a completely different story if you have paid full-timers working on luigi who can mentally do a good job of remembering what's going in. As @dlstadther said the releases are quite spread and seldom. Basically we'll rather put the existing maintainership power we have on responsive PRs over release processing. Sorry for any inconvenience and thanks for your contribution!

adaitche:
Thanks for the clarification. We will just pin the full version going forward.

I fully understand the issue about lack of full-time developers and the constraints this brings with it. Thanks a lot for developing and maintaining Luigi!

honnix:
Thanks for the contribution @adaitche and I agree with @Tarrasch . We (as paid full-timers) could have done better.

@Tarrasch I think we do have the opportunity to have better semantic versioning. I will bring this up internally and see how that goes. Thanks for maintaining Luigi!

mfcabrera:
LGTM, I am colleague of @adaitche. I honestly thing this change would have been hard to spot. This happen to me a lot in custom code when I added a new parameter and I was  using the class passing the parameters without explicitly naming it.  Maybe we could add that to the parameter documentation (i.e. that is recommended that parameter are passed explicitly as the parent classes might change). I know that's the whole reasoning behind the positional option but this just showed it can happen.

PR #2548, opened 2018-10-08
FIX accept pathlib based path as argument for LocalTarget

### LINKED ISSUE #2322: Support for Python 3 pathlib Paths in LocalTarget
It would be nice to be able to use `Path` based objects from Python 3's [`pathlib`](https://docs.python.org/3/library/pathlib.html) when defining paths for `LocalTarget`s. Currently, you need to cast them to strings or the target methods like `exists` and `open` won't work.

```
In [9]: lt = luigi.LocalTarget(Path('~') / 'kk')
In [10]: lt.exists()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-b13e17b73d35> in <module>()
----> 1 lt.exists()

~/.virtualenvs/bishop/lib/python3.6/site-packages/luigi/target.py in exists(self)
    238         """
    239         path = self.path
--> 240         if '*' in path or '?' in path or '[' in path or '{' in path:
    241             logger.warning("Using wildcards in path %s might lead to processing of an incomplete dataset; "
    242                            "override exists() to suppress the warning.", path)

TypeError: argument of type 'PosixPath' is not iterable
In [11]: lt = luigi.LocalTarget(str(Path('~') / 'kk'))
In [12]: lt.exists()
Out[12]: False

```

orsinium:
## Description

Accept pathlib.PosixPath, py._path.local.LocalPath and other path based objects as argument for FileSystemTarget.

## Motivation and Context

Fix #2322

I have included unit tests.


luigi/target.py
```diff
@@ -207,7 +207,8 @@ def __init__(self, path):
 
         :param str path: the path associated with this FileSystemTarget.
         """
-        self.path = path
+        # cast to str path objects like pathlib.PosixPath and py._path.local.LocalPath
```
dlstadther: Can we reword this in-line comment for clarity? `cast to str to allow path to be objects like pathlib....`

test/local_target_test.py
```diff
@@ -65,6 +65,21 @@ def test_exists(self):
         p.close()
         self.assertEqual(t.exists(), os.path.exists(self.path))
 
+    def test_pathlib(self):
+        """Test work with pathlib.Path"""
+        try:
+            import pathlib
+        except ImportError:
+            # skip test for environment without pathlib
+            return
```
dlstadther: Is there a way to use a skip decorator instead?

orsinium:
Added fixes from @dlstadther's review.

bjlange:
Amazing! Thanks @orsinium

orsinium:
Added usage example in docstring.

orsinium:
What about this PR? Does anyone have any changes requests?

josham:
@Tarrasch @dlstadther Any reason this isn't yet merged?

dlstadther:
@josham The code looks good, but the tests aren't passing

orsinium:
I couldn't tame the doctest, so I've just dropped usage example >.<

orsinium:
Travis is ok now

orsinium:
Wow, [Tarrasch](/Tarrasch), don't worry about it too much, please. I've returned it as you said. Thank you for a good idea :)

orsinium:
Ok, it doesn't work:
https://travis-ci.org/spotify/luigi/jobs/495852917

josham:
I think you want to remove `.. code-block:: python` from the docs

orsinium:
In this case it will be broken in the readthedocs api reference.

Tarrasch:
Just fyi you can test the docs locally too you know: https://github.com/spotify/luigi/blob/master/CONTRIBUTING.rst#writing-documentation

orsinium:
I know. BTW, I've made a few fixes for CONTRIBUTING.rst in another PR. However, sometimes it can't help because I can check locally only "*-core" tests, without hdfs.

I've pushed this code with broken doctest to show that your idea how to bypass this check doesn't work.

orsinium:
now it works

Tarrasch:
Thanks for this addition @orsinium!

orsinium:
Wow. Thank you, fellows. Sorry for long review process :)

bjlange:
🙌

PR #2549, opened 2018-10-09
FIX HdfsAtomicWriteDirPipe.close() when using snakebite and the file do not exist.

helkhalfi:
## Description
HdfsAtomicWriteDirPipe.close() throws a FileNotFoundException rather than a HDFSCliError when trying to remove a file that doesn't exist when `snakebite` is used.
See also => `https://github.com/spotify/luigi/pull/2119#issuecomment-307884052`

## Motivation and Context
We use snakebite (production and dev) in order to deal with hdfs (because it's much faster than hdfs :) right ) and as a side note we are a python shop.
When we migrated from 2.2.0 to 2.7.8 (we also noticed same thing in 2.7.5), jobs starts failing because luigi was trying to delete a directory that do not exist.

I ran my jobs with this code and it works for me


dlstadther:
Do we know why a non-existent directory is trying to be removed? Is _that_ behavior expected?

Is this something that can be unittested? If so, please include :)

luigi/contrib/hdfs/format.py
```diff
@@ -75,9 +77,13 @@ def abort(self):
     def close(self):
         super(HdfsAtomicWriteDirPipe, self).close()
         try:
-            remove(self.path)
-        except HDFSCliError:
-            pass
+            if exists(self.path):
+                remove(self.path)
```
helkhalfi: Hello @dlstadther : 
This is the line that ties to remove a path without checking if it exists.
A looking at how it was implemented try/except/pass the spirit is: it it fails do not bother.
I added a check to verify if the path exists before removing it.

helkhalfi:
I updated the PR with comments and code update.
Regarding the UT I believe they are disabled for this section as they fall into `TOXENV=py27-cdh` which is commented on travis.yml

dlstadther:
Makes sense to me.

If this section of unittests are disabled in Travis, why is the build failing?

fenimore:
Hmm I fixed the build failing with python 3 environments (exception syntax) but now the `flake8` lint is failing (even though it's failing where no changes have been made :thinking: )

PR #2551, opened 2018-10-10
Add Movio as part of the companies that use Luigi

utdemir:
We use Luigi extensively, and are really happy with it.

Thank you for creating it!


dlstadther:
Thanks @utdemir !

PR #2552, opened 2018-10-11
Added Redis as a state store backend

adilkhash:
## Description
Added Redis backend to store Luigi's tasks state

## Motivation and Context
We run our infrastructure inside containers hence are required to have centralized state store location

I ran my jobs with this code and it works for me. If community finds this PR useful I will add test.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -680,12 +680,13 @@ class SimpleTaskRedisState(SimpleTaskState):
         super(SimpleTaskRedisState, self).__init__(state_path)
         self._redis_url = state_path
         self._redis_key = state_key
-        self._conn = redis.StrictRedis.from_url(self._redis_url)
+        self.redis = redis
+        self._conn = self.redis.StrictRedis.from_url(self._redis_url)
 
     def load(self):
         try:
             state_obj = self._conn.get(self._redis_key)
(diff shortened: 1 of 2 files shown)


dlstadther:
This seems like a reasonable contribution. I'm not sure what the adoption rate would be. But i'm not immediately opposed to such a configurable Redis state store.

dhuckins:
+1 would like to have this feature

dhuckins:
am trying this in a distributed environment in front of a load balancer am getting:
`This progress looks :| because there were tasks that were not granted run permission by the scheduler`
is there a way to control how often the state is persisted?

dhuckins:
so using redis as the backend it would need to flush every server request, that's a lot of overhead

luigi/scheduler.py
```diff
@@ -668,6 +673,50 @@ def disable_workers(self, worker_ids):
             worker.tasks.clear()
 
 
+class SimpleTaskRedisState(SimpleTaskState):
```
Tarrasch: Is it still Simple? Maybe RedisTaskState?

adilkhash:
Closing PR because there is no reason to store dumped state, because it could be obsolete or corrupted (due to pickle)

PR #2558, opened 2018-10-25
Update for  AWS Batch module

yukihira1992:
Supported all parameters defined at the aws batch submit_job() api.

## Description
Added these parameters defined at the aws batch [submit_job()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/batch.html?highlight=batch#Batch.Client.submit_job) api.

- arrayProperties
- dependsOn
- containerOverrides
- retryStrategy
- timeout

At first I tried to use luigi.DictParameter class.
However boto3 cannot accept _FrozenOrderedDict class.
Therefore I used luigi.OptionalParameter class to handle Dictionary and List parameters as JSON strings.

## Motivation and Context
The original BatchTask class did not support the containerOverrides argument and it cannot change environment variables for docker containers at run-time.
In addition I support all parameters.

Tested this module in real AWS Batch environment.
And I updated test codes to use mocks at the [moto](https://github.com/spulec/moto) module.

### DIFF
--- a/luigi/contrib/batch.py
+++ b/luigi/contrib/batch.py
@@ -288,6 +288,10 @@ class BatchTask(luigi.Task):
     poll_time = luigi.IntParameter(default=POLL_TIME)
     queue = luigi.OptionalParameter(default=None)
     array_properties = luigi.OptionalParameter(default='{}')
+    # We selected OptionalParameter instead of DictParameter,
+    # since boto3(botocore) accepts only dict type
+    # and DictParameter (_FrozenOrderdDict) causes ParamValidationError.
+    # For the same reason, OptionalParameter is used for depends_on, container_overrides, and retry_strategy.
     depends_on = luigi.OptionalParameter(default='[]')
     container_overrides = luigi.OptionalParameter(default='{}')
     retry_strategy = luigi.OptionalParameter(default='{}')
--- a/luigi/contrib/batch.py
+++ b/luigi/contrib/batch.py
@@ -198,73 +198,91 @@ class BatchTask(luigi.Task):
     descriptions for how to issue the ``docker run`` command. This Luigi Task
     requires a pre-registered Batch jobDefinition name passed as a Parameter
 
-    :param job_definition (str): name (arn) of pre-registered jobDefinition
-        :: example
+    :param str job_definition: name (arn) of pre-registered jobDefinition.
+
+        .. code:: python
 
             job_definition = 'arn:aws:batch:<region>:<user_id>:job-definition/<job_definition_name>:<revision>'
 
-    :param job_name: name of specific job, for tracking in the queue and logs.
+    :param str job_name: name of specific job, for tracking in the queue and logs.
 
-    :param poll_time: interval in seconds to check a status of jobs.
+    :param int poll_time: interval in seconds to check a status of jobs.
 
-    :param queue: name (arn) of the job queue.
-        :: example
+    :param str queue: name (arn) of the job queue.
+
+        .. code:: python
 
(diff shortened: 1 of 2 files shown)


luigi/contrib/batch.py
```diff
@@ -164,6 +169,7 @@ def wait_on_job(self, job_id):
                 job_str = json.dumps(jobs, indent=4)
                 logger.debug('Job details:\n' + job_str)
 
+                print(jobs)
```
dlstadther: remove this before completing

luigi/contrib/batch.py
```diff
@@ -186,28 +192,121 @@ def register_job_definition(self, json_fpath):
...
 
     def run(self):
         bc = BatchClient(self.poll_time)
         job_id = bc.submit_job(
             self.job_definition,
             self.parameters,
-            job_name=self.job_name)
+            job_name=self.job_name,
+            queue=self.queue,
+            array_properties=json.loads(self.array_properties),
```
dlstadther: You say there is an issue here when using `DictParameter` due to type checking?

I've encountered something similar. I wonder what we can do to DictParameter to resolve the returned object type. Would you happen to have any suggestions?

It feels odd to be using `OptionalParameter` here (although functional). That said, I don't think this is a blocker. But perhaps at least worth an in-line comment for reason.

yukihira1992: @dlstadther Thanks for your comments!

> You say there is an issue here when using DictParameter due to type checking?

Yes, boto3 raises validation error when using DictParameter.
```
ParamValidationError: Parameter validation failed:
Invalid type for parameter containerOverrides, value: <FrozenOrderedDict OrderedDict([('vcpus', 123)])>, type: <class 'luigi.parameter._FrozenOrderedDict'>, valid types: <class 'dict'>
```

> It feels odd to be using OptionalParameter here (although functional). That said, I don't think this is a blocker. But perhaps at least worth an in-line comment for reason.

I think so too.
Since aws api frequently uses json format, I want to fix it conveniently if possible.

> I've encountered something similar. I wonder what we can do to DictParameter to resolve the returned object type. Would you happen to have any suggestions?

How about JSON compatible parameter?
1. User can set JSON object such as dictionary, list, string, numerical number, and boolean.
2. The parameter hold's value as json serialized string.
3. In the task class, parameters can be accessed as deserialized objects.

```
class FooTask(luigi.Task):
    param1 = luigi.JSONParameter()

    def run(self):
        print(self.param1)


foo = FooTask(param1={'key': 'value'})
foo.run()  # {'key': 'value'}
```

dlstadther: In that example, the only usage difference between JsonParameter and DictParameter is Dict's ability to allow an element to be a custom object?

There's definitely a tight relationship between the two.

I'd like to involve some other Luigi-contributors around this:
@Tarrasch 
@joemeszaros 
@nugend

yukihira1992: > In that example, the only usage difference between JsonParameter and DictParameter is Dict's ability to allow an element to be a custom object?

Yes, indeed.

If you want to develop based on DictParameter, I want a function to restore from _FrozenOrderedDict to pure python dictionary.
For example _recursively_defreeze(value).

nugend: Why is boto3 restricting to only values of type dict? Shouldn't it be looking for Mapping or MutableMapping from collections (depending on if it actually needs to modify or not).

nugend: Is it possible that's an upstream issue then?

yukihira1992: I opened new issue about DictParameter #2562 .
And I added in-line comments.

PR #2559, opened 2018-10-27
Add HdfsFlagTarget

fenimore:
## Description

Added an `HdfsFlagTarget` modeled off the class`S3FlagTarget` https://github.com/spotify/luigi/blob/master/luigi/contrib/s3.py#L680

## Motivation and Context
The problem this solves is the same as the `S3FlagTarget`.  We'll have `spark` jobs output to directories and use the `_SUCCESS` semaphore to check for job completion.

This is a fix we've used for a pipeline, and it runs :smile_cat: . It is also nearly identical to the `S3FlagTarget`, which also works. I included unit tests.


test/contrib/hdfs_test.py
```diff
@@ -386,6 +386,12 @@ def create_target(self, format=None):
             target.remove(skip_trash=True)
         return target
 
+    def create_flag_target(self, format=None):
```
dlstadther: the method name here is confusing. You aren't actually creating the flag target. Just the class object needed to later create the flag

fenimore: I'm not sure what you mean here, this method does create a `HdfsFlagTarget`, and the helper method is modeled off of the similar method called `create_target` line 385. 

But in any case I don't think the helper method is necessary (especially if only used in that once test) and can be removed and the FlagTarget created in the test

PR #2561, opened 2018-10-29
Added port property to CopyToTable

adilkhash:
## Description
Added abstract property `port` to CopyToTable

## Motivation and Context
Sometimes it is not comfortable to provide port inside host property, i.e. `localhost:port`, hence added optional `port` property to `CopyToTable` task in case when database listens custom port. Currently the only way to supply port is via `host`.


adilkhash:
Agree, did not pay attention to that, sorry. Fixed to optional property. 
Speaking about `host:port` option. I use `.env` files to store credentials, i.e:

```python
DB_HOST = 'localhost'
DB_NAME = 'database'
DB_USER = 'user'
DB_PASSWORD = 'password'
DB_PORT = 15432
```

`DB_HOST` variable is actively used everywhere and in order not to break things out, inside `CopyToTable` task I have something like this:

```python
class MyCopyToTable(CopyToTable):
    host = f'{config.DB_HOST}:15432'
```
Which looks a little bit ugly :)

adilkhash:
Is it possible to merge this one?

dlstadther:
Thanks for pinging @adilkhash . Merged!

PR #2564, opened 2018-10-30
Check type of column before migrating - fixes #2563

### LINKED ISSUE #2563: Postgres error if view depends on task_parameters.value
There is a problem related to #2240. Luigi runs `ALTER TABLE task_parameters ALTER COLUMN value TYPE TEXT` every time `luigi.db_task_history.DbTaskHistory` is instantiated.
Normally this would be fine because it will not change anything if the type of `value` is already `TEXT`.
However, if you have a Postgres view that depends on this column, Postgres will refuse to run this instruction with an error message like `DETAIL: rule _RETURN on view taskparameter_view depends on column "value"`.

There doesn't seem a sane way to make Postgres accept this statement, it [seems to be necessary](https://stackoverflow.com/questions/17989355/alter-column-ignoring-dependent-views) to drop the view (and all dependent views), run the `ALTER TABLE` statement and re-create them.

It would be much nicer, if Luigi could check if the migration is actually necessary before executing that statement, similar to the way it checks for *sqlite*:

```PYTHON
elif 'sqlite' in engine.dialect.name:
    # SQLite does not support changing column types. A database file will need
    # to be used to pickup this migration change.
    for i in conn.execute('PRAGMA table_info(task_parameters);').fetchall():
        if i['name'] == 'value' and i['type'] != 'TEXT':
            logger.warning(
                'SQLite can not change column types. Please use a new database '
                'to pickup column type changes.'
            )
```

jeverling:
This checks if the type of `task_parameters.value` is something else than `TEXT`, and only runs the migration in that case.
This fixes #2563

### DIFF
--- a/luigi/db_task_history.py
+++ b/luigi/db_task_history.py
@@ -263,7 +263,9 @@ def _upgrade_schema(engine):
         elif 'mssql' in engine.dialect.name:
             conn.execute('ALTER TABLE task_parameters ALTER COLUMN value TEXT')
         elif 'postgresql' in engine.dialect.name:
-            conn.execute('ALTER TABLE task_parameters ALTER COLUMN value TYPE TEXT')
+            if str([x for x in inspector.get_columns('task_parameters')
+                    if x['name'] == 'value'][0]['type']) != 'TEXT':
+                conn.execute('ALTER TABLE task_parameters ALTER COLUMN value TYPE TEXT')
         elif 'sqlite' in engine.dialect.name:
             # SQLite does not support changing column types. A database file will need
             # to be used to pickup this migration change.


luigi/db_task_history.py
```diff
@@ -263,7 +263,9 @@ def _upgrade_schema(engine):
         elif 'mssql' in engine.dialect.name:
             conn.execute('ALTER TABLE task_parameters ALTER COLUMN value TEXT')
         elif 'postgresql' in engine.dialect.name:
-            conn.execute('ALTER TABLE task_parameters ALTER COLUMN value TYPE TEXT')
+            if str([x for x in inspector.get_columns('task_parameters')
+                    if x['name'] == 'value'][0]['type']) != 'TEXT':
+                conn.execute('ALTER TABLE task_parameters ALTER COLUMN value TYPE TEXT')
```
dlstadther: what's up with the indentation? Something looks wrong. Maybe mix of spaces and tabs?

PR #2565, opened 2018-10-31
WIP: Detect parameter collisions in the @inherits decorator. See #2566 for more details.

KenyaDonDraper:
Detect parameter collisions while using the @inherits and the associated
@requires decorator. Collision detection for a particular parameter can
be opted out by adding it to the 'ignore_collisions=[]' list argument
of the decorator.

### DIFF
--- a/luigi/util.py
+++ b/luigi/util.py
@@ -283,21 +283,24 @@ class inherits(object):
                # ...
     """
 
-    def __init__(self, *tasks_to_inherit):
+    def __init__(self, *tasks_to_inherit, ignore_collisions=[]):
         super(inherits, self).__init__()
         if not tasks_to_inherit:
             raise TypeError("tasks_to_inherit cannot be empty")
 
         self.tasks_to_inherit = tasks_to_inherit
+        self.ignore_collisions = ignore_collisions
 
     def __call__(self, task_that_inherits):
-        # Get all parameter objects from each of the underlying tasks
+
+        # Check for parameter collisions and raise an exception if found.
+        self.check_for_parameter_collisions(task_that_inherits)
+
+        # Get all parameter objects from each of the underlying tasks and
+        # them to the inheriting task.
         for task_to_inherit in self.tasks_to_inherit:
             for param_name, param_obj in task_to_inherit.get_params():
-                # Check if the parameter exists in the inheriting task
-                if not hasattr(task_that_inherits, param_name):
-                    # If not, add it to the inheriting task
-                    setattr(task_that_inherits, param_name, param_obj)
+                setattr(task_that_inherits, param_name, param_obj)
 
         # Modify task_that_inherits by adding methods
         def clone_parent(_self, **kwargs):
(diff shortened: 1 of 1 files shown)


luigi/util.py
```diff
@@ -283,21 +283,24 @@ def run(self):
...
     """
 
-    def __init__(self, *tasks_to_inherit):
+    def __init__(self, *tasks_to_inherit, ignore_collisions=[]):
         super(inherits, self).__init__()
         if not tasks_to_inherit:
             raise TypeError("tasks_to_inherit cannot be empty")
 
         self.tasks_to_inherit = tasks_to_inherit
+        self.ignore_collisions = ignore_collisions
```
dlstadther: highly recommend
```python
def __init__(self, *tasks_to_inherit, ignore_collisions=None):
    ...
    self.ignore_collisions = ignore_collisions or []
```

luigi/util.py
```diff
@@ -283,21 +283,24 @@ def run(self):
...
 
     def __call__(self, task_that_inherits):
-        # Get all parameter objects from each of the underlying tasks
+
+        # Check for parameter collisions and raise an exception if found.
+        self.check_for_parameter_collisions(task_that_inherits)
+
+        # Get all parameter objects from each of the underlying tasks and
+        # them to the inheriting task.
```
dlstadther: maybe a typo... not sure what you are trying to say `and them to the inheriting task`

luigi/util.py
```diff
@@ -313,6 +316,41 @@ def clone_parents(_self, **kwargs):
 
         return task_that_inherits
 
+    def check_for_parameter_collisions(self, task_that_inherits):
+        error_msg = (
+            "Parameter name collision detected in tasks_to_inherit. "
+            "Parameter '{param}' in '{task}' duplicates "
+            "parameter '{param}' in {another_task}. "
```
dlstadther: could probably be simplified to `Parameter {param} in {task} is duplicated in {another_task}`. No need to continually output the name of the parameter if there is a collision due to the names being the same.

luigi/util.py
```diff
@@ -313,6 +316,41 @@ def clone_parents(_self, **kwargs):
...
+                    hasattr(task_that_inherits, param_name) and
+                    param_name not in self.ignore_collisions
+                ):
+                    raise AttributeError(error_msg.format(param=param_name,
+                                                          task=task_that_inherits,
+                                                          another_task=task_to_inherit))
+
+        # Check that the parameters from an inherited task don't mask the
+        # parameters from another inherited task.
+        for task_to_inherit in self.tasks_to_inherit:
```
dlstadther: As written, this loop is ran 3 times. 2 within this method and once afterward. Is there anyway to combine the looping of these `tasks_to_inherit` to avoid repetitive looping?

luigi/util.py
```diff
@@ -322,15 +360,16 @@ class requires(object):
...
     """
 
-    def __init__(self, *tasks_to_require):
+    def __init__(self, *tasks_to_require, ignore_collisions=[]):
         super(requires, self).__init__()
         if not tasks_to_require:
             raise TypeError("tasks_to_require cannot be empty")
 
         self.tasks_to_require = tasks_to_require
+        self.ignore_collisions = ignore_collisions
```
dlstadther: same here as mentioned above

luigi/util.py
KenyaDonDraper: I agree that was a bad practice. Assuming the mistake was using the mutable default argument, I actually prefer replacing the empty list with a tuple,

```Python
def __init__(self, *tasks_to_inherit, ignore_collisions=()):
    ...
    self.ignore_collisions = ignore_collisions
```
This way we remove the mutable default argument and at the same time allow the argument to communicate to a developer that the expected input is a sequence. The fewer lines the developer has to read before inferring the expected input the better.

luigi/util.py
KenyaDonDraper: yes typo, thanks. Was supposed to read: `and add them to the inheriting task`. Will make the change.

luigi/util.py
KenyaDonDraper: Agreed. Will make the change.

luigi/util.py
dlstadther: No complaint from me :)

TMaYaD:
FWIW, we found the name `inherits` very confusing for new users. We renamed it to `_include_params_from` and enhanced it to customise the parameters: https://gist.github.com/TMaYaD/1c5e0898179699bdc1885ff05cbea552

Let me know if I should open a PR for it.

PR #2571, opened 2018-11-05
Feature/override on requires function

nils-braun:
## Description
This PR adds another functionality to the `inherits` and `requires` decorator functions. The additional keyword arguments given during the call to `@inherits` (or `@requires`) will override the parameters in the base class, like when calling `self.clone(Task, **kwargs)` with the difference, that those overridden parameters will not be propagated to the inheriting class.

The changes in this PR allow to write something like

```python
class TaskA(luigi.Task):
    param_a1 = luigi.IntParameter()
    param_a2 = luigi.IntParameter()

@requires(TaskA, param_a1=3)
class TaskB(luigi.Task):
    param_b = luigi.IntParameter()
```
which defines a task `TaskB`, which does not have a parameter called `param_a1` (but still one with `param_a2` and `param_b`), but requires `TaskA` with `param_a1` set to 3 already.

This is especially useful if one wants to define a standard set of tasks, where a specific option is already known.

Additionally, I have refactored the `@requires` object a bit to inherit from the `@inherit` object (no pun intended).

## Motivation and Context
Example: you have a machine learning training and testing task. They both depend on one task to generate the data - with a parameter which data to generate. The training task will always require `Generation(training=True)` whereas the testing task will require `Generation(training=False)`. The new `@requires` call simplifies this.

I have included a small amount of unit tests for this.


luigi/util.py
```diff
@@ -281,30 +281,59 @@ def requires(self):
...
             for param_name, param_obj in task_to_inherit.get_params():
                 # Check if the parameter exists in the inheriting task
-                if not hasattr(task_that_inherits, param_name):
+                if not hasattr(task_that_inherits, param_name) and param_name not in self._non_copy_defaults.keys():
                     # If not, add it to the inheriting task
                     setattr(task_that_inherits, param_name, param_obj)
 
         # Modify task_that_inherits by adding methods
         def clone_parent(_self, **kwargs):
+            kwargs.update(**self._non_copy_defaults)
```
Tarrasch: Can this ever be passed by reference? If so do you modify the input?

nils-braun: The documentation says:

> If the form “**identifier” is present, it is initialized to a new ordered mapping receiving any excess keyword arguments, defaulting to a new empty mapping of the same type

(taken from https://docs.python.org/3/reference/compound_stmts.html#function-definitions), so I guess this is ok.

nils-braun:
I understand your concern.
The `@requires(functools.partial(TaskA, param_a1=3))` does unfortunately not work (the return from partial is not a task anymore), but would you like a free function for this more?

luigi/util.py
Tarrasch: Thanks for checking!

PR #2574, opened 2018-11-07
Make Worker parameter task_process_context an OptionalParameter (#2468)

josham:
Changed the `Worker` parameter `task_process_context` from `Parameter` to `OptionalParameter` to avoid unnecessary warning messages with it is not set.


Tarrasch:
Can you fix the build errors? I expect there to be a flake error.

Issue #2363, opened 2018-02-26
Missing parameters for BigQueryLoadTask


justjoheinz:
It seems that `bq load` now requires the parameter `autodetect` if one want to autodetect the schema from import csv or json files. Just ommitting the schema is not sufficient anymore.

It also provides a paramter for null values.

I created some kind of mixin which takes care of this in my local codebase:

```
        if self.null_marker:
            job['configuration']['load']['null_marker'] = self.null_marker

        if self.schema:
            job['configuration']['load']['schema'] = {'fields': self.schema}
        else:
            job['configuration']['load']['autodetect'] = True
```

but I am not a python wizard, so I do not want to create a PR. This changes work for me though and it would be nice to have them in the proper `BigQueryLoadTask`.

justjoheinz:
It would be nice if someone could look into this. I am not proficient enough in python to provide a PR. The issue is real however. the bigquery api has changed in regard to specifying that schemas should be autodetected.

dlstadther:
@drj42 Looks like you were the most recent BigQuery contributor. Is this something you've happened to already written for yourself? Or perhaps something you'd like to take a stab at?

If not, maybe @dmohns or @ulzha ?

justjoheinz:
Would be a pity to close this, if there is even a PR addressing the issue.

dlstadther:
Sorry @justjoheinz - that PR slipped through my notifications. Just reviewed it to get motion on this issue.

PR #2575, opened 2018-11-08
Add autodetect parameter to BigQueryLoadTask (#2363)

mklimasz:
## Description
When schema is not provided try to use `autodetect` parameter.
Removed comment that implicates that it doesn't work for JSON 
(docs linked below explicitly says it works for JSON)

## Motivation and Context
It solves: https://github.com/spotify/luigi/issues/2363

It also matches description of `configuration.load.autodetect`
in docs https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -443,7 +443,6 @@ class MixinBigQueryBulkComplete(object):
 
 class BigQueryLoadTask(MixinBigQueryBulkComplete, luigi.Task):
     """Load data into BigQuery from GCS."""
-
     @property
     def source_format(self):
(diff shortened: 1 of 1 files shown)


mklimasz:
Hey! Anybody will have time to check this one out or let's agree to close it? Regards.

mklimasz:
I have merged current master.

I don't think this change would be testable in unit tests, maybe in some integration tests.

dlstadther:
@honnix Do you/Spotify have BigQuery luigi experience to just say "LGTM"?

honnix:
👍 
This doc explains how it should work better I think: https://cloud.google.com/bigquery/docs/schema-detect

Issue #2578, opened 2018-11-14
DeprecationWarning: inspect.getargspec() is deprecated


josham:
The following warning is raised from the `inspect.getargspec` call in [scheduler.py](https://github.com/spotify/luigi/blob/06e3d9163c36f347cef09d9442aff55a10660f31/luigi/scheduler.py#L100).

A small example to show what I mean:

```
import warnings

from luigi.scheduler import rpc_method

warnings.simplefilter('module')

class Foo:
    @rpc_method()
    def foo(self):
        pass
```
Running the above will give:
```
luigi/scheduler.py:100: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
  fn_args = inspect.getargspec(fn)
```

I'm happy to submit a PR for this but let me know how you'd like me to fix it:

1.  Wrap the `inspect.getargspec` call with `warnings.catch_warnings` and ignore the warning
2.  Add a custom `getargspec` function to `luigi.six` (essentially the same as `inspect.getargspec` but without the warning being raised) and use that over `inspect.getargspec` if `PY3`
3. Some other way

dlstadther:
Given the recommended `getfullargspec` and `signature` don't exist in Py2.7, we may have to go the six.py route to maintain compatibility for Py2 and Py3.

I'd also invite a completely different way if you have a better suggestion.

PR #2579, opened 2018-11-14
Fix inspect.getargspec() DeprecationWarning in PY3 (#2578)

josham:
Adding a custom getargspec function to luigi.six (essentially the same as inspect.getargspec but without the warning being raised) and using that over inspect.getargspec if PY3


josham:
@dlstadther Added comment

dlstadther:
@josham could you merge in master to verify tests pass?

josham:
@dlstadther done

PR #2580, opened 2018-11-16
Silence warning: String parameter must have string default value.

lallea:
## Description

Silence a warning caused by luigi code.

No new test failures on my machine. The change itself did not seem worthy a test. The value is only read with "if self.context:" at https://github.com/spotify/luigi/blob/master/luigi/worker.py#L276


dlstadther:
I believe #2574 contains the more proper solution to this warning.

lallea:
I agree, closing this one.

Issue #2577, opened 2018-11-13
luigi.run() returns False on success


nreznik:
luigi.run() returns False when all tasks succeed with retries. Here's what we see in the log 

2018-11-13 03:19:26,569 luigi-interface INFO
===== Luigi Execution Summary =====

Scheduled 35 tasks of which:
* 35 ran successfully:
...
This progress looks :) because there were failed tasks but they all succeeded in a retry

===== Luigi Execution Summary =====

However luigi.run() returns False.  

We are using luigi==2.7.9.

dlstadther:
This is the current behavior. 

I've found this a very minor pain point. You can trace back from that method and see how the value is provided. (I did a couple months ago).

Feel free to submit a PR to resolve this [debatably] erroneous behavior in success response. :)

Tarrasch:
This isn't something I would like to change. Feel free to update docs though it the this behaviour isn't clear.

Tarrasch:
I'm afraid we'll break it for someone if we change the behaviour. Currently the luigi docs recommend `luigi.build()` or running it from the command line by the way.

I'd rather do a somewhat bigger effort and provide a new programmatic runner function that returns something like a `LuigiRunResult` that is more future proof than returning a bool.

As for your usage, wouldn't it be possible to cron using the `luigi` cmdline tool? It actually has some sort of error code reporting.

nreznik:
@Tarrasch How would luigi.build() fix the issue for us? It seems to return the same success value.  We are doing some other processing around the luigi call (metrics, etc.), so we call it from a script. Can you please point us to documentation on the luigi cmdline tool and error code reporting?

Tarrasch:
Sure, please see these docs:

 * https://luigi.readthedocs.io/en/stable/running_luigi.html
 * https://luigi.readthedocs.io/en/stable/configuration.html#retcode

Please let me know if this helps!

nreznik:
@Tarrasch Thank you! Can you please just confirm that even though we should use luigi.build(), it wouldn't fix the issue in question for us, since it would return the same value as luigi.run()?

Tarrasch:
Yes you're right. We're not exposing anything useful from ` build`/`run`. Perhaps one can implement something like I mentioned above (https://github.com/spotify/luigi/issues/2577#issuecomment-439631566)

dlstadther:
Closing due to #2630

PR #2582, opened 2018-11-17
Add some docs to interface.run

Tarrasch:
Slightly related to issue #2577.


Tarrasch:
I think (this was such a long time ago now):

 * Less entry points to maintain
 * It's not so nice that it parses the command line args I think.

PR #2583, opened 2018-11-17
Accept date_interval classes as arguments for relevant parameter types.

lallea:
## Description

Allow date_interval.{Month,Year} to be used as arguments for {Month,Year}Parameter.

## Motivation and Context

The date_interval classes are useful for date arithmetic. Allowing them to be used to fill parameter values saves a confusing '.date_a' when using them for calculating values in requires().

Added unit tests. Ran tests locally. No additional test failures.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -36,6 +36,7 @@ try:
 except ImportError:
     from configparser import NoOptionError, NoSectionError
 
+from luigi import date_interval
 from luigi import task_register
 from luigi import six
 from luigi import configuration
@@ -423,6 +424,9 @@ class DateParameter(_DateParameterBase):
         if value is None:
             return None
 
+        if isinstance(value, date_interval.Date):
+            value = value.date_a
+
         if isinstance(value, datetime.datetime):
             value = value.date()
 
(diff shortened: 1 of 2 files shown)


Tarrasch:
> Allowing them to be used to fill parameter values saves a confusing '.date_a' when using them for calculating values in requires().

Shouldn't one aim for "easy to read" code rather than "easy to write"? Won't a reader be a bit confused that it is possible to pass a pair of dates to something that expects a single date?

----

Also let's add docs if we decide to merge this.

lallea:
I agree, easy to read >> easy to write. See difference in example at end of comment. I find that the ".date_a" adds confusion for the reader. 

Parameters typically do not denote a point in time, but an interval, e.g. events collected during one day, active users during one month, etc. Looking again, however, I see that there is a DateIntervalParameter, and that the proposed change does not make sense for DateParameter.

I think that it does for MonthParameter and YearParameter, however. They match date_interval in the sense that they specify a specific, aligned date range.

I'll remove the change to DateParameter and add docs.


Before PR:

class MyRecursiveTask(Task):
  month = MonthParameter(default=Month.from_date(date.today()).prev().date_a)

  def requires(self):
    return {'previous_month": MyRecursive(Month.from_date(self.month).prev().date_a)}

After PR:

class MyRecursiveTask(Task):
  month = MonthParameter(default=Month.from_date(date.today()).prev())

  def requires(self):
    return {'previous_month": MyRecursive(Month.from_date(self.month).prev())}

lallea:
I corrected the docs a tiny bit.

dlstadther:
Could you merge master (for moto test success)?

lallea:
Merged master.

PR #2585, opened 2018-11-18
Support for Azure Blob Storage Target

manish:
Azure Blob Storage backend for Target. Transparently uploads and downloads text and binary objects to Azure Blob Storage

## Description

Provides a `AzureBlobTarget` and `AzureBlobClient`. When reading, `ReadableAzureBlobFile` is used and when writing `AtomicAzureBlobFile` is used.

Use account name + account key for authentication or use shared-access-signatures (like an OAuth token). For saving/loading file, provide the azure blob container and azure blob name. There is an option to not download the file and just get the specific bytes in memory including specific bytes range (advanced usage).

## Motivation and Context

While working on a project, I noticed that `S3Target` exists, but no such exists for Azure Blob Storage and I wanted to implement it since I wanted all my artifacts to remain under single cloud subscription (Azure)

I have written united tests which is `test/contrib/azureblob_test.py`. It requires azure account name + key or a SAS token to run. The values can be passed or can be exported as environment variables 

`ACCOUNT_NAME` - this is the [actual name of your azure blob storage account](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-dotnet?tabs=macos#copy-your-credentials-from-the-azure-portal).

`ACCOUNT_KEY` - this is the key (password) for the storage account. It has `root` like privileges, so not always a good idea to use it everywhere. [(Documentation)](https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1?toc=%2fazure%2fstorage%2fblobs%2ftoc.json)

`SAS_TOKEN` - A shared access signature (SAS) provides you with a way to grant limited access to objects in your storage account to other clients, without exposing your account key. [(Documentation)](https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1?toc=%2fazure%2fstorage%2fblobs%2ftoc.json)


luigi/contrib/azureblob.py
```diff
@@ -0,0 +1,286 @@
...
+                                                 account_key=self.options.get("account_key"),
+                                                 sas_token=self.options.get("sas_token"),
+                                                 protocol=self.kwargs.get("protocol"),
+                                                 connection_string=self.kwargs.get("connection_string"),
+                                                 endpoint_suffix=self.kwargs.get("endpoint_suffix"),
+                                                 custom_domain=self.kwargs.get("custom_domain"))
+
+    def upload(self, tmp_path, container, blob, **kwargs):
+        logging.debug("Uploading file '{tmp_path}' to container '{container}' and blob '{blob}'".format(
+            tmp_path=self.tmp_path, container=self.container, blob=self.blob))
```
dlstadther: shouldn't this be `tmp_path`, not `self.tmp_path`?

luigi/contrib/azureblob.py
```diff
@@ -0,0 +1,286 @@
...
+        lease_id = self.connection.acquire_blob_lease(container, blob)\
+            if self.exists("{container}/{blob}".format(container=container, blob=blob)) else None
+        try:
+            self.connection.create_blob_from_path(container, blob, tmp_path, lease_id=lease_id, progress_callback=kwargs.get("progress_callback"))
+        finally:
+            if lease_id is not None:
+                self.connection.release_blob_lease(container, blob, lease_id)
+
+    def download_as_bytes(self, container, blob, n=None):
+        start_range, end_range = (0, n-1) if n is not None else (None, None)
```
dlstadther: could `n` be renamed something more meaningful?

luigi/contrib/azureblob.py
```diff
@@ -0,0 +1,286 @@
...
+                destination_lease_id=destination_lease_id, source_lease_id=source_lease_id)
+        finally:
+            self.connection.release_blob_lease(source_container, source_blob, source_lease_id)
+            if destination_lease_id is not None:
+                self.connection.release_blob_lease(dest_container, dest_blob, destination_lease_id)
+
+    def rename_dont_move(self, path, dest):
+        self.move(path, dest)
+
+    def splitfilepath(self, filepath):
```
dlstadther: I wonder if you can use `urlsplit` like [s3 does](https://github.com/spotify/luigi/blob/master/luigi/contrib/s3.py#L524-L530). Not saying you have to, just know this similar idea is implemented

dlstadther: this should also be a `@staticmethod`

luigi/contrib/azureblob.py
```diff
@@ -0,0 +1,286 @@
...
+    def isdir(self, path):
+        """
+        Azure Blob Storage has no concept of directories. It always returns False
+        :param str path: Path of the Azure blob storage
+        :return: False
+        """
+        return False
+
+    def move(self, path, dest):
+        return self.copy(path, dest) and self.remove(path)
```
dlstadther: could `copy` succeed and `remove` fail? in that case, we have `path` in two locations

luigi/contrib/azureblob.py
```diff
@@ -0,0 +1,286 @@
...
+            raise FileAlreadyExists("The Azure blob path '{blob}' already exists under container '{container}'".format(
+                blob=blob, contaniner=container))
+
+    def isdir(self, path):
+        """
+        Azure Blob Storage has no concept of directories. It always returns False
+        :param str path: Path of the Azure blob storage
+        :return: False
+        """
+        return False
```
dlstadther: what value does this method provide?

luigi/contrib/azureblob.py
```diff
@@ -0,0 +1,286 @@
...
+            return self.connection.copy_blob(source_container, dest_blob, self.connection.make_blob_url(
+                source_container, source_blob),
+                destination_lease_id=destination_lease_id, source_lease_id=source_lease_id)
+        finally:
+            self.connection.release_blob_lease(source_container, source_blob, source_lease_id)
+            if destination_lease_id is not None:
+                self.connection.release_blob_lease(dest_container, dest_blob, destination_lease_id)
+
+    def rename_dont_move(self, path, dest):
+        self.move(path, dest)
```
dlstadther: Would it make more sense for this to be `move()` with some default option overwritten?

luigi/contrib/azureblob.py
```diff
@@ -0,0 +1,286 @@
...
+
+class ReadableAzureBlobFile(object):
+    def __init__(self, container, blob, client, download_when_reading, **kwargs):
+        self.container = container
+        self.blob = blob
+        self.client = client
+        self.closed = False
+        self.download_when_reading = download_when_reading
+        self.azure_blob_options = kwargs
+        self.download_file_location = os.path.join(tempfile.mkdtemp(prefix=datetime.datetime.now().__str__()), blob)
```
dlstadther: Why `now()` instead of `utcnow()`? Also, why `__str__()` instead of `str(datetime....)`?

test/contrib/azureblob_test.py
```diff
@@ -0,0 +1,164 @@
...
+Integration tests for azureblob module.
+"""
+
+import os
+import unittest
+
+import luigi
+from luigi.contrib.azureblob import AzureBlobClient, AzureBlobTarget
+
+client = AzureBlobClient(os.environ.get("ACCOUNT_NAME"), os.environ.get("ACCOUNT_KEY"), os.environ.get("SAS_TOKEN"))
```
dlstadther: Is there anyway to mock this for Travis testing?

luigi/contrib/azureblob.py
manish: I never thought it would make a difference as this is about creating a valid folder location on the disk.

test/contrib/azureblob_test.py
manish: Not to my knowledge. Azure Storage does come with an emulator but [unfortunately it only works on windows](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator)

There is an open source clone, but I was not very keen on adding yet another dependency to such a small test. I can DM you an account name/key if you want to try out these tests

luigi/contrib/azureblob.py
manish: `AzureBlobClient` implement `FileSystem` which has an method `def isdir(self, path):`, so I tried to implement it

luigi/contrib/azureblob.py
manish: Fixed this.

luigi/contrib/azureblob.py
manish: I checked the urlsplit, but most of the `?` part is not needed here. Given that blob storage is not hierarchal and has no inherent concept of folders, this method is basically splitting on first `/` encountered.

test/contrib/azureblob_test.py
```diff
@@ -0,0 +1,164 @@
...
+        m.update(datetime.datetime.now().__str__().encode())
+        container_name = m.hexdigest()
+
+        self.assertFalse(self.client.exists(container_name))
+        self.assertTrue(self.client.create_container(container_name))
+        self.assertTrue(self.client.exists(container_name))
+        self.client.delete_container(container_name)
+        self.assertFalse(self.client.exists(container_name))
+
+    def test_upload_copy_move_remove_blob(self):
```
dlstadther: Each method should have it's own tests. Why are they all being crammed into this one test?

test/contrib/azureblob_test.py
```diff
@@ -0,0 +1,164 @@
...
+        self.client.delete_container(container_name)
+        self.assertFalse(self.client.exists(container_name))
+
+
+class MovieScriptTask(luigi.Task):
+    def output(self):
+        return AzureBlobTarget("luigi-test", "movie-cheesy.txt", client, download_when_reading=False)
+
+    def run(self):
+        print(client.connection.create_container("luigi-test"))
```
dlstadther: why printing here?

test/contrib/azureblob_test.py
dlstadther: There look to be some some way to test this stuff without real accounts: https://github.com/Azure/azure-storage-python/tree/master/tests

And this is from the python SDK already being used by your additions.

luigi/contrib/azureblob.py
```diff
@@ -0,0 +1,291 @@
...
+        :return: False
+        """
+        return False
+
+    def move(self, path, dest):
+        try:
+            return self.copy(path, dest) and self.remove(path)
+        except:
+            self.remove(dest)
+            return False
```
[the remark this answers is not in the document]
manish: azure blob service has no API (to my knowledge) which moves a blob, so I had to implement it using un-atomic copy and remove.

luigi/contrib/azureblob.py
manish: this method `def rename_dont_move(self, path, dest):` overrides the method in `FileSystem` which is implemented in base class as

```python
warnings.warn("File system {} client doesn't support atomic mv.".format(self.__class__.__name__))
        if self.exists(dest):
            raise FileAlreadyExists()
        self.move(path, dest)
```

This implementation is not relevant to azure blob, so I had to override it.

test/contrib/azureblob_test.py
manish: I was trying to use it as a workflow since even if I move each to it's own test, they would still require each other

luigi/contrib/azureblob.py
dlstadther: Ohhhh. I didn't even look to see it was in Filesystem. Sorry about this unnecessary back and forth

:+1:

test/contrib/azureblob_test.py
dlstadther: understood

test/contrib/azureblob_test.py
manish: I am going to look into if it is possible to mock or create some kind of stubs. From the looks of the azure-storage-python tests, it looks like it uses some proxy, but I see no mocking.

I am going to try out https://github.com/azure/azurite but it would just end up introducing yet another dependency.

dlstadther: My main concern here is accepting a module which cannot be easily tested. It's one thing if it just can't be ran on Travis. But another if it can't be ran locally with tox. Of course, without an Azure account configured for these tests.

Just looking to minimize future maintenance burden :)

manish: Would making this addition to tox.ini work?

```
[testenv:azureblob]
passenv = ACCOUNT_NAME ACCOUNT_KEY SAS_TOKEN
deps =
  azure-storage
commands =
  python -m unittest -v test.contrib.azureblob_test
```

and append `azureblob` in 

```
[tox]
envlist = ..... azureblob
```

then run these tests with

```bash
tox -e azureblob
```

Pardon my inexperience. Python is not my primary language for bread and butter :)

manish: @dlstadther Would the above change in tox work?

dlstadther: It _would_ work, but that departs from our current test state. I'd really like to avoid requiring the use of actual credentials for testing

@Tarrasch Do you have any suggestions on best route to approve given the issues around testing without actual Azure credentials? I'd like not to keep blocking this PR, but I also wish to avoid maintaining code which requires manual setup to test.

shikharsg: @manish @dlstadther Re: "I'd really like to avoid requiring the use of actual credentials for testing"
Once the azure storage emulator is running(either [Azurite](https://github.com/Azure/Azurite) for linux or [emulator](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator) for windows), you just have to use the azure blob client like so:
```python
from azure.storage.blob import BlockBlobService

blob_client = BlockBlobService(is_emulated=True)  # will automatically use the default local connection string and keys
```
For both azurite and the windows emulator, the blob service runs on a default port and with a default key, which is built into the azure storage python client library(see [docs](https://azure-storage.readthedocs.io/ref/azure.storage.blob.blockblobservice.html)).

For an example of how to use azurite in travis, this is what I do [(here)](https://github.com/shikharsg/zarr/blob/abs_store/.travis.yml):
```
before_install:
  - docker pull arafato/azurite
  - mkdir ~/blob_emulator
  - docker run -e executable=blob -d -t -p 10000:10000 -v ~/blob_emulator:/opt/azurite/folder arafato/azurite
```

manish: I made the changes as you suggested, but I made it in tox so that I can run it on multiple python versions. Thanks a lot for your help @shikharsg

dlstadther:
The Azure tests, [denote that there were 17 tests run, 11 of them skipped](https://travis-ci.org/spotify/luigi/jobs/467570713#L664). So 6 actually ran which were the Azure tests. Do you know why the other 11 were attempted?

manish:
Those 11 tests were skipped as they did not satisfy the `NOSE_ATTR` value of `azureblob`. You can check that [many tests are skipped even in case of py37-aws](https://travis-ci.org/spotify/luigi/jobs/467570712#L840)

dlstadther:
Got it - Thanks @manish

manish:
Thank you a lot. Highly appreciate that you spent time to review and merge my PR.

PR #2588, opened 2018-11-22
Changed to buffered reads when using GCSTarget

Yarin78:
## Description
Buffers the file read operation after downloading a GCS file to the local file system.

## Motivation and Context
Reading unbuffered files can be veery slow.
This change is consistent with how the reading is implement in luigi.contrib.ftp.RemoteTarget.

### DIFF
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -32,6 +32,7 @@ from luigi.contrib import gcp
 import luigi.target
 from luigi import six
 from luigi.six.moves import xrange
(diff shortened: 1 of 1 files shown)


Yarin78:
Do I need to do anything here? I don't know how to make Travis kick in; it was supposed to happen automatically when the PR was open but for some reason it didn't. Ping @Tarrasch

Tarrasch:
I just did some high-tech hacking and now Travis seem to be running again. :)

If it passes let's just merge it.

PR #2591, opened 2018-11-28
contrib: Add ExternalDailySnapshot

brianmartin:
It is common to depend on the latest version of an ExternalTask, but not be concerned with
when that external task was generated.

I think this concern is generic enough to warrant having it upstream.

In addition to unit tests, this has been in wide use at Spotify for several years.

### DIFF
--- a/test/contrib/external_daily_snapshot_test.py
+++ b/test/contrib/external_daily_snapshot_test.py
@@ -1,0 +1,49 @@
+# Copyright (c) 2013 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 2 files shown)


test/contrib/external_daily_snapshot_test.py
```diff
@@ -0,0 +1,53 @@
...
+        MockTarget('data-pqr-zebra-Congo-2012-01-01').open('w').close()
+        d = DataDump.latest(date=datetime.date(2012, 1, 10), param='pqr', a='zebra', aa='Congo')
+        self.assertEquals(d.date, datetime.date(2012, 1, 1))
+
+        MockTarget('data-pqr-zebra-Congo-2012-01-05').open('w').close()
+        d = DataDump.latest(date=datetime.date(2012, 1, 10), param='pqr', aa='Congo', a='zebra')
+        self.assertEquals(d.date, datetime.date(2012, 1, 1))  # Should still be the same
+
+
+if __name__ == "__main__":
```
Tarrasch: I think you don't need this. See the docs for how we run the tests.

brianmartin: 👍 done

dlstadther:
We're good to go! Thanks for following up @brianmartin

PR #2593, opened 2018-11-29
Fix logging via toml

### LINKED ISSUE #2592: AttributeError during logging setup with Python 2.7
Hi,

I upgraded to luigi 2.8.0 and experience an `AttributeError` that occurs within the new `setup_logging.py`. Here is the important part of the stack trace:

```
  File ".../lib/python2.7/site-packages/luigi/cmdline.py", line 9, in luigi_run
    run_with_retcodes(argv)
  File ".../lib/python2.7/site-packages/luigi/retcodes.py", line 81, in run_with_retcodes
    InterfaceLogging.setup(env_params)
  File ".../lib/python2.7/site-packages/luigi/setup_logging.py", line 74, in setup
    configured = cls._section(opts)
  File ".../lib/python2.7/site-packages/luigi/setup_logging.py", line 42, in _section
    logging_config = cls.config['logging']
AttributeError: LuigiConfigParser instance has no attribute '__getitem__'
```

`cls.config` is a `ConfigParser` instance which does not implement `__getitem__` in Python 2.7 (which I'm forced to use -.-). This could be fixed by using `cls.config.options('logging')` instead. I checked, and this is the only place where items are used.

https://github.com/spotify/luigi/blob/0a098f6f99da0bad03af56a057b9a15254d1a957/luigi/setup_logging.py#L42

Adding @orsinium here who is maybe faster to fix this. I could create a PR on Friday.

orsinium:
## Description

Wrap AttributeError when try to configure logging via config section. Cfg file doesn't support inheritance enough for logging configuring, so we can skip logging configuring via config section when cfg-based config used.

## Motivation and Context

Fix #2592

Test included. Failed before fix, success after

### DIFF
--- a/luigi/setup_logging.py
+++ b/luigi/setup_logging.py
@@ -40,7 +40,7 @@ class BaseLogging(object):
         """Get logging settings from config file section "logging"."""
         try:
             logging_config = cls.config['logging']
-        except (TypeError, KeyError, NoSectionError):
+        except (TypeError, KeyError, NoSectionError, AttributeError):
             return False
         logging.config.dictConfig(logging_config)
         return True
--- a/luigi/setup_logging.py
+++ b/luigi/setup_logging.py
@@ -22,7 +22,7 @@ workers via command line arguments and options from config files.
 import logging
 import logging.config
 import os.path
-from luigi.configuration import get_config
+from luigi.configuration import get_config, LuigiConfigParser
 
 # In python3 ConfigParser was renamed
 # https://stackoverflow.com/a/41202010
@@ -38,9 +38,11 @@ class BaseLogging(object):
     @classmethod
     def _section(cls, opts):
         """Get logging settings from config file section "logging"."""
+        if isinstance(cls.config, LuigiConfigParser):
+            return False
         try:
             logging_config = cls.config['logging']
-        except (TypeError, KeyError, NoSectionError, AttributeError):
+        except (TypeError, KeyError, NoSectionError):
             return False
         logging.config.dictConfig(logging_config)
         return True
--- a/test/setup_logging_test.py
+++ b/test/setup_logging_test.py
@@ -1,5 +1,5 @@
 from luigi.setup_logging import DaemonLogging, InterfaceLogging
-from luigi.configuration import LuigiTomlParser, get_config
+from luigi.configuration import LuigiTomlParser, LuigiConfigParser, get_config
 from helpers import unittest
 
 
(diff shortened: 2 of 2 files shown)


luigi/setup_logging.py
```diff
@@ -40,7 +40,7 @@ def _section(cls, opts):
         """Get logging settings from config file section "logging"."""
         try:
             logging_config = cls.config['logging']
-        except (TypeError, KeyError, NoSectionError):
+        except (TypeError, KeyError, NoSectionError, AttributeError):
```
dlstadther: Do we want the attribute error to allow `logging_config = cls.config.options('logging')` like @riga suggested? Or is the best course of action to just return False?

orsinium: We can't configure logging via section in cfg. It have to be False for `LuigiConfigParser`.

orsinium: For section based logging configuring we need dicts in dists, but cfg format doesn't support it as opposite to toml, yaml, json and other good formats.

riga: > For section based logging configuring we need dicts in dists, but cfg format doesn't support it as opposite to toml, yaml, json and other good formats.

Is there a difference between Python 2 and 3?

orsinium: It's particular point of cfg format as is. There can be mistake only if cfg-based config has `logging` section, but there is no reason to do it.

riga: But don't you introduce a difference between 2 and 3 then? Might not be too important here, but still.

orsinium: Python version doesn't matter. In any python version cfg-based config doesn't contain `logging` section.

riga: Then I maybe just don't get why this block exists at all:

https://github.com/spotify/luigi/blob/8278773fe26bbd6cdb9d7b154a1e12fff254c81f/luigi/setup_logging.py#L74-L78

orsinium: It's True for toml-based config with `logging` section for any Python version.

orsinium: Also it will be true for YAML or JSON based configs in future. I hope, we will add it.

honnix:
As I commented https://github.com/spotify/luigi/issues/2592#issuecomment-446577334, 2.8.0 didn't have this issue before the bug was introduced in later PR. However we just release 2.8.1 yesterday that that became a broken version. Any progress on this PR?

luigi/setup_logging.py
honnix: I would also prefer not to catch `AttributeError` but handle this case explicitly.

orsinium: I've changed it

orsinium:
I've added explicit check for `LuigiConfigParser`.

honnix:
LGTM. @dlstadther @Tarrasch any thought? I wasn't following the original PR.

honnix:
@dlstadther please merge and release a hotfix, thanks!

dlstadther:
@honnix Done

honnix:
Awesome. Thanks!

Issue #2599, opened 2018-12-08
AtomicLocalFile: WindowsError 32 process cannot access the file because it is being used by another process


computercolin:
## Error
When AtomicLocalFile is removed via `__del__()` (or possibly through implicit `__del__` call from failed context manager exit #2087), `os.remove(...)` is called to delete the temp file, if it exists.

On windows, this call to `os.remove` fails with WindowsError
```
Exception WindowsError: (32, 'The process cannot access the file because it is being used by
another process', 'c:\users\user\appdata\local\temp\luigi-s3-tmp-6685635592') in
<bound method TestAtomicLocalFileExcepBehavior.del of > ignored
```
because the file is still open.. by python itself.. in our own `AtomicLocalFile` (doh!)

On linux, `os.remove` succeeds. It seems either `close()` is implicitly called, the linux py implementation detects that python-bin is the conflicting owner and handles it, or just linux is more flexible and lets you unlink the file, even though it is still open, or...

Anyway, solution is simple: make explicit call to `close()`. I have solution, makes sense to combine with patch for #2087 (more info there).

## Try it yourself
```python
import luigi
af = luigi.local_target.atomic_file('nopersist_foo')
del af
#  Produces above exception.
```

Issue #2087, opened 2017-04-07
luigi.target.AtomicLocalFile not so atomic after all


redhog:
## Symptom
The following code should not print "MOVE foo" due to the exception being thrown, but does when the interpreter exits:

    import luigi.target
    class X(luigi.target.AtomicLocalFile):
        def move_to_final_destination(self):
            print "MOVE %s" % self.path
    try:
        with X("foo") as f:
            f.write()
            raise Exception
    except:
        pass

## Cause
__del__ in the baseclass calls close() (see https://docs.python.org/2/library/io.html#io.IOBase.__del__), which uploads the file. The luigi authors have tried to fix this by overriding __del__. But the base class is implemented in C and the __del__ method of the baseclass is called regardless of the override.

## Solution
Here is a monkey-patch that fix this by setting a flag if an exception occures, and checking the flag in close():

    import luigi.target

    luigi.target.AtomicLocalFile.error = False
    def _AtomicLocalFile_close(self):
        super(luigi.target.AtomicLocalFile, self).close()
        if not self.error:
            self.move_to_final_destination()
    luigi.target.AtomicLocalFile.close = _AtomicLocalFile_close

    def _AtomicLocalFile__exit__(self, exc_type, exc, traceback):
        " Close/commit the file if there are no exception "
        if exc_type:
            self.error = True
            return
        return super(luigi.target.AtomicLocalFile, self).__exit__(exc_type, exc, traceback)
    luigi.target.AtomicLocalFile.__exit__ = _AtomicLocalFile__exit__

kalvdans:
It is better to not overload __del__ and instead force everyone to explicitly call close() or use context managers. I can prepare a patch if you agree with me.

redhog:
Please do not close. Please *fix*

dlstadther:
Hey @redhog Thanks for re-raising this issue! Our use of stale bot was not to ignore many PRs and Issues, but merely to identify those which still need our attention and have authors who will help push them through.

I'd be very happy to see a PR submitted to resolve the issue you've stated above! Feel free to be the author of that submission :)

redhog:
Oh no, not you again, stale bot ;)

computercolin:
I figured, since I'm looking at `AtomicLocalFile` as a "Target" (#2598) I would submit fix for this. But I had the misfortune of testing from Windows and quickly discovered more bugs! What a reward.

[...]

PR #2600, opened 2018-12-08
Fix AtomicLocalFile #2087 erroneous move (py2), #2599 dangling open file handle (win,nix?)

computercolin:
## Description
Fixes to AtomicLocalFile to fix 

## Motivation and Context
Bugs where AtomicLocalFile
  - erroneously commits failed data
  - fails to cleanup open file handles
  - raises "file open in other process" error and dies (on windows)

Only by hand. Tested with simple task pipe, and tested using reproduction code in #2087 and #2599

### DIFF
--- a/luigi/target.py
+++ b/luigi/target.py
@@ -316,18 +316,23 @@ class AtomicLocalFile(io.BufferedWriter):
     a temporary file in the local filesystem before
     moving it to its final destination.
 
-    This class is just for the writing part of the Target. See
-    :class:`luigi.file.LocalTarget` for example
+    This class handles only creation, cleanup of on-disk tmpfile.
+    Subclasses must implement `move_move_to_final_destination`.
+
+    :class:`luigi.local_target.atomic_file` example of local-endpoint atomic-file.
+    :class:`luigi.contrib.s3.AtomicS3File` example of remote-endpoint usage.
     """
 
     def __init__(self, path):
         self.__tmp_path = self.generate_tmp_path(path)
         self.path = path
+        self.unintended_close = False
         super(AtomicLocalFile, self).__init__(io.FileIO(self.__tmp_path, 'w'))
 
     def close(self):
         super(AtomicLocalFile, self).close()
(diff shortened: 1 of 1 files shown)


dlstadther:
@computercolin are there any tests which should already be covering this patch that weren't?

luigi/target.py
```diff
@@ -316,18 +316,23 @@ class AtomicLocalFile(io.BufferedWriter):
...
+    Subclasses must implement `move_move_to_final_destination`.
+
+    :class:`luigi.local_target.atomic_file` example of local-endpoint atomic-file.
+    :class:`luigi.contrib.s3.AtomicS3File` example of remote-endpoint usage.
     """
 
     def __init__(self, path):
         self.__tmp_path = self.generate_tmp_path(path)
         self.path = path
+        self.unintended_close = False
```
Tarrasch: Make this private?

luigi/target.py
```diff
@@ -336,6 +341,8 @@ def move_to_final_destination(self):
         raise NotImplementedError()
 
     def __del__(self):
+        self.unintended_close = True
+        self.close()
```
Tarrasch: This pattern is hard to read I think. If `close()` depends on the attribute `unintended_close` it make much more sense for it to be a parameter to `close()`

luigi/target.py
Tarrasch: Also this variable name is not descriptive I think. I don't understand what it does.

luigi/target.py
```diff
@@ -316,18 +316,23 @@ class AtomicLocalFile(io.BufferedWriter):
     a temporary file in the local filesystem before
     moving it to its final destination.
 
-    This class is just for the writing part of the Target. See
-    :class:`luigi.file.LocalTarget` for example
+    This class handles only creation, cleanup of on-disk tmpfile.
+    Subclasses must implement `move_move_to_final_destination`.
```
Tarrasch: is `move_move_to_final_destination` correctly spelled?

Issue #2604, opened 2018-12-11
cannot bind an existing table in a specific schema for sqla.CopyToTable when reflect is True


garaud:
Hi,

I would like to use `contrib.sqla.CopyToTable` for a table in a specific schema with the parameter `reflect=True`. SqlAlchemy can find by itself the column types and names since my table is already in my database. This table is not in the `public` schema. Thus, I use the class parameter `schema`.

Nevertheless, I got this traceback because `self.table_bound` is `None`. in other words, the reflection does not work.

```
Traceback (most recent call last):
  File "venv/lib/python3.6/site-packages/luigi/worker.py", line 199, in run
    new_deps = self._run_get_new_deps()
  File "venv/lib/python3.6/site-packages/luigi/worker.py", line 139, in _run_get_new_deps
    task_gen = self.task.run()
  File "venv/lib/python3.6/site-packages/luigi/contrib/sqla.py", line 391, in run
    for row in itertools.islice(rows, self.chunk_size)]
  File "venv/lib/python3.6/site-packages/luigi/contrib/sqla.py", line 391, in <listcomp>
    for row in itertools.islice(rows, self.chunk_size)]
AttributeError: 'MyTask' object has no attribute 'table_bound'
```

Here an example of my task:

```python

class MyTask(sqla.CopyToTable):
    schema = "jazz"
    table = "player"
    column_separator = ","
    connection_string = "postgresql:///jazzy"
    reflect = True

    def requires(self):
        return PlayerTaskFile()
```

I think I spot the problem.

* when the schema is not None, the `metadata` is built with the `schema` parameter https://github.com/spotify/luigi/blob/master/luigi/contrib/sqla.py#L343
* in this case, it's not necessary to specify the name of the schema at https://github.com/spotify/luigi/blob/master/luigi/contrib/sqla.py#L354

These lines work good for me:

```python
from sqlalchemy import create_engine, MetaData

eng = create_engine("postgresql:///jazzy")
meta = MetaData(schema="jazz")
meta.reflect(only=["player"], bind=eng)
# instead of meta.reflect(only=["jazz.player"], bind=eng)
meta.tables["jazz.player"]
```

Thanks
Damien G.

dlstadther:
Is this #2605 ?

I'd like to be sure that the patch works for both reflect true and reflect false. I don't utilize this module, so i'm hoping you can grab someone who has contributed to this module for review. Otherwise, I'll investigate further and see what i can understand and contribute.

Thanks!

[...]

PR #2605, opened 2018-12-11
Fix the table name when reflect is True in sqla.CopyToTable (#2604)

garaud:
When the schema is specified, MetaData is already parametrized with the right
name. No need to call "meta.reflect" with the name of the schema in this case.

### DIFF
--- a/luigi/contrib/sqla.py
+++ b/luigi/contrib/sqla.py
@@ -351,7 +351,7 @@ class CopyToTable(luigi.Task):
                         metadata.create_all(engine)
                     else:
                         full_table = '.'.join([self.schema, self.table]) if self.schema else self.table
-                        metadata.reflect(only=[full_table], bind=engine)
+                        metadata.reflect(only=[self.table], bind=engine)
                         self.table_bound = metadata.tables[full_table]
                 except Exception as e:
                     self._logger.exception(self.table + str(e))


garaud:
I launched

`tox -e core test/contrib/sqla_test.py` but 0 tests was ran.

dlstadther:
Sqla is in the [`contrib` tox attrib group](https://github.com/spotify/luigi/blob/master/test/contrib/sqla_test.py#L54)

so:
```shell
tox -e contrib test/contrib/sqla_test.py
```

dlstadther:
@sd2k I believe you added this bit of code. Could you review this small patch? Thanks!

sd2k:
Sure, I just tested this locally and it looks good (the old code failed), and it makes sense looking at the SQLAlchemy docs.

I don't think there's an easy way to add a test for it unfortunately since as far as I know SQLite (used in the tests) doesn't support schemas.

garaud:
Thanks @sd2k for the review and for testing this locally.

>  as far as I know SQLite (used in the tests) doesn't support schemas.

I think you're right.

garaud:
Thank you for the review. Do you think this can be merged in the future release? Thanks.

garaud:
Thanks!! :slightly_smiling_face:

PR #2615, opened 2019-01-03
Update postgres.py

dhuckins:
added port to `PostgresTarget`

## Description
 
added port to the postgres.Query

## Motivation and Context

was required to use `postgres.Query` to connect to a postgres db that did not use the default port

### DIFF
--- a/luigi/contrib/postgres.py
+++ b/luigi/contrib/postgres.py
@@ -392,5 +392,6 @@ class PostgresQuery(rdbms.Query):
             user=self.user,
             password=self.password,
             table=self.table,
-            update_id=self.update_id
+            update_id=self.update_id,
+            port=self.port
         )
--- a/test/contrib/postgres_test.py
+++ b/test/contrib/postgres_test.py
@@ -48,6 +48,7 @@ class DummyPostgresImporter(luigi.contrib.postgres.CopyToTable):
     date = luigi.DateParameter()
 
     host = 'dummy_host'
+    port = 1234
     database = 'dummy_database'
     user = 'dummy_user'
     password = 'dummy_password'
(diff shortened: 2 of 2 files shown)


dhuckins:
not sure what failed here

dlstadther:
That top codecov can be ignored. The important part is that Travis passed.

duckontheweb:
Hey, just FYI that this seems to have created a breaking change in a patch release. The previous way of defining ports was within the `host` property on the class, like so:

```python
class MyQueryTask(luigi.contrib.postgres.PostgresQuery):
    host = ':'.join([db_host, db_port])
    database = db_name
    user = db_user
    password = db_password
```

This now raises an exception like the following:

```console
  File "/usr/local/lib/python3.6/site-packages/luigi/worker.py", line 401, in check_complete
    is_complete = task.complete()
  File "/usr/local/lib/python3.6/site-packages/luigi/task.py", line 827, in complete
  File "/usr/local/lib/python3.6/site-packages/luigi/task.py", line 827, in <genexpr>
  File "/usr/local/lib/python3.6/site-packages/luigi/task.py", line 565, in complete
    outputs = flatten(self.output())
  File "/usr/local/lib/python3.6/site-packages/luigi/contrib/postgres.py", line 396, in output
    port=self.port
AttributeError: 'MyQueryTask' object has no attribute 'port'
```

It might be worth making this backwards compatible with the old format.

dlstadther:
@dhuckins  is this something you'd like to look into and patch?

dhuckins:
sure, added issue #2625
there's something similar https://github.com/spotify/luigi/blob/eadf7ab8a487740e38e6fe9f63b3cba6e31e6caf/luigi/contrib/mssqldb.py#L59-L64

honnix:
I think this PR is not really needed. I suggest we revert it.

honnix:
cc @NatashaL

NatashaL:
Fine by me. Then we close #2627 too

dhuckins:
shouldn't `Query` and `CopyToTable`  have the same parameters?

honnix:
They should, so #2627 adds that.

PR #2618, opened 2019-01-09
Use proper API call in bigtable.py's make_dataset

bardb:
## Description

There used to be two ways to make a dataset in the Google BigQuery API: the correct new way and the obsolete old way.  Of course our code used the obsolete old way.  Google seems to have taken the obsolete old way down.  This PR is to do it the correct new way instead.

## Motivation and Context

The BigQuery make_dataset API has two fields, "datasetId" and "id", which could be used to specify a dataset name.  The documentation for "id" says to use "datasetId" when creating a dataset. Our make_dataset method of course uses "id".  I conjecture that Google used to use "id" if "datasetId" were missing, but changed their impl to match the docco recently.

The make_dataset method in bigquery.py of course uses the "id" approach rather than the "datasetId", so the conjectured change would have broken it.

https://jira.spotify.net/browse/INCIDENT-8998

I confirmed this by cloning the relevant code and trying to create a dataset with it.  The old way fails with the same characteristic message that our users have been seeing.  The new way works.

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -187,7 +187,12 @@ class BigQueryClient(object):
             body = {}
 
         try:
-            body['id'] = '{}:{}'.format(dataset.project_id, dataset.dataset_id)
+            # Construct a message body in the format required by
+            # https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/python/latest/bigquery_v2.datasets.html#insert
+            body['datasetReference'] = {
+                'projectId': dataset.project_id,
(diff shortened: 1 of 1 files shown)


dlstadther:
@ulzha Could you take a look here - as I understand Spotify uses BigQuery?

The motivation and reasoning makes sense to me.

kanterov:
I can confirm that it works, and I just did exactly the same fix.

bardb:
Thanks! I don't have write access -- could someone merge it for me?

PR #2622, opened 2019-01-16
Add version cap to hypothesis[datetime]

dlstadther:
## Description

Lock hypothesis[datetime] version

## Motivation and Context

Hypothesis released version 4.0.0 and removed the import path which we use for tests.

Travis


Tarrasch:
Thanks!

PR #2623, opened 2019-01-16
setup.py: Support older setuptools (<=20.1.1)

brianmartin:
The `python_version` conditional was added fairly recently (<2 years).
We've run into a few cases where it would be very convenient for luigi
to support older setuptools (e.g. managed environments where it may be
difficult or not possible to upgrade).

The `python_version` conditional was introduced in luigi `2.8.x`. For now 
we're pinning to `<2.8.x` as a workaround.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@
 # the License.
 
 import os
+import sys
 
 from setuptools import setup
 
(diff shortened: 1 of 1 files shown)


brianmartin:
@dlstadther 👍 done and thanks for the review!

dlstadther:
Final verification from, does this code work for you in production?

brianmartin:
We haven't used in production (a bit tricky short of depending on the luigi git master), but I repro'd exactly the same error and verified this fix using the lower setuptools version locally.

brianmartin:
Gentle bump here. (It seems the failing codecov check may be spurious)

dlstadther:
In #2608, @honnix suggests we encourage people to update setuptools rather than coding around older versions

I'm on the fence with this issue. Upgrading setuptools is definitely encouraged, but understandably not everyone has access/permission to do so.

brianmartin:
Ah, I hadn't seen #2608, thanks for linking! I can see both sides as well.

In this case, I see the cost as quite low, which makes it worth it.

To provide a bit more detail, the managed environment I'm referring to is the Google Cloud AI/ML environment. Based on their [version list](https://cloud.google.com/ml-engine/docs/tensorflow/runtime-version-list) they're using the [Ubuntu-provided python-setuptools](https://packages.ubuntu.com/trusty/python-setuptools), which for trusty (14.04LTS) is 3.3 and for xenial (16.04LTS) is 20.7. That Google Cloud ML version list is underspecified as they don't specify which Ubuntu release, but based on the behavior we're seeing it must be trusty or older.

As for non-managed environments, I might be on the other side of the argument!

honnix:
@brianmartin Totally understand.

brianmartin:
Hey @honnix, are you saying you approve? cc @dlstadther

honnix:
Yeah. Sorry I was not clear about it.

PR #2624, opened 2019-01-16
Fix durations in D3 graph (fixes #2620)

### LINKED ISSUE #2620: Task visualiser: time values of D3 graph edges don't make sense?
When using the D3 graph visualization, the time values at the edges don't seem to correspond to the actual duration of that task. The [code](https://github.com/spotify/luigi/blob/b2a5759108eabf657c1495dac32d48801fb64e3a/luigi/static/visualiser/js/visualiserApp.js#L721) indicates that it's a duration time.

Here's an example script. It runs a `TestTask` which wraps 4 `DummyTask`s that each create a temporary file and sleep for 1 second.

So, running the whole pipeline should take a bit more than 4 seconds in total.

```python
import time

import luigi


class DummyTask(luigi.Task):
    tmp_path = luigi.Parameter()

    def output(self):
        return luigi.LocalTarget(path=self.tmp_path, is_tmp=True)

    def run(self):
        open(self.tmp_path, 'w').write("")
        time.sleep(1)


class TestTask(luigi.WrapperTask):
    def requires(self):
        for path in ["/tmp/a", "/tmp/b", "/tmp/c", "/tmp/d"]:
            yield DummyTask(tmp_path=path)


if __name__ == "__main__":
    luigi.build([TestTask()])
```

When I run a `luigid` daemon and then start the script:
```
time python test.py
```
`time` reports a bit more than 4 seconds as expected.

However, the D3 graph in the web viewer shows this:
screenshot from 2019-01-15 18-32-11

To add to the confusion, not even the `luigid` daemon was running as long as any of the shown times in this example.

Could this be a time conversion bug or am I understanding something wrong?

MichaelGrupp:
Fixes the duration calculation. Both date objects `startTime` and
`finishTime` are in milliseconds, so the additional multiplication with
1000 was wrong.

Also, the duration is now calculated with:
```
duration = last_updated - time_running
```
as it is done also in https://github.com/spotify/luigi/blob/eadf7ab8a487740e38e6fe9f63b3cba6e31e6caf/luigi/contrib/datadog_metric.py#L83

The display format was also problematic, because it only displayed the
seconds of the `Date` object, i.e. always less than 60 seconds. Now it
uses the `hh:mm:ss:f` format.

I decided to not display the time if the duration is longer than a day,
because then it gets messy with the Date objects (would require a
conversion to day of year, not day of month).

### DIFF
--- a/luigi/static/visualiser/js/visualiserApp.js
+++ b/luigi/static/visualiser/js/visualiserApp.js
@@ -631,19 +631,19 @@ function visualiserApp(luigi) {
         }
     }
 
-    function getFinishTime(tasks, listId){
-        var times = {};
+    function getDurations(tasks, listId){
+        var durations = {};
         for (var i = 0; i < listId.length; i++) {
             for (var j = 0; j < tasks.length; j++) {
                 if (listId[i] === tasks[j].taskId) {
-                    var finishTime = new Date(tasks[j].time_running*1000);
-                    var startTime = new Date(tasks[j].start_time*1000);
-                    var durationTime = new Date((finishTime - startTime)*1000).getSeconds();
-                    times[listId[i]] = durationTime;
+                    // The duration of the task from when it started running to when it finished.
+                    var finishTime = new Date(tasks[j].last_updated*1000);
+                    var startTime = new Date(tasks[j].time_running*1000);
+                    durations[listId[i]] = new Date(finishTime - startTime);
                 }
             }
         }
-        return times;
+        return durations;
     }
 
     function getParam(tasks, id){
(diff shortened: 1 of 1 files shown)


luigi/static/visualiser/js/visualiserApp.js
```diff
@@ -718,11 +718,21 @@ function visualiserApp(luigi) {
...
+                                var oneDayInMilliseconds = 24 * 60 * 60 * 1000
+                                if (duration.getTime() < oneDayInMilliseconds) {
+                                    // Label task duration in stripped ISO format (hh:mm:ss:f)
+                                    g.setEdge(task.inputQueue[i], task.taskId, {
+                                        label: duration.toISOString().substr(11, 12),
+                                        width: 40
+                                    });
+                                } else {
+                                    g.setEdge(task.inputQueue[i], task.taskId, {
+                                        label: "> 24h",
```
Tarrasch: Can you remove this repetition?

MichaelGrupp: Sorry for the slow response. Fixed in the latest commit.

MichaelGrupp:
We tested this patch with a bunch of pipelines and it worked fine.

In the long run, I think it makes sense to refactor the code so that the durations are displayed in the node box - but I suggest to merge this first so we don't have wrong values anymore :)

PR #2627, opened 2019-01-17
Add port field for PostgresQuery (fixes #2625)

### LINKED ISSUE #2625: PR on postgres.py broke backwards compat
reported by @duckontheweb on #2615 
```python
class MyQueryTask(luigi.contrib.postgres.PostgresQuery):
    host = ':'.join([db_host, db_port])
    database = db_name
    user = db_user
    password = db_password
```
This now raises an exception like the following:
```
  File "/usr/local/lib/python3.6/site-packages/luigi/worker.py", line 401, in check_complete
    is_complete = task.complete()
  File "/usr/local/lib/python3.6/site-packages/luigi/task.py", line 827, in complete
  File "/usr/local/lib/python3.6/site-packages/luigi/task.py", line 827, in <genexpr>
  File "/usr/local/lib/python3.6/site-packages/luigi/task.py", line 565, in complete
    outputs = flatten(self.output())
  File "/usr/local/lib/python3.6/site-packages/luigi/contrib/postgres.py", line 396, in output
    port=self.port
AttributeError: 'MyQueryTask' object has no attribute 'port'
```

NatashaL:
Addresses https://github.com/spotify/luigi/issues/2625.

## Description

Init an empty port field for PostgresQuery. This will maintain default behaviour of the PostgresQuery and PostgresTarget classes, and users can keep using it by providing the port only as part of the host string.

## Motivation and Context

Addresses https://github.com/spotify/luigi/issues/2625 so as to have a non-breaking version update when using postgres.

Updated existing unit tests and added one more test.

### DIFF
--- a/luigi/contrib/postgres.py
+++ b/luigi/contrib/postgres.py
@@ -363,9 +363,6 @@ class PostgresQuery(rdbms.Query):
 
     To customize the query signature as recorded in the database marker table, override the `update_id` property.
     """
-
-    port = None
-
     def run(self):
         connection = self.output().connect()
         connection.autocommit = self.autocommit
--- a/luigi/contrib/postgres.py
+++ b/luigi/contrib/postgres.py
@@ -364,6 +364,8 @@ class PostgresQuery(rdbms.Query):
     To customize the query signature as recorded in the database marker table, override the `update_id` property.
     """
 
+    port = None
+
     def run(self):
         connection = self.output().connect()
         connection.autocommit = self.autocommit
--- a/luigi/contrib/rdbms.py
+++ b/luigi/contrib/rdbms.py
@@ -288,6 +288,13 @@ class Query(luigi.task.MixinNaiveBulkComplete, luigi.Task):
     def host(self):
         return None
 
+    @property
+    def port(self):
+        """
+        Override to use port separately from host. Used in PostgresQuery.
+        """
+        return None
+
     @abc.abstractproperty
     def database(self):
         return None
--- a/luigi/contrib/rdbms.py
+++ b/luigi/contrib/rdbms.py
@@ -343,4 +343,3 @@ class Query(luigi.task.MixinNaiveBulkComplete, luigi.Task):
         Override with an RDBMS Target (e.g. PostgresTarget or RedshiftTarget) to record execution in a marker table
         """
         raise NotImplementedError("This method must be overridden")
-
--- a/luigi/contrib/rdbms.py
+++ b/luigi/contrib/rdbms.py
@@ -277,7 +277,9 @@ class Query(luigi.task.MixinNaiveBulkComplete, luigi.Task):
 
         Optionally override:
 
+        * `port`,
         * `autocommit`
+        * `update_id`
 
         Subclass and override the following methods:
 
@@ -291,7 +293,7 @@ class Query(luigi.task.MixinNaiveBulkComplete, luigi.Task):
     @property
     def port(self):
         """
-        Override to use port separately from host. Used in PostgresQuery.
+        Override to specify port separately from host.
         """
         return None
 
--- a/luigi/contrib/rdbms.py
+++ b/luigi/contrib/rdbms.py
@@ -283,11 +283,16 @@ class Query(luigi.task.MixinNaiveBulkComplete, luigi.Task):
 
         Subclass and override the following methods:
 
+        * `run`
         * `output`
     """
 
     @abc.abstractproperty
     def host(self):
+        """
+        Host of the RDBMS. Implementation should support `hostname:port`
+        to encode port.
+        """
         return None
 
     @property
@@ -321,6 +326,13 @@ class Query(luigi.task.MixinNaiveBulkComplete, luigi.Task):
     def autocommit(self):
         return False
 
+    @property
+    def update_id(self):
+        """
+        Override to create a custom marker table 'update_id' signature for Query subclass task instances
+        """
+        return self.task_id
+
     @abc.abstractmethod
     def run(self):
         raise NotImplementedError("This method must be overridden")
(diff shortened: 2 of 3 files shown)


luigi/contrib/postgres.py
```diff
@@ -364,6 +364,8 @@ class PostgresQuery(rdbms.Query):
     To customize the query signature as recorded in the database marker table, override the `update_id` property.
     """
 
+    port = None
```
honnix: Why is `port` not defined by the base class? https://github.com/spotify/luigi/blob/master/luigi/contrib/rdbms.py#L263

honnix: OK host includes both hostname and port.

honnix: We should handle this similarly like: https://github.com/dhuckins/luigi/blob/de80e083c09221652420cc643d2cd817fff6028e/luigi/contrib/postgres.py#L128

Also please update the docstring to include `port`.

Actually I would prefer we revert https://github.com/spotify/luigi/pull/2615 because it is not needed. Encoding port number as part of `host` would achieve the same, right?

NatashaL: yes, looks like that indeed

honnix: Or we can define `port` property in `rdbms.Query`, which seems more consistent to me. Something like: https://github.com/spotify/luigi/pull/2561/files#diff-1e59aaa5da6736162dabbdec5c5e17c2R185

honnix:
@NatashaL I will take over this PR.

honnix:
@dhuckins PTAL

dhuckins:
not sure how this fixes https://github.com/spotify/luigi/issues/2625 though? the problem comes up when combining the port and hostname `host = 'localhost:5432'

honnix:
@dhuckins Does that answer your question? Removing of `port = 1234` in test cases ensures the backward compatibility.

dhuckins:
LGTM then

honnix:
@dlstadther Could you please take a look and approve it if looks good? Thanks.

luigi/contrib/rdbms.py
```diff
@@ -277,7 +277,9 @@ class Query(luigi.task.MixinNaiveBulkComplete, luigi.Task):
 
         Optionally override:
 
+        * `port`,
         * `autocommit`
+        * `update_id`
```
dlstadther: I'm not able to comment on the specific line, i want. But since we're updating this docstring for other portions which should have already been included (e.g `update_id`), shouldn't we add `run` to the `Subclass and override the following methods:`?

luigi/contrib/rdbms.py
```diff
@@ -288,6 +290,13 @@ class Query(luigi.task.MixinNaiveBulkComplete, luigi.Task):
     def host(self):
         return None
 
+    @property
+    def port(self):
+        """
+        Override to specify port separately from host.
```
dlstadther: Should we be more clear here about `host:port` format?

test/contrib/postgres_test.py
```diff
@@ -125,6 +135,14 @@ def test_bulk_complete(self, mock_connect):
         self.assertFalse(task.complete())
 
 
+@attr('postgres')
+class PostgresQueryWithPortTest(unittest.TestCase):
```
dlstadther: Why does this need its own `TestCase`? I feel it belongs within `PostgresQueryTest` since we aren't doing any `setUp()` logic with initializing the desired DummyPostgresQuery class of choice.

luigi/contrib/rdbms.py
honnix: That was actually something very confusing to me. The abstract property here doesn't say anything about that format and basically leaves it to subclass (even worse the actual Target such as https://github.com/spotify/luigi/blob/198dfa1c20acdacd150cf7818f716c2dd31936ef/luigi/contrib/postgres.py#L128) to interpret. It feels a bit weird if we disclose the format here while don't somehow enforce it in the base class.

honnix: Tried to "enforce" it in https://github.com/spotify/luigi/pull/2627/commits/0f77c2336039312263e4b8e4b98aaa1d51df726c

test/contrib/postgres_test.py
```diff
@@ -124,6 +126,10 @@ def test_bulk_complete(self, mock_connect):
         ])
         self.assertFalse(task.complete())
 
+    def test_override_port(self):
+        output = DummyPostgresQueryWithPort(date=datetime.datetime(1991, 3, 24)).output()
+        self.assertEquals(output.port, 1234)
```
dlstadther: Could we also add a test for the `host:port` parsing? I'm not seeing one here

honnix: Should that be part of this PR? I think it might fit better to test https://github.com/spotify/luigi/blob/master/luigi/contrib/postgres.py#L101 instead, because that's where parsing happens.

dlstadther: I'm fine with that :+1:

honnix: I gave it a try. PTAL again. https://github.com/spotify/luigi/pull/2627/commits/22a6566043155f43c803e6694c0d56d60fea7201

test/contrib/postgres_test.py
```diff
@@ -124,6 +130,14 @@ def test_bulk_complete(self, mock_connect):
...
+
+    def test_port_encoded_in_host(self):
+        output = DummyPostgresQueryWithPortEncodedInHost(date=datetime.datetime(1991, 3, 24)).output()
+        self.assertEquals(output.port, '1234')
```
honnix: This is rather unpleasant...

dlstadther: Agreed.

The parsing is just done as a string split.. 

I'd suggest we leave this test here for this branch (for the sake of current state test success). Then address the fact that port could be a str if interpreted or an int/str if provided manually in a separate PR.

honnix: Will merge then. Thanks for reviewing.

dlstadther: Thank you for pushing back against the quick fix and taking over the more appropriate solutions! I greatly appreciate your involvement and look forward to the future aid in maintaining Luigi! :)

PR #2434, opened 2018-05-30
Add Datadog contrib for monitoring purpose

cabouffard:
## Description
Datadog is a tool that allows you to send metrics that you create
dashboard and add alerting on specific behaviors.

Adding this contrib will allow for users of this tool to log their pipeline
information to Datadog.

## Motivation and Context
Based on the status change of a task, we log that information to Datadog
with the parameters that were used to run that specific task.

This allows us to easily create dashboards to visualize the health. For
example, we can be notified via Datadog if a task has failed, or we can
graph the execution time of a specific task over a period of time.

The implementation idea was strongly based on the stale PR
https://github.com/spotify/luigi/pull/2044.

## Have you tested this? If so, how?
We've been using this contrib for multiple months now (maybe a year?), at Glossier. This is the main point of reference to see the health of our pipeline.

PR #2628, opened 2019-01-17
Add Prometheus contrib for monitoring purpose

victoriaalee:
## Description
Prometheus is a tool that scrapes exposed metrics endpoints and allows for you to visualize and alert on the metrics. Adding this contrib will allow for users to monitor Luigi tasks via Prometheus.

## Motivation and Context
Based on the status change of a task, Prometheus metrics are updated with the parameters that were used to run that specific task.

The implementation idea was strongly based on the merged PR for Datadog #2434.

I have included unit tests.


luigi/server.py
```diff
@@ -274,6 +275,17 @@ def get(self):
         self.redirect("/static/visualiser/index.html")
 
 
+class MetricsHandler(tornado.web.RequestHandler):
+    def initialize(self, scheduler):
+        self._scheduler = scheduler
+
+    def get(self):
+        metrics = self._scheduler._state._metrics_collector.generate_latest()
```
Tarrasch: Is the `generate_latest` assuming we use Prometheus? If so let's add some docs?

luigi/server.py
```diff
@@ -287,7 +299,8 @@ def app(scheduler):
         (r'/history', RecentRunHandler, {'scheduler': scheduler}),
         (r'/history/by_name/(.*?)', ByNameHandler, {'scheduler': scheduler}),
         (r'/history/by_id/(.*?)', ByIdHandler, {'scheduler': scheduler}),
-        (r'/history/by_params/(.*?)', ByParamsHandler, {'scheduler': scheduler})
+        (r'/history/by_params/(.*?)', ByParamsHandler, {'scheduler': scheduler}),
+        (r'/metrics', MetricsHandler, {'scheduler': scheduler})
```
Tarrasch: Wow, how does this look like? Can you add screenshots?

Tarrasch:
Cool. I'm fine to merge this in 1-2 weeks then I think, so other's have a chance to try it out or give input.

dlstadther:
@victoriaalee  Has this integration still be stable for your uses?

victoriaalee:
Hi all!

I've discovered some discrepancies between the exported Prometheus metric and the task history runtimes displayed in the `/tasklist/:name` endpoint.

I noticed that the `_update_task_history` method is called at different places than `set_status` (which also handles calling the functions to update metrics). Is there any documentation or could I receive some guidance on why this is the case? Intuitively, it seems that updating task history and setting status of a task should go hand-in-hand.

Thanks in advance.

Tarrasch:
> or could I receive some guidance on why this is the case?

Unfortunately that's one part of the code bases I don't know well and frankly I think it could use some love. Also I think it's written before it was an obvious thing to submit your code with rigorous unit tests.

Were you to refactor/rewrite any part of it (doesn't need to be perfect, just better) I'm happy to merge.

victoriaalee:
Gotcha. I'll be on vacation for the next two weeks, but I'll take some time afterward to delve into the logic behind task status transitions and see if I can resolve the discrepancies and get this PR merged!

victoriaalee:
Thank you all for your patience. I did some investigation with the jobs being run at my company and their expected runtimes and found that this Prometheus contrib produces accurate results whereas the `/tasklist` endpoint does not.

Please let me know what else needs to be done to get this merged. Thanks!

Tarrasch:
Can you find anyone to review this (just for style, as it seems to work already)? Maybe even at your own company?

Otherwise I'm happy to merge.

victoriaalee:
@Tarrasch The code has been reviewed by a colleage at my company.

kwilcox:
We have an internal “thing” that puts our Luigi task history into Prometheus, mind if I take a look at this next week before merge? @jessicaaustin

dbolshak:
We are looking forward it as well.

kwilcox:
Looks good, :+1: for a merge.

What do you see as the best path to support a run-time defined `MetricCollector`?  I was thinking about allowing the path of a `MetricCollector` subclass as input to the config parameter (like `my.own.MetricsCollector`) and in the `else` of the `MetricsCollectors.get` we could try to load the passed in config option as a class using `importlib`. That would open up the option for custom metrics classes external from the luigi codebase.

dbolshak:
@victoriaalee, @kwilcox Hello, which plans do you have related to the PR?
Can I help you somehow?

I believe it would be awesome to get prometheus support in next release!

Thank you for your contributions and make life easier.

victoriaalee:
@kwilcox I love your idea to allow the use of custom metrics classes external to the Luigi codebase. I know at my company we've been running off of a fork with this contrib for a while, and this would've would've been a more elegant solution.

@dbolshak As far as I'm aware this PR is good to merge. @Tarrasch?

Tarrasch:
Thanks for everyone's input and this contribution @victoriaalee! Perhaps we (I) should finally make a new release of luigi after Easter.

luigi/server.py
```diff
@@ -45,6 +45,7 @@
 import time
 
 import pkg_resources
+from prometheus_client import CONTENT_TYPE_LATEST
```
honnix: This breaks because we are missing `prometheus_client` as a dependency in `setup.py`. So instead of pulling the dependency, we can hardcode the string.

honnix: Or better if we could somehow let contrib module set additional headers...

luigi/server.py
```diff
@@ -274,6 +275,17 @@ def get(self):
...
 
+class MetricsHandler(tornado.web.RequestHandler):
+    def initialize(self, scheduler):
+        self._scheduler = scheduler
+
+    def get(self):
+        metrics = self._scheduler._state._metrics_collector.generate_latest()
+        if metrics:
+            self.write(metrics)
+            self.set_header('Content-Type', CONTENT_TYPE_LATEST)
```
honnix: If this header only for Prometheus?

PR #2629, opened 2019-01-17
Add support for specifying kubernetes namespace

jsteigs642:
## Description
Add support for specifying kubernetes namespace

## Motivation and Context
This change will allow users to submit jobs to a specific namespace in a k8s cluster rather than having them all run in the default namespace.

I ran my jobs with this code and it works for me.  The current unit tests for this module don't cover any different parameters so I don't think additional tests are needed for this change.

### DIFF
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -59,7 +59,7 @@ class kubernetes(luigi.Config):
     max_retrials = luigi.IntParameter(
         default=0,
         description="Max retrials in event of job failure")
-    kubernetes_namespace = luigi.Parameter(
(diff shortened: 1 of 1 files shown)


luigi/contrib/kubernetes.py
```diff
@@ -59,6 +59,9 @@ class kubernetes(luigi.Config):
     max_retrials = luigi.IntParameter(
         default=0,
         description="Max retrials in event of job failure")
+    kubernetes_namespace = luigi.Parameter(
```
dlstadther: This should be an `OptionalParameter` if you want the default to be `None` and not show a warning

jsteigs642: Thanks, I've updated the branch.

PR #2634, opened 2019-01-24
[contrib.hdfs.target.HdfsFlagTarget] fixed argument forwarding to the…

dgronskij:
There is a mismatch between `HdfsTarget.__init__` signature and how it's called from the `HdfsFlagTarget.__init__`

### DIFF
--- a/luigi/contrib/hdfs/target.py
+++ b/luigi/contrib/hdfs/target.py
@@ -214,7 +214,7 @@ class HdfsFlagTarget(HdfsTarget):
         if path[-1] != "/":
             raise ValueError("HdfsFlagTarget requires the path to be to a "
                              "directory.  It must end with a slash ( / ).")
-        super(HdfsFlagTarget, self).__init__(path, format, client)
+        super(HdfsFlagTarget, self).__init__(path, format, fs=client)
         self.flag = flag
 
     def exists(self):


luigi/contrib/hdfs/target.py
```diff
@@ -214,7 +214,7 @@ def __init__(self, path, format=None, client=None, flag='_SUCCESS'):
         if path[-1] != "/":
             raise ValueError("HdfsFlagTarget requires the path to be to a "
                              "directory.  It must end with a slash ( / ).")
-        super(HdfsFlagTarget, self).__init__(path, format, client)
+        super(HdfsFlagTarget, self).__init__(path, format, is_tmp=False, fs=client)
```
honnix: There is not need to set `is_tmp=False` which is the default value in `HdfsTarget`.

honnix:
Looks like a good fix. How come this was not caught by test case? It was introduced by this PR: https://github.com/spotify/luigi/pull/2559

luigi/contrib/hdfs/target.py
dgronskij: I thought that for sake of human readability I better should name-mention parameters that seem being out of context / have different naming but if it doesn't fit, I'll change that

dgronskij:
@honnix the test case covers only `path` / `format` parameters. Not sure if we have to check `fs(client)` / `is_tmp` too

luigi/contrib/hdfs/target.py
honnix: I meant there is no need to pass in is_tmp, fs=client is sufficient.

dgronskij: Agreed, fixed

honnix:
It would be nicer we can add test case to cover the change. But I don't have strong opinion in this case. What do you say @dlstadther ?

dlstadther:
I'm not overly familiar with these HDFS classes. @dgronskij What is the naming convention which is broken?

I agree with @honnix that test coverage would be great. I often rely on Travis tests in areas of Luigi to which i'm much less familiar.

dgronskij:
@dlstadther, well by "broken" I meant using `client` in the module where `fs` keyword is already used with the same meaning. Nevermind:)

BTW, it seems to me (please, correct me if it's not true, I have no previous experience with Travis CI / Nose) that HDFS related stuff is not tested at all: https://github.com/spotify/luigi/blob/55ca81d77677f19f654d868b2bc160d9c3a6197e/.travis.yml#L31
How is it meant to be tested? Manual runs against forked repos (by the way, is it possible)?

Moreover, `HdfsTargetTestMixin` that test `HdfsFlagTarget` via `test_flag_target` seems to be broken since `format` variable not being defined anywhere.

PR #2636, opened 2019-01-25
Fix links to SQLAlchemy documentation

illagrenan:
The original links in https://luigi.readthedocs.io/en/stable/central_scheduler.html#enabling-task-history referred  to 404.


dlstadther:
What happened to flake test state?

I see you've updated the links to `latest`. While it may be preferable to link to released content, latest will indefinitely be available. I can be ok with linking to `latest` docs.

illagrenan:
Link to 0.9 is dead and it looks that SQLAlchemy Documentation supports only versioning for these versions:

2019-01-28 23_32_54-window

Issue #2644, opened 2019-02-01
Tasks can be run several times under certain conditions


GoodDok:
Hi,
I encountered a problem of repeatedly running tasks in my pipeline running on Luigi 2.8.0. 
Basically, it consists of several tasks:

- `A(input_dt)` — the lowest level luigi task, it's result is stored in DB, so `exists` check on target takes some time;
- `B(input_dt)` — has `A(input_dt)` as it's dependency;
- `C(input_dt)` — triggered by cron once in 10 min, depends on `B(input_dt)` and `B(input_dt - 10 min)`.

As you may see, each `B` task is checked twice, and sometimes it leads to duplicate runs.
I tried to reproduce the bug for a long time and finally I wrote a test that reproduces it locally:
 
```
import logging
import shutil
import unittest
from multiprocessing import Process
from time import sleep

import luigi
import luigi.server
from luigi.task import logger

existence_check_delay = 2
parent_sleep_time = 5
child_sleep_time = 4

logging.basicConfig(format='[%(asctime)s: %(name)-5s %(levelname)-5s %(filename)s:%(lineno)s] '
                           '[%(processName)s:%(process)d %(threadName)s] %(message)s',
                    filename="test_result.log")


class SleepyLocalTarget(luigi.LocalTarget):

    def exists(self):
        sleep(existence_check_delay)
        return super().exists()

    def touch(self):
        self.open("w").close()


class Parent(luigi.Task):
    output_dir = luigi.Parameter(default="")

    def requires(self):
        return Child(output_dir=self.output_dir)

    def output(self):
        return SleepyLocalTarget("{}/Parent".format(self.output_dir))

    def run(self):
        logger.info("============= PARENT IS RUNNING =============")
        sleep(parent_sleep_time)
        self.output().touch()


class Child(luigi.Task):
    output_dir = luigi.Parameter(default="")

    def output(self):
        return SleepyLocalTarget("{}/Child".format(self.output_dir))

    def run(self):
        logger.info("============= CHILD IS RUNNING =============")
        sleep(child_sleep_time)
        self.output().touch()


class ConcurrencyTest(unittest.TestCase):

    @staticmethod
    def test_async_run():
        temp_dir = "_temp_concurrent"
        server_process = Process(target=luigi.server.run)
        process1 = Process(target=luigi.build,
                           kwargs={"tasks": [Parent(output_dir=temp_dir)]})
        process2 = Process(target=luigi.build,
[...]

PR #2645, opened 2019-02-04
Issue 2644: Tasks can be run several times under certain conditions

GoodDok:
## Description
Wrote an additional check in `worker.py` after all dependencies of the task were validated,
this should prevent the situation when the real status of the task was changed to DONE,
but the worker sends PENDING to the scheduler.

## Motivation and Context
There may be a situation when a worker has stale state of a task and this leads to duplicate runs.
luigi diagram

For the specific example see the corresponding issue: https://github.com/spotify/luigi/issues/2644

There is a test provided in the PR, it fails with the current implementation and PR resolves it.


test/worker_test.py
```diff
@@ -2135,3 +2136,61 @@ def run(self):
...
+
+    def run(self):
+        time.sleep(self.time_to_sleep)
+        self.output().touch()
+
+
+class ConcurrentTest(unittest.TestCase):
+
+    def setUp(self):
+        self.p = tempfile.mkdtemp()
```
Tarrasch: Maybe add `suffix='-luigi-testcase'` for traceability?

test/worker_test.py
```diff
@@ -2135,3 +2136,61 @@ def run(self):
                 self.assertEqual(0, self.sch._state.get_task(s1.task_id).failures.num_failures())
                 self.assertEqual(self.per_task_retry_count, self.sch._state.get_task(e2.task_id).failures.num_failures())
                 self.assertEqual(self.default_retry_count, self.sch._state.get_task(e1.task_id).failures.num_failures())
+
+
+class SleepyLocalTarget(luigi.LocalTarget):
```
Tarrasch: Can we reduce the complexity of the testcase you're building? I'm not sure I fully understand what you're solving and that maybe could be easier if the test case was simpler (if possible).

test/worker_test.py
```diff
@@ -2135,3 +2136,61 @@ def run(self):
...
+    time_to_sleep = 0.5
+
+    def exists(self):
+        time.sleep(self.time_to_sleep)
```
Tarrasch: I'm a bit worried about this. We reserve the right to check `complete()` multiple times in luigi, and if an "refactoring" changes this (checks more often or for example caches and checks less), then this test-case might start to break.

test/worker_test.py
```diff
@@ -2135,3 +2136,61 @@ def run(self):
...
+
+    def tearDown(self):
+        shutil.rmtree(self.p)
+
+    def test_sending_same_task_twice_does_not_lead_to_double_run(self):
+        process = Process(target=luigi.build,
+                          kwargs={"tasks": [Parent(output_dir=self.p)], "local_scheduler": True})
+        process.start()
+        time.sleep(SleepyLocalTarget.time_to_sleep * 2 + Parent.time_to_sleep + Child.time_to_sleep)
+        self.assertTrue(luigi.build([Parent(output_dir=self.p)], local_scheduler=True))
```
Tarrasch: I have little idea what this assertTrue shows. Did this fail before?? Did it fail because nothing was executed or what?

There is a (very) new thing called LuigiRunResult which you can use here to make it clearer what the test case is checking (it also adds an enum that could improve clarity further)

test/worker_test.py
```diff
@@ -2135,3 +2136,61 @@ def run(self):
...
+
+class ConcurrentTest(unittest.TestCase):
+
+    def setUp(self):
+        self.p = tempfile.mkdtemp()
+
+    def tearDown(self):
+        shutil.rmtree(self.p)
+
+    def test_sending_same_task_twice_does_not_lead_to_double_run(self):
```
Tarrasch: Thanks for the descriptive test name. Luigi could use more of these.

luigi/worker.py
```diff
@@ -842,7 +842,14 @@ def _add(self, task, is_complete):
                     task.trigger_event(Event.DEPENDENCY_DISCOVERED, task, d)
                     yield d  # return additional tasks to add
 
-                deps = [d.task_id for d in deps]
+                # if dependencies check took too long, the parent task might have been completed by other worker
+                if status == PENDING and task.complete():
```
Tarrasch: Why is `status == PENDING` needed?  The comment doesn't cover that right?

test/worker_test.py
GoodDok: Let me clarify, there's a `SleepyLocalTarget` that does not allow to run the task twice: it fails with an exception. So if the `Parent` task is run twice, `luigi.build` will return False. This test fails without the changes in `worker.py` in this PR since luigi tries to run `Parent` task twice.
Do you think this can be expressed using `LuigiRunResult` in a clearer way? If yes, I'll give it a try.

luigi/worker.py
GoodDok: Well, I wanted to filter out `DISABLED` status to call `task.complete()` only when it's really needed. WDYT?

test/worker_test.py
GoodDok: The test checks that the `Parent` task won't be run twice, so it shouldn't fail even if we add more or less `complete()` checks in the future

GoodDok:
@Tarrasch thank you very much for starting reviewing this PR! :+1:
I attached a picture describing the problem that I'm facing: if worker A finishes the task right before worker B sends tasks status, we have duplicate runs. Please take a look and let me know if it's clear.
I have a working case described in the issue: https://github.com/spotify/luigi/issues/2644, if you need any other details please let me know. I suspect these guys faced the same problem: https://groups.google.com/forum/#!topic/luigi-user/YCEQIEz0jLo
After discovering this issue I think it has somewhat deeper roots. This PR will resolve **most** of the problem cases, but the race condition will still exist, even with the only one task: 
- worker A runs some task
- worker B starts the same task and checks it's output does not exist
- worker A finishes task right after that
- workers A and B send status to the scheduler, worker A appears to be the first one - status DONE
- then worker B - status PENDING
- worker B asks for more work - runs the task for the second time

I think to completely resolve this issue there must be something like "timestamp of status check" to be sent to scheduler along with the status itself. And scheduler should resolve what to do if the status timestamp is older than the one it has in it's state. Just my thoughts.

GoodDok:
@Tarrasch, can you please take a look at the changes I submitted? This is just the draft of an idea how this issue can be resolved with the help of distrust period.
At the moment, I don't see any way to "rerun" the check of the task status without rewriting a big part of the basic logic regarding the interaction between workers and the scheduler. So I chose to keep the task that was DONE very recently in the same state.
E. g., let's take a look at the code that sends the task metadata to the scheduler:
https://github.com/spotify/luigi/blob/3701c68d7f31c543f6f871ed141a7ba170fd9e8c/luigi/worker.py#L561-L583
If I'm not wrong, it does not assume anything wrong may happen during `self._scheduler.add_task`, it even writes the record to the task history earlier than `add_task` itself was called. The interaction happens only from a worker to the scheduler at this point, not vice verca. So we cannot ask the worker here "wait, then retry sending the same task with the updated status again" here without significant code changes.

In the other hand, once we send the task in PENDING state, the scheduler treats is as the one that can be given to the worker unless we change the task's status on-the-fly. Maybe we could rerun check when a worker asks for work, but then we'll have to detect which of the tasks need this "status recheck" and which don't need it.

Actually, I'm not a big fan of the approach selected, but the main goal for me was to reduce the impact of these changes.


Please let me know if you have any thoughts regarding this!

luigi/scheduler.py
```diff
@@ -150,6 +150,8 @@ class scheduler(Config):
 
     metrics_collector = parameter.EnumParameter(enum=MetricsCollectors, default=MetricsCollectors.default)
 
+    rerun_task_blocking_period = parameter.IntParameter(default=10)
```
Tarrasch: let's come up with a better name for this? I don't think we can come up with a name people will understand with a few words, but it should still make sense.

 * Maybe `stable_done_cooldown_secs`?
 * Whichever we choose we maybe want `_secs` in the name. There's a TimeDeltaParameter but I think we should avoid those fancy stuff in the scheduler. I dream that one day that the scheduler is written in another language even.

Also add docs to this parameter.

luigi/scheduler.py
```diff
@@ -840,6 +842,10 @@ def add_task(self, task_id=None, status=PENDING, runnable=True,
         if task is None or (task.status != RUNNING and not worker.enabled):
             return
 
+        # if the task was completed only recently, do not run the same task again
+        if status == PENDING and task.status == DONE and (time.time() - task.updated) < self._config.rerun_task_blocking_period:
+            status = DONE
```
Tarrasch: What happens if you just return here instead???

luigi/scheduler.py
```diff
@@ -840,6 +842,10 @@ def add_task(self, task_id=None, status=PENDING, runnable=True,
         if task is None or (task.status != RUNNING and not worker.enabled):
             return
 
+        # if the task was completed only recently, do not run the same task again
+        if status == PENDING and task.status == DONE and (time.time() - task.updated) < self._config.rerun_task_blocking_period:
```
Tarrasch: Yes, pretty much what I expected. Great!

luigi/scheduler.py
```diff
@@ -840,6 +842,10 @@ def add_task(self, task_id=None, status=PENDING, runnable=True,
         if task is None or (task.status != RUNNING and not worker.enabled):
             return
 
+        # if the task was completed only recently, do not run the same task again
```
Tarrasch: Describe what you actually do, the implementation, not why you do it. :)

     # Ignore claims that the task is PENDING if it very recently was marked as DONE.

Maybe?

test/scheduler_api_test.py
```diff
@@ -46,6 +46,7 @@ def get_scheduler_config(self):
             'disable_window': 10,
             'retry_count': 3,
             'disable_hard_timeout': 60 * 60,
+            'rerun_task_blocking_period': 0
```
Tarrasch: Good start. This is fine as other tests didn't assume this.

Tarrasch: Also of course you should add a few other unit tests with this flag set :)

luigi/scheduler.py
GoodDok: Thanks, changed to the proposed version

luigi/scheduler.py
GoodDok: As far as I understand, the same thing happens -- the scheduler rejects giving task to the worker. Changed to simple "return".

GoodDok:
@Tarrasch @honnix @dlstadther psst, guys, probably you forgot about this one

test/scheduler_test.py
```diff
@@ -281,3 +286,65 @@ def test_get_pending_tasks_with_many_done_tasks(self):
...
+class StableDoneCooldownSecsTest(unittest.TestCase):
+
+    def setUp(self):
+        self.p = tempfile.mkdtemp()
+
+    def tearDown(self):
+        shutil.rmtree(self.p)
+
+    def run_task(self):
+        return luigi.build([FailingOnDoubleRunTask(output_dir=self.p)],
```
Tarrasch: Can you implement this with LuigiRunResult? Test will become more readable.

test/scheduler_test.py
```diff
@@ -281,3 +286,65 @@ def test_get_pending_tasks_with_many_done_tasks(self):
...
+        process = Process(target=self.run_task)
+        try:
+            # scheduler is started
+            server_process.start()
+            # first run is started
+            process.start()
+            time.sleep(FailingOnDoubleRunTask.time_to_run_secs + FailingOnDoubleRunTask.time_to_check_secs)
+            # second run of the same task is started
+            luigi_run_result = self.run_task()
+            self.assertEqual(luigi_run_result, not failure_expected)
```
Tarrasch: This line in particular is very hard to read.

test/scheduler_test.py
GoodDok: Thank you, done

test/scheduler_test.py
GoodDok: OMG, of course :) 
thank you, fixed

GoodDok:
@Tarrasch thank you for the thorough review again!
I've rebased + followed your recommendations to improve the readability of the tests.
Please let me know if we need any other changes here.

eloyfelix:
many thanks @GoodDok et al for this work, I'm afraid that I'm also having this issue :P Hope it gets merged soon!

GoodDok:
Hi there, @Tarrasch @honnix @dlstadther,
Just wanted to let you know that I'm still here once you are ready to review my final changes.

Seems like I'm not the only one who has faced this issue.

dlstadther:
It looks like @Tarrasch already approved the changes. I'll give him a little to re-confirm, otherwise i'll merge

Tarrasch:
My approval still stands. Sorry for being away for so long and not merging your rebase. I or any another maintainer may merge this once you rebase. Sorry!

GoodDok:
@dlstadther @Tarrasch guys, I've rebased the branch, please feel free to merge :)

PR #2649, opened 2019-02-07
Add first iteration of a Snowflake contrib

cabouffard:
## Description
This will add a first version of the Snowflake contrib as we currently use it. The only functionality it has is to copy data from S3. It doesn't support other actions that the PSQL or Redshift integration would. 

As we use Snowflake more and more, we might contribute more to this contrib.

I feel like we should this contribute this to the community and the community could add their own functionality as needed.

## Motivation and Context
We're currently investigating migrating from Redshift to Snowflake and to our surprise there was no Snowflake integration in Luigi. Since we still needed it, we've decided to build our own following our simple use case which is to load data from S3 into Snowflake.

This is currently running in our staging environment and tested through unit tests. The unit tests could be better as there are some edge cases that aren't tested, but Redshift and PSQL integrations suffer from the same problem.

Hopefully, someone else that uses Snowflake can vouch this contrib and make sure that isn't too flakey.


luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,411 @@
...
+import logging
+
+import luigi
+from luigi.contrib import rdbms
+from luigi.contrib import redshift
+from luigi.contrib import postgres
+
+logger = logging.getLogger('luigi-interface')
+
+# Snowflake's library do not have constant for their error code.
```
dlstadther: `library do not` -> `library does not`

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,411 @@
...
+
+try:
+    import snowflake
+    import snowflake.connector
+except ImportError:
+    logger.warning("Loading snowflake module without snowflake-connector-python installed. "
+                   "Will crash at runtime if snowflake functionality is used.")
+
+
+class _SettingsMixins():
```
dlstadther: do we need to inherit from `object`?

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,411 @@
...
+        """
+        Fetch the warehouse configuration from the luigi configuration file under:
+
+        [snowflake]
+        warehouse=value
+        """
+        return luigi.configuration.get_config().get(self.configuration_section, 'warehouse', default=None)
+
+
+class _CredentialsMixin(redshift._CredentialsMixin):
```
dlstadther: Snowflake's website suggests multiple cloud providers are available. Your implementation auth is specific for AWS. Should AWS be an option which happens to be the default?

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,411 @@
...
+        """
+        Args:
+            host (str): Snowflake server address.
+            warehouse (str): Snowflake warehouse name.
+            database (str): Database name
+            user (str): Database user
+            password (str): Password for specified user
+            update_id (str): An identifier for this data set
+        """
+        self.host = host
```
dlstadther: does Snowflake not require a `port`?

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,411 @@
...
+        * Create the table if it doesn't already exists
+        * Truncate the table if the flag "do_truncate_table" is set
+
+        * Copy the data from S3 to the Snowflake table
+
+        * Run any potential "post_copy" queries
+
+        * Save the state of this task to Snowflake as "been run"
+        """
+        if not (self.table):
```
dlstadther: `table` is already an `abstractproperty`. this check feels excessive

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,411 @@
...
+
+        Do not re-run the task if it has been previously run
+        """
+        new_connection = connection is None
+        if new_connection:
+            connection = self.connect()
+            connection.autocommit = True
+        cursor = connection.cursor()
+        try:
+            cursor.execute("""SELECT 1 FROM {marker_table} WHERE update_id = %s LIMIT 1""".format(marker_table=self.marker_table),
```
dlstadther: can we avoid mixing string formatting types? Can we just use `.format()`?

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,411 @@
...
+        # commit and clean up
+        connection.close()
+
+    def copy(self, cursor, f):
+        """
+        COPY data from from S3 into Snowflake.
+
+        This currently only works for key-base credentials.
+        """
+        logger.info("Inserting file: %s", f)
```
dlstadther: `.format()` please

tox.ini
```diff
@@ -48,6 +49,7 @@ setenv =
   hdp: HADOOP_DISTRO=hdp
   hdp: HADOOP_HOME={toxinidir}/.tox/hadoop-hdp
   contrib: NOSE_ATTR=contrib
+  snowflake: NOSE_ATTR=snowflake
```
dlstadther: if snowflake is assed as its own tox type, we'll also want to `and not snowflake` from `core` (below)

test/contrib/snowflake_test.py
```diff
@@ -0,0 +1,373 @@
...
+    def s3_load_path(self):
+        return 's3://%s/%s' % (BUCKET, KEY)
+
+
+class DummyS3CopyToTableWithKeys(DummyS3CopyToTableBase):
+    aws_access_key_id = 'property_key'
+    aws_secret_access_key = 'property_secret'
+
+
+@attr('snowflake', 'contrib')
```
dlstadther: not sure this needs to be added to multiple test groups

luigi/contrib/snowflake.py
cabouffard: From what I understand `object` would only be required for Python 2.X while it's not necessary for Python 3.X unless we're trying to be python version agnostic here. More information around that [concept here](https://stackoverflow.com/a/45062077).

luigi/contrib/snowflake.py
cabouffard: Yes, you are right, it supports multiple cloud providers. We've only been using AWS so I took the code from the Redshift contrib. I like the idea of the option, I'll see what I can do.

luigi/contrib/snowflake.py
cabouffard: No, it does not!

test/contrib/snowflake_test.py
cabouffard: Do you have a preference between setting it to `contrib` vs `snowflake`?

dlstadther: 🤔 

This is a good question bc it starts to break my original line of thinking. I had included an `AWS` test section, which this _could_ fit into. However, it's not exclusive to AWS.

Let's go with `contrib` for now

luigi/contrib/snowflake.py
dlstadther: for now, Luigi still supports (and tests) 2.7

luigi/contrib/snowflake.py
cabouffard: I am having a problem and I'd like to have some feedback. Luigi is able to hydrate a property for a given class if that property exists during the initialization of the class.

Given this use case:

```python
MyMixin(object):
    @property
    def aws_access_key_id(self):
        """
        Override to return the key id.
        """
        return luigi.configuration.get_config().get(self.configuration_section, 'aws_access_key_id')

    @property
    def aws_secret_access_key(self):
        """
        Override to return the secret access key.
        """
        return luigi.configuration.get_config().get(self.configuration_section, 'aws_secret_access_key')


MyTask(rdms.CopyTable, MyMixins):
    pass
```

If there is an environment variable named `aws_secret_access_key`, then calling `MyTask.aws_secret_access_key` will yield that value, else it will execute the function defined in the mixin.

My concern with your ask is that we would need to have one Mixin that contains all the different properties. The ones from AWS and the ones from Azure. 

If we do it differently, then hydrating the properties using the environment variable isn't supported anymore and that flow breaks. 

I've added a new commit on this pull request named "--wip--" that illustrate what I'm trying to go for but the specs using "TestInternalCredentials" are obviously failing, because it can't hydrate the object since we can only tell which credentials to use after the class has already been initialized.

Let me know if the problem I'm raising is unclear, I'd need some guidance to solve the issue!
Thanks!

dlstadther:
hi @cabouffard ! Thanks so much for coming back to this PR. I'll try to get to reviewing _soon_ - i'm a bit overloaded at the moment, but if too much time passes and i have re-reviewed yet please `@` me!

cabouffard:
@dlstadther 10 days seems to be a good amount for me to start pinging you again! :)

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,523 @@
...
+    Generic mixins that defines how to fetch the database and warehouse
+    configuration for a given snowflake integration.
+    """
+    @property
+    def configuration_section(self):
+        """
+        Override to change the configuration section used
+        to obtain default credentials.
+        """
+        return 'snowflake'
```
dlstadther: what is the motivation behind allowing both the` _SettingsMixin` and `SnowflakeAuth` to separately set their `configuration_section`s? Why would you want them to differ?

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,523 @@
...
+        """
+        Fetch the warehouse configuration from the luigi configuration file under:
+
+        [snowflake]
+        warehouse=value
+        """
+        return luigi.configuration.get_config().get(self.configuration_section, 'warehouse', default=None)
+
+    @property
+    def role(self):
```
dlstadther: This doesn't exist in `configuration.rst`

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,523 @@
...
+
+
+class AwsKeyAuth(SnowflakeAuth):
+    def credentials(self):
+        """
+        Return an AWS credential string for the provided task.
+
+        If no valid credentials are set, raise a NotImplementedError.
+        """
+        if self.aws_access_key_id and self.aws_secret_access_key:
```
dlstadther: i understand this is still WIP. Does AWS snowflake only accept key-based auth? Does it support role-based?

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,523 @@
...
+
+
+class SnowflakeTarget(postgres.PostgresTarget):
+    """
+    Target for a resource in Snowflake.
+
+    Snowflake connection string is similar to postgres with a few adjustments.
+    """
+    marker_table = luigi.configuration.get_config().get(
+        'snowflake',
```
dlstadther: just a note.... at this point, we've allowed the section to be configured for settings and auth, but are hardcoding the db section here. I'd suspect someone would likely store all these configuration variables in the same cfg section.

Do we require they all exist in `snowflake`? Allow them to be individually configurable? Or set once, use in all places?

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,523 @@
...
+            connection = self.connect()
+            connection.autocommit = True
+        cursor = connection.cursor()
+        try:
+            cursor.execute("""SELECT 1 FROM {marker_table} WHERE update_id = '{update_id}' LIMIT 1""".format(marker_table=self.marker_table,
+                                                                                                             update_id=self.update_id))
+            row = cursor.fetchone()
+        except snowflake.connector.errors.ProgrammingError as e:
+            # Snowflake doesn't seem to have constants for their error code
+            if e.errno == SNOWFLAKE_GENERIC_PROGRAMMING_ERROR:
```
dlstadther: what is the rationale for not letting this error raise?

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,523 @@
...
+        * `host`,
+        * `database`,
+        * `user`,
+        * `password`,
+        * `table`,
+        * `columns`,
+        * `s3_load_path`.
+    """
+    configuration_section = 'snowflake'
+    snowflake_copy_auth = luigi.EnumParameter(enum=SnowflakeSupportedAuth, default=SnowflakeSupportedAuth.default)
```
dlstadther: am i correct that the `default` (value 0) will raise an exception bc it is not currently implemented in the `SnowflakeSupportedAuth`?

dlstadther: Is this a value that can be pulled from the `cfg`? I suppose someone who is using snowflake s3copytotable in multiple instances is going to be referencing the same cluster (and thus auth method) each time.

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,523 @@
...
+
+        query = ("""COPY INTO {table} from {source} CREDENTIALS=({creds}) {options};""".format(
+            table=self.table,
+            source=f,
+            creds=self.credentials(),
+            options=self.copy_options)
+        )
+        cursor.execute(query)
+
+    def credentials(self):
```
dlstadther: in `redshift.py`, credentials is a hidden method. What do you think is the appropriate type? hidden vs public? method vs property?

luigi/contrib/snowflake.py
```diff
@@ -0,0 +1,523 @@
...
+    Subclass and override the required attributes:
+        * `host`,
+        * `database`,
+        * `user`,
+        * `password`,
+        * `table`,
+
+    Note:
+        This uses execute_string which is not safe for SQL injections
+        https://docs.snowflake.net/manuals/user-guide/python-connector-api.html#execute_string
```
dlstadther: given this caveat, what was the motivation for using this method?

PR #2652, opened 2019-02-10
Simplify implementation of temporary_path()

Tarrasch:
## Description

The context manager FileSystemTarget.temporary_path() rolled its own manager, it was overcomplicated so i rewrote using contextlib.

Unit tests already existed. Since they were really poorly written I also rewrote them a bit.


Tarrasch:
Thanks for the quick review!

PR #2654, opened 2019-02-12
Added TTL to keep_alive

TMaYaD:
## Description
This adds an additional parameter `keep_alive_ttl` which determines how long a worker is kept alive for new tasks to be available. 

## Motivation and Context
This will prevent a worker from waiting indefinitely when another worker is running the pending tasks and this one has nothing else to do.

We have been using these changes in our production for a while now. The code paths around the changes are a little too hairy to be sure we got all edge cases correctly.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -308,6 +308,12 @@ wait_jitter
   workers do not ask the scheduler for another job at the same time.
   Default: 5.0
 
+max_keep_alive_idle_duration
+  .. versionadded:: 2.8.4
+
+  Maximum duration to keep worker alive while in idle state.
+  Default: 0 (Indefinitely)
+
 max_reschedules
   Maximum number of times to reschedule a failed task.
   Default: 1
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -412,8 +412,6 @@ class worker(Config):
                                    config_path=dict(section='core', name='worker-ping-interval'))
     keep_alive = BoolParameter(default=False,
                                config_path=dict(section='core', name='worker-keep-alive'))
-    keep_alive_ttl = TimeDeltaParameter(default=datetime.timedelta(0),
-                                        config_path=dict(section='core', name='worker-keep-alive-ttl'))
     count_uniques = BoolParameter(default=False,
                                   config_path=dict(section='core', name='worker-count-uniques'),
                                   description='worker-count-uniques means that we will keep a '
@@ -426,6 +424,8 @@ class worker(Config):
     wait_interval = FloatParameter(default=1.0,
                                    config_path=dict(section='core', name='worker-wait-interval'))
     wait_jitter = FloatParameter(default=5.0)
+
+    max_keep_alive_idle_duration = TimeDeltaParameter(default=datetime.timedelta(0))
 
     max_reschedules = IntParameter(default=1,
                                    config_path=dict(section='core', name='worker-max-reschedules'))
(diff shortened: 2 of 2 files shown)


luigi/worker.py
```diff
@@ -411,6 +412,8 @@ class worker(Config):
                                    config_path=dict(section='core', name='worker-ping-interval'))
     keep_alive = BoolParameter(default=False,
                                config_path=dict(section='core', name='worker-keep-alive'))
+    keep_alive_ttl = TimeDeltaParameter(default=datetime.timedelta(0),
```
dlstadther: Feels a little odd to maintain the legacy config format here and allow both `core -> worker-keep-alive-ttl` and `worker -> keep_alive_ttl` for a brand new config arg.

I'm not against it. Just seems odd to allow. But I'll let @Tarrasch or @honnix chime in on that

dlstadther: could this also be added to Worker arg documentation?

luigi/worker.py
```diff
@@ -1127,8 +1132,16 @@ def _keep_alive(self, get_work_response):
             return get_work_response.n_pending_last_scheduled > 0
         elif self._config.count_uniques:
             return get_work_response.n_unique_pending > 0
+        elif get_work_response.n_pending_tasks <= 0:
```
dlstadther: how can n_pending_task ever be less than 0?

luigi/worker.py
```diff
@@ -1127,8 +1132,16 @@ def _keep_alive(self, get_work_response):
...
+            return False
+        elif not self._config.keep_alive_ttl:
```
dlstadther: does the default value drop you into this elif block?

luigi/worker.py
```diff
@@ -1127,8 +1132,16 @@ def _keep_alive(self, get_work_response):
...
+            return True
+        elif not self._idle_since:
+            return True
         else:
-            return get_work_response.n_pending_tasks > 0
+            time_to_shutdown = self._idle_since + self._config.keep_alive_ttl - datetime.datetime.now()
+            logger.debug("%s until shutdown", time_to_shutdown)
```
dlstadther: should this log the worker id? is the worker_id accessible here? Just seems like a valuable thing to know within the logging.

luigi/worker.py
TMaYaD: I'm lost here. Is there another place I should define this config option?

luigi/worker.py
TMaYaD: Good question. I guess I didn't pay enough attention to the context.

luigi/worker.py
TMaYaD: Yes! Yes, it does.

luigi/worker.py
dlstadther: I was providing some historical context.

i.e.
```
[worker]
keep_alive = True
```
used to be required to be
```
[core]
worker-keep-alive: True
```

Then Config classes came along and now
```
class worker(Config):
    keep_alive = BoolParameter(default=False)
```
knows to look at `[worker] keep_alive` for its possible override value.

----

So what i was mentioning here is that it's odd to add a new Worker cfg parameter and allow it to be included in the legacy `[core] worker-whatever-param` format.

----

Hopefully, that was less unclear (kinda hard to clearly articulate) :)

doc/configuration.rst
```diff
@@ -308,6 +308,12 @@ wait_jitter
   workers do not ask the scheduler for another job at the same time.
   Default: 5.0
 
+max_keep_alive_idle_duration
+  .. versionadded:: 2.8.4
```
Tarrasch: oh I didn't know of this! Cool!

Tarrasch:
Thanks!!

PR #2655, opened 2019-02-12
Added worker-id commandline parameter

TMaYaD:
Re-opening https://github.com/spotify/luigi/pull/2267 with a rebased branch.

## Description
This change adds a `--worker-id` which will be used for worker to identify itself. The user is expected to make sure these are unique.

## Motivation and Context
We needed to start and (safely) shutdown workers based on resource availability. In order to use `disable-worker` rpc call we needed the worker id. But there is no way to know the same. We can't rely on the included host name since we are using ECS to run the workers inside containers
This change allows us to set the worker-id using a known uuid and other details and use the same to disable it later.

This is already available as a function parameter (which doesn't seem to be used anywhere). We tested it in our own setup without any issues.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -407,7 +407,7 @@ def check_complete(task, out_queue):
 class worker(Config):
     # NOTE: `section.config-variable` in the config_path argument is deprecated in favor of `worker.config_variable`
 
-    id = Parameter(default=None,
+    id = Parameter(default='',
                    description='Override the auto-generated worker_id')
     ping_interval = FloatParameter(default=1.0,
                                    config_path=dict(section='core', name='worker-ping-interval'))
@@ -517,7 +517,7 @@ class Worker(object):
 
         self._config = worker(**kwargs)
 
-        worker_id = worker_id or self._config.id or 'Worker(%s)' % ', '.join(['%s=%s' % (k, v) for k, v in self._worker_info])
+        worker_id = worker_id or self._config.id or self._generate_worker_id(self._worker_info)
 
         assert self._config.wait_interval >= _WAIT_INTERVAL_EPS, "[worker] wait_interval must be positive"
         assert self._config.wait_jitter >= 0.0, "[worker] wait_jitter must be equal or greater than zero"
@@ -628,6 +628,10 @@ class Worker(object):
         except BaseException:
             pass
         return args
+
+    def _generate_worker_id(self, worker_info):
+        worker_info_str = ', '.join(['{}={}'.format(k, v) for k, v in worker_info])
+        return 'Worker({})'.format(worker_info_str)
 
     def _validate_task(self, task):
         if not isinstance(task, Task):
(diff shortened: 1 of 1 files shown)


luigi/worker.py
```diff
@@ -407,6 +407,8 @@ def check_complete(task, out_queue):
 class worker(Config):
     # NOTE: `section.config-variable` in the config_path argument is deprecated in favor of `worker.config_variable`
 
+    id = Parameter(default=None,
```
dlstadther: `Parameter`s should have a default of `''`. If you truly want default `None`, consider `OptionalParameter`

luigi/worker.py
```diff
@@ -511,11 +513,10 @@ def __init__(self, scheduler=None, worker_id=None, worker_processes=1, assistant
         self.worker_processes = int(worker_processes)
         self._worker_info = self._generate_worker_info()
 
-        if not worker_id:
-            worker_id = 'Worker(%s)' % ', '.join(['%s=%s' % (k, v) for k, v in self._worker_info])
-
         self._config = worker(**kwargs)
 
+        worker_id = worker_id or self._config.id or 'Worker(%s)' % ', '.join(['%s=%s' % (k, v) for k, v in self._worker_info])
```
dlstadther: since you've already modified this line, any objection to going ahead and replacing that ugly `%` string interpolation with `.format()`? :)

luigi/worker.py
Tarrasch: Nice catch @dlstadther, but for this case certainly use `''` and just do `if` on it.

luigi/worker.py
Tarrasch: I would also not do a one liner for this, so it's more readable

luigi/worker.py
TMaYaD: Changing this to `''` should be good enough. We are already doing `_config.id or` anyway.

luigi/worker.py
```diff
@@ -625,6 +626,10 @@ def _generate_worker_info(self):
             pass
         return args
 
+    def _generate_worker_id(self, worker_info):
+        worker_info_str = ', '.join(['{}={}'.format(k, v) for k, v in worker_info])
```
TMaYaD: I don't think `format` looks any less uglier. f-strings would be a different matter but i don't think they're supported on older python.

Tarrasch: We promote `format` since it's more pythonic. It's implemented without magic unlike `%`. Luigi only supports 2.7+ (other versions are long deprecated)

Tarrasch:
Can you fix the merge conflict? Sorry about this, I would do it in the browser but I'm afraid I'll break the line length limit if I do it in the browser.

TMaYaD:
@Tarrasch Done. You can merge it in.

dlstadther:
Looks like this is ready! Merging

PR #2656, opened 2019-02-13
Prevented range-tasks from autoloading

TMaYaD:
Reopening https://github.com/spotify/luigi/pull/2265

This change makes --help-all a lot less noisy.
Range tasks can be loaded if required through the usual --module
parameter.

Existing behaviour is not changed except for adding a deprecation warning which can be suppressed.

### DIFF
--- a/luigi/__init__.py
+++ b/luigi/__init__.py
@@ -75,5 +75,5 @@ if not configuration.get_config().has_option('core', 'autoload-range'):
     warnings.warn(warning_message, DeprecationWarning)
 
 if configuration.get_config().getboolean('core', 'autoload-range', True):
-    from .tools import range  # just makes the tool classes available from command line
+    from .tools import range  # noqa: F401    just makes the tool classes available from command line
     __all__.append('range')
--- a/luigi/__init__.py
+++ b/luigi/__init__.py
@@ -48,14 +48,11 @@ from luigi.execution_summary import LuigiStatusCode
 from luigi import event
 from luigi.event import Event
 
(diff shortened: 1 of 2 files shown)


luigi/__init__.py
```diff
@@ -48,17 +48,32 @@
...
 ]
+
+if not configuration.get_config().has_option('core', 'autoload-range'):
+    import warnings
+    warning_message = '''
+        Autoloading range tasks by default has been deprecated and will be removed in a future version.
+        To get the behavior now add an option to luigi.cfg:
+
+          [core]
+            autoload-range: false
```
dlstadther: If this is a new config option, i'd expect to see a docs update.

I have no qualms with this as it doesn't change the default behavior (aside from a warning), but allows powerusers to reduce some noise

Tarrasch:
Another reason we really want this:

```
~/repos/luigi master  
❯ PYTHONPATH=. ./bin/luigi examples/foo.py
  File "./bin/luigi", line 5, in <module>
    import luigi.cmdline
  File "/home/rouhani/repos/luigi/luigi/__init__.py", line 51, in <module>
    from .tools import range  # just makes the tool classes available from command line
  File "/home/rouhani/repos/luigi/luigi/tools/range.py", line 37, in <module>
    from dateutil.relativedelta import relativedelta
ImportError: No module named dateutil.relativedelta
```

I think that without range one should be able to download and run luigi without needing any other libraries.

TMaYaD:
Added documentation for the config option.

PR #2657, opened 2019-02-13
Introduced process_requires

TMaYaD:
Reopens https://github.com/spotify/luigi/pull/2316 

## Description
Adds a `process_requires` method to `Task` class. Similar to `_requires` and `process_resources`, this allows a Template task to inspect and modify dependencies before passing them to Luigi worker.

This is different from `_requires` or `deps` in that this works with dynamic dependencies yielded from `run` as well.

## Motivation and Context
We have a template task that needed to pass some parameters along the dependency tree. This change could be used to obviate the need for using `clone` in every Task:
```
def process_requires(self, requires):
    for req in luigi.task.flatten(requires):
        req.job_id = self.job_id
    return requires
```

"We ran our jobs with this code and it works for us."
I could use some help in adding unit tests!

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -600,6 +600,18 @@ class Task(object):
         """
         return []  # default impl
 
+    def process_requires(self, requires):
+        """
+        Override in "template" tasks which themselves are supposed to be
+        subclassed and thus have their requires() overridden (name preserved to
+        provide consistent end-user experience), yet need to introduce
+        custom behaviour into dependencies.
+
+        This defers from _requires in that this will not directly call task.requires
+        and this is being used by dependencies yielded by run
+        """
+        return requires
+
     def requires(self):
         """
         The Tasks that this Task depends on.
@@ -652,7 +664,7 @@ class Task(object):
         Returns the flattened list of requires.
         """
         # used by scheduler
-        return flatten(self._requires())
+        return flatten(self.process_requires(self._requires()))
 
     def run(self):
         """
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -607,8 +607,30 @@ class Task(object):
         provide consistent end-user experience), yet need to introduce
         custom behaviour into dependencies.
 
-        This defers from _requires in that this will not directly call task.requires
-        and this is being used by dependencies yielded by run
+        Differences with :py:meth:`_requires`:
+          * method signature: :py:meth:`_requires` doesn't take params and calls
+            :py:meth:`requires` directly where as this method is expected to work only on
+            requirements passed via parameters.
+          * Code paths: :py:meth:`_requires` is run only during the statically defined
+            requirements where as this method works with dynamic dependencies yielded from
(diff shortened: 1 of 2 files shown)


luigi/task.py
```diff
@@ -600,6 +600,25 @@ def output(self):
...
+            class TemplateTask(luigi.Task):
+                log_level = DEBUG
+
+                def process_requires(self, requirements):
+                    for(task in flatten(requirements)):
+                        task.log_level = self.log_level # Propagate log level up the dependency chain.
+
+                    return requirements
+        """
+        return requires
```
dlstadther: if this method is replacing `_requires()`, shouldn't the return be `flatten(requires)`?

luigi/task.py
```diff
@@ -647,12 +661,15 @@ def input(self):
...
         Returns the flattened list of requires.
+
+        Override in "template" tasks which themselves are supposed to be
+        subclassed and thus have their requires() overridden (name preserved to
+        provide consistent end-user experience), yet need to introduce
+        (non-input) dependencies.
         """
         # used by scheduler
-        return flatten(self._requires())
+        return flatten(self.process_requires(self.requires()))
```
dlstadther: or this be `flatten(self.process_requires(flatten(self.requires())))`

luigi/task.py
```diff
@@ -600,6 +600,25 @@ def output(self):
...
+        provide consistent end-user experience), yet need to introduce
+        custom behaviour into dependencies.
+
+        ex:
+            class TemplateTask(luigi.Task):
+                log_level = DEBUG
+
+                def process_requires(self, requirements):
+                    for(task in flatten(requirements)):
+                        task.log_level = self.log_level # Propagate log level up the dependency chain.
```
Tarrasch: There is a Task cache so I would be careful aboud doing mutations. I also said this in previous PR I think. :)

luigi/task.py
```diff
@@ -647,12 +661,15 @@ def input(self):
 
     def deps(self):
         """
-        Internal method used by the scheduler.
```
Tarrasch: Wait why?

Can we start with a smaller PR just renaming `deps` to `_deps` or something? Or maybe even inline it at places it is used (if apporpirate).

luigi/task.py
TMaYaD: Assumption for process_requires is that it can handle any valid structured requirements and doesn't have any side effects like flattening the result. So `self.process_requires(self.requires())` should return a structure similar to what you expect from `self.requires()`. It's the job of `dep` flatten it as expected by it's users.

luigi/task.py
TMaYaD: The main purpose of `deps` seems to be overridden in `Wrapped` class to handle sub-task dependencies.

TMaYaD: @Tarrasch Raised #2685 to address this. Can you take a look?

PR #2658, opened 2019-02-13
Fix broken aws tests

dlstadther:
## Description

Update `botocore` minimum version to align with `s3transfer=0.2.0` to avoid `ImportError: cannot import name ReadTimeoutError`

## Motivation and Context

Master tests are failing for AWS

Travis

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,6 +14,7 @@ deps=
   unittest2<2.0
   boto<3.0
   boto3>=1.4.4
+  botocore>=1.12.36
   sqlalchemy<2.0
   elasticsearch<2.0.0
   psutil<4.0
--- a/tox.ini
+++ b/tox.ini
@@ -14,7 +14,7 @@ deps=
   unittest2<2.0
   boto<3.0
   boto3>=1.4.4
-  botocore>=1.12.36
+  s3transfer==0.1.13
   sqlalchemy<2.0
   elasticsearch<2.0.0
   psutil<4.0


dlstadther:
Current patch removes `ReadTimeoutError`, but introduces an issue with moto actually mocking boto3 services.

Altered local solution to lock `s3transfer==0.1.13` and tests pass. Need to find out why `s3transfer==0.2.0` cause these issues.

dlstadther:
Not saying we have to lock s3transfer version, but want to verify while i return to other tasks.

honnix:
So my interpretation of this PR is still WIP?

dlstadther:
@honnix This PR fixes the failed tests, but it locks us into the previous version of s3transfer. I looked through its commits between version 0.1.13 and 0.2.0 and didn't immediately see anything major.

I'm not convinced this is the best solution. But it does show that something about s3transfer==0.2.0 causes our `moto` tests to break

TMaYaD:
@dlstadther Should I include this in my PRs to make sure tests pass? FWIW, the PRs have nothing to do with AWS directly.

dlstadther:
@TMaYaD I would not include this yet. I'd like to get a better understanding of why this fixes the tests and provide a more sustainable solution.

If your PR(s) don't touch AWS stuff and those 2 are the only failing test suites, there's no reason to prevent that merge.

Tarrasch:
I merged as I think being green is important - I hope you guys don't mind!

PR #2659, opened 2019-02-13
Add Textpert to companies using Luigi

evictor:
LMK if more information needed. :) Thanks for the great OSS.


dlstadther:
Merging as test failures are same as master (and not patched yet)

Thanks for using Luigi!

PR #2661, opened 2019-02-14
Provide automatic URL tracking for Spark applications

GoodDok:
## Description
A new string parameter `tracking_url_pattern` was added to `ExternalProgramTask` task in order to provide a mechanism for parsing the task output and update tracking URL in Luigi web UI on the run. Also the new parameter is set for `SparkSubmitTask` according to the log message structure: https://github.com/apache/spark/blob/0a4c03f7d084f1d2aa48673b99f3b9496893ce8d/core/src/main/scala/org/apache/spark/ui/WebUI.scala#L133

## Motivation and Context
One of the most missing things to me in Luigi, while I'm working with Spark applications, is automatic detection of the link for Spark Web UI. It is very uncomfortable to look for Spark jobs e.g. in Yarn UI among thousands of other jobs, having only the name of the Luigi task plus it's parameters.
The existence of such feature in Luigi would hugely simplify the process of tracking Spark jobs that are running at the moment.

### DIFF
--- a/luigi/contrib/external_program.py
+++ b/luigi/contrib/external_program.py
@@ -36,6 +36,7 @@ import signal
 import subprocess
 import sys
 import tempfile
+from contextlib import contextmanager
 from multiprocessing import Process
 from time import sleep
 
@@ -118,10 +119,8 @@ class ExternalProgramTask(luigi.Task):
 
         try:
             if self.tracking_url_pattern:
-                with _TrackingUrlContext(pattern=self.tracking_url_pattern, main_proc_args=args,
-                                         main_proc_kwargs=kwargs, set_tracking_url=self.set_tracking_url) as proc:
-                    with ExternalProgramRunContext(proc):
-                        proc.wait()
+                with self.proc_with_tracking_url_context(proc_args=args, proc_kwargs=kwargs) as proc:
+                    proc.wait()
             else:
                 proc = subprocess.Popen(args, **kwargs)
                 with ExternalProgramRunContext(proc):
@@ -149,46 +148,42 @@ class ExternalProgramTask(luigi.Task):
                 tmp_stderr.close()
                 tmp_stdout.close()
 
+    @contextmanager
+    def proc_with_tracking_url_context(self, proc_args, proc_kwargs):
+        time_to_sleep = 0.5
 
-class _TrackingUrlContext(object):
-    time_to_sleep = 0.5
+        file_to_write = proc_kwargs.get('stdout')
+        proc_kwargs.update({'stdout': subprocess.PIPE})
+        main_proc = subprocess.Popen(proc_args, **proc_kwargs)
+        pipe_to_read = main_proc.stdout
 
-    def __init__(self, main_proc_args, main_proc_kwargs, pattern, set_tracking_url):
-        # replace process output with PIPE for scanning
-        self.file_to_write = main_proc_kwargs.get('stdout')
-        main_proc_kwargs.update({'stdout': subprocess.PIPE})
-        self.main_proc = subprocess.Popen(main_proc_args, **main_proc_kwargs)
-        self.pipe_to_read = self.main_proc.stdout
+        def _track_url_by_pattern():
+            """
+            Scans the pipe looking for a passed pattern, if the pattern is found, `set_tracking_url` callback is sent.
+            If tmp_stdout is passed, also appends lines to this file.
+            """
+            pattern = re.compile(self.tracking_url_pattern)
+            for new_line in iter(pipe_to_read.readline, ''):
+                if new_line:
+                    if file_to_write:
+                        file_to_write.write(new_line)
+                    match = re.search(pattern, new_line.decode('utf-8'))
+                    if match:
+                        self.set_tracking_url(match.group(1))
+                else:
+                    sleep(time_to_sleep)
 
-        self.pattern = pattern
-        self.set_tracking_url = set_tracking_url
-        self.track_proc = Process(target=self._track_url_by_pattern)
-
-    def __enter__(self):
-        self.track_proc.start()
-        return self.main_proc
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        # need to wait a bit to let the subprocess read the last lines
-        sleep(self.time_to_sleep * 2)
-        self.track_proc.terminate()
-        self.pipe_to_read.close()
-
-    def _track_url_by_pattern(self):
-        """
-        Scans the pipe looking for a passed pattern, if the pattern is found, `set_tracking_url` callback is sent.
-        If tmp_stdout is passed, also appends lines to this file.
-        """
-        pattern = re.compile(self.pattern)
-        for new_line in iter(self.pipe_to_read.readline, ''):
-            if new_line:
-                if self.file_to_write:
-                    self.file_to_write.write(new_line)
-                match = re.search(pattern, new_line.decode('utf-8'))
-                if match:
-                    self.set_tracking_url(match.group(1))
-            else:
-                sleep(self.time_to_sleep)
+        track_proc = Process(target=_track_url_by_pattern)
+        try:
+            track_proc.start()
+            with ExternalProgramRunContext(main_proc):
+                yield main_proc
+        finally:
+            # need to wait a bit to let the subprocess read the last lines
+            track_proc.join(time_to_sleep * 2)
+            if track_proc.is_alive():
+                track_proc.terminate()
+            pipe_to_read.close()
 
 
 class ExternalProgramRunContext(object):
--- a/luigi/contrib/external_program.py
+++ b/luigi/contrib/external_program.py
@@ -31,10 +31,13 @@ from one command to the next, you're probably better off using something like
 
 import logging
 import os
+import re
 import signal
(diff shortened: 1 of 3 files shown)


luigi/contrib/external_program.py
```diff
@@ -132,6 +140,47 @@ def run(self):
                 tmp_stdout.close()
 
 
+class TrackingUrlContext(object):
```
Tarrasch: Can we make this private? `_TrackingUrlContext` - then you don't need to add documentation

luigi/contrib/external_program.py
```diff
@@ -132,6 +140,47 @@ def run(self):
...
+        self.file_to_write = main_proc_kwargs.get('stdout')
+        main_proc_kwargs.update({'stdout': subprocess.PIPE})
+        self.main_proc = subprocess.Popen(main_proc_args, **main_proc_kwargs)
+        self.pipe_to_read = self.main_proc.stdout
+
+        self.pattern = pattern
+        self.set_tracking_url = set_tracking_url
+        self.track_proc = Process(target=self._track_url_by_pattern)
+
+    def __enter__(self):
```
Tarrasch: Is it possible to use contextlib? This reminds me of #2652. :)

luigi/contrib/external_program.py
```diff
@@ -132,6 +140,47 @@ def run(self):
...
+        self.track_proc.start()
+        return self.main_proc
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        # need to wait a bit to let the subprocess read the last lines
```
Tarrasch: Is this the right way to do it? Is there no signalling or waiting for termination way?

test/contrib/external_program_test.py
```diff
@@ -184,6 +184,44 @@ def test_non_mocked_task_run(self):
             # clean up temp files even if assertion fails
             shutil.rmtree(tempdir)
 
+    def test_tracking_url_pattern_works_with_capture_output_disabled(self):
+        temp_dir = tempfile.mkdtemp()
+
+        # since set_tracking_url is called in a separate process
+        # we need to do something like creating a temp file
```
Tarrasch: Maybe you want https://docs.python.org/3/library/multiprocessing.html#sharing-state-between-processes ?

luigi/contrib/external_program.py
```diff
@@ -60,6 +63,7 @@ class ExternalProgramTask(luigi.Task):
     """
 
     capture_output = luigi.BoolParameter(default=True, significant=False, positional=False)
+    tracking_url_pattern = luigi.Parameter(default="", significant=False, positional=False)
```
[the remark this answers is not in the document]
GoodDok: Added description + documentation below

luigi/contrib/external_program.py
GoodDok: Thanks, done

luigi/contrib/external_program.py
GoodDok: Thanks, done, now looks cleaner

luigi/contrib/external_program.py
GoodDok: replaced `sleep` with `track_proc.join(timeout)`, not sure if I have any other option, what do you think?

test/contrib/external_program_test.py
GoodDok: Thank you, changed, now looks much better

luigi/contrib/external_program.py
```diff
@@ -60,6 +64,17 @@ class ExternalProgramTask(luigi.Task):
...
+        default="", significant=False, positional=False,
+        description="Regex pattern used for searching URL in the logs of the external program")
+    """
+    Regex pattern used for searching URL in the logs of the external program.
+
+    If a log line matches the regex, the first group in the matching is set as the tracking URL
+    for the job in the web UI. Example: 'Job UI is here: (https?://.*)'.
+
+    Default value is an empty string, so URL tracking is not performed.
+    """
```
Tarrasch: Does this docs render nice?

luigi/contrib/external_program.py
```diff
@@ -131,6 +148,43 @@ def run(self):
                 tmp_stderr.close()
                 tmp_stdout.close()
 
+    @contextmanager
+    def proc_with_tracking_url_context(self, proc_args, proc_kwargs):
```
Tarrasch: You made it public again -.-

luigi/contrib/external_program.py
```diff
@@ -131,6 +148,43 @@ def run(self):
...
+        try:
+            track_proc.start()
+            with ExternalProgramRunContext(main_proc):
+                yield main_proc
+        finally:
+            # need to wait a bit to let the subprocess read the last lines
+            track_proc.join(time_to_sleep * 2)
+            if track_proc.is_alive():
+                track_proc.terminate()
+            pipe_to_read.close()
```
Tarrasch: Thanks for updating this part of the code. Since it's always easy to make a mistake, would you mind trying to run this for a week in production for you?

GoodDok: Yup, did it with the latest changes, I'll let you know how it goes

luigi/contrib/external_program.py
GoodDok: My bad, fixed

luigi/contrib/external_program.py
GoodDok: Hmm, how can I check it?

GoodDok:
@Tarrasch thank you for being so responsive! 
While deploying these changes for the working pipeline I figured out that by default Spark apps write all the logs into `stderr` (before I was refactoring the code on my local machine and it seems like my Spark config is somewhat unusual writing to `stdout`). Also, the format of output logs for spark depends on `deploy-mode` used in `spark-submit`.
So I introduced a new parameter `track_url_in_stderr` that is False by default + added a test for it. Please let me know if anything else should be added.
Now it's working with the real pipeline, let's take some time, I'll be back with the update.

luigi/contrib/external_program.py
Tarrasch: I recently wrote about it here: https://github.com/spotify/luigi/pull/2548#issuecomment-465763391

luigi/contrib/external_program.py
```diff
@@ -60,6 +65,20 @@ class ExternalProgramTask(luigi.Task):
...
+        description="Regex pattern used for searching URL in the logs of the external program")
+    """
+    Regex pattern used for searching URL in the logs of the external program.
+
+    If a log line matches the regex, the first group in the matching is set as the tracking URL
+    for the job in the web UI. Example: 'Job UI is here: (https?://.*)'.
+
+    Default value is an empty string, so URL tracking is not performed.
+    """
+    track_url_in_stderr = luigi.BoolParameter(
```
Tarrasch: Maybe you want `track_url_in_stderr = luigi.ChoiceParameter`?

GoodDok:
Hi @Tarrasch, just wanted to notify that the changes were applied to the real pipeline and no issues were met during the period of testing.
Also, I was thinking about the changed you suggested regarding `luigi.parameter.ChoiceParameter` and committed yet another version that seems to me a bit easier to understand. 
How it was before:
```
tracking_url_pattern = luigi.Parameter(
        default="", significant=False, positional=False, visibility=ParameterVisibility.HIDDEN,
        description="Regex pattern used for searching URL in the logs of the external program")

track_url_in_stderr = luigi.BoolParameter(
        default=False, significant=False, positional=False, visibility=ParameterVisibility.HIDDEN,
        description="When set to True, 'tracking_url_pattern' is searched in stderr instead of stdout")
```
How it is now:
```
stream_for_searching_tracking_url = luigi.parameter.ChoiceParameter(
        var_type=str, choices=['none', 'stdout', 'stderr'], default='none',
        significant=False, positional=False, visibility=ParameterVisibility.HIDDEN,
        description="Stream for searching tracking URL")

tracking_url_pattern = luigi.OptionalParameter(
        default=None, significant=False, positional=False, visibility=ParameterVisibility.HIDDEN,
        description="Regex pattern used for searching URL in the logs of the external program")
```
How do you think, is it a better option?
The changes are contained in the last commit (they haven't been tested in prod), and the thing that surprises me a lot is a failing **codecov/patch**: the lines related to the context manager are marked as not covered although there are several tests covering this part...

Tarrasch:
Feel free to ignore CodeCov.

I would maybe have preferred

    tracking_url_pattern = luigi.Parameter(...)

But it's not worth blocking this PR more for that. :)

GoodDok:
@Tarrasch thank you very much for a thorough review of my changes! :+1:

Tarrasch:
Please send a fix, also with a test if it makes sense to avoid future breakages.

GoodDok:
@Tarrasch here it is: https://github.com/spotify/luigi/commit/ae065b308b85846aa474b49bf3633cbd8c8a73f4
The changes contain the correctly named parameter + two tests for cluster and client modes.
I tested the changes locally, but I wanted to run the remote ones also. Should I create a new PR to test these changes remotely?

Tarrasch:
I can only merge something (conveniently) if you send it as a Pull Request. That way Travis will also run.

Issue #2570, opened 2018-11-05
KubernetesJobTask "No pod scheduled" error


StasDeep:
I often see the following error when using `KubernetesJobTask`:
```
[28:luigi-interface:224@18:33] [ERROR] [pid 28] Worker Worker(salt=476402868, workers=20, host=something-host-1541289600-zntd7, username=root, pid=1) failed    Something(local_execution=Falser)
Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/site-packages/luigi/worker.py", line 205, in run
    new_deps = self._run_get_new_deps()
  File "/opt/conda/lib/python3.6/site-packages/luigi/worker.py", line 142, in _run_get_new_deps
    task_gen = self.task.run()
  File "/usr/src/app/phpipeline/luigi_util.py", line 231, in run
    super(PHKubernetesJobTask, self).run()
  File "/opt/conda/lib/python3.6/site-packages/luigi/contrib/kubernetes.py", line 355, in run
    self.__track_job()
  File "/opt/conda/lib/python3.6/site-packages/luigi/contrib/kubernetes.py", line 199, in __track_job
    while not self.__verify_job_has_started():
  File "/opt/conda/lib/python3.6/site-packages/luigi/contrib/kubernetes.py", line 261, in __verify_job_has_started
    assert len(pods) > 0, "No pod scheduled by " + self.uu_name
AssertionError: No pod scheduled by something-20181104183346-b51d371a5bfd4197
[1:luigi-interface:570@18:33] [INFO] Informed scheduler that task   Something_42b6a6d55a   has status   FAILED
```
It is hard to reproduce, but seems like sometimes pod needs a bit more time to be created, but task does not wait for it to be created and ends up raising the error. Task gets FAILED status, but a pod is still created and is run without Luigi control.

When the task is restarted and not failed, a pod is run under Luigi control, but there's still that uncontrolled pod, so we end up having two pods making the same thing.

I've managed to fix it the most naive way. When getting pods to verify that they are started, instead of:
```python
pods = self.__get_pods()
```
I do the following:
```python
for _ in range(3):
    pods = self._KubernetesJobTask__get_pods()

    if pods:
        break

    # If pod was not returned, sleep to wait for pod to be created.
    sleep(15)
```
This is not a beautiful way of fixing the issue, but it works (though it will for sure fail if pod needs >45 seconds to be started).

keisuke-umezawa:
I have same issues.

StasDeep:
[...]

PR #2664, opened 2019-02-25
Add max_retrials_to_get_pods

keisuke-umezawa:
## Description
I added `max_retrials_to_get_pods`.

## Motivation and Context
To fix https://github.com/spotify/luigi/issues/2570 .

### DIFF
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -62,9 +62,12 @@ class kubernetes(luigi.Config):
     kubernetes_namespace = luigi.OptionalParameter(
         default=None,
         description="K8s namespace in which the job will run")
-    max_retrials_to_get_pods = luigi.IntParameter(
+    max_retries = luigi.IntParameter(
         default=0,
-        description="Max retrials to get pods' informations")
+        description="Max retries to get pods' informations")
+    poll_time_for_retry = luigi.IntParameter(
+        default=5,
+        description="Poll time for retrying to get pods' informations")
 
 
 class KubernetesJobTask(luigi.Task):
@@ -176,11 +179,18 @@ class KubernetesJobTask(luigi.Task):
         return self.kubernetes_config.max_retrials
 
     @property
-    def max_retrials_to_get_pods(self):
+    def max_retries(self):
         """
-        Maximum number of retrials to get pods' informations.
+        Maximum number of retries to get pods' informations.
         """
-        return self.kubernetes_config.max_retrials_to_get_pods
+        return self.kubernetes_config.max_retries
+
+    @property
+    def poll_time_for_retry(self):
+        """
+        Poll time for retrying to get pods' informations.
+        """
+        return self.kubernetes_config.poll_time_for_retry
 
     @property
     def backoff_limit(self):
@@ -249,7 +259,7 @@ class KubernetesJobTask(luigi.Task):
         pass
 
     def __get_pods(self):
-        for _ in range(self.max_retrials_to_get_pods + 1):
+        for _ in range(self.max_retries + 1):
             pod_objs = Pod.objects(self.__kube_api, namespace=self.kubernetes_namespace) \
                 .filter(selector="job-name=" + self.uu_name) \
                 .response['items']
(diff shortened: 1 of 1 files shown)


luigi/contrib/kubernetes.py
```diff
@@ -239,10 +249,19 @@ def signal_complete(self):
...
+            pod_objs = Pod.objects(self.__kube_api, namespace=self.kubernetes_namespace) \
+                .filter(selector="job-name=" + self.uu_name) \
+                .response['items']
+            pods = [Pod(self.__kube_api, p) for p in pod_objs]
+
+            if pods:
+                break
+
+            # If pod was not returned, sleep to wait for pod to be created.
+            time.sleep(self.__POLL_TIME)
```
dlstadther: if we're going to allow the number of retries to be configurable, shouldn't we also allow the wait time to be configurable as well? You may want to try a lot in a short period or only a few times over a longer duration

luigi/contrib/kubernetes.py
```diff
@@ -239,10 +249,19 @@ def signal_complete(self):
...
+
+        return pods
```
dlstadther: am i correct if the pod isn't available this will return `None`? (just want to be sure i understand)

luigi/contrib/kubernetes.py
```diff
@@ -62,6 +62,9 @@ class kubernetes(luigi.Config):
     kubernetes_namespace = luigi.OptionalParameter(
         default=None,
         description="K8s namespace in which the job will run")
+    max_retrials_to_get_pods = luigi.IntParameter(
```
dlstadther: `retrials` isn't the word you intend. I think you mean `retries`

and to that end, can the var name just be `max_retries`?

keisuke-umezawa:
@dlstadther 
I fixed them! Please review it.

luigi/contrib/kubernetes.py
```diff
@@ -172,6 +178,20 @@ def max_retrials(self):
...
 
+    @property
+    def max_retries(self):
+        """
+        Maximum number of retries to get pods' informations.
+        """
+        return self.kubernetes_config.max_retries
+
+    @property
+    def poll_time_for_retry(self):
```
dlstadther: Do you think `poll_time_for_retry` should replace `__POLL_TIME`?

luigi/contrib/kubernetes.py
keisuke-umezawa: Error logs:

```
  File "/Users/kumezawa/.pyenv/versions/anaconda3-4.3.0/lib/python3.6/site-packages/luigi/worker.py", line 200, in run
  File "/Users/kumezawa/.pyenv/versions/anaconda3-4.3.0/lib/python3.6/site-packages/luigi/worker.py", line 140, in _run_get_new_deps
    task_gen = self.task.run()
  File "/Users/kumezawa/.pyenv/versions/anaconda3-4.3.0/lib/python3.6/site-packages/luigi/contrib/kubernetes.py", line 400, in run
    self.__track_job()
  File "/Users/kumezawa/.pyenv/versions/anaconda3-4.3.0/lib/python3.6/site-packages/luigi/contrib/kubernetes.py", line 233, in __track_job
    while not self.__verify_job_has_started():
  File "/Users/kumezawa/.pyenv/versions/anaconda3-4.3.0/lib/python3.6/site-packages/luigi/contrib/kubernetes.py", line 302, in __verify_job_has_started
    pods = self.__get_pods()
  File "/Users/kumezawa/.pyenv/versions/anaconda3-4.3.0/lib/python3.6/site-packages/luigi/contrib/kubernetes.py", line 274, in __get_pods
    return pods
UnboundLocalError: local variable 'pods' referenced before assignment
```

plizmol:
It seems that the PR was abandoned, however this issue is steel persist and we're using ugly hacks to mitigate it. I would like to own this PR to resolve all remaining code review comments and deliver this fix to next release. Does anyone have any objections? @keisuke-umezawa @dlstadther @Tarrasch

dlstadther:
Feel free to submit a PR resolving this issue. I can provide feedback as my time allows.

PR #2669, opened 2019-03-12
Enable automatic URL tracking for Spark applications

GoodDok:
## Description
https://github.com/spotify/luigi/pull/2661 has provided the mechanism of tracking URL in external programs.
So now we can enable it for Spark applications:
If spark application is run in cluster mode, the string matching the pattern `tracking URL: (https?://.*)\s` is searched.
If spark application is run in client mode, the string matching the pattern `Bound (?:.*) to (?:.*), and started at (https?://.*)\s` is searched.
In both cases the searching is performed in `stderr` stream.

## Motivation and Context
To enable URL tracking in Spark applications.

The PR contains some new unit tests + existing tests were run locally

### DIFF
--- a/luigi/contrib/spark.py
+++ b/luigi/contrib/spark.py
@@ -54,7 +54,7 @@ class SparkSubmitTask(ExternalProgramTask):
     # Only log stderr if spark fails (since stderr is normally quite verbose)
     always_log_stderr = False
     # Spark applications write its logs into stderr
-    track_url_in_stderr = True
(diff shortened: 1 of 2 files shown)


GoodDok:
Ok, I'll be back when it's been working long enough.

GoodDok:
Hi @Tarrasch, just wanted to inform that the pipeline with these changes applied has been working for last ~2 weeks with no issues observed.

Tarrasch:
Awesome. Let's merge then!

PR #2673, opened 2019-03-14
Wildcard support in GCSClient.exists

kagrze:
## Description

Add wildcard support to `GCSClient.exists`.

## Motivation and Context

To be able to use wildcards in the last part of the `GCSTargets` path, e.g.:
```
def output(self):
    return GCSTarget('gs://my_bucket_id/my_dir/*.txt.gz')
```

I've tested this change by calling `GCSClient.exists` from the python console and also by running a simple pipeline with one `ExternalTask` with a wildcard `GCSTarget` output.

### DIFF
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -193,6 +193,11 @@ class GCSClient(luigi.target.FileSystem):
         return response
 
     def exists(self, path):
+        if '*' in path:
+            for _ in self.list_wildcard(path):
+                return True
+            return False
+
         bucket, obj = self._path_to_bucket_and_key(path)
         if self._obj_exists(bucket, obj):
             return True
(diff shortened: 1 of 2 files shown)


kagrze:
I've added a simple unit test.

dlstadther:
For reference, I approved it given a past need I've had to return true for the existence of a file prefix in s3.

dlstadther:
@honnix Which side of the fence on you on here? Is this a bad practice we should avoid supporting? Or is this a valid use case for `exists()`?

lallea:
I think that extending GCSTarget to wildcards would add complexity with little benefit. A wildcard target will not support some operations that GCSTargets do, e.g. reading and writing from run(). Wildcards and dynamic targets can be useful, but it should IMHO either be a GCSWIldTarget or a function that takes a wildcard and emits a list of GCSTarget.

It is a somewhat philosophical question, but we have been bitten by magic in Luigi in the past, and over time moved towards less magic and more explicit behaviour. IMHO that trend is beneficial. Luigi's main strength is simplicity. :-)

Tarrasch:
Closing as of previous discussion.

Issue #2218, opened 2017-08-24
Windows Error: Cannot open output with format parameter specified


shaynewang:
My output definition is:
```
    def output(self):
        fn = os.path.join(self.output_dir,
                          'output_files','export.tsv')
        return luigi.LocalTarget(fn, format=luigi.format.Gzip)
```
But when I try to open the output file to write I get the error below:
```
Traceback (most recent call last):
  File "C:\Users\levinjos\AppData\Local\Continuum\Anaconda2\lib\site-packages\luigi\worker.py", line 192, in run
    new_deps = self._run_get_new_deps()
  File "C:\Users\levinjos\AppData\Local\Continuum\Anaconda2\lib\site-packages\luigi\worker.py", line 144, in _run_get_ne
w_deps
    requires = task_gen.send(next_send)
  File "c:\users\levinjos\classifications_pipeline\classifications\retrieval.py", line 256, in run
    with self.output().open('w') as fl:
  File "C:\Users\levinjos\AppData\Local\Continuum\Anaconda2\lib\site-packages\luigi\file.py", line 134, in open
    return self.format.pipe_writer(atomic_file(self.path))
  File "C:\Users\levinjos\AppData\Local\Continuum\Anaconda2\lib\site-packages\luigi\format.py", line 492, in pipe_writer

    return OutputPipeProcessWrapper(args, output_pipe)
  File "C:\Users\levinjos\AppData\Local\Continuum\Anaconda2\lib\site-packages\luigi\format.py", line 183, in __init__
    close_fds=True)
  File "C:\Users\levinjos\AppData\Local\Continuum\Anaconda2\lib\subprocess.py", line 667, in __init__
    raise ValueError("close_fds is not supported on Windows "
ValueError: close_fds is not supported on Windows platforms if you redirect stdin/stdout/stderr
```
Line here is the cause to the problem, do we need to close file descriptors if the platform is mswindows?:
https://github.com/spotify/luigi/blob/bb9c8f94314c91306475cb568a62d106788a106e/luigi/format.py#L103-L107

luigi (2.6.2)
Machine: Windows 7
Python 2.7.11 |Anaconda 4.0.0 (64-bit)| (default, Feb 16 2016, 09:58:36) [MSC v.1500 64 bit (AMD64)] on win32

atharvai:
I've had this issue on Windows 10 64 bit, Python2.7.14 (64bit), luigi>=2.3.0.
Is there guidance on a workaround until this is fixed?

[...]

PR #2674, opened 2019-03-15
- Create subprocess without preexec_fn or close_fds arguments in Windows

tjrileywisc:
## Description
Added a check for Windows os when creating a subprocess.

## Motivation and Context
I ran into this problem when trying to use ssh on Windows. The ssh example task will create warnings about Windows not supporting preexec_fn arguments to subprocess.Popen. 

This may fix https://github.com/spotify/luigi/issues/2218

These changes will let me run ssh tasks on Windows (ssh and scp both work for me after these changes).

### DIFF
--- a/luigi/format.py
+++ b/luigi/format.py
@@ -23,6 +23,7 @@ import re
 import locale
 import tempfile
 import warnings
+import sys
 
 from luigi import six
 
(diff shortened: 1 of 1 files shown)


luigi/format.py
```diff
@@ -95,6 +96,12 @@ def create_subprocess(self, command):
         http://www.chiark.greenend.org.uk/ucgi/~cjwatson/blosxom/2009-07-02-python-sigpipe.html
```
Tarrasch: Should this be updated as well?

tjrileywisc: I'm unfortunatley not that familiar with *nix so I can't really comment on whether or not this is still relevant.

Tarrasch:
If you can find anyone (worst case it can also be someone from your org to approve this change I think we can merge.

charlesoblack:
Manually implementing this commit causes some "FileNotFound" errors to be thrown for me... It looks like this is an issue with subprocess in that you have to pass `shell=True` (see [here](https://stackoverflow.com/questions/24306205/file-not-found-error-when-launching-a-subprocess-containing-piped-commands)); but then this fails for me as well with an EOFError on a pickled `pandas.DataFrame`.

tjrileywisc:
@charlesoblack have you had any further time to test this?

tjrileywisc:
@charlesoblack have you had any further time to test this?

Issue #2665, opened 2019-03-01
Add a generic DataflowTask to luigi/contrib


ClaireMcGinty:
cc @daikeshi -- we were discussing how much code duplication there is between our Scala/Scio-based Dataflow Task and his Python/Beam-based Dataflow Task, mostly in the [Dataflow argument passing](https://cloud.google.com/dataflow/docs/guides/specifying-exec-params#setting-other-cloud-pipeline-options) and some input/output URI parsing. I think it would make sense to contribute a generic `DataflowTask` to the `contrib` module of Luigi that we can extend for different SDKs.

What do you think? @daikeshi and I could probably work together to contribute this.

(for reference, here's the python/beam task: https://github.com/spotify/spotify-tensorflow/blob/master/spotify_tensorflow/luigi/python_dataflow_task.py)

PR #2675, opened 2019-03-18
add beam_dataflow_task to luigi/contrib

ClaireMcGinty:
#2665 

Adds a task for Beam Dataflow jobs to luigi/contrib, mostly as a wrapper for [DF execution params](https://cloud.google.com/dataflow/docs/guides/specifying-exec-params) and formatting of input/output URIs, with callbacks on successful/unsuccessful execution. Also provides an (overridable) runner to monitor job execution after submitting the job.

CC @daikeshi

### DIFF
--- a/luigi/contrib/beam_dataflow.py
+++ b/luigi/contrib/beam_dataflow.py
@@ -218,7 +218,7 @@ class BeamDataflowJobTask(MixinNaiveBulkComplete, luigi.Task):
     dataflow_params = None
     output_uris = None
 
-    def __init(self):
+    def __init__(self):
         if not isinstance(self.dataflow_params, DataflowParams):
             raise ValueError("dataflow_params must be of type DataflowParams")
 
--- a/luigi/contrib/beam_dataflow.py
+++ b/luigi/contrib/beam_dataflow.py
@@ -139,7 +139,6 @@ class _CmdLineRunner(object):
         if exit_code:
             output = "".join(output_lines)
             raise subprocess.CalledProcessError(exit_code, cmd, output=output)
-        return exit_code
 
 
 @six.add_metaclass(ABCMeta)
@@ -392,9 +391,9 @@ class BeamDataflowJobTask(MixinNaiveBulkComplete, luigi.Task):
         input_args = []
 
         for (name, targets) in job_input.items():
-            uri_targets = luigi.task.flatten(targets)
-            uris = [self._targets_to_uri_getter.get(
-                uri_target.__class__)(uri_target) for uri_target in uri_targets]
+            uris = [
+              self.get_target_path(uri_target) for uri_target in luigi.task.flatten(targets)
+            ]
             if isinstance(targets, dict):
                 """
                 If targets is a dict that means it had multiple outputs.
@@ -423,19 +422,18 @@ class BeamDataflowJobTask(MixinNaiveBulkComplete, luigi.Task):
         output_args = []
 
         for (name, target) in job_output.items():
-            getter = self._targets_to_uri_getter.get(target.__class__)
-            uri = getter(target)
+            uri = self.get_target_path(target)
 
             self.output_uris[name] = uri
             output_args.append("--%s=%s" % (name, uri))
 
         return output_args
 
-    @property
-    def _targets_to_uri_getter(self):
-        return dict([
-            (luigi.LocalTarget, lambda t: t.path),
-            (gcs.GCSTarget, lambda t: t.path),
-            (bigquery.BigQueryTarget, lambda t: "{}:{}.{}".format(
-                t.project_id, t.dataset_id, t.table_id))
-        ])
+    @staticmethod
+    def get_target_path(target):
+        if isinstance(target, luigi.LocalTarget) or isinstance(target, gcs.GCSTarget):
+            return target.path
+        elif isinstance(target, bigquery.BigQueryTarget):
+            "{}:{}.{}".format(target.project_id, target.dataset_id, target.table_id)
+        else:
+            raise ValueError("Target not supported")
--- a/luigi/contrib/beam_dataflow.py
+++ b/luigi/contrib/beam_dataflow.py
@@ -382,6 +382,15 @@ class BeamDataflowJobTask(MixinNaiveBulkComplete, luigi.Task):
         if isinstance(job_input, luigi.Target):
             job_input = {"input": job_input}
 
+        elif isinstance(job_input, tuple):
+            job_input = {job_input[0]: job_input[1]}
+
+        elif isinstance(job_input, list):
+            if all(isinstance(item, tuple) for item in job_input):
+                job_input = dict(job_input)
+            else:
+                job_input = {"input": job_input}
+
         elif not isinstance(job_input, dict):
             raise ValueError("Input (requires()) must be dict type")
 
@@ -400,14 +409,20 @@ class BeamDataflowJobTask(MixinNaiveBulkComplete, luigi.Task):
                 Make the input args in that case "<input key>-<task output key>"
                 """
                 names = ["%s-%s" % (name, key) for key in targets.keys()]
+
             else:
                 names = [name] * len(uris)
+
+            input_dict = {}
+
             for (arg_name, uri) in zip(names, uris):
                 pattern = self.file_pattern().get(name, 'part-*')
-                input_args.append(
-                    "--%s=%s" %
-                    (arg_name, uri.rstrip('/') + '/' + pattern)
-                )
+                input_value = input_dict.get(arg_name, [])
+                input_value.append(uri.rstrip('/') + '/' + pattern)
+                input_dict[arg_name] = input_value
+
+            for (key, paths) in input_dict.items():
+                input_args.append("--%s=%s" % (key, ','.join(paths)))
 
         return input_args
 
@@ -416,7 +431,7 @@ class BeamDataflowJobTask(MixinNaiveBulkComplete, luigi.Task):
         if isinstance(job_output, luigi.Target):
             job_output = {"output": job_output}
         elif not isinstance(job_output, dict):
-            raise ValueError("Input (requires()) must be dict type")
+            raise ValueError("Task output must be Target or dict type")
 
         self.output_uris = {}
         output_args = []
--- a/luigi/contrib/beam_dataflow.py
+++ b/luigi/contrib/beam_dataflow.py
@@ -31,7 +31,7 @@ logger = logging.getLogger('luigi-interface')
 
 
 @six.add_metaclass(abc.ABCMeta)
-class DataflowParams(object):
+class DataflowParamKeys(object):
     """
     Defines the naming conventions for Dataflow execution params.
     For example, the Java API expects param names in lower camel case, whereas
@@ -219,8 +219,8 @@ class BeamDataflowJobTask(MixinNaiveBulkComplete, luigi.Task):
     output_uris = None
 
     def __init__(self):
-        if not isinstance(self.dataflow_params, DataflowParams):
-            raise ValueError("dataflow_params must be of type DataflowParams")
+        if not isinstance(self.dataflow_params, DataflowParamKeys):
+            raise ValueError("dataflow_params must be of type DataflowParamKeys")
 
     @abstractmethod
     def dataflow_executable(self):
@@ -228,7 +228,7 @@ class BeamDataflowJobTask(MixinNaiveBulkComplete, luigi.Task):
         Command representing the Dataflow executable to be run.
         For example:
 
-        return ['java com.spotify.luigi.MyClass', '-Xmx256m']
+        return ['java', 'com.spotify.luigi.MyClass', '-Xmx256m']
         """
         pass
 
@@ -274,14 +274,14 @@ class BeamDataflowJobTask(MixinNaiveBulkComplete, luigi.Task):
         """
         return {}
 
-    def on_output_validation(self):
+    def on_successful_output_validation(self):
         """
         Callback that gets called after the Dataflow job has finished
         successfully if validate_output returns True.
         """
         pass
 
-    def cleanup_on_error(self):
+    def cleanup_on_error(self, error):
         """
         Callback that gets called after the Dataflow job has finished
         unsuccessfully, or validate_output returns False.
@@ -298,21 +298,17 @@ class BeamDataflowJobTask(MixinNaiveBulkComplete, luigi.Task):
             self.cmd_line_runner.run(cmd_line, self)
         except subprocess.CalledProcessError as e:
             logger.error(e, exc_info=True)
-            self.cleanup_on_error()
-            """
-            Exit Luigi with the same exit code as the Dataflow job process, so
-            users can easily exit the job with code 50 to avoid Styx retries
-            https://github.com/spotify/styx/blob/master/doc/design-overview.md#workflow-state-graph
-            """
+            self.cleanup_on_error(e)
             os._exit(e.returncode)
 
         self.on_successful_run()
 
         if self.validate_output():
-            self.on_output_validation()
+            self.on_successful_output_validation()
         else:
-            self.cleanup_on_error()
-            raise ValueError("Output validation failed")
+            error = ValueError("Output validation failed")
+            self.cleanup_on_error(error)
+            raise error
 
     def _mk_cmd_line(self):
         cmd_line = self.dataflow_executable()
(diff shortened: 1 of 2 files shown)


test/contrib/beam_dataflow_test.py
```diff
@@ -0,0 +1,314 @@
...
+    def network(self):
+        return "network"
+
+    @property
+    def subnetwork(self):
+        return "subnetwork"
+
+    @property
+    def labels(self):
+        return "labels"
```
Tarrasch: Is it possible to just do `labels = "labels"`?`

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+    service_account = None
+    zone = None
+    region = None
+    labels = {}
+
+    cmd_line_runner = _CmdLineRunner
+    dataflow_params = None
+    output_uris = None
+
+    def __init(self):
```
daikeshi: `__init` -> `__init__`?

test/contrib/beam_dataflow_test.py
ClaireMcGinty: yes! thanks for the suggestion 👍

luigi/contrib/beam_dataflow.py
ClaireMcGinty: 🤦‍♀️ thanks!

honnix:
@Tarrasch Thanks for looking at this. We will take over and continue with the review.

daikeshi:
LGTM! 👍  
@ClaireMcGinty thanks for putting it into luigi.contrib!

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+        if not isinstance(self.dataflow_params, DataflowParams):
+            raise ValueError("dataflow_params must be of type DataflowParams")
+
+    @abstractmethod
+    def dataflow_executable(self):
+        """
+        Command representing the Dataflow executable to be run.
+        For example:
+
+        return ['java com.spotify.luigi.MyClass', '-Xmx256m']
```
honnix: Shouldn't it be `['java', 'com.spotify.luigi.MyClass', '-Xmx256m']` instead?

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+                )
+
+        return input_args
+
+    def _format_output_args(self):
+        job_output = self.output()
+        if isinstance(job_output, luigi.Target):
+            job_output = {"output": job_output}
+        elif not isinstance(job_output, dict):
+            raise ValueError("Input (requires()) must be dict type")
```
honnix: Copy-paste? Guessing it should be `Output must be ...`

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+                    (arg_name, uri.rstrip('/') + '/' + pattern)
+                )
+
+        return input_args
+
+    def _format_output_args(self):
+        job_output = self.output()
+        if isinstance(job_output, luigi.Target):
+            job_output = {"output": job_output}
+        elif not isinstance(job_output, dict):
```
honnix: `luigi.Task` also supports `Task`, `list` and `tuple` as return types of `output()`. Consider doing the same?

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+            raise ValueError("Input (requires()) must be dict type")
+
+        self.output_uris = {}
+        output_args = []
+
+        for (name, target) in job_output.items():
+            getter = self._targets_to_uri_getter.get(target.__class__)
+            uri = getter(target)
+
+            self.output_uris[name] = uri
```
honnix: Is this dict really needed, apart from in tests? We are able to verify the uri values through the method return value `output_args`.

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+    job_name = None
+    worker_disk_type = None
+    service_account = None
+    zone = None
+    region = None
+    labels = {}
+
+    cmd_line_runner = _CmdLineRunner
+    dataflow_params = None
+    output_uris = None
```
honnix: What is this supposed to be used for?

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+            getter = self._targets_to_uri_getter.get(target.__class__)
+            uri = getter(target)
+
+            self.output_uris[name] = uri
+            output_args.append("--%s=%s" % (name, uri))
+
+        return output_args
+
+    @property
+    def _targets_to_uri_getter(self):
```
narape: I find this clever but obscure way to get the path of a Target.
I think that it will be simpler and more readable to write this function this way:
```python
  @property
  def _get_target_path(self, target):
    if isinstance(target, luigi.LocalTarget) or isinstance(target, gcs.GCSTarget):
      return target.path
    elif isinstance(target, bigquery.BigQueryTarget):
       "{}:{}.{}".format(target.project_id, target.dataset_id, target.table_id))
    else:
      raise ValueError("Target not supported")
```

Then when invoking this, lets say in line 406 you will write it like this:

```python
uris = [self._get_target_path(uri_target) for uri_target in uri_targets]
```

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+
+    def _get_dataflow_args(self):
+        def f(key, value):
+            return '--{}={}'.format(key, value)
+
+        output = []
+
+        runner = self._get_runner()
+        if runner:
+            output.append(f(self.dataflow_params.runner, runner))
```
honnix: `runner` always has a value (defaults to `DirectRunner`). If a user forgets to overwrite `dataflow_params`, then this will fail with an NPE.

Seems like `dataflow_params` should be a required property, or have a default implementation as a fallback.

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+import luigi
+from luigi import six
+from luigi.contrib import bigquery, gcs
+from luigi.task import MixinNaiveBulkComplete
+
+logger = logging.getLogger('luigi-interface')
+
+
+@six.add_metaclass(abc.ABCMeta)
+class DataflowParams(object):
```
honnix: It would be clearer if this is named something like `DataflowParamKeys` or `DataflowParamNames` otherwise it sounds like it wraps the param values, instead of the param names/keys.

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+            "BlockingDataflowPipelineRunner"
+        ]:
+            logger.warning("Using deprecated runner %s. Consider upgrading to "
+                           "Beam 2.x." % self.runner)
+            return self.runner
+
+        else:
+            logger.warning("Found unsupported runner %s. Defaulting to "
+                           "DirectRunner." % self.runner)
+            return "DirectRunner"
```
honnix: Let's just fail loudly. The warning will probably be missed by the users, and if the behaviour is not what they want/except, it can be source of confusion and longer debugging cycles.

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+        ]:
+            return self.runner
+
+        elif self.runner in [
+            "InProcessPipelineRunner",
+            "BlockingDataflowPipelineRunner"
+        ]:
+            logger.warning("Using deprecated runner %s. Consider upgrading to "
+                           "Beam 2.x." % self.runner)
+            return self.runner
```
honnix: Maybe we don't need to support deprecated beam runners?

luigi/contrib/beam_dataflow.py
honnix: If we don't default to `DirectRunner` then the default implementation for `DataflowParams` might not be needed.

test/contrib/beam_dataflow_test.py
```diff
@@ -0,0 +1,261 @@
...
+
+        task.before_run = MagicMock()
+        task.validate_output = MagicMock()
+        task.on_successful_run = MagicMock()
+        task.on_output_validation = MagicMock()
+        task.cleanup_on_error = MagicMock()
+
+        task.run()
+
+        task.before_run.assert_called_once_with()
```
honnix: Just `assert_called_once()` should be enough https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once

luigi/contrib/beam_dataflow.py
ClaireMcGinty: hey, I agree that this method is a bit obscure as is. What I had in mind was that people might want to extend this method for custom Targets. For example, spotify's internal scio luigi extension supports some internally defined target types. can you think of a good way to extend the base target types here?

danielnorberg: Could users then just override `_get_target_path` ?

Also, should the name of a method that is expected to be overridden by subclasses should start with underscore?

test/contrib/beam_dataflow_test.py
ClaireMcGinty: Hm, I used `assert_called_once_with` because it's also a method that's also available in the Python 2 unittest library (`assert_called_once` isn't). It seems like other unit test in the luigi project are doing the same thing

honnix: Thanks. Good to know.

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+            output.append(f(self.dataflow_params.labels, json.dumps(self.labels)))
+
+        return output
+
+    def _format_input_args(self):
+        job_input = self.input()
+        if isinstance(job_input, luigi.Target):
+            job_input = {"input": job_input}
+        elif not isinstance(job_input, dict):
+            raise ValueError("Input (requires()) must be dict type")
```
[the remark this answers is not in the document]
ClaireMcGinty: thanks! it looks like [input()](https://github.com/spotify/luigi/blob/62fbaf20775b69995cacb3e266a8b50a6959fd92/luigi/task.py#L636) does convert Tasks to a `list` of their respective paths, so I definitely ned to add that. do you know where `tuple` output support is implemented?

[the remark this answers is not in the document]
ClaireMcGinty: also -- do you have any thoughts on how a flat `list` of Targets should get represented? for example if our `requires()` is a `Task` like this: 

```
class TestRequires(luigi.ExternalTask):
    def output(self):
        return [
          luigi.LocalTarget(path='some-input-dir'),
          luigi.LocalTarget(path='some-input-dir2')
        ]
```

Should that get formatted as something like `input1=some-input-dir, input2=some-input-dir2`, or `input=some-input-dir1,some-input-dir2`?

luigi/contrib/beam_dataflow.py
ClaireMcGinty: yeah.... this is kind of a result of porting over needed features from our internal implementation, where we use them to publish output in the successful run callback :( @daikeshi also requested them for his team's [use case](https://github.com/spotify/spotify-tensorflow/blob/master/spotify_tensorflow/luigi/python_dataflow_task.py). I guess they could be re-derived in the implementation, though. what do you think?

luigi/contrib/beam_dataflow.py
ClaireMcGinty: good catch, thanks!

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,451 @@
...
+            https://github.com/spotify/styx/blob/master/doc/design-overview.md#workflow-state-graph
+            """
+            os._exit(e.returncode)
+
+        self.on_successful_run()
+
+        if self.validate_output():
+            self.on_output_validation()
+        else:
+            self.cleanup_on_error()
```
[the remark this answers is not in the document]
ClaireMcGinty: Thanks. I modified the method to take in an `error` as input.

luigi/contrib/beam_dataflow.py
ClaireMcGinty: thx! in the `__init__` for `BeamDataflowTask` there's a check that `self.dataflow_params` is an instance of `DataflowParamKeys`. that should be enough of a check, right?

I modified the behavior so that if `self.runner` isn't specified, it defaults to `DirectRunner` with a warning. Otherwise, if any runner that isn't `DirectRunner`/`DataflowRunner` is specified, an exception is raised.

luigi/contrib/beam_dataflow.py
ClaireMcGinty: works for me!

luigi/contrib/beam_dataflow.py
ClaireMcGinty: you're right, i was probably overthinking it. Updated it to be a static method `get_target_path` that can be overridden.

luigi/contrib/beam_dataflow.py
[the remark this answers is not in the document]
ClaireMcGinty: hey @honnix just wanted to ping this discussion, looks like it's the main outstanding issue on the PR?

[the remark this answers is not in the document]
honnix: Sorry for replying late. We will take a look at this.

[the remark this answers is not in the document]
honnix: `tuple` would fall into this branch: https://github.com/spotify/luigi/blob/master/luigi/task.py#L883

Regarding how to represent `input`, I'm not very familiar with Dataflow options. I think either way if fine as long as it can be interpreted correctly.

luigi/contrib/beam_dataflow.py
honnix: This needs to be fixed.

luigi/contrib/beam_dataflow.py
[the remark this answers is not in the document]
ClaireMcGinty: cool, I updated it to support a list of `Target`s, a list of `(key, Target)` tuples, and a single tuple. The `Task` type is handled in the background in the [getpaths()](https://github.com/spotify/luigi/blob/master/luigi/task.py#L838) method. For a flat list of `Targets` I chose to represent it as `--input=some-input-dir1,some-input-dir2`.

luigi/contrib/beam_dataflow.py
ClaireMcGinty: I'm not sure if those types make sense for Dataflow job outputs the same way they do for inputs... maybe `tuple` could be converted to a `dict` but IMO `Task`/`list` are too ambiguous to infer how the user wants them represented as Dataflow job args. what do you think?

luigi/contrib/beam_dataflow.py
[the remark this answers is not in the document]
honnix: I think now this error message is not accurate anymore. How about just `"Invalid input (requires())"? Or we could give a list of supported options in the message.

luigi/contrib/beam_dataflow.py
honnix: `self.output_uris` shadows the class variable, is that intended?

luigi/contrib/beam_dataflow.py
honnix: I not very sure of the use case. But if you believe that is more reasonable, we need to document somehow because that is rather specialized than a normal Luigi `output()`.

luigi/contrib/beam_dataflow.py
```diff
@@ -0,0 +1,454 @@
...
+            output.append(f(self.dataflow_params.job_name, self.job_name))
+        if self.service_account:
+            output.append(f(self.dataflow_params.service_account, self.service_account))
+        if self.labels:
+            output.append(f(self.dataflow_params.labels, json.dumps(self.labels)))
+
+        return output
+
+    def _format_input_args(self):
+        job_input = self.input()
```
honnix: I think we can apply https://github.com/spotify/luigi/blob/master/luigi/task.py#L856 on `job_input` to get a flattened list so that we don't need to handle it ourselves.

honnix: Scratch that. I guess it doesn't fit our use case here.

luigi/contrib/beam_dataflow.py
ClaireMcGinty: yeah, I think it's overall going to be a bit awkward because the `output()` is used both in Luigi itself to check `Task` completion, as well as in the Dataflow java/python job (ie in Scio user code, a user could save to two output locations passed in as args, like  `data.saveAsTextFile(args("textOutput")); data.saveAsBigquery(args("bqOutput"))`). I did add some documentation and a unit test to verify that `Task.complete()` works as expected for dict type outputs.

luigi/contrib/beam_dataflow.py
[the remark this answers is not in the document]
ClaireMcGinty: thanks for catching this, fixed.

luigi/contrib/beam_dataflow.py
ClaireMcGinty: it is intended--do you foresee an issue with that? even if that method is accidentally called twice it's idempotent so should end up with the same value.

luigi/contrib/beam_dataflow.py
ClaireMcGinty: cc @daikeshi, used in implementations of the post-run callbacks.

luigi/contrib/beam_dataflow.py
honnix: Why wouldn't an instance variable like this be good enough? What is the use case to have both class and instance variables?

ClaireMcGinty: the original intent is to avoid having to recompute output URIs twice in any callback implementations but..... it's a pretty cheap computation tbh. Personally i'm ok taking it out. @daikeshi can you comment if you're ok with this? output uris can be recomputed in a one liner like `output_uris = dict(((name, self.get_target_path(target)) for (name, target) in self.output().items()))`

ClaireMcGinty: ended up removing `self.output_uris` for the time being to move this PR forward.

daikeshi: Sorry, somehow I missed this message.. yeah, this is totally fine! We can deal with it in our job class.

PR #2678, opened 2019-03-19
Check if the GCS directory object exists

andrusha:
## Description
`_obj_exists` lists all the objects with given prefix, so in case if you have those objects in your bucket:
```
gs://bucket/dir/obj1
```
And you ask if `gs://bucket/dir/` exists it'll return `True` while no `gs://bucket/dir/` object is present. This broke `mkdir` method, as it would not create directory object if some contents already had been present inside of it.

This fixes that behaviour checking if literal path is present in the list.

```
# Create file in the bucket
$ python
>>> from luigi.contrib.gcs import GCSClient
>>> gcs_client = GCSClient()
>>> gcs_client.put_string(b"", "gs://bucket/testsomething5/test", mimetype='text/plain')

# Verify that only file is there
$ gsutil ls gs://bucket/testsomething5/
gs://bucket/testsomething5/test

# Try to create directory
>>> gcs_client.mkdir("gs://bucket/testsomething5/")

# Nothing happen, still only file object is there
$ gsutil ls gs://bucket/testsomething5/
gs://bucket/testsomething5/test

# Why does it fail? Because this directory already "exists"
>>> gcs_client.exists("gs://bucket/testsomething5/")
True
```

### DIFF
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -136,13 +136,13 @@ class GCSClient(luigi.target.FileSystem):
 
     def _obj_exists(self, bucket, obj):
         try:
-            self.client.objects().get(bucket=bucket, object=obj).execute()
+            objects = self.client.objects().get(bucket=bucket, object=obj).execute()
         except errors.HttpError as ex:
             if ex.resp['status'] == '404':
                 return False
             raise
         else:
-            return True
+            return obj in objects
 
     def _list_iter(self, bucket, prefix):
         request = self.client.objects().list(bucket=bucket, prefix=prefix)


honnix:
Thanks for the PR! Could you please add some test cases in https://github.com/spotify/luigi/blob/master/test/contrib/gcs_test.py#L78 ?

Tarrasch:
Thanks for reviewing @honnix. Can we update the title to clarify this is about gcs?

andrusha:
@honnix @Tarrasch luigi is confused about what directory in Google Storage is

[Google Storage is a _flat file_ storage](https://cloud.google.com/storage/docs/gsutil/addlhelp/HowSubdirectoriesWork), it has no concept of directory, objects are addressed by their full path (which might have `/` in them, but they carry no special meaning).

Some tools like `gsutil` for convenience will split paths on `/` and present them to the user as directories (which do not necessary exist on Google Storage as separate objects). Eg:

```
gs://bucket/long/object/path.avro

$ gsutil ls bucket/
gs://bucket/long

$ gsutil ls bucket/long
gs://bucket/long/object

...
```

This is how `exists` https://github.com/spotify/luigi/blob/master/luigi/contrib/gcs.py#L137-L145 work. It assumes that any part of the path to existing object "exists" the same way `gsutil` does.

Now `mkdir` https://github.com/spotify/luigi/blob/master/luigi/contrib/gcs.py#L294-L303 knows that there are no directories in Cloud Storage, so in order to actually "make a directory" it'll create an empty object on that path.

If you try to emulate hierarchical fs, then `exists` + `mkdir` together make sense, however none of them separately make sense in context of flat fs Google Storage is.

Correct implementation would be:
1. `exists` checks only for presence of actual object.
2. `mkdir`
2.1. `mkdir` is noop
2.2. `mkdir` creates empty object, but why? if it's for https://github.com/GoogleCloudPlatform/gcsfuse support, then it should create the whole chain of objects on the path, but maybe luigi shouldn't be concerned about it at all.

Are there hidden assumptions that underlying fs must be hierarchical or do you have context on why those two methods were implemented in this way?

PR #2679, opened 2019-03-21
Require at least python-dateutil version 2.7.5 instead of only 2.7.5 …

### LINKED ISSUE #2662: Is there a reason python-dateutil is pinned to v2.7.5?
In this [commit](https://github.com/spotify/luigi/commit/ca0aa9afedecda539339e51974ef38cecf180d4b), I can see that python-dateutil has been pinned to version 2.7.5 -  is this strictly necessary? Version 2.8.0 was released a couple of weeks ago and It's causing `ContextualVersionConflict` errors for us.

soxofaan:
Fixes #2662 

## Description
See setup.py diff: replaces `python-dateutil==2.7.5` dependency with `python-dateutil>=2.7.5`.

## Motivation and Context
Luigi as a library should not pin down a dependency on an exact version, but specify an "at least" constraint.

python-dateutil 2.8.0 was released on Feb 5 2019

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ install_requires = [
     'tornado>=4.0,<5',
     # https://pagure.io/python-daemon/issue/18
     'python-daemon<2.2.0',
-    'python-dateutil==2.7.5',
+    'python-dateutil>=2.7.5,<3',
 ]
 
 # Note: To support older versions of setuptools, we're explicitly not


lallea:
I wrote the original line of code, and agree that this PR is a beneficial change. :-)

honnix:
Reasonable change. Thank you for contribution. Could you please put upper boundary limiting to major version? We used to have issues with some dependency, e.g tornado.

soxofaan:
@honnix done

PR #2681, opened 2019-03-25
Fix azureblob splitfilepath to work for windows file paths

jefftalb:
## Description
Just a small change to the splitfilepath function so it will now also work with file paths that have backslashes. 

## Motivation and Context
The splitfilepath function would not work with windows file paths with '\\' in them. 

I have ran my own jobs with this change and have also included a test for paths with a backslash.

### DIFF
--- a/luigi/contrib/azureblob.py
+++ b/luigi/contrib/azureblob.py
@@ -164,7 +164,10 @@ class AzureBlobClient(FileSystem):
 
     @staticmethod
     def splitfilepath(filepath):
-        splitpath = filepath.split("/")
+        if ("/" in filepath):
+            splitpath = filepath.split("/")
+        else:
+            splitpath = filepath.split("\\")
         container = splitpath[0]
         blobsplit = splitpath[1:]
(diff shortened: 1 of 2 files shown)


Tarrasch:

jefftalb:
@manish

luigi/contrib/azureblob.py
```diff
@@ -164,7 +164,10 @@ def rename_dont_move(self, path, dest):
 
     @staticmethod
     def splitfilepath(filepath):
-        splitpath = filepath.split("/")
+        if ("/" in filepath):
```
manish: while this looks good and none of the azureblob tests failed (to my knowledge, would it be possible to use `os.sep` instead of hardcoding the path here?

jefftalb: That is definitely a better idea. I also updated the test paths to be separated by os.sep

jefftalb: Oops, turns out this doesn't work. When I tested it, splitfilepath is being sent paths that are sometimes '/' separated and sometimes '\' separated.

PR #2685, opened 2019-04-03
Rename _requires

TMaYaD:
This is a part of #2657. 

1. Rename `_requires` to `process_requires` to be consistent with other similar methods.
2. Replace calls to `deps` with `process_requires`. Based on previous discussions, it was never meant to be a public method and it's only duplicating the same function as `process_requires`.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -613,16 +613,25 @@ class Task(object):
         """
         return []  # default impl
 
-    def _requires(self):
+    def process_requires(self):
         """
         Override in "template" tasks which themselves are supposed to be
         subclassed and thus have their requires() overridden (name preserved to
         provide consistent end-user experience), yet need to introduce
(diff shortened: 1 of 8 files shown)


TMaYaD:
@tarrasch @dlstadther ping

luigi/task.py
```diff
@@ -613,16 +613,25 @@ def requires(self):
...
         the superclass.
         """
+
+        _requires = getattr(self, '_requires', None)
+        if callable(_requires):
+            warnings.warn("Use of _requires has been deprecated. To customize "
+                          "dependencies in a child class, use `process_requires`.",
+                          DeprecationWarning)
+
+            return _requires()
```
dlstadther: Have you confirmed that this block works as expected? (for those who do override `_requires()`)

PR #2689, opened 2019-04-07
aws batch : job queue as parameter

dav009:
## Description
allowing batch tasks to pass `job_queue` as a parameter when scheduling a task.
default behaviour is left untouched.

## Motivation and Context
- in environments with multiple compute engines and job queues this is a useful parameter.

- locally scheduled a task in a particular queue


dlstadther:
@jfeala Given you added aws batch support, could you confirm this looks good to you? Thanks

jfeala:
Looks good to me

Issue #2688, opened 2019-04-04
Inconsistent dash/underscore handling and usage in configuration parameters


soxofaan:
I was reading through the documentation on Luigi configuration https://luigi.readthedocs.io/en/stable/configuration.html and noticed that there is inconsistent usage of dashes and underscores in the config params. For example under `[core]` there is:
- `default-scheduler-host`, `hdfs-tmp-dir`, `parallel-scheduling-processes`, ...
- `log_level`, `max_reschedules`, `parallel_scheduling`

I found #2133  and #2160 which replaced a lot of dash-cases with underscore-style, but not everything. However, when I started digging into this I noticed that it's not as simple as "underscores is recommended, dashes are deprecated".

The luigi core features (e.g. `worker.py` and `scheduler.py`) use the `luigi.task.Config` approach: they make a `Config` subclass with a bunch of luigi parameters to hold the configuration. With this approach both the underscore-style and dash-style keys in configuration files are supported. Dash-style entries in your configuration will however cause DeprecationWarnings.

On the other hand, a lot of contrib modules (e.g. hadoop, spark, sqla, ...) use `luigi.configuration.get_config().get()` directly which requires the correct spelling of the parameter name, practically dash-style in all cases. If you would put the parameter in underscore-style in your config file, it would not be picked up.

To sum up: officially (core luigi features) undercore-style configs are recommended, but for contrib modules only dash-style works in practice.
This creates a very confusing experience when you are trying to get your config working as expected.

dlstadther:
@soxofaan As noted, we encourage Config-class-style usage. I'd argue instances of the older style should be updated.

soxofaan:
> instances of the older style should be updated.

I agree, and that was what I started doing. 
However, as I tried to explain: `luigi.configuration.get_config().get()` (which is **not** forgiving about config style errors) is used in a lot of contrib modules instead of  the `luigi.task.Config` pattern (which does allow dash/underscore style errors). So just updating the older style instances would however break a lot of existing contrib related configs for a lot of users. 

To avoid backwards incompatible config issues, we should also make sure that `luigi.configuration.get_config().get()` accepts dash/underscore style errors

soxofaan:
no need for wontfix label as the issue is fixed by PR #2691

PR #2691, opened 2019-04-08
Reconcile underscore/dash config style handling (#2688)

soxofaan:
As discussed in #2688:
Underscore config style is recommended Luigi, but `luigi.task.Config` also  helpfully allows dash style in configuration files (with a deprecation warning).

In a lot of places in contrib modules however, `configuration.get_config().get()`  is used, with dash-style e.g. in the spark contrib module:
```
    return configuration.get_config().get(self.spark_version, "deploy-mode", None)
```
`configuration.get_config().get()` does not provide the dash/underscore translation feature, so one is forced to use `deploy-mode` in their config file. Using the seemingly recommended `deploy_mode` alias will not work. This leads to inconsistently styled and confusing configuration files.

This PR reconciles the underscore/dash handling of `luigi.task.Config` and `configuration.get_config().get()`, so that both use the same logic: both underscores and dashes are allowed, but dash-style triggers a DeprecationWarning.

Added some tests about this too.

### DIFF
--- a/luigi/configuration/cfg_parser.py
+++ b/luigi/configuration/cfg_parser.py
@@ -162,7 +162,18 @@ class LuigiConfigParser(BaseParser, ConfigParser):
         Raises an exception if the default value is not None and doesn't match the expected_type.
         """
         try:
-            return method(self, section, option, **kwargs)
+            try:
+                # Underscore-style is the recommended configuration style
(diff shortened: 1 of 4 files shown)


Tarrasch:
Thanks @soxofaan. Quality and correctness is what luigi needs the most I think. :)

PR #2695, opened 2019-04-16
Fix flake errors after moving to python 3

GoodDok:
## Description
Fixed flake errors on Python 3.
Regarding `builtins = unicode`: see https://www.dangtrinh.com/2018/02/how-to-fix-f821-undefined-name-unicode.html

## Motivation and Context
From https://travis-ci.org/spotify/luigi/jobs/520087734:
> Using the default Python version 2.7. Starting on 2019-04-16 the default will change to 3.6. If you wish to keep using this version beyond this date, please explicitly set the Python value in configuration.

Starting from 2019-04-16 builds started to fail (for example, https://travis-ci.org/spotify/luigi/jobs/520844168) showing the errors that haven't been recognized before and preventing other PRs to be built successfully. 

Automatic Travis tests + locally

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -178,6 +178,7 @@ Some more companies are using Luigi but haven't had a chance yet to write about 
 * `Starsky Robotics <https://www.starsky.io/>`_
 * `BaseTIS <https://www.basetis.com/>`_
 * `Hopper <https://www.hopper.com/>`_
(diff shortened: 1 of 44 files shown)


Tarrasch:
Thanks for looking after our build status!

PR #2704, opened 2019-05-08
add callback to metric collector

honnix:
So that they can configure http handler. This is to fix #2628.

## Description

#2628 introduced dependency of `prometheus_client` and that breaks `server.py`. This PR
adds a callback for the metrics collector to further customise http handler.

## Motivation and Context

I'm not entirely sure exposing http handler to contrib is a good idea, but this is to avoid pulling
in dependency or hardcoding a header value only for the sake of a certain metrics collector.


honnix:
@Tarrasch pr @dlstadther PTAL. The latest release of Luigi is broken.

dlstadther:
Screen Shot 2019-05-08 at 1 15 46 PM
> commented 3 hours from now
uhhhh...bug?

honnix:
Indeed, wrong order.

PR #2705, opened 2019-05-08
Fix BigQueryTarget parsing in beam_dataflow module

ClaireMcGinty:
With added unit test coverage.


honnix:
Fyi, you can pip install from github master to do locally test, so no need to get a release.

Issue #2708, opened 2019-05-15
prometheus metrics collector broken


Globegitter:
I have a luigi.conf as follows used by the latest luigi release 2.8.5

```
[scheduler]
retry-delay: 1
retry_count: 0
record_task_history = False
metrics_collector = prometheus

[worker]
keep_alive: false
wait_interval: 1
```

then starting up `luigid --port 8005` and making a request to `localhost:8005/metrics` I get the following exception:

```
etl_1  | 2019-05-15 12:10:07,587 tornado.application[1] ERROR: Uncaught exception GET /metrics (172.22.0.1)
etl_1  | HTTPServerRequest(protocol='http', host='local.ecosia.org:8005', method='GET', uri='/metrics', version='HTTP/1.1', remote_ip='172.22.0.1', headers={'Host': 'local.ecosia.org:8005', 'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64; rv:67.0) Gecko/20100101 Firefox/67.0', 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8', 'Accept-Language': 'de-DE,en-US;q=0.7,en;q=0.3', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive', 'Cookie': 'ECFG=a=1:as=1:cs=0:dt=pc:f=i:fr=0:fs=0:l=en:lt=1557921956:mc=de-de:nf=1:nt=0:t=275:tt=0:tu=auto:wu=auto:ma=1; ecosia_sp2id.3d5d=1c68e633-f199-431a-9723-809c8bc2a816.1552574687.164.1557921957.1557913598.fc7b2121-8875-49ec-88f2-12b92923c01f; lastReadNotificationTimestamp=1557756263397; _ga=GA1.2.869378598.1553858530; _fbp=fb.1.1554110834265.1486831124; lastDiscardedNotification={%22banner%22:1556538435552%2C%22popup%22:1556540594641}; ecosia_sp2ses.3d5d=*', 'Upgrade-Insecure-Requests': '1'})
etl_1  | Traceback (most recent call last):
etl_1  |   File "/usr/local/lib/python3.7/site-packages/tornado/web.py", line 1510, in _execute
etl_1  |     result = method(*self.path_args, **self.path_kwargs)
etl_1  |   File "/usr/local/lib/python3.7/site-packages/luigi/server.py", line 284, in get
etl_1  |     metrics.configure_http_handler(self)
[...]

PR #2709, opened 2019-05-15
fix calling method on wrong object

honnix:
## Description

This fixes bug introduce by #2704. Address #2708.

Unit test


honnix:
PTAL @Globegitter @dbolshak

Issue #2714, opened 2019-05-23
SendGrid integration is deprecated


charlesoblack:
Just like #1956 . Somehow I can't reopen the old issue.

I can work on fixing this, but would appreciate some pointers as to how the config files work exactly.

dlstadther:
The contribution/bugfix is appreciated

What is/are your question(s) about config files?

charlesoblack:
I was under the impression the config worked in a far more complicated manner. Please see PR #2715 . Let me know if any changes need to be done :)

Issue #1956, opened 2016-12-06
Sendgrid email notifications broken by new Sendgrid API


bjlange:
The v3+ version of the `sendgrid-python` package gets rid of SendGridClient in favor of SendGridAPIClient and changes a number of things, so `send_email_sendgrid` (https://github.com/spotify/luigi/blob/master/luigi/notifications.py#L236) now throws an error:
```
Traceback (most recent call last):
  File "/home/ubuntu/.virtualenvs/janrain/local/lib/python2.7/site-packages/luigi/retcodes.py", line 74, in run_with_retcodes
    worker = luigi.interface._run(argv)['worker']
  File "/home/ubuntu/.virtualenvs/janrain/local/lib/python2.7/site-packages/luigi/interface.py", line 237, in _run
    return _schedule_and_run([cp.get_task_obj()], worker_scheduler_factory)
  File "/home/ubuntu/.virtualenvs/janrain/local/lib/python2.7/site-packages/luigi/interface.py", line 196, in _schedule_and_run
    success &= worker.run()
  File "/home/ubuntu/.virtualenvs/janrain/local/lib/python2.7/site-packages/luigi/worker.py", line 1025, in run
    self._handle_next_task()
  File "/home/ubuntu/.virtualenvs/janrain/local/lib/python2.7/site-packages/luigi/worker.py", line 908, in _handle_next_task
    self._email_task_failure(task, expl)
  File "/home/ubuntu/.virtualenvs/janrain/local/lib/python2.7/site-packages/luigi/worker.py", line 540, in _email_task_failure
    headline="A task failed when running. Most likely run() raised an exception.",
  File "/home/ubuntu/.virtualenvs/janrain/local/lib/python2.7/site-packages/luigi/worker.py", line 547, in _email_error
    notifications.send_error_email(formatted_subject, message, task.owner_email)
  File "/home/ubuntu/.virtualenvs/janrain/local/lib/python2.7/site-packages/luigi/notifications.py", line 356, in send_error_email
    recipients=recipients
  File "/home/ubuntu/.virtualenvs/janrain/local/lib/python2.7/site-packages/luigi/notifications.py", line 328, in send_email
    email_sender(sender, subject, message, recipients, image_png)
  File "/home/ubuntu/.virtualenvs/janrain/local/lib/python2.7/site-packages/luigi/notifications.py", line 241, in send_email_sendgrid
    client = sendgrid_lib.SendGridClient(
[...]

PR #2715, opened 2019-05-24
Adapt SendGrid integration to use the updated libraries.

charlesoblack:
## Description

SendGrid changed their API about three years ago to use API tokens instead of username/password combos. The `sendgrid` library adapted, but Luigi's code that depends on SendGrid did not.

## Motivation and Context

This re-instates SendGrid integration for sending out emails when Luigi Tasks fail. It solves #2714 and #1956 (since they are duplicate issues).

I ran a simple Task that force-failed by calling `sys.exit()`, together with a config file using an `apikey` value in `[sendgrid]` and a `method=sendgrid` on `[email]`. I received the "Task failed" emails, even with multiple email addresses set up.

### DIFF
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -232,20 +232,21 @@ def send_email_ses(sender, subject, message, recipients, image_png):
 def send_email_sendgrid(sender, subject, message, recipients, image_png):
     import sendgrid as sendgrid_lib
     client = sendgrid_lib.SendGridAPIClient(sendgrid().apikey)
+
+    to_send = sendgrid_lib.Mail(
+            from_email=sender,
+            to_emails=recipients,
+            subject=subject)
+
     if email().format == 'html':
-        to_send = sendgrid_lib.Mail(from_email=sender,
-            to_emails=recipients,
-            subject=subject,
-            html_content=message)
+        to_send.add_content(message,'text/html')
     else:
-        to_send = sendgrid_lib.Mail(from_email=sender,
-            to_emails=recipients,
-            subject=subject,
-            plain_text_content=message)
+        to_send.add_content(message,'text/plain')
+
     if image_png:
         to_send.add_attachment(image_png)
 
-    response = client.send(to_send)
+    client.send(to_send)
 
 
 def _email_disabled_reason():
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -126,12 +126,9 @@ class smtp(luigi.Config):
 
 
 class sendgrid(luigi.Config):
-    username = luigi.parameter.Parameter(
-        config_path=dict(section='email', name='SENDGRID_USERNAME'),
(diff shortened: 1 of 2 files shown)


luigi/notifications.py
```diff
@@ -234,20 +231,21 @@ def send_email_ses(sender, subject, message, recipients, image_png):
...
+        to_send = sendgrid_lib.Mail(from_email=sender,
+            to_emails=recipients,
+            subject=subject,
+            html_content=message)
     else:
-        to_send.set_text(message)
+        to_send = sendgrid_lib.Mail(from_email=sender,
+            to_emails=recipients,
+            subject=subject,
+            plain_text_content=message)
```
dlstadther: The `Mail()` generation could be more DRY (similar to how it was before):
```python
sendgrid_mail = sendgrid_lib.Mail()

sendgrid_mail.from_email = sender
sendgrid_mail.to = receipients
sendgrid_mail.subject = subject

if email().format == 'html':
    sendgrid_mail.add_content(message, 'text/html')
else:
    sendgrid_mail.add_content(message, 'text/plain')

if image_png:
    sendgrid_mail.add_attachment(image_png)
```

luigi/notifications.py
```diff
@@ -234,20 +231,21 @@ def send_email_ses(sender, subject, message, recipients, image_png):
...
     if image_png:
         to_send.add_attachment(image_png)
 
-    client.send(to_send)
+    response = client.send(to_send)
```
dlstadther: no reason to set `response` bc it is not returned

luigi/notifications.py
charlesoblack: how about initializing `Mail()` with the sender/recipients/subject already? would this be enough?

I originally had it that way with `add_content` dependent on html or not, but I wasn't sure `add_content` would accept `'text/html'` as the mime type (and have no way of testing this currently).

dlstadther: sure, you can initialize mail with sender/receipients/subject before the conditional

As far as the `add_content`, i traced the source code and provided the same values they were are.

charlesoblack: ah, perfect. I guess I was one level too shallow. I'll fix this up and commit again. Is there a way to squash the commits after i've submitted the PR or do I have to start a new PR? (sorry, haven't had much experience with this)

dlstadther: you can just push new commits. We tend to squash merge PRs

dlstadther:
Please take a look at failed tests. Looks like you'll need to fix the flake8 and update the test cfg for `apikey`

charlesoblack:
jeez, that took way too long for me to understand what was going on.

tests should all pass now. sorry for the back-and-forth - i'm not super familiar with travis/tox and unit testing in general. does this look good or should any more changes be done?

PR #2716, opened 2019-05-28
Expandable Namespace Folders for the Visualiser Sidebar

stuson:
## Description

Added expandable folders based off namespace to the sidebar on the task visualiser. Additionally added a clear selection button to avoid having to re-expand a folder to deselect a task family.

Updated sidebar screenshot

## Motivation and Context

We have a very long list of tasks on our machine and would end up scrolling through that list trying to find the task we wanted. I originally wrote this as a JS bookmarklet for our team, but I've adapted it for the source code.

I have ran the visualiser locally and it works.


dlstadther:
What does this look like if you do not explicitly specify namespaces for your tasks? Are they all grouped into one namespace - the default?

stuson:
Tasks with no namespace are grouped into "Other"

stuson:
Updated: The "Others" section is now expanded by default, avoiding extra clicks when no namespaces are specified.

stuson:
Another update: The folder containing the currently selected family should be expanded on page load too I think. If there's no selected family expand the "Others" folder.

dlstadther:
@stuson Can you confirm that this code is working as expected? And that you've been using it in production for awhile?

dlstadther:
Would you also be able to replace the README image for central scheduler image with this version?

Tarrasch:
> Would you also be able to replace the README image for central scheduler image with this version?

That would be great - but I wouldn't block this PR on this. I recall we even had a screenshot the old UI for over a year. :)

Tarrasch:
This feature looks great by the way. I'm totally supporting an effort on evangelizing usage of namespaces.

stuson:
> @stuson Can you confirm that this code is working as expected? And that you've been using it in production for awhile?

We've been using this for about a month with no issues so far for us.

Tarrasch:
Thanks!!

PR #2720, opened 2019-05-31
Documentation: Running Luigi on Windows

tfiers:
## Description
- Added section "Running on Windows" to `doc/running_luigi.rst`
- Made "Running Luigi" a proper item in the documentation side-nav, with sub-items.
  (Before, all subsections of the `doc/running_luigi.rst` file appeared separately as top-level items in the side-nav. This became too long).

## Motivation and Context
I test [my Luigi-based software](https://github.com/tfiers/sharp/) locally on Windows (for later deployment on Linux). That's how I discovered these Windows issues.
I have spent a little time figuring out these issues.
This will spare other Luigi users the same puzzlement.

A simple test to show the multiprocessing issue on Windows:
```py
import luigi

class Burrito(luigi.WrapperTask):
    def requires(self):
        return (Madamme(prot=put) for put in range(30))


class Madamme(luigi.Task):
    prot = luigi.IntParameter()

    def output(self):
        return luigi.LocalTarget(f"{self.prot}.txt")

    def run(self):
            f.write('prot ' * self.prot)

luigi.build([Burrito()], local_scheduler=True, workers=2)
```
Works on Unix, but on Windows only when `workers=1`

Tested locally with `tox -e docs`

### DIFF
--- a/doc/running_luigi.rst
+++ b/doc/running_luigi.rst
@@ -127,16 +127,17 @@ Response of luigi.build()/luigi.run()
          print(luigi_run_result.summary_text)
 
 
-Running Luigi on Windows
-^^^^^^^^^^^^^^^^^^^^^^^^
+Luigi on Windows
+^^^^^^^^^^^^^^^^
 
 Most Luigi functionality works on Windows. Exceptions:
 
-- Specifying multiple worker processes using the `workers` argument for `luigi.build`,
-  or using `--workers` on the command line. (Similarly for specifying
-  `--worker-force-multiprocessing`). For most programs, this will result in
-  failure (a common sight is `BrokenPipeError`). The reason is that worker processes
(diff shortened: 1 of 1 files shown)


tfiers:
@dlstadther, this doc change look good?

mostaphaRoudsari:
I believe you can fix this issue by adding `freeze_support` in Windows. This code should work on Windows:

```python
import luigi
from multiprocessing import freeze_support

class Burrito(luigi.WrapperTask):
    def requires(self):
        return (Madamme(prot=put) for put in range(30))


class Madamme(luigi.Task):
    prot = luigi.IntParameter()

    def output(self):
        return luigi.LocalTarget(f"{self.prot}.txt")

    def run(self):
            f.write('prot ' * self.prot)


if __name__ == '__main__':
    freeze_support()
    luigi.build([Burrito()], local_scheduler=True, workers=2)
```

Issue #2721, opened 2019-06-03
RedisTarget is incompatible with redis-py >= 3


KenyaDonDraper:
From the redis-py v3.0.0 [release notes]*(https://github.com/andymccurdy/redis-py/blob/9b03af26dc829beea232a3248768de933f4c3b67/CHANGES#L9-L15):

> Only bytes, strings and numbers (ints, longs and floats) are acceptable
for keys and values. Previously redis-py attempted to cast other types
to str() and store the result. This caused must confusion and frustration
when passing boolean values (cast to 'True' and 'False') or None values
(cast to 'None'). It is now the user's responsibility to cast all
key names and values to bytes, strings or numbers before passing the
value to redis-py.

The current version of RedisTarget is, however, still trying to set a datetime value - see https://github.com/spotify/luigi/blob/4ab2a909f42f2507310b6d902f2ed9ed0b7eae83/luigi/contrib/redis_store.py#L90. This is causing the following error for those using redis-py >= 3.0.0;

> redis.exceptions.DataError: Invalid input of type: 'datetime'. Convert to a byte, string or number first.

I think we should cast the datetime to a str as a fix. Will be sending a PR shortly.

PR #2722, opened 2019-06-03
Cast datetime to str before setting 'date' key in RedisTarget

KenyaDonDraper:
Fixes  https://github.com/spotify/luigi/issues/2721


dlstadther:
@Gr1f0n6x , given you added Redis support, could you confirm this shouldn't negatively impact existing uses of this module?

Gr1f0n6x:
@dlstadther , unfortunately, I'm not an author of this module. The real author is @pbamba (https://github.com/spotify/luigi/pull/2489). 
But, as I think, this change should not has negative impact. For existing code it will be equal to double call to str. It may add some overhead (probably, too small for worrying).

PR #2727, opened 2019-06-14
sdist upload has been deprecated for a while

honnix:
## Description

New release process.

## Motivation and Context

`python setup.py sdist upload` has been deprecated for a while.

By following it and getting a new version released.


honnix:
I wish we had an automated way. :D No, all manually crafted.

How could this PR fail all CI builds?

honnix:
Does this seem useful? https://github.com/apps/release-drafter

Not super powerful, but at least could automate some.

PR #2732, opened 2019-06-27
Change GET request to POST requests

peter-volkov:
GET requests should not have http body according to RFC.
Works well with POST here, and fails with GET for some cases:
2019-03-22 20:51:24,589 tornado.application[19] ERROR: Uncaught exception GET /api/add_task ...

  File "contrib/python/tornado/tornado/web.py", line 1510, in _execute
  File "contrib/python/luigi/luigi/server.py", line 120, in get
  File "contrib/python/luigi/luigi/scheduler.py", line 819, in add_task
    assert worker is not None
AssertionError

By the way: urllib fetcher uses POST here (urlopen(full_url, body, timeout).)

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -92,7 +92,7 @@ class RequestsFetcher(object):
 
     def fetch(self, full_url, body, timeout):
         self.check_pid()
-        resp = self.session.get(full_url, data=body, timeout=timeout)
+        resp = self.session.post(full_url, data=body, timeout=timeout)
         resp.raise_for_status()
         return resp.text
 


honnix:
Since this will be a breaking change, shall we support both methods for a few more versions? Of course we can mark the `GET` method as deprecated.

peter-volkov:
It is not a breaking change as I can see.
Im already using this fix in production. Any other info is needed from me?

Tarrasch:
Makes sense I suppose - honestly I don't know.

Can you just clarify when one gets the stack trace? For particular version of libraries or is it your task that is special?

peter-volkov:
I guess the thing is in third-party http-balancer that stands between luigi client and scheduler in my infrastructure. Maybe the balancer drops the http body. As far as I know it does not reproduce with http connection.
Anyway I would merge it for the sake of expected behavior.

Tarrasch:
Thanks for this patch @peter-volkov !

PR #2735, opened 2019-07-10
Bump tornado dependency to >=5.0,<6 version

kozako2:
## Description
Bump of tornado version because there is a dependency mismatch when  installing this in `jupyter/minimal-notebook:d4cbf2f80a2a` docker image. 

Tested using `tox -e py37`. Nevertheless more tests needed,

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ with open('README.rst') as fobj:
     long_description = readme_note + fobj.read()
 
 install_requires = [
-    'tornado>=4.0,<5',
+    "tornado>=5,<6",
     # https://pagure.io/python-daemon/issue/18
     'python-daemon<2.2.0',
     'python-dateutil>=2.7.5,<3',


dlstadther:
We attempted to update Tornado once before and ran into issues (#2490). Not sure if you were aware. Happy to chat about your problems and proposed solutions!

daritter:
Hi all,

After looking in #2490, Would it be possible to loosen the tornado requirements if the python version is >=2.7.9 or >=3.4?

```
# Tornado>5 requires updated ssl module so we only allow it for recent enough versions of python: 3.4+ and 2.7.9+
if sys.version_info[:2] >= (3, 4) or (sys.version_info[:2] == (2, 7) and sys.version_info[2]>=9):
     install_requires.append('tornado>=4.0,<6')
else:
     install_requires.append('tornado>=4.0,<5')
```

(I would not mind to not care about 2.7 and just relax the requirement for 3.4+). 

I'd be happy to make a pr if requested ...

arc12:
+1 for resolving this; it makes it impossible for me to get a mutually compatible environment.

honnix:
I believe this is covered by https://github.com/spotify/luigi/pull/2761. Please feel free to reopen if this PR intends to cover more.

Issue #2738, opened 2019-07-12
Datadog metrics collector sends all params as tags


tophers42:
Was trying to use the datadog metrics collector as added in: https://github.com/spotify/luigi/pull/2434.
It's a great plugin, but there is one major issue that makes it unsuitable for us to use and we've had to monkey patch it.
The collector sends all params for a task to the datadog as tags. This is problematic for two reasons:
        1. Params tend to have high cardinality which is not compatible with datadog tags (see number 6: https://docs.datadoghq.com/tagging/#defining-tags)
        2. Params can contain secrets and there doesn't seem to be any filtering of which
        params are sent to DD. At the very least, the datadog collector should respect the visibility of params. 

For reference here's how we monkey-patched it in our `__init__.py`
```
from luigi.contrib import datadog_metric


def override_format_task_params_to_tags(self, task):
    """
    Monkey patch an override to DatadogMetricsCollector._format_task_params_to_tags
    In general, params should not be sent as tags to DD for multiple reasons:
        1. Params tend to have high cardinality which is not compatible with datadog tags
        2. Params can contain secrets and there doesn't seem to be any filtering of which
        params are sent to DD.
    """
    return []


datadog_metric.DatadogMetricsCollector._format_task_params_to_tags = override_format_task_params_to_tags  # noqa: E501
```

PR #2740, opened 2019-07-12
Only send public params to datadog

tophers42:
* Update DatadogMetricsCollector to respect HIDDEN and PRIVATE param visibility

## Description
This PR updates the DatadogMetricsCollector to respect HIDDEN and PRIVATE params and _not_ send those as tags to Datadog.

Debatably, params should not be used as tags at all (or at least not by default) as params tend to have high cardinality (dates, ids, etc) which is not recommended by datadog: https://docs.datadoghq.com/tagging/#defining-tags (see item 6)

## Motivation and Context
Per my reported issue: https://github.com/spotify/luigi/issues/2738 using the datadog metrics collector has the potential to leak secrets to datadog as tags

* Added unit tests

### DIFF
--- a/luigi/contrib/datadog_metric.py
+++ b/luigi/contrib/datadog_metric.py
@@ -109,7 +109,7 @@ class DatadogMetricsCollector(MetricsCollector):
 
     def _format_task_params_to_tags(self, task):
         params = []
-        for key, value in task.params.items():
+        for key, value in task.public_params.items():
             params.append("{key}:{value}".format(key=key, value=value))
 
         return params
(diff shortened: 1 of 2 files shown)


dlstadther:
@cabouffard given your contribution of Luigi's datadog integration, perhaps you can discuss the use of params as tags. If this is a deficiency of the implementation, do either of you have suggestions on how to allow tag assignment to be configurable?

tophers42:
I agree the ideal solution seems to be to make the tags configurable. Perhaps another visibility enum entry: `ParameterVisibility.METRIC`, which behaves the same as `PUBLIC`, but is also intended to be visible to metrics collectors?

PR #2745, opened 2019-07-15
Fix SendGrid email API documentation.

### LINKED ISSUE #2741: SendGrid documentation on Config documentation page is incorrect
When doing PR #2715 , I forgot to update the configuration documentation for SendGrid. Specifically, the `username`/`password` options should be replaced with `apikey`.

charlesoblack:
## Description
Fixes the SendGrid documentation to mention the API key instead of username/password combination. Closes #2741 .

## Motivation and Context
After PR #2715 , which updated the usage of SendGrid to comply with the new API format, the documentation was forgotten. This fixes the RST file.

No tests required.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -363,7 +363,7 @@ check_unfulfilled_deps
 
 force_multiprocessing
   By default, luigi uses multiprocessing when *more than one* worker process is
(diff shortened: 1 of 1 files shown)


charlesoblack:
Thanks dlstadther!

PR #2751, opened 2019-07-25
Adds "Force Commit" button in UI to set tasks to DONE

brandonAIQ:
## Description
While using Luigi, we've frequently come into the situation where members of our team need to "force-commit" a task, i.e. convince Luigi that a task is done despite that it's failed or still running, in order to let other important upstream tasks run. Maybe it didn't actually fail but threw an exception while cleaning up, or what the task was supposed to do was fixed manually -- we don't want the task to run again, since that'd take too much time or do the operation again, we just want Luigi to allow its dependents to run.

This PR adds such a button!

I built the project, went to the UI with a dummy task, and verified that the button does what it's supposed to. I tried both with "a task that's running too long" (i.e. the `RUNNING -> DONE` example) and "a task that failed" (i.e. the `FAILED -> DONE` case).

This is my first time contributing to Luigi, so please be open with the feedback and let me know if there's anything else I should do to make this merge-able!


brandonAIQ:
Can someone have a look at this and let me know if there's something else I should do?

dlstadther:
@riga It looks like you're one of the most recent folks to add new rpc methods for interaction via the GUI. Could you take a look here?

@brandonAIQ Are you able to add any new tests for this new rpc method?

brandonAIQ:
Good idea! I'll add a test now

brandonAIQ:
Ok, added a couple unit tests 🙂

riga:
Looks good to me.

I'm just wondering if "force commit" is the right name here. If I was new to luigi and saw a "force commit" button on the scheduler interface, I wouldn't know what to expect. Maybe sth like "mark as done" is more verbose?

brandonAIQ:
Oh, yeah "force commit" is just what we call it internally, could be clearer -- I'll update

luigi/static/visualiser/index.html
```diff
@@ -35,14 +35,14 @@
...
                 {{^statusMessage}}
                   {{#progressPercentage}}<button class="btn btn-primary btn-xs statusMessage" title="Status message" data-toggle="tooltip" data-task-id="{{taskId}}" data-display-name="{{displayName}}"><i class="fa fa-comment"></i></button>
                   {{/progressPercentage}}
                 {{/statusMessage}}
-                {{#acceptsMessages}}<button class="btn btn-success btn-xs schedulerMessage" title="Send message" data-toggle="tooltip" data-task-id="{{taskId}}" data-display-name="{{displayName}}" data-worker="{{workerIdRunning}}"><i class="fa fa-paper-plane"></i></button>{{/acceptsMessages}}
+                {{#acceptsMessages}}<button class="btn btn-default btn-xs schedulerMessage" title="Send message" data-toggle="tooltip" data-task-id="{{taskId}}" data-display-name="{{displayName}}" data-worker="{{workerIdRunning}}"><i class="fa fa-paper-plane"></i></button>{{/acceptsMessages}}
             </div>
```
riga: Yeah, with increasing number of buttons there might be a need for custom colors ;)

brandonAIQ:
Alright, made the updates. Only "weird" thing I did was that I kinda stole the `.btn-success` style from the "send message" button, since it'd be confusing to have two green buttons sitting right next to each other, and the only bootstrap style we _haven't_ used yet is `.btn-default`, which makes just no sense for a "force commit" button to use :P 

Let me know if you have a better approach on this one!

luigi/static/visualiser/index.html
brandonAIQ: Oh, hah yeah commented on this down below :) Yeah I was dying for `.btn-yellow` or something but it seems like bootstrap doesn't have anything like that

brandonAIQ:
Bump. Does anyone have any more feelings about this PR?

dlstadther:
@riga seems to approve. You added tests and tests are passing.

If i don't hear anything else, i'll merge

PR #2759, opened 2019-08-10
Show task history link in visualizer when recording.

riga:
## Description

This PR adds a link to the task history page (`/history`) when `record_task_history` is enabled:

history link

To achieve that I
- added a new rpc method to `Scheduler` that returns the `record_task_history` setting,
- added a template for top navigation items,
- added the link to the top navigation after an initial check using the rpc method.

## Motivation and Context

In some projects, easy browsing of the task history might be an important feature, so having a link in the top navigation bar seems like a good thing.

## Tests

I added two simple tests that invoke the added rpc method and expect either True or False, depending on the `record_task_history` setting.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -993,6 +993,10 @@ class Scheduler(object):
         return {"response": response}
 
     @rpc_method()
+    def has_task_history(self):
+        return self._config.record_task_history
+
+    @rpc_method()
     def is_pause_enabled(self):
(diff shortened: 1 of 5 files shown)


dlstadther:
I'm not sure i'm totally following the appearance of the change. Could you fix the link to the image? (I'm getting 401)

Glad to see new tests which are passing!

riga:
@dlstadther I fixed the image link.

PR #2760, opened 2019-08-10
Add internal version info

riga:

This PR adds version (and meta) infos to the luigi package itself, and by extension provides them to packages built on top of luigi. E.g. the version can be queried now by doing 

```python
import luigi
luigi.__version__
```

which is standard in many other packages as well.

The current luigi version is only known in `setup.py` at the time of building the package. Therefore, 
I added `__meta__.py` which contains meta package information like version number, author, license etc. This file is used by both the setup file and the luigi package itself, so that there is only one place where versions are bumped.

Also, this allows for having a `--version` parameter on the command line. I think it is sufficient if either `luigid` or `luigi` has that feature, and since I didn't want to interfere with the argument parsing in `cmdline_parser.py`, I added it to `luigid`.

## Tests

I added a command line test that calls `luigid --version` and checks the output.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,14 +58,19 @@ if os.environ.get('READTHEDOCS', None) == 'True':
     install_requires.remove('python-daemon<2.2.0')
     install_requires.append('sphinx>=1.4.4')  # Value mirrored in doc/conf.py
 
+# load meta package infos
+meta = {}
+with open("luigi/__meta__.py", "r") as f:
+    exec(f.read(), meta)
+
 setup(
     name='luigi',
-    version='2.8.7',
-    description='Workflow mgmgt + task scheduling + dependency resolution',
+    version=meta['__version__'],
+    description=meta['__doc__'],
     long_description=long_description,
-    author='The Luigi Authors',
-    url='https://github.com/spotify/luigi',
-    license='Apache License 2.0',
+    author=meta['__author__'],
+    url=meta['__contact__'],
+    license=meta['__license__'],
     packages=[
         'luigi',
         'luigi.configuration',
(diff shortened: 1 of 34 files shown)


setup.py
```diff
@@ -58,14 +58,19 @@ def get_static_files(path):
     install_requires.remove('python-daemon<2.2.0')
     install_requires.append('sphinx>=1.4.4')  # Value mirrored in doc/conf.py
 
+# load meta package infos
+meta = {}
+with open("luigi/__meta__.py", "r") as f:
+    exec(f.read(), meta)
```
riga: One could also do sth like

```python
sys.path.insert(0, "luigi")
import __meta__ as meta
sys.path.pop(0)
```

I have no strong preference here.

dlstadther:
Could you also update the note in `release-process.rst` if we're moving where the `version` is stored?

honnix:
I agree with @Tarrasch . Is it needed?

riga:
Sorry for the long pause.

Personally, I only need to access the version information on the `luigi` module itself, so I'm fine of course with removing it again from `luigid`.

riga:
Kind ping =)

honnix:
@dlstadther LGTM. We will adapt the release process a bit.

honnix:
Sorry for backing to this PR. Not entirely sure it is caused by the change here, but since https://pypi.org/project/luigi/2.8.11/ we have lost all metadata in pypi, and 2.8.11 is the first release including this change. @riga Would you mind taking a look at this? Thanks.

riga:
@honnix Sure, I will look into this later today.

Issue #2734, opened 2019-07-08
enum34 dependency breaks install using poetry on python 3.7.3


Timost:
Hi,
I've several projects that depend on luigi, I use poetry to manage their dependencies and when trying to build them on python `3.7.3` the install fails because of `enum34`.
The issue with `enum34` has impacted many well known projects (see https://github.com/sdispater/poetry/issues/1122 for more details):
- https://github.com/sdispater/poetry/issues/881
- https://github.com/sdispater/poetry/issues/844
- https://github.com/tensorflow/tensorflow/issues/15136

It seems these projects had to modify their `setup.py` to properly use environment markers as specified in https://www.python.org/dev/peps/pep-0508/#environment-markers.

I've taken a quick look at the `setup.py` of luigi and there is a comment dating back from 2016 guarding against the use of those environment markers for `enum34`. I was wondering if this was still needed and if not if it could be be modified to conform to PEP508.

Tarrasch:
Hi @Timost, feel free to do a PR making us conform to PEP508 - most maintainers here (including me) are not really packaging gurus. Help to make us conform to the standards would be much appreciated.

Timost:
Hi @Tarrasch, Thank you for your answer.

I'm not a packaging guru either. I've dug into the issue a bit more and it seems that this issue is only present when using poetry with python 3.7.3. Using pip or pipenv, the install works with python 3.7.3.
There are actually two severity degrees to this issue. When installing luigi using poetry on python 3.7.3 the `enum34` package gets installed (although it shouldn't) but the installation succeeds. When using poetry's path dependencies together with poetry and python 3.7.3 the install actually crashes (see https://github.com/sdispater/poetry/issues/1122#issuecomment-500124981)

From what I gather from https://github.com/grpc/grpc/issues/18013#issuecomment-476763524 and https://github.com/sdispater/poetry/issues/328#issuecomment-407472719 poetry does not parse the setup.py file so cannot take into account the `> python 3.4` guard that is implemented in python. Poetry  only parses the data coming from the built package provided by pipy (probably `luigi-2.8.7/luigi.egg-info/requires.txt`), In which `enum34` is listed with no `<python3.4` environment marker.

I've taken a look at https://github.com/spotify/luigi/pull/2623 which actually reverts the changes that are needed for the install to work when using poetry. Since this is a pretty recent PR and the issue seems present only when using poetry, python 3.7.3. I'm not sure what the best decision is.
[...]

PR #2761, opened 2019-08-12
Update tornado requirement for new enough python versions

daritter:
## Description

Increase the tornado version requirement to allow tornado 5 but only on systems with python 3.4+ or 2.7.9+

This is a replacement for #2735 

## Motivation and Context

This has been tried before for all supported python version in #2490 but had to be reverted in #2503  due to incompatibilities in the ssl module for old python versions. 

However the latest version of jupyter notebook (starting with 6.0) now requires [tornado >=5](https://github.com/jupyter/notebook/commit/729183b14830fb32063b145348c875ebfef278c7) so it is no longer possible to have one consistent stack with both jupyter and luigi.

same changes as tested in #2490 but no additional tests by me

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,11 +38,18 @@ with open('README.rst') as fobj:
     long_description = readme_note + fobj.read()
 
 install_requires = [
-    'tornado>=4.0,<5',
     # https://pagure.io/python-daemon/issue/18
     'python-daemon<2.2.0',
     'python-dateutil>=2.7.5,<3',
 ]
+
+# Tornado >=5 requires updated ssl module so we only allow it for recent enough
+# versions of python (3.4+ and 2.7.9+).
+if sys.version_info[:2] >= (3, 4) or (sys.version_info[:2] == (2, 7) and
+                                      sys.version_info[2] >= 9):
+    install_requires.append('tornado>=4.0,<6')
+else:
+    install_requires.append('tornado>=4.0,<5')
 
 # Note: To support older versions of setuptools, we're explicitly not
 #   using conditional syntax (i.e. 'enum34>1.1.0;python_version<"3.4"').


setup.py
```diff
@@ -38,12 +38,19 @@ def get_static_files(path):
...
     # https://pagure.io/python-daemon/issue/18
     'python-daemon<2.2.0',
     'python-dateutil>=2.7.5,<3',
 ]
 
+# Tornado >=5 requires updated ssl module so we only allow it for recent enough
+# versions of python (3.4+ and 2.7.9+).
+if sys.version_info[:2] >= (3, 4) or (sys.version_info[:2] == (2, 7)
+                                      and sys.version_info[2]>=9):
+     install_requires.append('tornado>=4.0,<6')
```
dlstadther: wouldn't you want this one to be between 5 and 6?

daritter: I assumed that keeping it at  4.0 as minimum requirement is the correct thing to do as luigi does work with tornado 4.0 and I didn't want to introduce any possible problems that the minimum version is different depending on the python version.

Increasing the minimum version only would make sense if luigi would start using features from tornado 5 which will be complicated due to the split requirement.

dlstadther: That's fair. I'm ok with that.

tomzx:
Would it be possible to also support tornado 6?

dlstadther:
@tomzx [Tornado 6 notes lack of support for Python 2.7 and 3.4](https://www.tornadoweb.org/en/stable/releases/v6.0.0.html)

So the addition of Tornado 6 support would require another conditional check for Python version >= Python 3.5.2

Additionally, you'd need to verify no functional support has been removed for how Luigi calls Tornado.

Timost:
Hi everyone, 
I think the latest best practice for specifying dependencies  is to follow [PEP508](https://www.python.org/dev/peps/pep-0508/#environment-markers) and use environment markers. I've opened #2734 regarding the `enum34` dependency where using the `sys.version_info` conditional breaks the install when using poetry.

That being said I'm not entirely sure how this specific conditional would translate into environment markers '^^. Maybe something like:
```
'tornado>=4.0,<6; python_version >= "3.4" or (python_version == "2.7" and python_full_version >= "2.7.9")'
'tornado>=4.0,<5; (python_version > "2.7" and python_version < "3.4") or (python_full_version < "2.7.9")'
```

setup.py
honnix: @dlstadther Sorry for replying late. I was on a long vacation. Yes I'm OK with this approach.

honnix:
@Timost Thanks for firing the issue. Shall we merge this PR as it is  and follow up in #2734, which might take a bit more time to conclude? What do you think?

Timost:
@honnix sounds like a reasonable compromise to me yes.

honnix:
@Timost Thanks for understanding. @dlstadther Do you agree to merge?

dlstadther:
I agree

PR #2764, opened 2019-08-14
Add worker locks for (optional) inter-task synchronization

riga:
#  Description

This PR adds a multiprocessing-safe `Lock()` object to workers. The lock is passed to task processes and forwarded to running tasks to be able to synchronize between multiple tasks.

# Motivation

At the end of some of our tasks, we print summaries that are quite useful in our use-cases. When running with multiple worker processes, however, the summaries are often mixed and become rather impractical. This can be solved by using multiprocessing-aware locks.

# Tests

I added a new test case that runs two tasks with each of them adding two lines to the same log file. Using some delays (not optimal, but it's hard so simulate (a)synchronicity) I forced the two lines to be mixed between the tasks. When the worker lock is used, the lines are written into the file one task after another.

### DIFF
--- a/doc/luigi_patterns.rst
+++ b/doc/luigi_patterns.rst
@@ -353,3 +353,48 @@ messages:
 
 Messages can be sent right from the scheduler UI which also displays responses (if any). Note that
 this feature is only available when the scheduler is configured to send messages (see the :ref:`scheduler-config` config), and the task is configured to accept them.
+
+Synchronization between tasks
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some use cases might require the synchronization of certain parts in the ``run()`` method of tasks. A good example is the writing of a summary to a shared file or terminal output. Consider the following task:
+
+.. code-block:: python
+
+    class ASyncExample(luigi.Task):
+
+        name = luigi.Parameter()
+
+        def output(self):
+            ...
+
+        def run(self):
+            # do some computation
+            data = ...
+
+            # write a long-ish summary to stdout
+            print("summary of task " + self.name)
+            for obj in data:
+                print(obj)
+
+When two ``ASyncExample`` tasks with different ``--name`` parameters are run simultaneously by a worker with 2 worker processes (``--workers 2``) and the point of printing the summary is reached at the same time, the terminal output is likely to be mixed between the two tasks. This way, the respective summaries become rather impractical.
+
+To ensure that the summaries are printed one after another, you can make use the ``worker_lock`` that is known to tasks within their ``run()`` method:
+
+.. code-block:: python
+
+    class SyncExample(luigi.Task):
+
+        ...
+
+        def run(self):
+            # do some computation
+            data = ...
+
+            # write a long-ish summary to stdout, avoid interference using the worker lock
+            with self.worker_lock:
+                print("summary of task " + self.name)
+                for obj in data:
+                    print(obj)
+
+Please note that lock objects are defined per worker (as the name suggests), so synchronization can only be achieved between tasks run by the same worker.
--- a/test/worker_test.py
+++ b/test/worker_test.py
@@ -27,7 +27,7 @@ import threading
 import time
 
 import psutil
-from helpers import (unittest, with_config, skipOnTravis, LuigiTestCase,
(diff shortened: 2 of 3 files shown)


riga:
Mh, a lot of test appear to fail but when I try to reproduce them (python 3.7) they all pass. Might be travis issues. Can someone maybe trigger the tests again?

dlstadther:
I haven't reviewed the PR yet, but i've restarted the failing Travis tests

test/worker_test.py
```diff
@@ -1042,6 +1042,62 @@ def test_stop_getting_new_work_assistant_keep_alive(self):
...
+        log_file = luigi.Parameter()
+
+        def write_line(self, line):
+            with open(self.log_file, "a") as f:
+                f.write(line + "\n")
+
+        def run(self):
+            time.sleep(self.delay1)
+            if self.use_locking:
+                self.worker_lock.acquire()
```
dlstadther: So if you don't acquire and release the lock, there are no functional changes? (i.e. none impacting to existing users who don't explicitly use this feature)

test/worker_test.py
```diff
@@ -1042,6 +1042,62 @@ def test_stop_getting_new_work_assistant_keep_alive(self):
...
+            self.write_line("start " + self.name)
+            time.sleep(self.delay2)
+            self.write_line("end " + self.name)
+            if self.use_locking:
+                self.worker_lock.release()
```
dlstadther: Suppose someone uses this feature, but doesn't release the lock. That wouldn't be good. Perhaps provide some documentation on the feature and possible use cases.

test/worker_test.py
riga: Exactly.

test/worker_test.py
riga: Yup, I will add a section to the "Luigi Patterns" once the tests look fine.

riga: Also, I would encourage to use the lock context guard in the docs to avoid trouble for users forgetting to release the lock in most cases.

riga:
There seems to be memory issues since `os.fork()` is called too often and the memory of the main processes is copied every time.

I digged a bit deeper and it turns out that the previous approach of calling `multiprocessing.Manager().Queue()` actually resulted in better garbage collection, though I don't understand why the queue object still worked. I will add "[WIP]" label to the PR title as long as I debug.

riga:
Ok, looks like the `multiprocessing.Manager()` is intended to be used such that it can be garbage collected while the created queue objects still work properly, so memory is not an issue. I changed the PR description and removed the [WIP] label again. Ready for review.

lallea:
+1 to what @Tarrasch said. A suggestion: Split out the printouts to a reporting Task that runs with the computation Task as a dependency. Let each reporting Task use a Luigi resource of which there is only one. That would effectively serialise the report steps.

doc/luigi_patterns.rst
```diff
@@ -353,3 +353,48 @@ messages:
...
+            data = ...
+
+            # write a long-ish summary to stdout
+            print("summary of task " + self.name)
+            for obj in data:
+                print(obj)
+
+When two ``ASyncExample`` tasks with different ``--name`` parameters are run simultaneously by a worker with 2 worker processes (``--workers 2``) and the point of printing the summary is reached at the same time, the terminal output is likely to be mixed between the two tasks. This way, the respective summaries become rather impractical.
+
+To ensure that the summaries are printed one after another, you can make use the ``worker_lock`` that is known to tasks within their ``run()`` method:
```
tonnamb: Small grammatical suggestion here for ‘you can make use **of** the ...’

riga:
Sorry for the long pause.

@lallea Thanks for the suggestion. While this would definitely make sense for workflows that anyway consist of a larger number of tasks, I'm afraid that smaller setups might become overly complicated.

But actually I think @Tarrasch is right here, and there is no issue with doing this in user land, i.e., by using a globally created lock/semaphore instead of using subprocess initializer args/kwargs. This wouldn't work on Windows though due to process forking, but I would be fine with that.

Feel free to close.

Tarrasch:
Thanks for coming back to us @riga! Furthemore, your own codebase might become more explicit (readable) if this locking functionality is in user land, rather than people needing to also read the luigi core to make sense of your user code. :)

Issue #2769, opened 2019-08-19
SQLServer incompatibility with task history: Operand type clash: datetime2 is incompatible with timestamp


shan224:
Error occurs when accessing the /history endpoint and when using pyodbc+mssql. 

Root cause: 
The TaskEvent class creates a column of type [TimeStamp.](https://github.com/spotify/luigi/blob/0f31f07ecc58acda864e14ae563484865ff3128a/luigi/db_task_history.py#L213) Which is incorrect for mssql because the [mssql implementation of TimeStamp](https://github.com/zzzeek/sqlalchemy/blob/7e0dfd2ad4141289a992adc65d2a6641f46dc82d/lib/sqlalchemy/dialects/mssql/base.py#L1039) is different from the SQL standard 🙄 .

Solution: 
This column needs to be defined as of type DateTime2 if connecting to mssql

shan224:
Update: 

I have tested that the column type is DATETIME2. 
However, I get another error when accessing the /history endpoint 

```
sqlalchemy.exc.ProgrammingError: (pyodbc.ProgrammingError) ('42000', "[42000] [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Column 'tasks.task_id' is invalid in the select list because it is not contained in either an aggregate function or the GROUP BY clause. (8120) (SQLExecDirectW); [42000] [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Statement(s) could not be prepared. (8180)")
[SQL: SELECT tasks.id AS tasks_id, tasks.task_id AS tasks_task_id, tasks.name AS tasks_name, tasks.host AS tasks_host
FROM tasks JOIN task_events ON tasks.id = task_events.task_id
WHERE task_events.ts >= ? GROUP BY tasks.id, task_events.event_name, task_events.ts ORDER BY task_events.ts DESC]
[parameters: (datetime.datetime(2019, 8, 19, 10, 53, 33, 459352),)]

```

I tried using an explicit select ([previous line](https://github.com/spotify/luigi/blob/0f31f07ecc58acda864e14ae563484865ff3128a/luigi/db_task_history.py#L163))  which cleared a few errors.

 ``` 
return session.query(TaskRecord.id, TaskRecord.name, TaskRecord.host, TaskEvent.event_name,TaskEvent.ts).\
                join(TaskEvent).\
                filter(TaskEvent.ts >= yesterday).\
[...]

PR #2770, opened 2019-08-20
override timestamp type for mssql connections

shan224:
## Description
Overriding the TIMESTAMP type because mssql uses a completely different type.
***This assumes that you will never want to use mssqls [timestamp](https://docs.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-2017)*** which is probably a safe assumption IMO

https://docs.sqlalchemy.org/en/13/core/custom_types.html#overriding-type-compilation

## Motivation and Context
https://github.com/spotify/luigi/issues/2769#issue-482514807

I have tested that the decorated function correctly creates a datetime2 column. 
I still need to rebuild the package and do an end to end test tomorrow. 

Update: 
See linked issue for update

### DIFF
--- a/luigi/db_task_history.py
+++ b/luigi/db_task_history.py
@@ -50,6 +50,8 @@ import sqlalchemy.ext.declarative
 import sqlalchemy.orm
 import sqlalchemy.orm.collections
 from sqlalchemy.engine import reflection
+from sqlalchemy.ext.compiler import compiles
+
 Base = sqlalchemy.ext.declarative.declarative_base()
 
 logger = logging.getLogger('luigi-interface')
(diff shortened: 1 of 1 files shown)


shan224:
Update: 

I have tested that the column type is correctly created as DATETIME2. 
However, I get another error when accessing the /history endpoint 

```
sqlalchemy.exc.ProgrammingError: (pyodbc.ProgrammingError) ('42000', "[42000] [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Column 'tasks.task_id' is invalid in the select list because it is not contained in either an aggregate function or the GROUP BY clause. (8120) (SQLExecDirectW); [42000] [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Statement(s) could not be prepared. (8180)")
[SQL: SELECT tasks.id AS tasks_id, tasks.task_id AS tasks_task_id, tasks.name AS tasks_name, tasks.host AS tasks_host
FROM tasks JOIN task_events ON tasks.id = task_events.task_id
WHERE task_events.ts >= ? GROUP BY tasks.id, task_events.event_name, task_events.ts ORDER BY task_events.ts DESC]
[parameters: (datetime.datetime(2019, 8, 19, 10, 53, 33, 459352),)]

```


Im not sure if we need to explicitly select all the columns. I've played around with little success. 
Also, it looks like the attributes form the tasks object being render in server.py and recent.html dont match up with the query. 

I don't have much time to go into this further

honnix:
Please feel free to reopen it.

PR #2767, opened 2019-08-17
Add ConfigParser as the base class of LuigiTomlParser

fullflu:
## Description

This PR adds `ConfigParser` as the base class of `LuigiTomlParser` to use TOML configuration with S3 Client. Two files are changed:
- I modify `luigi/configuration/toml_parser.py`.
- I add `examples/foo_s3.py` to check the validity of my modification.

## Motivation and Context

It is pointed out that S3 Client doesn't work with TOML configuration #2698 .
Fixing the issue would help a lot of users because TOML configuration and S3 Client are useful.
Now I propose to fix the issue by a simple modification.

## Have you tested this? If so, how?
 

I do not add any unit tests.
Instead of unit tests, I add an example file (`examples/foo_s3.py`) to check the validity of my modification.
I confirm that my modification worked with S3 Client.
Please note that I have not confirmed that my modification does not cause other unknown errors.
I would be grateful that someone could review this PR.

## Details of tests

The details of the results of running `examples/foo_s3.py` are as follows.

Current LuigiTmolParser raises AttributeError (as described in #2698 ).

```
************:luigi *******$ LUIGI_CONFIG_PARSER=toml LUIGI_CONFIG_PATH=examples/config.toml PYTHONPATH=. python3 -m luigi --module examples.foo_s3 examples.Foo --local-scheduler
INFO     2019-08-16 23:36:38 setup_logging:86 logging configured via config section
WARNING  2019-08-16 23:36:38 worker:658 Will not run examples.Foo() or any dependencies due to error in complete() method:
Traceback (most recent call last):
  File "/Users/*******/projects/luigi/luigi/worker.py", line 409, in check_complete
    is_complete = task.complete()
  File "/Users/*******/projects/luigi/luigi/task.py", line 564, in complete
    outputs = flatten(self.output())
  File "/Users/*******/projects/luigi/examples/foo_s3.py", line 12, in output
    return S3Target("s3://ryft-public-sample-data/hoge")
  File "/Users/*******/projects/luigi/luigi/contrib/s3.py", line 673, in __init__
    self.fs = client or S3Client()
  File "/Users/*******/projects/luigi/luigi/contrib/s3.py", line 106, in __init__
    options = self._get_s3_config()
  File "/Users/*******/projects/luigi/luigi/contrib/s3.py", line 507, in _get_s3_config
    defaults = dict(configuration.get_config().defaults())
AttributeError: 'LuigiTomlParser' object has no attribute 'defaults'

[...]

Issue #2698, opened 2019-04-29
S3 Client doesn't work with TOML configuration


liamlundy:
When using the TOML configuration, the s3 client breaks.

To reproduce, create luigi.toml configuration file and set the `LUIGI_CONFIG_PARSER` environment variables to `toml` and running the following code:

```
from luigi.contrib.s3 import S3Target
S3Target('test_path')
```

Error:
```
ae_luigi_worker/tasks/pipeline_mta_model_task.py:43: in output
    return self.upload_task.output()
ae_luigi_worker/tasks/pipeline_mta_model_task.py:160: in output
    for output_file in self.run_model_task.output_files]
ae_luigi_worker/tasks/pipeline_mta_model_task.py:160: in <listcomp>
    for output_file in self.run_model_task.output_files]
../../../.local/share/virtualenvs/ae/lib/python3.7/site-packages/luigi/contrib/s3.py:673: in __init__
    self.fs = client or S3Client()
../../../.local/share/virtualenvs/ae/lib/python3.7/site-packages/luigi/contrib/s3.py:106: in __init__
    options = self._get_s3_config()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
key = None

    @staticmethod
    def _get_s3_config(key=None):
>       defaults = dict(configuration.get_config().defaults())
E       AttributeError: 'LuigiTomlParser' object has no attribute 'defaults'
```

This error is occurring at: https://github.com/spotify/luigi/blob/c5eca1c3c3ee2a7eb612486192a0da146710a1e9/luigi/contrib/s3.py#L507 and appears to be due to the fact that the `LuigiTomlParser` doesn't have a defaults method like the standard configuration object. This could be fixed in the s3 client code, but it seems like the interface of the two configuration objects should be standard.

tfiers:
PR for a very similar issue: #2702

PR #2182, opened 2017-07-13
Update S3's _get_s3_config() to not include DEFAULT section

dlstadther:
## Description

Update _get_s3_config to only return s3 section with interpolation, excluding defaults. This will allow the use of ConfigParser's DEFAULT section without S3/boto throwing errors.

## Motivation and Context

Resolves #2180 

## Have you tested this? If so, how?
 

Tests pass locally and remotely. Example code in linked issue now execute without error.

PR #2772, opened 2019-08-21
S3 tomlparser temporary fix

fullflu:
## Description

This PR adds a simple conditional branch to use TOML configuration with S3 Client. Two files are changed:

I modify luigi/contrib/s3.py.
I add examples/foo_s3.py to check the validity of my modification.

## Motivation and Context

It is pointed out that S3 Client doesn't work with TOML configuration #2698 .
Fixing the issue would help a lot of users because TOML configuration and S3 Client are useful.
Now I propose to fix the issue by a simple modification.

I do not add any unit tests because there is not test of `S3Client._get_s3_config` method.
Instead of unit tests, I add an example file (examples/foo_s3.py) to check the validity of my modification.
I confirm that my modification worked with S3 Client.
Please note that I have not confirmed that my modification does not cause other unknown errors.
I would be grateful that someone could review this PR.

Instead of #2767 , I propose to fix S3Client.
Although this PR might not be the best solution as @orsinium said, it would be a better solution because we can avoid the parsing error temporarily.

## Details of tests

```
********:luigi ************$ LUIGI_CONFIG_PARSER=toml LUIGI_CONFIG_PATH=examples/config.toml PYTHONPATH=. python3 -m luigi --module examples.foo_s3 examples.Foo --local-scheduler
INFO     2019-08-21 15:56:19 setup_logging:86 logging configured via config section
INFO     2019-08-21 15:56:23 worker:595 Informed scheduler that task   examples.Foo__99914b932b   has status   PENDING
INFO     2019-08-21 15:56:23 interface:172 Done scheduling tasks
INFO     2019-08-21 15:56:23 worker:1182 Running Worker with 1 processes
INFO     2019-08-21 15:56:23 worker:165 [pid 61882] Worker Worker(salt=792552536, workers=1, host=********.local, username=************, pid=61882) running   examples.Foo()
Running Foo that outputs S3Target
INFO     2019-08-21 15:56:23 worker:217 [pid 61882] Worker Worker(salt=792552536, workers=1, host=********.local, username=************, pid=61882) done      examples.Foo()
INFO     2019-08-21 15:56:23 worker:595 Informed scheduler that task   examples.Foo__99914b932b   has status   DONE
INFO     2019-08-21 15:56:23 worker:493 Worker Worker(salt=792552536, workers=1, host=********.local, username=************, pid=61882) was stopped. Shutting down Keep-Alive thread
INFO     2019-08-21 15:56:23 interface:175 

Scheduled 1 tasks of which:
* 1 ran successfully:
    - 1 examples.Foo()

This progress looks :) because there were no failed tasks or missing dependencies

```

### DIFF
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -504,21 +504,23 @@ class S3Client(FileSystem):
 
     @staticmethod
     def _get_s3_config(key=None):
-        defaults = dict(configuration.get_config().defaults())
-        try:
-            config = dict(configuration.get_config().items('s3'))
-        except (NoSectionError, KeyError):
-            return {}
-        # So what ports etc can be read without us having to specify all dtypes
-        for k, v in six.iteritems(config):
+        if os.environ.get("LUIGI_CONFIG_PARSER") != "toml":
+            defaults = dict(configuration.get_config().defaults())
             try:
-                config[k] = int(v)
-            except ValueError:
-                pass
-        if key:
-            return config.get(key)
-        section_only = {k: v for k, v in config.items() if k not in defaults or v != defaults[k]}
-
+                config = dict(configuration.get_config().items('s3'))
+            except (NoSectionError, KeyError):
+                return {}
+            # So what ports etc can be read without us having to specify all dtypes
+            for k, v in six.iteritems(config):
+                try:
+                    config[k] = int(v)
+                except ValueError:
+                    pass
+            if key:
+                return config.get(key)
+            section_only = {k: v for k, v in config.items() if k not in defaults or v != defaults[k]}
+        else:
+            return {}
         return section_only
 
     @staticmethod
--- a/examples/foo_s3.py
+++ b/examples/foo_s3.py
@@ -1,0 +1,12 @@
+import luigi
+from luigi.contrib.s3 import S3Target
+
+
+class Foo(luigi.Task):
+    task_namespace = 'examples'
+
+    def run(self):
+        print("Running Foo that outputs S3Target")
+
+    def output(self):
+        return S3Target("s3://ryft-public-sample-data/hoge")


fullflu:
@dlstadther 
Could you tell me an example where `_get_s3_config` method does not return an empty dict?
I'm sorry that I cannot follow the context of this method completely.

Then I'll try to support a default section in TOML parser.

dlstadther:
Let's suppose you have the following in your INI cfg:
```ini
[DEFAULT]
home_path: /opt/luigi
env: production

[s3]
aws_access_key_id: <your-key>
aws_secret_access_key: <your-secret>
calling_format: boto.s3.connection.OrdinaryCallingFormat
# ^ this was needed back before luigi upgraded to boto3, but the idea that additional boto3 kwargs could be supplied still stands
```

`s3.S3Client()`'s `.__init__()` will set `options = self._get_s3_config()` and later supply any kwarg which weren't used for authentication to [`boto3.resource(..., **options)`](https://github.com/spotify/luigi/blob/master/luigi/contrib/s3.py#L159-L163)

So prior to the "removing default section vars from s3 section", boto3.resource would blow up when it received an unknown kwarg `home_path` (going back to the example ini above).

[Here was my original PR which added this functionality](https://github.com/spotify/luigi/pull/2182).

---

I hope that provides a more clear explanation for why that code exists.

luigi/contrib/s3.py
```diff
@@ -504,21 +504,23 @@ def list(self, path, start_time=None, end_time=None, return_key=False):  # backw
...
     @staticmethod
     def _get_s3_config(key=None):
-        defaults = dict(configuration.get_config().defaults())
-        try:
-            config = dict(configuration.get_config().items('s3'))
-        except (NoSectionError, KeyError):
-            return {}
-        # So what ports etc can be read without us having to specify all dtypes
-        for k, v in six.iteritems(config):
+        if os.environ.get("LUIGI_CONFIG_PARSER") != "toml":
```
orsinium: If I remain correctly, the config type can be specified not only from environment variables, but also through passing from the code. More stable way is check the instance of the config: `if isinstance(configuration.get_config(), ...)`

PR #2779, opened 2019-09-09
PySparkTask: handle special characters in name (#2778)

### LINKED ISSUE #2778: PySparkTask fails if name contains invalid characters
PySparkTask derives run_path and run_pickle from name property, but it do not escape special characters:
1. '/' in name causes FileNotFoundError raised by mkdtemp
2. ',' is separator of files in spark-submit '--files' arg, and run_pickle is passed to 'files' arg in cluster mode.
3. ':' causes java.net.URISyntaxException

To reproduce: create PySparkTask subclass with name property containing above special characters and try to run it.

mrk-its:
Fixes #2778 

## Description
It cleans _name_ property before it is used to derive _run_path_ and _run_pickle_ properties.

## Motivation and Context
If name property contains some special characters PySparkTask fails: #2778 

Unit test is included

### DIFF
--- a/luigi/contrib/spark.py
+++ b/luigi/contrib/spark.py
@@ -299,9 +299,9 @@ class PySparkTask(SparkSubmitTask):
         return [self.app, pickle_loc] + self.app_options()
 
     def run(self):
-        name = re.sub(r'[^\w]', '_', self.name)
-        self.run_path = tempfile.mkdtemp(prefix=name)
-        self.run_pickle = os.path.join(self.run_path, '.'.join([name, 'pickle']))
+        path_name_fragment = re.sub(r'[^\w]', '_', self.name)
+        self.run_path = tempfile.mkdtemp(prefix=path_name_fragment)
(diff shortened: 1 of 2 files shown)


luigi/contrib/spark.py
```diff
@@ -298,8 +299,9 @@ def app_command(self):
         return [self.app, pickle_loc] + self.app_options()
 
     def run(self):
-        self.run_path = tempfile.mkdtemp(prefix=self.name)
-        self.run_pickle = os.path.join(self.run_path, '.'.join([self.name.replace(' ', '_'), 'pickle']))
+        name = re.sub(r'[^\w]', '_', self.name)
```
GoodDok: I'd rename variable `name` to something like `run_path_prefix` (maybe some better option) due to two reasons: 
- it is actually a path, not a name anymore
- we must be having a hard time dealing with `name` and `self.name` in the same place

mrk-its: Good point! Renamed.

mrk-its:
@Tarrasch @dlstadther could you take a look? Thanks!

Issue #2670, opened 2019-03-12
Luigi not working behind reversproxy with baseurl


zwotzie:
I would like to run Luigi behind haproxy with baseurl /luigi. 

client.cfg contains:
default-scheduler-url: http**s**://domain.tld/luigi

Normaly i would expect, that this would work, because this configuration is semantically a url not a hostname...

Luigi is running well until _urljoin(base, url) in rpc.py is called.

The "funny" thing is, that the parameter variable basecontains the right default-scheduler-url and the url contains at first time '/api/ping'. The result ends in the wrong url: https://domain.tld/api/ping but i would expect https://domain.tld/luigi/api/ping. The _urljoin function does not join the urls parts but rather replaces parts of it.

At least I do not know why the scheme is twice replaced back and forth in the same function.

greysonevins:
@zwotzie did you find a solution to this one? I am not able to solve this one either. Want to host a scheduler on a sub directory of a url but no luck either.

zwotzie:
yes i find a solution and build my own luigi.: https://github.com/zwotzie/luigi 
There are two differences between here and there. First i use snakebite-py3 and therefor i have to remove the code to stop and for the reverseproxy usage (haproxy) see the changes in  https://github.com/zwotzie/luigi/commit/e0175d682f47c19d55ffbdff11d36a0f2c49aaf3

Hope that helps

tashrifbillah:
To add a bit more, I tried the following nginx.conf with default `luigi.cfg`:

```vim
server {
    location / {
        proxy_pass http://localhost:8082/static/visualiser/;
    }
}
```

Is that what @greysonevins failed to solve?

zwotzie:
I had to explain the setup better, sorry for that.

The client.cfg from above is for the luigi workers and not for the scheduler.

In my case, the setup looks like:

`
haproxy (ssl offloading) -> nginx (revers proxy) -> luigi-central-scheduler
                                  -> nginx ...
`

The haproxy redirects everythin from /luigi to nginx. Nginx rewrites all urls so that luigi (with standard configuration) can understand it.

tashrifbillah:
Thanks @zwotzie . This is how I solved this: issue https://github.com/spotify/luigi/issues/2972

tashrifbillah:
Hello again @zwotzie , I was able to proxy the Luigi server against `https://yourdomain.xy/` using the following:
```vim
location / {
  proxy_pass http://localhost:8082;
}
```
[...]

PR #2785, opened 2019-09-14
don't use absolute path in redirect for visualizer

timkpaine:
In reverse proxy setups, you sometimes want to setup sites behind a sub path e.g. my.site.com/luigi/
This PR makes sure the initial redirect is relative to the base_url, not absolute.


honnix:
Thanks for the PR. Have you tested this will work for both behind reverse proxy and direct access?

timkpaine:
will get back to this this week, got distracted

timkpaine:
@honnix works fine for both, @Tarrasch it might be possible to modify this test to test explicitly
https://github.com/spotify/luigi/blob/865cc4e97540b5adc19898cbb5f31f39bff791de/test/server_test.py#L101

honnix:
Something like this would work https://github.com/tornadoweb/tornado/blob/master/tornado/test/httpclient_test.py#L293

timkpaine:
some further activity

timkpaine:
finally got around to this and added a test that specifically checks there is no leading slash:

```python
        response = self.fetch("/", follow_redirects=False)
        self.assertEqual(response.code, 302)
        self.assertEqual(response.headers['Location'], 'static/visualiser/index.html')  # assert that doesnt beging with leading slash !
```

timkpaine:
thanks!

tashrifbillah:
This work missed my eyes. I have reported the need in many places before including issue https://github.com/spotify/luigi/issues/2981. A similar issue is https://github.com/spotify/luigi/issues/2670. Thanks for the modification. I shall test it and let you know.

tashrifbillah:
Unfortunately, the PR didn't work for me. See below a comparison between `plotly` server and `luigi` server:

# Launching luigid

> luigid --port 8083

# Confirming luigi is functional w/o proxy

> curl -L http://localhost:8083

# Configuring nginx proxy

```vim
    server {

        root /var/www/html/;


        location /luigi/ {
            proxy_pass http://localhost:8083;
        }

        location /dash/ {
            proxy_pass http://localhost:8084;
        }


    }

```


# Issue
```
# works gracefully
curl -L http://myIP/dash/

# fails :(
curl -L http://myIP/luigi/
```

Failure logs:

<details><summary>console log</summary>

```vim
<html><title>404: Not Found</title><body>404: Not Found</body></html>
```

</details>

<details><summary>nginx/access.log</summary>

```vim
170.223.221.217 - [13/Nov/2020:18:54:43 -0500] "GET /luigi/ HTTP/1.1" 404 69 "-" "curl/7.29.0" "-"
```

</details>

<details><summary>nginx/error.log</summary>

```vim
2020/11/13 19:05:09 [notice] 106027#0: signal process started
```

</details>

@timkpaine @honnix

timkpaine:
The logs you posted are nginx telling you that the route is missing. Nothing to do with Luigi, you haven't even hit the Luigi process.

timkpaine:
also you installed from master right?

timkpaine:
don't you need a trailing slash on `proxy_pass http://localhost:8083;`? otherwise you're redirecting to localhost:8083/luigi/, where whereas this is a valid dash route, its not a valid luigi route

tashrifbillah:
> also you installed from master right?

Upon release 3.0.2, I applied your change, which is just one `/` removal.

Can you see my edited comment above for `Tornado log`?

timkpaine:
https://serverfault.com/questions/562756/how-to-remove-the-path-with-an-nginx-proxy-pass

tashrifbillah:
Thanks Tim, it looks like the trailing slash is required. So is it the trailing slash that makes sure the redirection is towards `http://localhost:8083`?

That said, I am going to do a few more tests tonight with other tabs such as `/history`.

timkpaine:
yeah all of this has nothing to do with luigi, its all nginx configuration. The trailing slash ensures that myip/luigi/a/path -> localhost:8083/a/path. You just happened to get unlucky with the url as dash directs its traffic through /dash anyway.

tashrifbillah:
Hello again @timkpaine , unfortunately, we are not set yet. Have you tried to run a job through your modification? See the issue I figured:

### nginx.conf

```vim
    server {
        root /var/www/html/;
        location /luigi/ {
            proxy_pass http://localhost:8083;
        }
    }
```

### luigi.cfg:

```vim
[core]
default-scheduler-url = http://myIP/luigi/

[scheduler]
record_task_history = True
state_path = ${HOME}/luigi-state.pickle

[task_history]
db_connection = sqlite:///${HOME}/luigi-task-hist.db
```


### Error

Then if you try my [simple task](https://gist.github.com/tashrifbillah/2bca276c7a6e2d5c85e9206d6522b37e), it should run into the following error:

```vim
WARNING: Failed connecting to remote scheduler 'http://myIP/luigi'
  File "/tmp/new_luigi/miniconda3/lib/python3.7/site-packages/luigi/rpc.py", line 163, in _fetch
  File "/tmp/new_luigi/miniconda3/lib/python3.7/site-packages/luigi/rpc.py", line 117, in fetch
    resp.raise_for_status()
  File "/tmp/new_luigi/miniconda3/lib/python3.7/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://myIP/api/add_task
```

As apparent from `http://myIP/api/add_task`, I think my nginx URL rewriting isn't right. `http://myIP/api/add_task` isn't even hitting Luigi. It should be `http://localhost:8083/api/add_task`.

Any suggestion about the URL rewriting for `@proxy_pass` in the above `nginx.conf`?

timkpaine:
I don't use nginx for this but I think you'll have a lot easier time if you set it up as a sub domain rather than a path, e.g. luigi.myip rather than myip/luigi. When you reverse proxy under paths, often times you need a lot more library modifications.

tashrifbillah:
Did you test the job scheduling part other than visualizer?

timkpaine:
I run jobs from the same machine so my scheduler url is local host. But it's reasonable to assume that the client would still use absolute paths. This would still work fine with a sub domain instead of a path

tashrifbillah:
I shall try what you suggested. More info for you--the `/history` tab in the seemingly functional visualizer is missing style sheets. Did you notice that?

tashrifbillah:
Hi @timkpaine , could you at least check if style sheets load for you in the `/history` tab?

tashrifbillah:
I solved the style sheet loading issue by another location block:

```vim    
    server {
        root /var/www/html/;
        location /luigi/ {
            proxy_pass http://localhost:8083/;
        }

        location /static/ {
            proxy_pass http://localhost:8083/static/;
        }

    }
```

It is a hack that does not make me so happy but it works.

Issue #2788, opened 2019-09-20
Scheduler endpoint for checking status/health


Valleo4:
When using Docker Swarm for example, a healthcheck command can be used to check whether services in containers are up. Right now the best way to check health of the luigi scheduler is by GET request to root "/" which redirects and returns the landing page, which is not fitting when just checking the service.

Perhaps the service could accept HEAD requests to the root "/" for this purpose?

Valleo4:
closed in #2789

PR #2789, opened 2019-09-20
Add HEAD endpoint to scheduler server for status/health checks

Valleo4:
Add status/health endpoint for the luigi scheduler by letting it accept HEAD requests to "/". Issue  described in #2788.

### DIFF
--- a/luigi/server.py
+++ b/luigi/server.py
@@ -273,6 +273,10 @@ class RootPathHandler(BaseTaskHistoryHandler):
     def get(self):
         self.redirect("/static/visualiser/index.html")
 
+    def head(self):
+        self.set_status(204)
+        self.finish()
+
 
 class MetricsHandler(tornado.web.RequestHandler):
     def initialize(self, scheduler):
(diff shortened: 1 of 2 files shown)


honnix:
Thanks for the PR. Could you please add unit test?

Valleo4:
Done, thanks for the quick reply :)

luigi/server.py
```diff
@@ -273,6 +273,10 @@ class RootPathHandler(BaseTaskHistoryHandler):
     def get(self):
         self.redirect("/static/visualiser/index.html")
 
+    def head(self):
```
Tarrasch: Is the name "head" a convention for health checking? 

Regardless it caught me by surprise, maybe add some docs to this function on why it exist? It should also say what 204.

Valleo4: "head" is for the HEAD http request, and it's not a convention i would say, but it's what fits the purpose in the protocol. 
I can add some docs if you want, but i don't know where to put it, there's only autogenerated docs for the web API.

Tarrasch: What about `""" function docs """`?

Valleo4:
@honnix @dlstadther can you take a look at this?

PR #2798, opened 2019-10-02
Added optional argument 'aws_session_token' to S3Client

Bonsanto:
## Description
Added optional named argument 'aws_session_token' to luigi.contrib.s3.S3Client class. If this argument is provided then during the `s3` method usage it will be used.

## Motivation and Context
- Some organizations provide you with `aws_access_key_id`, `aws_secret_access_key` and some times they also provide you with the `aws_session_token` which is mandatory.

I ran many of my jobs locally and it works perfectly.


Bonsanto:
@dlstadther I'm not sure how to test this nevertheless I tried my best adding some tests.

PR #2799, opened 2019-10-09
More stronger random tmp filenames.

peykar:
It's better to have more stronger random tmp filenames, so duplication chance will be decreased.

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -28,9 +28,9 @@ Be aware that normal ftp does not provide secure communication.
 import datetime
 import ftplib
 import os
-import random
 import tempfile
 import io
+from uuid import uuid4
 
 import luigi
 import luigi.file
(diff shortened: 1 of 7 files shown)


honnix:
Thanks for the PR. Have you encountered duplication in practice?

peykar:
> Thanks for the PR. Have you encountered duplication in practice?

You're welcome. Nope, not really.

peykar:
> I think there's a risk we're causing "too long file name" or broken cleanup-regexps for our users. I however don't think our users have any issue with duplicate names currently. Hmm...

Is there any test to check the aforementioned risk?

honnix:
I would prefer we don't make this change if nothing currently known as broken, due to the risk @Tarrasch mentioned. And it is probably not worth the effort trying to mitigate the risk.

peykar:
> I would prefer we don't make this change if nothing currently known as broken, due to the risk @Tarrasch mentioned. And it is probably not worth the effort trying to mitigate the risk.

ok

PR #2801, opened 2019-10-09
EnumListParameter

evictor:
## Description

A parameter type complementary to EnumParameter that allows for an arbitrarily sized list (0 to *n*) of enum values in a str-delimited format.

## Motivation and Context

Convenient addition to the Luigi parameter type inventory.

I have included unit tests that match the style of existing `EnumParameter` tests.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -892,7 +892,9 @@ class EnumParameter(Parameter):
 
 class EnumListParameter(Parameter):
     """
-    A parameter whose value is a comma-separated list of :class:`~enum.Enum`.
+    A parameter whose value is a comma-separated list of :class:`~enum.Enum`. Values should come from the same enum.
+
+    Values are taken to be a list, i.e. order is preserved, duplicates may occur, and empty list is possible.
 
     In the task definition, use
 
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -892,7 +892,7 @@ class EnumParameter(Parameter):
 
 class EnumListParameter(Parameter):
     """
-    A parameter whose value is a str-separated (default ",") list of :class:`~enum.Enum`.
+    A parameter whose value is a comma-separated list of :class:`~enum.Enum`.
 
     In the task definition, use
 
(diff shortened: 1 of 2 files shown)


luigi/parameter.py
```diff
@@ -893,6 +893,51 @@ def serialize(self, e):
         return e.name
 
 
+class EnumListParameter(Parameter):
+    """
+    A parameter whose value is a str-separated (default ",") list of :class:`~enum.Enum`.
+
```
Tarrasch: Maybe add that the values have to come from the same enum.

evictor:
OK, I removed the delim param and renamed the underlying variable to `sep` to follow Python's underlying var name for the similar concept.

luigi/parameter.py
Tarrasch: Did you consider this?

luigi/parameter.py
```diff
@@ -893,6 +893,52 @@ def serialize(self, e):
         return e.name
 
 
+class EnumListParameter(Parameter):
+    """
+    A parameter whose value is a comma-separated list of :class:`~enum.Enum`.
+
+    In the task definition, use
```
Tarrasch: Also add one line explaining that this is a list in the sense that order matters and there can be duplicates and it can be empty.

test/parameter_test.py
```diff
@@ -437,6 +451,13 @@ class Foo(luigi.Task):
         p = luigi.parameter.EnumParameter(enum=MyEnum)
         self.assertEqual(hash(Foo(args=MyEnum.A).args), hash(p.parse('A')))
 
+    def test_enum_list(self):
```
Tarrasch: Can you also add a test case where you pass a default value to the new Parameter?

luigi/parameter.py
evictor: Will note that

luigi/parameter.py
evictor: OK

Tarrasch:
Travis for 2.7 was red, but it's gone now. :D

PR #2802, opened 2019-10-10
Update task.flatten to check if struct is instance of Task or Target

maurosilber:
## Description
Change `task.flatten` function to stop recursion when a `Task` or `Target` instance is found.

## Motivation and Context
It allows for `Task` or `Target` to implement `__getitem__` or `__iter__`. I'd like to use this to present a simpler interface for subclasses of `Target`s or `WrappedTask`s. The latter I sometimes use as `subtasks` in downstream `Task`s.

This function is used to create a flat list of `Task`s or `Target`s (depending where it's used) from a structured input of `Task`s or `Target`s.
Currently, it has a section when it tries to convert the input to an iterator, that would yield either `Task`s, `Target`s or another structured input. If a subclass of `Task` or `Target` implements `__getitem__` or `__iter__`, it gets iterated and its items (which wouldn't be `Task`s or `Target`s, in my case) flattened.

I did a simple test by checking that `Task._requires()` outputted the same list of `Tasks`,
and checked for uses of flatten in the codebase, which is always used to get a flat list of `Task`s or `Target`s.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -40,6 +40,7 @@ from luigi import six
 from luigi import parameter
 from luigi.task_register import Register
 from luigi.parameter import ParameterVisibility
+from luigi.target import Target
 
 Parameter = parameter.Parameter
 logger = logging.getLogger('luigi-interface')
@@ -875,7 +876,7 @@ def flatten(struct):
         for _, result in six.iteritems(struct):
             flat += flatten(result)
         return flat
-    if isinstance(struct, six.string_types):
+    if isinstance(struct, (six.string_types, Task, Target)):
         return [struct]
 
     try:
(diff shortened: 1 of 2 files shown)


Bonsanto:
It'd be nice to add a unit test to specifically check that it works with those type of objects

maurosilber:
I agree that it's not a good idea to add complexity. But I think this doesn't. Instead, it prevents bugs. Otherwise, we should add documentation that one isn't supposed to implement neither `__iter__` nor `__getitem__` in `Task` or `Target`. I don't know if there's a way to prevent subclasses from implementing a method.

Tarrasch:
I'm happy to add the docs for now.

Issue #2749, opened 2019-07-23
TypeError because task.retry is None


BenBlaisdell:
This issue has popped up twice for my org in 2 different environments, but I'm not sure how it's getting in this state.

It appears that `task.retry` is `None` in the state pickle.

```
2019-07-23 14:12:23,723 luigi.scheduler INFO     Removing task 'JobA__2019_07_22T11300_347472d5b5'
2019-07-23 14:12:23,723 luigi.scheduler INFO     Removing task 'JobB_530b2afff0'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobC_79d10cad66'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobA__2019_07_22T18300_4978b6fc52'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobB_9e8056c664'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobC_485505f19d'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobA__2019_07_22T19000_3ac92efe3c'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobC_0cd4a17227'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobB_6e2688e055'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobA__2019_07_22T20300_c34dde6ea0'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobC_39bb92e304'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobB_0a559bef01'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobA__2019_07_22T21300_3bffc126d4'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobB_f3519a299f'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobC_715cfc8f19'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobA__2019_07_22T22000_b6b1cce74f'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobC_e9c261dcb9'
2019-07-23 14:12:23,724 luigi.scheduler INFO     Removing task 'JobB_3636409089'
[...]

PR #2803, opened 2019-10-10
Fixing comparison in update_status when task.retry is None #2749

kefayati:
## Description

This PR addresses a carry-over code from Python 2 that fails in Python 3; namely, `None` cannot be compared to `Float` and hence the test results in an exception that stops luigi daemon and stalls tasks depending on it.

I also addressed another similar `None` case in `prometheus_metrics` that stops luigi daemon when tasks are already done and hence `task.time_running` is `None`.
## Motivation and Context

Quick test for `task.retry` and `task.time_running` -- which are `None` by default.
 

### DIFF
--- a/luigi/contrib/prometheus_metric.py
+++ b/luigi/contrib/prometheus_metric.py
@@ -46,12 +46,16 @@ class PrometheusMetricsCollector(MetricsCollector):
         self.task_execution_time.labels(family=task.family)
 
     def handle_task_failed(self, task):
-        self.task_failed_counter.labels(family=task.family).inc()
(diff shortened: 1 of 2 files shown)


kefayati:
@victoriaalee Please review the contrib part. Thanks!

kefayati:
@robmoore I think the current one is more clear on possible values for `task.retry`, also would not be prone to potentially unintended consequences when `task.retry` is `0` or anything else that evaluates as `False`.

victoriaalee:
@kefayati lgtm

PR #2807, opened 2019-10-16
Properly quote program arguments in ExternalProgramTask logs

arteymix:
This is a big convenience when you need to rerun a failing command in a shell.

### DIFF
--- a/luigi/contrib/external_program.py
+++ b/luigi/contrib/external_program.py
@@ -42,6 +42,7 @@ from time import sleep
 
 import luigi
 from luigi.parameter import ParameterVisibility
+from luigi import six
 
 logger = logging.getLogger('luigi-interface')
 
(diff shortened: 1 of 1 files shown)


luigi/contrib/external_program.py
```diff
@@ -121,7 +122,7 @@ def _clean_output_file(self, file_object):
     def run(self):
         args = list(map(str, self.program_args()))
 
-        logger.info('Running command: %s', ' '.join(args))
+        logger.info('Running command: %s', ' '.join(six.shlex_quote(arg) for arg in args))
```
honnix: this needs to be `six.moves.shlex_quote` i think.

arteymix: It seems that you were right.

honnix:
@GoodDok Could you please take a look as well?

Tarrasch:
Sorry. This PR looks interesting but I think it might need to be rebased. Is `six` still required?

arteymix:
`shlex.quote` was introduced in Python 3.3 and it was not backported in 2.7, so yes, it is still required.

Tarrasch:
We dropped Python 2 support I recall..

honnix:
The Python 2 deprecation PR has not been merged. I will get it done asap. Sorry for taking so long.

PR #2808, opened 2019-10-17
Raise correct Batch exception when job fails before start

relvacode:
## Description

  - Use the container reason from the Batch job attempt if the statusReason is `Task failed to start`
  - Use the most recent attempt when evaluating errors

## Motivation and Context

If a Batch job fails to start (due to an invalid task definition for example) no log group is created for that task so the actual Exception raised by the BatchClient is 

```
ResourceNotFoundException
An error occurred (ResourceNotFoundException) when calling the GetLogEvents operation: The specified log stream does not exist.
```

This change fixes it so that if the task has failed to start the reason from the container is used instead. This makes the actual Exception raised by the Batch client:

```
BatchJobException
Job [REDACTED] failed: CannotPullContainerError: Error response from daemon: manifest for [REDACTED] not found
````

Additionally the Batch client was only looking at the first container attempt. This change uses the latest container attempt in any status reason or log gathering.

I have included an additional unit test for this case and have also run both the original version and my fixed version with a broken Batch task definition (results above).

(Note that the existing `test_submit_job_non_existant_queue` test case always fails for me in both the original and new version)

### DIFF
--- a/test/contrib/batch_test.py
+++ b/test/contrib/batch_test.py
@@ -62,6 +62,7 @@ class MockBotoBatchClient(object):
                     'status': 'SUCCEEDED',
                     'attempts': [
                         {
+                            'statusReason': 'Essential container in task exited',
                             'container': {
                                 'logStreamName': 'test_job_abcd_log_stream'
                             }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/contrib/batch_test.py [hunk 1]
```diff
@@ -162,6 +163,46 @@ def test_wait_on_job_failed(self):
             self.bc.wait_on_job(job_id)
             self.assertTrue('log line 1' in context.exception)
 
+    def test_wait_on_job_failed_before_start(self):
```
honnix: According to travis, this test case doesn't cover the change. Could you please take a look? Thanks.

PR #2809, opened 2019-10-17
fix parameter exposure in error emails

steveims:
A task error results in an email sent to associated addresses.  The email currently includes non-public parameters in the subject and body.  That was a problem for us because our HIDDEN and PRIVATE parameters (e.g. passwords) were exposed.

We made two changes to fix the problem:
1. serialize only PUBLIC parameters to the body of the error email
2. serialized only significant and PUBLIC parameters in the task representation, which is used to generate the subject for error emails (e.g. https://github.com/spotify/luigi/blob/865cc4e97540b5adc19898cbb5f31f39bff791de/luigi/worker.py#L686)

Note: #2 makes sense to us because we never want HIDDEN or PRIVATE parameters exposed (e.g. logs, etc).

We've been running this change in production for two weeks: error emails are now as desired and not seeing any regression.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -544,7 +544,7 @@ class Task(object):
         repr_parts = []
         param_objs = dict(params)
         for param_name, param_value in param_values:
-            if param_objs[param_name].significant and param_objs[param_name].visibility == ParameterVisibility.PUBLIC:
+            if param_objs[param_name].significant:
                 repr_parts.append('%s=%s' % (param_name, param_objs[param_name].serialize(param_value)))
 
         task_str = '{}({})'.format(self.get_task_family(), ', '.join(repr_parts))
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -544,7 +544,7 @@ class Task(object):
         repr_parts = []
         param_objs = dict(params)
         for param_name, param_value in param_values:
-            if param_objs[param_name].significant:
+            if param_objs[param_name].significant and param_objs[param_name].visibility == ParameterVisibility.PUBLIC:
                 repr_parts.append('%s=%s' % (param_name, param_objs[param_name].serialize(param_value)))
 
         task_str = '{}({})'.format(self.get_task_family(), ', '.join(repr_parts))
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -405,7 +405,7 @@ def format_task_error(headline, task, command, formatted_exception=None):
         </html>
         ''')
 
-        str_params = task.to_str_params()
+        str_params = task.to_str_params(only_public=True)
         params = '\n'.join('<tr><th>{}</th><td>{}</td></tr>'.format(*items) for items in str_params.items())
         body = msg_template.format(headline=headline, name=task.task_family, param_rows=params,
                                    command=command, traceback=formatted_exception)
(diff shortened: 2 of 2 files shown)


luigi/task.py
```diff
@@ -544,7 +544,7 @@ def __repr__(self):
         repr_parts = []
         param_objs = dict(params)
         for param_name, param_value in param_values:
-            if param_objs[param_name].significant:
+            if param_objs[param_name].significant and param_objs[param_name].visibility == ParameterVisibility.PUBLIC:
```
dlstadther: I think this may bring up a valid question. Should `significant` parameters be allowed to be anything other than `public`?
I'm inclined to think if a parameter is defined as significant, it should be included in its task's `__repr__`.

@nryanov You added the parameter visibility feature. What are your thoughts on my comment above?

steveims: Good question.  I'm not a heavy user of Luigi, but so far I'd be fine with a boolean on parameters to control whether they are exposed (logs, email, etc).

dlstadther: Could you validate if adding `significant=False` to the Parameter instantiation correctly removes it from the error email's subject?

The [docstring for this parameter configuration](https://github.com/spotify/luigi/blob/master/luigi/parameter.py#L145-L149) continues to lead me to believe that non-public parameters should not be significant.

nryanov: Sorry for delayed answer

@dlstadther this is a good question. I think that  significant parameters may be public, hidden and private in terms of significance. Maybe we should reconsider current description of 'significant' and declare it as flag important only for task_id generation.  If `significant` will be important only for task_id generation, then implementation of `Task.__repr__`  will be easier as it will be able to include only `public` parameters.

But currently, parameters significance could control its exposure in some cases

dlstadther: Sorry for my absence here on this item.

I think for the matter of changing the definition of "significant parameters" and representation of a Task's `__repr`, we should ensure to get feedback from @honnix and @Tarrasch .

Tarrasch: I think what @nryanov said makes a lot of sense. However I think we've misused `__repr__` here. It should probably show *all* parameters. The issue however here is that we've totally misused it when we put it in the email header.

Maybe the email header could use `str()` instead of `repr()`. And we would implement

    def __str__() = return self._get_task_as_string(show_only_public_parameters=True)

 and

    def __repr__() = return self._get_task_as_string()

maybe.

Tarrasch: Thinking about it... using `str()` at that call site would also not be explicit, there we probably want to say we want it to be "give me string representation with PUBLIC params only' somehow.

steveims:
Thank you for the comments.

Giving the caller explicit control over parameter exposure makes sense.  Not sure how many places that might affect?  (I'd prefer to know that secrets (passwords, etc) will not appear anywhere (logs, email, etc).)

Apologies for not providing tests.  I'm not familiar with luigi; I was just asked by our team to solve this problem.  I'd be happy to add tests if someone could suggest them.

Tarrasch:
@steveims can you refactor this pull request as I previously asked? (remove changes to task.py and add a test for the other change)

We have a contributing file about testing: https://github.com/spotify/luigi/blob/master/CONTRIBUTING.rst

steveims:
@Tarrasch I removed changes to task.py.

Apologies, but I will not be able to submit any tests. I'm not familiar with Luigi and not able to learn right now.

Please advise if you'd rather that I close this PR and open an issue instead.

Tarrasch:
Feel free to take your time on writing the test. But we do expect that on all changes to core parts of this repository.

PR #2813, opened 2019-10-28
Add pod_creation_wait_interal

plizmol:
## Description

Add pod_creation_wait_interal property to give Kubernetes cluster some time to create pods for Job.

## Motivation and Context

Fix https://github.com/spotify/luigi/issues/2570, also aimed to replace abandoned https://github.com/spotify/luigi/pull/2664

### DIFF
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -66,6 +66,7 @@ class kubernetes(luigi.Config):
 
 class KubernetesJobTask(luigi.Task):
     __DEFAULT_POLL_INTERVAL = 5  # see __track_job
+    __DEFAULT_POD_CREATION_INTERVAL = 5
     _kubernetes_config = None  # Needs to be loaded at runtime
 
     def _init_kubernetes(self):
@@ -215,8 +216,8 @@ class KubernetesJobTask(luigi.Task):
 
     @property
     def pod_creation_wait_interal(self):
-        """Delay for initial pod creation for just submitted job"""
-        return 5
+        """Delay for initial pod creation for just submitted job in seconds"""
+        return self.__DEFAULT_POD_CREATION_INTERVAL
 
     def __track_job(self):
         """Poll job status while active"""
@@ -286,6 +287,7 @@ class KubernetesJobTask(luigi.Task):
                 'No pods found for {}, waiting for cluster state to match the job definition', self.uu_name
             )
             time.sleep(self.pod_creation_wait_interal)
+            pods = self.__get_pods()
 
         assert len(pods) > 0, "No pod scheduled by " + self.uu_name
         for pod in pods:
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -284,7 +284,7 @@ class KubernetesJobTask(luigi.Task):
         pods = self.__get_pods()
         if not pods:
             self.__logger.debug(
-                'No pods found for {}, waiting for cluster state to match the job definition', self.uu_name
+                'No pods found for %s, waiting for cluster state to match the job definition' % self.uu_name
             )
             time.sleep(self.pod_creation_wait_interal)
             pods = self.__get_pods()
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -213,10 +213,15 @@ class KubernetesJobTask(luigi.Task):
         """How often to poll Kubernetes for job status, in seconds."""
         return self.__DEFAULT_POLL_INTERVAL
 
+    @property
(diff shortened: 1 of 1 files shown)


plizmol:
@honnix @Tarrasch @dlstadther Could someone please review?

luigi/contrib/kubernetes.py
```diff
@@ -213,10 +213,15 @@ def poll_interval(self):
...
+    @property
+    def pod_creation_wait_interal(self):
+        """Delay for initial pod creation for just submitted job"""
+        return 5
+
     def __track_job(self):
         """Poll job status while active"""
         while not self.__verify_job_has_started():
-            time.sleep(self.__DEFAULT_POLL_INTERVAL)
+            time.sleep(self.poll_interval)
```
honnix: 👍

luigi/contrib/kubernetes.py
```diff
@@ -276,6 +281,11 @@ def __verify_job_has_started(self):
 
         # Verify that the pod started
         pods = self.__get_pods()
+        if not pods:
```
honnix: Not sure I follow the logic here. How would the `sleep` impact the assertion afterwards?

dlstadther: I'm with @honnix . Seems like you'd want to re-eval `pods=self.__get_pods()` before asserting

luigi/contrib/kubernetes.py
```diff
@@ -213,10 +213,15 @@ def poll_interval(self):
         """How often to poll Kubernetes for job status, in seconds."""
         return self.__DEFAULT_POLL_INTERVAL
 
+    @property
+    def pod_creation_wait_interal(self):
+        """Delay for initial pod creation for just submitted job"""
```
honnix: Please use a constant similar as `__DEFAULT_POLL_INTERVAL`, also in docstring please describe the unit so it would be easier to follow.

luigi/contrib/kubernetes.py
plizmol: Good catch, thanks

plizmol:
@honnix @dlstadther I believe your comments are resolved, could you please review again?

luigi/contrib/kubernetes.py
```diff
@@ -213,10 +214,15 @@ def poll_interval(self):
...
+    @property
+    def pod_creation_wait_interal(self):
+        """Delay for initial pod creation for just submitted job in seconds"""
+        return self.__DEFAULT_POD_CREATION_INTERVAL
+
     def __track_job(self):
         """Poll job status while active"""
         while not self.__verify_job_has_started():
-            time.sleep(self.__DEFAULT_POLL_INTERVAL)
+            time.sleep(self.poll_interval)
```
dlstadther: I think we might have steered you wrong... Or maybe i misunderstood the code the first time.

I'm looking at this more closely and have noticed that now each iteration of this `while not self.__verify_job_has_started()` we'll wait both `pod_creation_wait_interval` seconds AND `poll_interval` seconds.

This doesn't seem desirable. It also isn't clear from the user perspective - for 2 intervals to both impact the same wait time.

Am i missing something?

luigi/contrib/kubernetes.py
pisymbol: I believe you have a formatting bug on L287 of this patch. You want to use '%s' instead of '{}' and '%' instead of comma to be stylistically consistent with the rest of the file.

I too am running into this on one of our more established pipelines btw.

plizmol: Thanks, i fixed it

pisymbol: Please note that with your patch, things run better but now I am getting an issue with dependencies (Unfulfilled exceptions) when I raise workers > 1. I have no idea why.

plizmol:
@honnix @dlstadther @pisymbol ping

luigi/contrib/kubernetes.py
honnix: I think I understand the explanation. Most likely this new added `sleep` would only happen once. Either afterwards we see pods and will always see while waiting for container running state, or we won't and assertion will be made.

honnix:
LGTM. It would be nicer if we could separate waiting for pod to show up and container to be running.

plizmol:
> LGTM. It would be nicer if we could separate waiting for pod to show up and container to be running.

Sorry, could you please clarify is this a change request or just a comment? Also what are the odds that this PR will be merged anytime soon?

honnix:
@plizmol It's more as a comment. I'm OK with this change. What do you think @dlstadther ?

dlstadther:
I'm ok with this. If someone doesn't like it in the future, they can submit another PR to separate the delays.

Issue #2814, opened 2019-10-28
test/contrib/kubernetes_test.py fails on master


plizmol:
```
FEDEBUG: Checking if SuccessJob() is complete
/home/.../luigi/luigi/worker.py:409: UserWarning: Task SuccessJob() without outputs has no custom complete() method
  is_complete = task.complete()
INFO: Informed scheduler that task   SuccessJob__99914b932b   has status   PENDING
INFO: Done scheduling tasks
INFO: Running Worker with 1 processes
DEBUG: Asking scheduler for work...
DEBUG: Pending tasks: 1
INFO: [pid 14613] Worker Worker(salt=..., workers=1, host=..., username=..., pid=14613) running   SuccessJob()
DEBUG: Kubernetes auth method: kubeconfig
INFO: Submitting Kubernetes Job: success-20191028221352-86fbccc971ff4748
INFO: Start tracking Kubernetes Job: success-20191028221352-86fbccc971ff4748
DEBUG: Waiting for Kubernetes job success-20191028221352-86fbccc971ff4748 to start
INFO: To stream Pod logs, use:
INFO: `kubectl logs -f pod/success-20191028221352-86fbccc971ff4748-xlv8q`
INFO: Kubernetes job success-20191028221352-86fbccc971ff4748 succeeded
INFO: [pid 14613] Worker Worker(salt=..., workers=1, host=..., username=..., pid=14613) done      SuccessJob()
DEBUG: 1 running tasks, waiting for next task to finish
INFO: Informed scheduler that task   SuccessJob__99914b932b   has status   DONE
DEBUG: Asking scheduler for work...
DEBUG: Done
DEBUG: There are no more tasks to run at this time
INFO: Worker Worker(salt=..., workers=1, host=..., username=..., pid=14613) was stopped. Shutting down Keep-Alive thread
INFO: 
===== Luigi Execution Summary =====

Scheduled 1 tasks of which:
* 1 ran successfully:
    - 1 SuccessJob()

This progress looks :) because there were no failed tasks or missing dependencies

===== Luigi Execution Summary =====

.
======================================================================
ERROR: test_output (__main__.TestK8STask)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/.../.virtualenvs/luigi_temp/lib/python3.7/site-packages/mock/mock.py", line 1330, in patched
    return func(*args, **keywargs)
  File "test/contrib/kubernetes_test.py", line 105, in test_output
[...]

PR #2815, opened 2019-10-28
Fix failing tests for KubernetesJob

plizmol:
## Description

Fix failing tests for KubernetesJob
## Motivation and Context

Fix https://github.com/spotify/luigi/issues/2814
 

I have included unit tests

### DIFF
--- a/test/contrib/kubernetes_test.py
+++ b/test/contrib/kubernetes_test.py
@@ -91,7 +91,7 @@ class TestK8STask(unittest.TestCase):
         self.assertTrue("failed" in job.obj["status"])
         time.sleep(10)
         self.assertTrue(
-            len(Pod.objects(kube_api).filter(selector=f"job-name={fail.uu_name}").response['items']) ==
+            len(Pod.objects(kube_api).filter(selector="job-name={}".format(fail.uu_name)).response['items']) ==
             fail.max_retrials
         )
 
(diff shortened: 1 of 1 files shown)


Tarrasch:
Isn't tests running by Travis? Maybe we should mention it in the test file?

jbw:
I also can't run these tests locally. I'm not sure if they are getting ran as it doesn't appear in the coverage report (https://codecov.io/gh/spotify/luigi/tree/master/luigi/contrib).

Tarrasch:
Oh, this is an old PR. If anyone can confirm this fix makes it run locally then let's merge. @jbw , can you check?

test/contrib/kubernetes_test.py
```diff
@@ -81,19 +81,25 @@ def test_success_job(self):
...
         kube_api = HTTPClient(KubeConfig.from_file("~/.kube/config"))  # assumes minikube
         jobs = Job.objects(kube_api).filter(selector="luigi_task_id="
                                                      + fail.job_uuid)
         self.assertEqual(len(jobs.response["items"]), 1)
         job = Job(kube_api, jobs.response["items"][0])
         self.assertTrue("failed" in job.obj["status"])
-        self.assertTrue(job.obj["status"]["failed"] > fail.max_retrials)
+        time.sleep(10)
+        self.assertTrue(
+            len(Pod.objects(kube_api).filter(selector="job-name={}".format(fail.uu_name)).response['items']) ==
```
jbw, line 94: @Tarrasch this should still be > rather than == I think the sleep(10) is hiding this. I think the behaviour should be 1 greater than max retries since the filter will also include the original failed job. I can make that change and raise it in another PR.

jbw:
@Tarrasch are able to check the references PR #2930 

As a separate PR, I'd like to add in some extra unit tests so we can have it run as part of CI

PR #2818, opened 2019-11-06
[luigi.contrib.spark] pyspark python options added

drowoseque:
## Description
Added `hadoop_user_name`, `pyspark_python`, `pyspark_driver_python` options for SparkSubmitTask

## Motivation and Context
It's a usual usecase when you want to submit your spark application with a specific python environment on spark workers or under specific user

I have included a unit test for that
That approach works under my production environment

### DIFF
--- a/luigi/contrib/spark.py
+++ b/luigi/contrib/spark.py
@@ -97,7 +97,7 @@ class SparkSubmitTask(ExternalProgramTask):
 
     @property
     def hadoop_user_name(self):
-        return os.getlogin()
+        return None
 
     @property
     def spark_version(self):
--- a/luigi/contrib/spark.py
+++ b/luigi/contrib/spark.py
@@ -35,6 +35,12 @@ from luigi import configuration
 
 logger = logging.getLogger('luigi-interface')
 
+_DEFINED_SYSTEM_SETTINGS = (
+    'HADOOP_CONF_DIR',
+    'HADOOP_USER_NAME',
+    'PYSPARK_PYTHON',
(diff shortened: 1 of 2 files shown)


drowoseque:

drowoseque:
@mrk-its 
@GoodDok 
could you review this PR please ?

drowoseque:
@GoodDok
 your turn dude :)

luigi/contrib/spark.py
```diff
@@ -168,15 +192,13 @@ def archives(self):
     def hadoop_conf_dir(self):
         return configuration.get_config().get(self.spark_version, "hadoop-conf-dir", None)
 
-    def get_environment(self):
```
honnix: This seems to be a breaking change though.

drowoseque: @honnix 
it's an internal method used by `SparkSubmitTask` in `program_environment` only. 
But I agree that it's better not to change

drowoseque: @honnix 
done

honnix: Thanks. It's unfortunate we leaked this method as 'public'.

mfcabrera:
I wanted to say that this was already possible using the `conf` property of the Task. 

```python
 @property
    def conf(self):
        conf = super().conf or {}

        conf["spark.pyspark.driver.python"] = "bar"
        conf["spark.pyspark.python"] =  "foo"
        return conf
``` 

I prefer this way as I find it more flexible but I don't think having those additional properties hurt though. 

Cheers :).

PR #2819, opened 2019-11-06
Past date parameter

drowoseque:
## Description
Added a new parameter for launching tasks for past dates only
If task is submitted for today or later then exception raises

## Motivation and Context
It's a common use case to submit tasks that do some calculations over closed period of time; most common case is yesterday - we compute data that came yesterday
In this case if we call our task accidentally over today data which are yet not full (since today is not completed), we'll get incomplete output which is still present. Next time our task is submitted (by cron for example at 00:10 ) our task will be considered as completed and our pipelines that use this incomplete data will be incomplete as well. In order to avoid that situation we introduce `PastDateParameter` which will raise `FutureDateException` if we accidentally called our task over not yet closed date.

I've provided a couple of unit tests and also I have two year experience of using that parameter in production


drowoseque:

drowoseque:
@soxofaan 
@dlstadther 
@Tarrasch 

have a looke pls at this pr

luigi/parameter.py
```diff
@@ -497,6 +506,19 @@ def normalize(self, value):
         return datetime.date(year=value.year - delta, month=1, day=1)
 
 
+class PastDateParameter(DateParameter):
+    """
+    caveat: if your host default timezone is different from your `business` timezone
+    that might not work for you
+    """
+
+    def normalize(self, value):
```
dlstadther: by overriding the parent normalize, you lose the assurance that the param is a date and not a datetime, etc.

luigi/parameter.py
```diff
@@ -497,6 +506,19 @@ def normalize(self, value):
...
+        if value >= datetime.date.today():
```
dlstadther: It feels a little weird to evaluate "today" using system time. (Not saying this makes sense, but) suppose your system time is local but you eval based on UTC or visa versa.

dlstadther:
After thinking more about this and looking at `parameter.py`, i notice/remembered [`NumericalParameter`](https://github.com/spotify/luigi/blob/master/luigi/parameter.py#L1122) where it's sort of this merger between IntParameter or FloatParameter where you can also specific upper and lower bounds.

It got me thinking if it started to make sense (not saying it does) for there to be properties on a DateParameter (or really any parameter) to aid in data validation. I'm thinking like min/max bounds in a similar way to NumericalParameter. 

If so, then there might not need to be a new parameter type, but rather optional data validation controls on existing parameter types.

Again, not saying we_should_. Just wanted to put my thoughts here for conversation purposes.

@honnix or @Tarrasch Do y'all have thoughts on this?

luigi/parameter.py
drowoseque: @dlstadther
okay, that's fair
updated pr with possibility of choosing timezone - local or utc

drowoseque:
@honnix ^

honnix:
@dlstadther I think that is an idea worth exploring.

@drowoseque There are many changes not related to this PR. I would prefer keeping the PR small and do cleanup in a separated PR, if needed.

luigi/parameter.py
honnix: I believe this is resolved?

drowoseque: @honnix yes

drowoseque:
> @dlstadther I think that is an idea worth exploring.
> 
> @drowoseque There are many changes not related to this PR. I would prefer keeping the PR small and do cleanup in a separated PR, if needed.

@honnix 
which ones for example?

honnix:
e.g https://github.com/spotify/luigi/pull/2819/files#diff-ebc26d15ceb44f6010f69d513d8f3829R770

drowoseque:
> e.g https://github.com/spotify/luigi/pull/2819/files#diff-ebc26d15ceb44f6010f69d513d8f3829R770

@honnix 
fixed

drowoseque:
@honnix 
@dlstadther 
i've fixed comments
could you please take a look at it?

drowoseque:
@honnix 
@dlstadther 
ping

Tarrasch:
> if we call our task accidentally over today data which are yet not full (since today is not completed), we'll get incomplete output which is still present. 

I find this to be a very non-luigic pattern. Why don't your task's dependencies not do `complete()` checks a if all data isn't complete? Generally one shouldn't rely on that "it's been 12 hours data is probably in", because whatever produced that data could've been broken.

----

I don't think the motivation you have is a valid use case enough for it to be worth to maintaining this extra parameter type.

PR #2820, opened 2019-11-07
[luigi.contrib.spark] tracking_url_pattern as a property

drowoseque:
## Description
`tracking_url_pattern` setting took out from method `run` and was redefined as a property

## Motivation and Context
It's normally no a very good practice to put a variable setting logic inside other method
So I took it out

All tests are passed in `test/contrib/spark_test.py`
Particularly `test_tracking_url_is_found_in_stderr_cluster_mode` tests correctness of `tracking_url_pattern` work

### DIFF
--- a/luigi/contrib/spark.py
+++ b/luigi/contrib/spark.py
@@ -24,6 +24,7 @@ import shutil
 import importlib
 import tarfile
 import inspect
+
 try:
     import cPickle as pickle
 except ImportError:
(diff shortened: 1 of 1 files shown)


drowoseque:

drowoseque:
@dlstadther 
> You've removed the run method and its call to super().run(). That doesn't seem desirable.
Why?
In previous variant method `run` was the same as its parent's one except the fact that `tracking_url_pattern` was set inside it.
I just put definition of `tracking_url_pattern` out of `run` method into a property so now we don't need to redefine `run` method in `SparkSubmitTask`, so the logic stays the same (test runs prove that), just code is little bit cleaner

drowoseque:
@dlstadther how many approves do i need  for merging it?

PR #2822, opened 2019-11-11
[luigi.contrib.external_program.ExternalProgramTask] logs_output_pattern_to_url provided

drowoseque:
## Description
Method `logs_output_pattern_to_url` added to `ExternalProgramTask`

## Motivation and Context
Sometimes our external tasks do not contain url in output logs but one easily can be restored from output
For example if we run Tez Task which outputs YARN application id only
we can restore tracking url by appending a resource manager address + endpoint
That could be done by using `logs_output_pattern_to_url` 

I've provided unit tests for that


drowoseque:

luigi/contrib/external_program.py
```diff
@@ -180,7 +183,9 @@ def _track_url_by_pattern():
                         file_to_write.write(new_line)
                     match = re.search(pattern, new_line.decode('utf-8'))
                     if match:
-                        self.set_tracking_url(match.group(1))
+                        self.set_tracking_url(
+                            self.logs_output_pattern_to_url(match.group(1))
```
dlstadther: i don't see how this has any different impact. you're just returning the same value you're passing.

drowoseque: @dlstadther
Basically current behaviour is the same, but if I need to change logic of line found in logs transforming I can redefine logs_output_pattern_to_url in my inheritor class

drowoseque: @dlstadther

dlstadther: Ok makes sense now. I'm only now struggling with the name of the method; it doesn't seem clear right now. Thoughts on `format_tracking_url()` or something like that?

drowoseque: agree, name is weird
maybe `convert_to_url()` ?

dlstadther: specifically, it's for converting something to a tracking_url

drowoseque: then i can suggest

- `convert_to_tracking_url()`
- `make_tracking_url()`
- `make_tracking_url()`
- `tracking_url_from()`

drowoseque: @dlstadther 
@honnix 

what do you think ?

honnix: I will throw one more option. :D

`extract_tracking_url`

drowoseque: I'm not sure that's about exctracting _url_
that's more about matching _sth_  from logs into url

honnix: Isn't the matching part done before calling this method? To me this method will further look into the string and extract `tracking_url`.

What do you have in mind to do in this method exactly? Maybe just call it `make/build_tracking_url`?

drowoseque: yes, build is a perfect match

luigi/contrib/external_program.py
```diff
@@ -118,6 +118,9 @@ def _clean_output_file(self, file_object):
         file_object.seek(0)
         return ''.join(map(lambda s: s.decode('utf-8'), file_object.readlines()))
 
+    def build_tracking_url(self, logs_output):
```
honnix: Sorry for going back and forth. Adding docstring here would be very helpful for others to understand the need of this method.

drowoseque: @honnix done

honnix:
LGTM now.

drowoseque:
@honnix 
okay, can we go on with that?

honnix:
@drowoseque Could you please take a look why this broke master CI? Thanks.

drowoseque:
I see similar failures in https://travis-ci.org/spotify/luigi/jobs/618083277 but in next build this test was okay https://travis-ci.org/spotify/luigi/builds/619552243
It seems that `worker_test.DynamicDependenciesWithMultipleWorkersTes.test_dynamic_dependenciest` is flaky
```
FAIL: test_dynamic_dependencies (worker_test.DynamicDependenciesWithMultipleWorkersTest)
  File "/home/travis/build/spotify/luigi/test/worker_test.py", line 1066, in test_dynamic_dependencies
    self.assertTrue(time.time() - t0 < self.timeout)
AssertionError: False is not true
```

drowoseque:
@honnix could you please rerun ci ?

Issue #2756, opened 2019-08-09
Warnings conflict in initialization


dyens:
Two deprecaction warnings conflicts to each other.

Simple to reproduce:

```python
# DeprecationWarning ignored by default, unless triggered by code in __main__
import warnings 
warnings.simplefilter('default')
import luigi
```

After running this code a have a warning:

```
DeprecationWarning:
Autoloading range tasks by default has been deprecated and will be removed in a future version. 
To get the behavior now add an option to luigi.cfg:

          [core]
            autoload-range: false
```

Ok, i  create luigi.cfg in root:

```
▶ /bin/cat luigi.cfg
[core]
  autoload-range: false
```

Run script again. And i have another Warning:

```
DeprecationWarning: Configuration [core] autoload-range (with dashes) should be avoided. Please use underscores: autoload_range.
  warnings.warn(warn, DeprecationWarning)
```
Ohh. Ok, i change my config:
```
▶ /bin/cat luigi.cfg
[core]
  autoload_range: false
```
But now i have again previous DeprecationWarning...

PR #2823, opened 2019-11-11
Check for `autoload_range` istead of `autoload-range`

JensGM:
Using the cfg parser would either cause a deprecation warning without specifying

```
[core]
autoload-range=false
```

or shout at you for using a dash instead of an underscore.

This fix let's you specify autoload_range.

Resolves https://github.com/spotify/luigi/issues/2756

### DIFF
--- a/luigi/__init__.py
+++ b/luigi/__init__.py
@@ -61,19 +61,19 @@ __all__ = [
     'NumericalParameter', 'ChoiceParameter', 'OptionalParameter', 'LuigiStatusCode'
 ]
 
-if not configuration.get_config().has_option('core', 'autoload-range'):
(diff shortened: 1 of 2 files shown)


honnix:
@dlstadther Could you please take a look at this as you reviewed [the original PR](https://github.com/spotify/luigi/pull/2656)?

JensGM:
I'll try it tomorrow, but they should just get shouted at for using it, as they already do

JensGM:
@dlstadther You were right. I've updated the pull request to add an override of has_option, it now supports dashes, but raises the same warning as `_get_with_default`.

JensGM:
I think it's fair to point out that the current solution is incorrect, and does not work if you do as the warning suggests

PR #2826, opened 2019-11-12
[luigi.contrib.hive] WarehouseHiveClient

drowoseque:
## Description
I've added WarehouseHiveClient which allows to determine if table\partition exists when it's files exist on hdfs

## Motivation and Context
Sometimes it's not enough to have an entry in hive metastore that table\partition exists. 
In particular cases we want to be sure if files for that table/partition actually exists
Using of this client is suitable if you want to be sure that your task doesn't write zero rows.
It can be suitable for tasks like "insert overwrite table (...) partition (...)"

I've provided several unit tests for that

### DIFF
--- a/luigi/contrib/hive.py
+++ b/luigi/contrib/hive.py
@@ -16,6 +16,7 @@
 #
 
 import abc
+import collections
 import logging
 import operator
 import os
@@ -95,6 +96,23 @@ def run_hive_script(script):
     if not os.path.isfile(script):
         raise RuntimeError("Hive script: {0} does not exist.".format(script))
     return run_hive(['-f', script])
+
+
+class AmbiguousLocationException(Exception):
+    pass
+
+
+def _validate_partition(partition):
+    """
+    If partition is set and it's size more than one and it's not ordered
+    than we're unable to restore its path in warehouse definitely
+    """
+    if (
+            partition
+            and len(partition) > 1
+            and not isinstance(partition, collections.OrderedDict)
+    ):
+        raise AmbiguousLocationException()
 
 
 @six.add_metaclass(abc.ABCMeta)
@@ -298,6 +316,7 @@ class WarehouseHiveClient(HiveClient):
         return False
 
     def partition_spec(self, partition):
+        _validate_partition(partition)
         return '/'.join([
             '{}={}'.format(k, v) for (k, v) in six.iteritems(partition or {})
         ])
@@ -535,8 +554,10 @@ class ExternalHiveTask(luigi.ExternalTask):
     def output(self):
         if len(self.partition) != 0:
             assert self.partition, "partition required"
-            return HivePartitionTarget(table=self.table,
-                                       partition=self.partition,
-                                       database=self.database)
+            return HivePartitionTarget(
+                table=self.table,
+                partition=self.partition,
+                database=self.database
+            )
         else:
             return HiveTableTarget(self.table, self.database)
--- a/luigi/contrib/hive.py
+++ b/luigi/contrib/hive.py
@@ -22,6 +22,7 @@ import operator
 import os
 import re
 import subprocess
+import sys
 import tempfile
 import warnings
 
@@ -98,6 +99,16 @@ def run_hive_script(script):
     return run_hive(['-f', script])
 
 
+def _is_ordered_dict(dikt):
+    if isinstance(dikt, collections.OrderedDict):
+        return True
+
+    if sys.version_info >= (3, 7):
+        return isinstance(dikt, dict)
+
+    return False
+
+
 def _validate_partition(partition):
     """
     If partition is set and its size is more than one and not ordered,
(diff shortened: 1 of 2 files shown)


drowoseque:

luigi/contrib/hive.py
```diff
@@ -87,6 +98,19 @@ def run_hive_script(script):
     return run_hive(['-f', script])
 
 
+def _validate_partition(partition):
+    """
+    If partition is set and it's size more than one and it's not ordered
```
dlstadther: `If partition is set and its size is more than one and not ordered, then we're unable to restore its path in the warehouse`

luigi/contrib/hive.py
```diff
@@ -87,6 +98,19 @@ def run_hive_script(script):
...
+    than we're unable to restore its path in warehouse definitely
+    """
+    if (
+            partition
+            and len(partition) > 1
+            and not isinstance(partition, collections.OrderedDict)
```
dlstadther: how does this handle the fact that python 3.7+ dictionaries guarantee insertion order?

luigi/contrib/hive.py
```diff
@@ -248,12 +272,60 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        return os.path.join(
+            self.warehouse_location,
+            database + '.db',
+            table,
+            self.partition_spec(partition)
+        )
+
+    def table_exists(self, table, database='default', partition=None):
+        """
+        We consider table/partition as existing if corresponding path in hdfs exists
```
dlstadther: Can we remove the "we" wording?

`The table/partition is considered existing if corresponding path in hdfs exists and ....`

luigi/contrib/hive.py
drowoseque: Agree.
I should treat that case separately
sth like
`def _is_ordered_dict(dikt):
    import sys
    if sys.version_info >= (3, 7):
        return isinstance(dikt, dict)
    return isinstance(dikt, collections.OrderedDict)


def _validate_partition(partition):
    """
    If partition is set and its size is more than one and not ordered,
    then we're unable to restore its path in the warehouse
    """
    if (
            partition
            and len(partition) > 1
            and _is_ordered_dict(partition)`

Agree?
@dlstadther

dlstadther: I think collections.OrderedDict still exists in py3.7 so we'd still want to support that. If we want to allow both, i'd think check for ordereddict first then elif 3.7 and dict. else fail.

drowoseque: @dlstadther
I added handling of dicts in py3.7+ and modified a unit-test `test_table_exists_ambiguous_partition` for that
Could you please have a look at that?

drowoseque:
@rizzatti
@stroykova 
could you have a look at this pr?

luigi/contrib/hive.py
stroykova: Do we need to check if order exists when we require one or less elements?

drowoseque: @stroykova this code in comments is sketch, have a look at "Files changed" tab, please

luigi/contrib/hive.py
```diff
@@ -248,12 +283,60 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        """
+        The table/partition is considered existing if corresponding path in hdfs exists
+        and contains file except those which match pattern set in  `ignored_file_masks`
+        """
+        path = self.table_location(table, database, partition)
+        if self.hdfs_client.exists(path):
+            ignored_files = get_ignored_file_masks()
+            filenames = self.hdfs_client.listdir(path)
+            if ignored_files is None:
+                return True
```
stroykova: Is it possible to check ignored_files before calling listdir? It will be a little faster.
I am not familiar with hive warehouse, so I do not know is it needed to call listdir for existence.

drowoseque: absolutely agree

luigi/contrib/hive.py
```diff
@@ -248,12 +283,60 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                if not pattern.match(filename):
+                    return True
+
+        return False
+
+    def partition_spec(self, partition):
+        _validate_partition(partition)
+        return '/'.join([
+            '{}={}'.format(k, v) for (k, v) in six.iteritems(partition or {})
+        ])
```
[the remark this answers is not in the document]
drowoseque: Thank you for suggestion
In parent class this method is declared abstract
I agreee that this code looks very similar to other inheritors of base class

drowoseque:
@stroykova thank you for review so much

drowoseque:
> I think this makes sense. I've never used this part of luigi and i don't use a hive warehouse.
> 
> Could you ping the other contrib/hive contributors for review?

@dlstadther

done

PR #2827, opened 2019-11-12
Add Python version-agnostic get_writer_schema.

ra1861:
We got an error while running with Python3: ```AttributeError("'DatumReader' object has no attribute 'writers_schema'",)```

## Description
The name of the attribute is changed between versions 2 & 3.
[PY2](https://github.com/apache/avro/blob/master/lang/py/src/avro/io.py#L634)
[PY3](https://github.com/apache/avro/blob/master/lang/py3/avro/io.py#L421)
```
-def __init__(self, writers_schema=None, readers_schema=None):
+def __init__(self, writer_schema=None, reader_schema=None):
```
## Motivation and Context
Programs calling BigQueryLoadAvro fails.

Added a unit-test.

### DIFF
--- a/luigi/contrib/bigquery_avro.py
+++ b/luigi/contrib/bigquery_avro.py
@@ -1,7 +1,7 @@
 """Specialized tasks for handling Avro data in BigQuery from GCS.
 """
 import logging
-import sys
+import six
 
 from luigi.contrib.bigquery import BigQueryLoadTask, SourceFormat
 from luigi.contrib.gcs import GCSClient
@@ -77,10 +77,10 @@ class BigQueryLoadAvro(BigQueryLoadTask):
 
     @staticmethod
     def get_writer_schema(datum_reader):
-        if sys.version_info >= (3, 0):
-            return datum_reader.writer_schema
-        else:
+        if six.PY2:
             return datum_reader.writers_schema
+        else:
+            return datum_reader.writer_schema
 
     def _set_output_doc(self, avro_schema):
         bq_client = self.output().client.client
--- a/luigi/contrib/bigquery_avro.py
+++ b/luigi/contrib/bigquery_avro.py
@@ -61,7 +61,7 @@ class BigQueryLoadAvro(BigQueryLoadTask):
             # requiring the remainder of the file...
             try:
                 reader = avro.datafile.DataFileReader(fp, avro.io.DatumReader())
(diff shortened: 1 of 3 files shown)


luigi/contrib/bigquery_avro.py
```diff
@@ -74,6 +75,13 @@ def read_schema(fp):
             raise exception_reading_schema[0]
         return schema[0]
 
+    @staticmethod
+    def get_writer_schema(datum_reader):
+        if sys.version_info >= (3, 0):
```
drowoseque: I think that that `if six.PY3L:` is much more clear

ra1861: Sure.

dlstadther:
@ra1861 could you look into why most of the tests are now failing?

honnix:
Might need to install avro package in tox.

luigi/contrib/bigquery_avro.py
```diff
@@ -74,6 +75,13 @@ def read_schema(fp):
             raise exception_reading_schema[0]
         return schema[0]
 
+    @staticmethod
+    def get_writer_schema(datum_reader):
```
honnix: I think we can underscore this method.

luigi/contrib/bigquery_avro.py
```diff
@@ -60,7 +61,7 @@ def read_schema(fp):
             # requiring the remainder of the file...
             try:
                 reader = avro.datafile.DataFileReader(fp, avro.io.DatumReader())
-                schema[:] = [reader.datum_reader.writers_schema]
+                schema[:] = [self.get_writer_schema(reader.datum_reader)]
```
honnix: Calling a static method via `self` seems weird.

test/contrib/bigquery_avro_test.py
```diff
@@ -0,0 +1,35 @@
...
+
+from nose.plugins.attrib import attr
+
+@attr('gcloud')
+class BigQueryAvroTest(unittest.TestCase):
+
+    def test_writer_schema_method_existence(self):
+        reader = DatumReader(None, None)
+        schema = BigQueryLoadAvro.get_writer_schema(reader)
+        self.assertEqual(schema, None, "writer(s) schema attribute not found")
```
bardb: Would it be possible to use a value other than `None`?  `None` shows up a lot as a not-there value.

ra1861:
I think I am done with a code changes. :)

PR #2828, opened 2019-11-14
Handle both ‘writers_schema’ and ‘writer_schema’ from a DatumReader

bardb:
The Py2 version of a DatumReader has an attribute `writers_schema` which `contrib/bigquery_avro.py` references.  The Py3 version calls the same thing `writer_schema`, dropping one `s`.  We (Spotify) had trouble with a program (bq-load) built on Luigi which loads avro files into bigquery, because of this problem.

https://github.com/apache/avro/blob/master/lang/py/src/avro/io.py#L634-L641
https://github.com/apache/avro/blob/master/lang/py3/avro/io.py#L421-L428

This PR adds machinery to handle both versions.  It checks for one attribute then the other.   (As opposed to, say, branching on language version)

Unfortunately there aren't any tests for bigquery_avro, and I don't know enough to write them.

I did, at least, run bq-load with this version of Luigi and confirm that it solved my particular problem.

### DIFF
--- a/luigi/contrib/bigquery_avro.py
+++ b/luigi/contrib/bigquery_avro.py
@@ -54,13 +54,26 @@ class BigQueryLoadAvro(BigQueryLoadTask):
         schema = []
         exception_reading_schema = []
 
+        def extract_writer_schema(datum_reader):
+            # Extracts the writer's schema from datum_reader.
+            # The Py2 version of the code calls the field 'writers_schema'
+            # The Py3 version of the code calls the field 'writer_schema'
+            # See:
+            # https://github.com/apache/avro/blob/master/lang/py/src/avro/io.py#L634-L641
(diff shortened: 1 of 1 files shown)


dlstadther:
Is this a duplicate of #2827 ?

honnix:
I believe so, and I prefer the solution in #2827.

honnix:
@bardb Do you agree? Shall we close this PR and follow up in #2827 ?

bardb:
Either way.

I definitely like  #2827 <https://github.com/spotify/luigi/pull/2827> 's
tests.

I wrote mine as "use whichever attribute is available" rather than "use
writers for py2, writer for other py" because I don't 100% trust Avro not
to come up with a py2 release which uses writer, or even a py3 release
which goes back to writers.

But that's not very likely, with Py2 going away and all.

So sure, whichever.

bardb:
(Closed because https://github.com/spotify/luigi/pull/2827)

Issue #2830, opened 2019-11-22
Get ready for Python2 deprecation


honnix:
@dlstadther @Tarrasch It's almost there. We will need to give people heads-up before dropping Python2 CI.

dlstadther:
How should we handle versioning for this support deprecation? While day 0 will just be the removal of Python 2 testing, 2.7 will continue to work for a short time. Do we go ahead and major increase at that time?

honnix:
I think I makes sense to do a major release on day 0 even there is nothing as major. It is a promise that we no longer support py2 and will start introducing py3 only features. I also think we should remove six and all the other try import catch thingy. Wdyt?

drowoseque:
That sounds great
I would like to participate in it if you don't mind

honnix:
@drowoseque Of course not.

drowoseque:
I finally came with following idea about `luigi.six`

For each method in this module we 
- replace one with raising exception sth like `SixUsageException`
- eliminate method from `luigi`
- passing tests guarantee that no method invocation is performing

after that we remove `luigi.six`

drowoseque:
I've found a lot of crunches in luigi for working on python 3.2-3.5
Does Luigi correctly work on those versions if it's not covered by travis?
Should we remove those crunches or should we leave them?
@honnix 
@dlstadther

drowoseque:
I think I'm done with erasing py2-support in luigi
any other ideas to do for that?
@honnix 
@dlstadther

Tarrasch:
Nice to see it finally happening!

honnix:
Removing six and related docs then I think we are good. Thanks a lot for your great contribution @drowoseque . Good job!

Next step I'm thinking to release a few beta versions on that branch before merging to master (we should hold on till 2020 at least). What do you think, folks? We should also try to find a few beta testers, including Spotify as one of them.

honnix:
We might as well stop supporting old versions of py3.

drowoseque:
@honnix 

> Removing six and related docs then I think we are good. Thanks a lot for your great contribution @drowoseque . Good job!
> 
> Next step I'm thinking to release a few beta versions on that branch before merging to master (we should hold on till 2020 at least). What do you think, folks? We should also try to find a few beta testers, including Spotify as one of them.

we in cian.ru will gladly test this beta-version

drowoseque:
@honnix 
@dlstadther 
@Tarrasch 
[...]

PR #2831, opened 2019-11-24
#2830: Python 2 deprecation

drowoseque:
## Description
Python 2 deprecation
https://github.com/spotify/luigi/issues/2830
- six-functions replaced with py3-functions
 - `luigi.six` removed (including tests)
- `luigi.file` removed ("Delete this file any time after 7 Feb 2017")
- `luigi.contrib.hdfs.snakebite_client` removed as it does not support py3 (https://github.com/spotify/snakebite)
- py27 env removed from `tox.ini` and `.travis.yml`

https://www.python.org/dev/peps/pep-0373

All tests are passed


honnix:
Good job. You have been very productive!

I have a few suggestions might be useful to move forward more smoothly. Maybe we can make multiple PRs instead of a giant one. For example:

* One to reformat `core` code (we could expand that to `contrib` later after we clean up the pending PRs). We won't mind introducing `black` and `isort` in `tox` and `flake8` check it. I have some ideas and examples if you would like to see.
* One to clean up code for all sorts of things, docstring, comment, line breaks, legacy files, ...
* One to remove `future`
* One to remove six (could also be done in a few separated PRs)
* One to remove legacy code after py2 cleanup
* ...

And it is totally OK we do PR on PR style so at least we can review separately.

With this I believe it would be a lot easier for others to review code changes and a lot easier for us to ship code gradually (and early on) instead of a big bang. What do you think? Thanks again for the PR and it looks amazing!

drowoseque:
@honnix 
I've already thought about it after had a look at what I've done
I think I can break this giant one onto small ones

honnix:
@drowoseque Shall we close this PR?

PR #2836, opened 2019-11-25
no py27 in travis

drowoseque:
## Description
- py27 tests removed from `.travis.yml`

## Motivation and Context
#2830 

Tests are passing since several envs were removed

### DIFF
--- a/luigi/format.py
+++ b/luigi/format.py
@@ -481,3 +481,4 @@ Bzip2 = Bzip2Format()
 
 def get_default_format():
     return Text
+
--- a/luigi/format.py
+++ b/luigi/format.py
@@ -481,4 +481,3 @@ Bzip2 = Bzip2Format()
 
 def get_default_format():
     return Text
-
--- a/luigi/format.py
+++ b/luigi/format.py
@@ -20,11 +20,7 @@ import subprocess
 import io
 import os
 import re
-import locale
 import tempfile
-import warnings
-
-from luigi import six
 
 
 class FileWrapper(object):
--- a/luigi/format.py
+++ b/luigi/format.py
@@ -20,7 +20,11 @@ import subprocess
 import io
 import os
 import re
+import locale
 import tempfile
+import warnings
+
+from luigi import six
 
 
 class FileWrapper(object):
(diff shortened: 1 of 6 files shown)


drowoseque:
@honnix @dlstadther 
can't figure out why coverage become so low on `luigi.contrib.s3`  after removing py27 env in travis
could you please have a quick look through?
have any ideas?

honnix:
> can't figure out why coverage become so low on `luigi.contrib.s3`

What makes you believe this is low? https://codecov.io/gh/spotify/luigi/compare/34c80af2d96b7cfc1a96727a575ed6057352b416...d7acc5e120dfbafc9a86b309c4e3b689fbed74cb/diff looks fine to me.

drowoseque:
actually there are unexpected coverage changes in the report

https://codecov.io/gh/spotify/luigi/compare/34c80af2d96b7cfc1a96727a575ed6057352b416...d7acc5e120dfbafc9a86b309c4e3b689fbed74cb/changes

luigi/format.py
```diff
@@ -510,13 +477,7 @@ def pipe_writer(self, output_pipe):
 SysNewLine = NewlineFormat()
 Gzip = GzipFormat()
 Bzip2 = Bzip2Format()
-MixedUnicodeBytes = MixedUnicodeBytesFormat()
```
honnix: What is the replacement of this?

drowoseque: @honnix 
no replacement as this format was introduced in #786 with comment
> Accept to write unicode in Files for compat  Only accept on python2 Fix #786

honnix: Yeah it was added fo py2 compatibility, but could be abused for other purpose :( If for whatever reason, `lines` is a mixture of `str` and `bytes` and people decide to use this.

drowoseque: I don't think people use `luigi.format.*` in their mundane lives outside `luigi`
do you think we should leave it here?

honnix: I'm very sure people do and that is what these formatters for. I would suggest we leave it there and revamp the test case to reflect the "correct" usage: dealing with mixed of bytes and str. Wdyt?

drowoseque: sounds good.
+ maybe we should add deprecation warning for that

honnix: We don't need to deprecate it I think. It serves a use case and we have no replacement that we can recommend user to migrate to.

PR #2849, opened 2019-11-28
Add a JSON config parser

adonig:
## Description

I've added a JSON config parser analogous to the TOML one.

## Motivation and Context

JSON is a nice format to write config files. We needed lists/arrays, so INI was not an option and found the TOML tooling quite unpleasant to use (i.e. fmt-toml, TOML VSCode extension).

I have included unit tests and I ran my jobs with this code and it works for me.

If this gets merged, maybe you guys can add a note to the luigi docs, so people know how to use it (add a luigi.json file and set LUIGI_CONFIG_PARSER=json). A sample config can be found in the tests directory.


adonig:
I recently found out it is not sufficient to extend `luigi.configuration.BaseParser`. Sadly some modules, i.e. `luigi.contrib.s3`, rely on methods only defined in Python's `configparser.RawConfigParser`. Maybe we should put all methods that we rely on into `luigi.configuration.BaseParser`.

Tarrasch:
> Maybe we should put all methods that we rely on into luigi.configuration.BaseParser

That would be a good start.

> If this gets merged, maybe you guys can add a note to the luigi docs, so people know how to use it (add a luigi.json file and set LUIGI_CONFIG_PARSER=json). A sample config can be found in the tests directory.

Ideally you include documentation in this PR itself. :)

PR #2852, opened 2019-12-02
py 2.7 mentions removed from README and setup.py

drowoseque:
## Description
py 2.7 mentions removed from `README.md` and `setup.py`

## Motivation and Context
#2830


dlstadther:
Oh wait, could you also update `Contributing.rst` examples to not reference py27 tests?

PR #2856, opened 2019-12-03
[luigi.six.getargspec] removed

drowoseque:
## Description
`luigi.six.getargspec` removed

## Motivation and Context
#2830

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -96,12 +96,8 @@ def rpc_method(**request_args):
         # If request args are passed, return this function again for use as
         # the decorator function with the request args attached.
         args, varargs, varkw, defaults, kwonlyargs, kwonlydefaults, ann = inspect.getfullargspec(fn)
-        if kwonlyargs or ann:
-            raise ValueError("Function has keyword-only parameters or annotations"
-                             ", use getfullargspec() API which can support them")
-        (first_arg, *all_args) = args
-
         assert not varargs
+        first_arg, *all_args = args
         assert first_arg == 'self'
         defaults = dict(zip(reversed(all_args), reversed(defaults or ())))
         required_args = frozenset(arg for arg in all_args if arg not in defaults)
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -30,14 +30,13 @@ from luigi.batch_notifier import BatchNotifier
 import pickle
 import functools
 import hashlib
+import inspect
 import itertools
 import logging
 import os
 import re
 import time
 import uuid
-
-from luigi import six
 
 from luigi import configuration
 from luigi import notifications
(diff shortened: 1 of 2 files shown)


drowoseque:
@honnix 
@dlstadther 
PTAL
thanks

luigi/scheduler.py
```diff
@@ -96,12 +95,15 @@ def rpc_method(**request_args):
...
-
-        assert not fn_args.varargs
-        assert fn_args.args[0] == 'self'
-        all_args = fn_args.args[1:]
-        defaults = dict(zip(reversed(all_args), reversed(fn_args.defaults or ())))
+        args, varargs, varkw, defaults, kwonlyargs, kwonlydefaults, ann = inspect.getfullargspec(fn)
+        if kwonlyargs or ann:
+            raise ValueError("Function has keyword-only parameters or annotations"
+                             ", use getfullargspec() API which can support them")
+        (first_arg, *all_args) = args
```
dlstadther: are these parentheses necessary? (genuine question, i'm unsure)

luigi/scheduler.py
```diff
@@ -96,12 +95,15 @@ def rpc_method(**request_args):
...
+
+        assert not varargs
```
dlstadther: Should this be checked before unpacking `args`?

drowoseque: yes, makes sense
thanks

luigi/scheduler.py
drowoseque: no, they are not necessary
just it looks more clean with them

luigi/scheduler.py
```diff
@@ -96,12 +95,15 @@ def rpc_method(**request_args):
...
         # the decorator function with the request args attached.
-        fn_args = six.getargspec(fn)
-
-        assert not fn_args.varargs
-        assert fn_args.args[0] == 'self'
-        all_args = fn_args.args[1:]
-        defaults = dict(zip(reversed(all_args), reversed(fn_args.defaults or ())))
+        args, varargs, varkw, defaults, kwonlyargs, kwonlydefaults, ann = inspect.getfullargspec(fn)
+        if kwonlyargs or ann:
+            raise ValueError("Function has keyword-only parameters or annotations"
```
honnix: Not due to this PR, but I'm not sure what this error tries to tell. Here we already use `getfullargspec`. Do I miss anything?

drowoseque: @honnix 
I think that's related to the fact, this is going on inside of function `getargspec` which is not supposed to deal with annotations and keyword-only parameters
so this exception seems to be some sort of code-consistency among python version
but I'm still unsure.
I think we can remove this raising

drowoseque:
@honnix 
@dlstadther 
PTAL
thanks

cianru:
@honnix
@dlstadther
ping

PR #2857, opened 2019-12-03
[luigi.six.text type] removed

drowoseque:
## Description
`luigi.six.text type` removed

## Motivation and Context
#2830


drowoseque:
@honnix 
@dlstadther 
PTAL
thanks

PR #2860, opened 2019-12-06
isort everything and enforce it

honnix:
## Description

This is extremely existing-PRs unfriendly, but I started to feel bad about those imports. So this is up to discussion.


honnix:
Hmm this somehow broke everything. 🤔

honnix:
We have circular import .... :(

honnix:
Too much to fix I think. It's unfortunately some modules loaded by `luigi/__init__.py` relies on other things loaded by `luigi/__init__.py`.

honnix:
I think submodules should not rely on parent init.

Tarrasch:
This is an awesome change. I'm sorry to hear that it wasn't an easy change. Kinda weird that import order turns out to matter here :((

honnix:
@Tarrasch Yeah it worked eventually but I think the better fix is to avoid utilising `luigi/__init__.py` exports in core part of the code, and that is why we created a circular imports in this case.

honnix:
This probably has too much impact on existing PRs. Close for now.

PR #2861, opened 2019-12-09
[luigi.contrib.pyspark_runner] SparkSession support in PySparkTask

drowoseque:
## Description
`luigi.contrib.spark.PySparkTask` now supports `spark_session` as a first argument in `main()`.
To use it just enable `pyspark_runner.use_spark_session = True` in `luigi.cfg`
This change allows users to work with `spark_sesion` it in their jobs on Apache Spark 2+ version.

## Motivation and Context
Apache Spark 2.0.0 have introduced a new entrypoint object which is `SparkSession` instead of `SparkContext` in previous releases. 
`luigi.contrib.spark.PySparkTask` in its `main` function supports only `sc` argument which doesn't allow you to use spark session object.

* I've added unit tests into my PR
* I've checked locally that this works as expected (for both old-style and new-style syntax)


drowoseque:
this branch was checked out from wrong branch

PR #2862, opened 2019-12-09
[luigi.contrib.pyspark_runner] SparkSession support in PySparkTask

drowoseque:
## Description
`luigi.contrib.spark.PySparkTask` now supports `spark_session` as a first argument in `main()`.
To use it just enable `pyspark_runner.use_spark_session = True` in `luigi.cfg`
This change allows users to work with `spark_sesion` it in their jobs on Apache Spark 2+ version.

## Motivation and Context
Apache Spark 2.0.0 have introduced a new entrypoint object which is `SparkSession` instead of `SparkContext` in previous releases. 
`luigi.contrib.spark.PySparkTask` in its `main` function supports only `sc` argument which doesn't allow you to use spark session object.

* I've added unit tests into my PR
* I've checked locally that this works as expected (for both old-style and new-style syntax)


drowoseque:
@dlstadther 
@honnix 
@ntim 
please, take a look at that

Tarrasch:
You probably want other Spark users to comment on the usefulness of this PR. :)

drowoseque:
> You probably want other Spark users to comment on the usefulness of this PR. :)

yes, but none of them (`PysparkTask` contributors) seem to be online last year except @ntim and you @Tarrasch

luigi/contrib/pyspark_runner.py
```diff
@@ -36,11 +36,66 @@
...
 import os
 
+from luigi import configuration
+
 # this prevents the modules in the directory of this script from shadowing global packages
 sys.path.append(sys.path.pop(0))
 
 
-class PySparkRunner(object):
+class SparkEntryPoint(object):
```
Tarrasch: Can you make this private? (rename to `_SparkEntryPoint`?

luigi/contrib/pyspark_runner.py
```diff
@@ -36,11 +36,66 @@
...
+    def __init__(self, conf):
+        self.conf = conf
+
+    def __enter__(self):
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
Tarrasch: Aren't you overriding these anyway? Do you need this class? Can't you just do the `self.conf = conf` in both of the two classes below?

luigi/contrib/pyspark_runner.py
drowoseque: thanks for your response, I think this makes sense

luigi/contrib/pyspark_runner.py
drowoseque: surely I could, but I think that having `SparkEntryPoint` as an abstract class will add some clarity

drowoseque:
@b2arn
PTAL

drowoseque:
@Tarrasch done

drowoseque:
@honnix 
@dlstadther 
PTAL

luigi/contrib/pyspark_runner.py
```diff
@@ -51,14 +106,32 @@ def __init__(self, job, *args):
...
+def _pyspark_runner_with(name, entry_point_class):
+    return type(name, (AbstractPySparkRunner,), {'_entry_point_class': entry_point_class})
+
+
+PySparkRunner = _pyspark_runner_with('PySparkRunner', SparkContextEntryPoint)
+PySparkSessionRunner = _pyspark_runner_with('PySparkSessionRunner', SparkSessionEntryPoint)
+
+
+def _use_spark_session():
+    return bool(configuration.get_config().get('pyspark_runner', "use_spark_session", False))
```
dlstadther: Is this a new config class section? If so, how would you support the ability for users to decide on their own section name?

luigi/contrib/pyspark_runner.py
```diff
@@ -51,14 +106,32 @@ def __init__(self, job, *args):
         self.args = args
 
     def run(self):
-        from pyspark import SparkContext, SparkConf
+        from pyspark import SparkConf
         conf = SparkConf()
         self.job.setup(conf)
-        with SparkContext(conf=conf) as sc:
+        with self._entry_point_class(conf) as (entry_point, sc):
```
dlstadther: Any specific reason why you removed the kwarg reference?

luigi/contrib/pyspark_runner.py
```diff
@@ -36,11 +36,66 @@
...
 import os
 
+from luigi import configuration
+
 # this prevents the modules in the directory of this script from shadowing global packages
 sys.path.append(sys.path.pop(0))
 
 
-class PySparkRunner(object):
+class _SparkEntryPoint(object):
```
dlstadther: any reason why this isn't a child of `abc.ABCMeta`?

luigi/contrib/pyspark_runner.py
```diff
@@ -36,11 +36,66 @@
...
+                '''.format(spark_version)
+            )
+
+    def __enter__(self):
+        self._check_major_spark_version()
+        from pyspark.sql import SparkSession
+        self.spark = SparkSession \
+            .builder \
+            .config(conf=self.conf) \
+            .enableHiveSupport() \
```
dlstadther: Does it make sense to always enable Hive Support? Was that a default for Spark 1.x SparkContexts?

luigi/contrib/pyspark_runner.py
```diff
@@ -51,14 +106,32 @@ def __init__(self, job, *args):
...
     def run(self):
-        from pyspark import SparkContext, SparkConf
+        from pyspark import SparkConf
         conf = SparkConf()
         self.job.setup(conf)
-        with SparkContext(conf=conf) as sc:
+        with self._entry_point_class(conf) as (entry_point, sc):
             self.job.setup_remote(sc)
-            self.job.main(sc, *self.args)
+            self.job.main(entry_point, *self.args)
```
dlstadther: So when using `use_spark_session`, the entrypoint here is a SparkSession object rather than a SparkContext. How can this be documented clearly to users? Or is it even necessary? Will `job.main()` just know how to handle its first arg input based on its Spark version?

luigi/contrib/pyspark_runner.py
drowoseque: @dlstadther 
In spark 1.x hive support was provided in `HiveContext` - object differing from `SparkContext`
This object was obtained and managed inside client's `PySparkTask` inheritor
Now this two objects are merged into single `SparkSession`
But the fact is that when we create a `SparkSession` object once (inside of `SparkSessionEntryPoint`) clients are not able to enableHiveSupport in their client code since session is already started.
I think we can add a bool field `enable_hive_support` into `SparkSessionEntryPoint` which could be passed from `PySparkTask` inheritor

luigi/contrib/pyspark_runner.py
drowoseque: @dlstadther 
I didn't get it
Could you please explain?

luigi/contrib/pyspark_runner.py
drowoseque: @dlstadther 
Short answer is that client code supposed to be aware of spark version it uses.
Long answer is:

when we want to use spark1 syntax we remain everything without any changes
when we decide to move onto  spark2+ we rewrite all our `PySparkTask` inheritors from `SparkContext` usage to `SparkSession` usage
e.g.

```
class X(PySparkTask):
    def main(self, sc, *args):
        sc.textFile('...').saveAsTextFile('...')
```

onto


```
class X(PySparkTask):
    def main(self, spark, *args):
        spark.read.text('...').write ....
```

and set `pyspark_runner.use_spark_session`option to True

luigi/contrib/pyspark_runner.py
dlstadther: Got it. I'm ok starting with it enabled and someone can disable in the future if desired.

luigi/contrib/pyspark_runner.py
dlstadther: Ah i see. So there's more to be done than just adding the config setting. SparkSession and SparkContext do not provide identical signatures.

luigi/contrib/pyspark_runner.py
dlstadther: Nevermind here. I looked through some other examples and we don't generally allow configurability of section names.

drowoseque:
@GoodDok 
@mrk-its 
PTAL as well

drowoseque:
@Tarrasch @honnix 
please up

PR #2863, opened 2019-12-09
add to LocalTarget read and write from text, json, pickle

H4dr1en:
This PR aims to enrich LocalTarget with read and write functions for text, json and pickle files.

## Description
Add following functions to LocalTarget:

1. read_text
2. read_json
3. read_pickle
4. write_text
5. write_json
6. write_pickle

## Motivation and Context
It will remove a lot of boilerplate code:
```
class TaskA(luigi.Task):

    def output(self):
        return luigi.LocalTarget("foo.json")

    def run(self):
        output = dict(foo="bar")
        self.output().write_json(output)

class Task(luigi.Task):

    def requires(self):
        return TaskA()

    def output(self):
        return luigi.LocalTarget("foo.pickle")

   def run(self):
        input = self.input().read_json()
       # Fancy computations
        self.output().write_pickle(input)
```

### DIFF
--- a/luigi/local_target.py
+++ b/luigi/local_target.py
@@ -26,6 +26,8 @@ import tempfile
 import io
 import warnings
 import errno
+import pickle
+import json
 
 from luigi.format import FileWrapper, get_default_format
 from luigi.target import FileAlreadyExists, MissingParentDirectory, NotADirectory, FileSystem, FileSystemTarget, AtomicLocalFile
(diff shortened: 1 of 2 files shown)


Tarrasch:
I think using those methods in your code would make your code less readable no?

H4dr1en:
> I think using those methods in your code would make your code less readable no?

Well, I would argue that it is not: I have tons of Tasks where I have to write the following boilerplate code.

```
def run(self):
    with open(self.input().path, 'r') as f:
        json.load(f)

    # Fancy job

    with open(self.output().path, 'w') as f:
        json.dump(result, f)
```

So that I ended up declaring simple functions to do that:

```
def write_json(obj, file):
    with open(file, 'w') as f:
        json.dump(obj, f)

def read_json(file):
    with open(file, 'r') as f:
        json.load(f)
```

Which eventually I called as follows:

```
def run(self):
     inp = json.load(self.input().path)

    # Fancy job

    write_json(self.output().path, result)
```

Which is less readable, in the end, than having these functions in the LocalPath object.

honnix:
My view on this is those methods are too specific to add into core. And what's more, the implementation of those methods are generic and it doesn't make much sense to name them so specifically (think about later on if someone wants to add read_jpeg, read_pdf...). I would suggest subclassing LocalTarget and adding those methods there in your code.

honnix:
Sorry I was too quick. A correction, the implementation of pickle and json is not that generic. But still, my point is there.

H4dr1en:
@honnix Well I would be fine with having utils functions like `read_jpeg` / `read_pdf`. I worked on different project with Luigi I each time I had the need to code these functions (read/write pickle/json files). Therefore I think it makes sense to have them in `LocalTarget`, since this class already contains utils functions like `makedirs`, `open`or `move`.

drj42:
IMHO, these might be better implemented as custom format classes (like the ones defined in `luigi.format`).  Formatters provide a generic interface for wrapping target IO streams, can be chained together, and added to any target object that can support basic file IO (e.g LocalTarget, S3Target, etc).

PR #2865, opened 2019-12-10
Read parameter values from parent classes in configuration files

### LINKED ISSUE #2864: MissingParameterException with Config parent class
## Context

Consider the very simple following code:

```
import luigi

class cfg(luigi.Config):
    a = luigi.Parameter()

class TaskA(cfg):
    pass

if __name__ == '__main__':
    luigi.build([TaskA()], local_scheduler=True)
```

Together with the following `luigi.cfg` file:
```
[cfg]
a = foo
```

## Problem

This leads during execution to the following error:

```
File "/Users/xxx/test.py", line 16, in <module>
    luigi.build([TaskA()], local_scheduler=True)
File "/Users/xxx/luigi/luigi/task_register.py", line 88, in __call__
    param_values = cls.get_param_values(params, args, kwargs)
File "/Users/xxx/luigi/luigi/task.py", line 420, in get_param_values
    raise parameter.MissingParameterException("%s: requires the '%s' parameter to be set" % (exc_desc, param_name))
luigi.parameter.MissingParameterException: TaskA[args=(), kwargs={}]: requires the 'a' parameter to be set
```

Indeed, luigi tries to get value from parameter a in luigi.cfg only on `TaskA` section, not on parent `cfg` section.

## Suggestion

Luigi should try to get parent default configuration in `luigi.cfg`

H4dr1en:
## Description
Fix #2864 

Currently, during instantiation of a task, values of parameters are only checked in the following order:

- Positional arguments
- Keywords arguments
- defaults of task

How I changed this:

- Positional arguments
- Keywords arguments
- defaults of task
- **defaults of parents task**

"I ran my jobs with this code and it works for me.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -198,15 +198,15 @@ class Parameter(object):
 
         return self.parse(value)
 
-    def _get_value(self, task_name, param_name):
-        for value, warn in self._value_iterator(task_name, param_name):
+    def _get_value(self, task_name, param_name, allow_default=True):
+        for value, warn in self._value_iterator(task_name, param_name, allow_default):
             if value != _no_value:
                 if warn:
                     warnings.warn(warn, DeprecationWarning)
                 return value
         return _no_value
 
-    def _value_iterator(self, task_name, param_name):
+    def _value_iterator(self, task_name, param_name, allow_default=True):
         """
         Yield the parameter values, with optional deprecation warning as second tuple value.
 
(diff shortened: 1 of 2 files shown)


Tarrasch:
As I commented in the issue, I'm not sure what you're doing there is something we see as a valid use case. At least not the way I understood it. 

But I can see why one would want this too. Regardless this change would eventually require tests and docs. I'm also thinking if one should have this behavior configurable or not.


By the way by _defaults of task_ you meant _defaults of task (based on what's read in config)_ right?

H4dr1en:
> As I commented in the issue, I'm not sure what you're doing there is something we see as a valid use case. At least not the way I understood it.

In what is this not a "valid" use case?

> But I can see why one would want this too. Regardless this change would eventually require tests and docs. I'm also thinking if one should have this behavior configurable or not.

Configurable? What do you mean?

> By the way by _defaults of task_ you meant _defaults of task (based on what's read in config)_ right?

Yes, exactly!

Tarrasch:
> In what is this not a "valid" use case?

Nevermind, I understand what you're doing now. It became clear when I saw the docs you linked to from the issue.

> Configurable? What do you mean?

I think I meant (1) setting through cmdline and (2) setting through .cfg file.

> I'll add tests and doc as soon as I undersood the test architect

Sounds good! Also remember to use a more descriptive PR title. :)

tashrifbillah:
Unfortunately, I am unable to give this PR a go. It still doesn't solve the issue I reported in [this ](https://github.com/spotify/luigi/issues/2864#issuecomment-632927416)comment.

PR #2866, opened 2019-12-11
Add m3

vaaaaanquish:
Hi. M3 are using luigi!

We are making a luigi wrapper and tool kit. And we are using luigi's eco system.
- https://github.com/m3dev/gokart
- https://github.com/m3dev/thunderbolt

We will reduce our knowledge to luigi.
Thank you for the great library :)


Tarrasch:
Cool!

PR #2870, opened 2019-12-14
object inheritance removed in class declaration

drowoseque:
## Description
`class Xyz(object)` -> `class Xyz`

## Motivation and Context
#2830

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -8,6 +8,7 @@ participating, you are expected to honor this code.
 Running the tests
 -----------------
 
+
 We are always happy to receive Pull Requests. When you open a PR, it will
 automatically build on Travis. So you're not strictly required to test the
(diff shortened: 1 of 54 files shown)


CONTRIBUTING.rst
```diff
@@ -8,6 +8,7 @@ participating, you are expected to honor this code.
 Running the tests
 -----------------
 
+
```
Tarrasch: er why?

drowoseque: nothing special, but first commit went unsuccessfully with coveio arguing  on diffs
I know this is a special case for codeio to bark the wrong tree, so another additional commit with just blank line forces it to rerun it successfully

Tarrasch: Oh, I you've separated the commits. Thank you very much!

PR #2872, opened 2019-12-17
[luigi.contrib.hive] HiveTableTarget inherits HivePartitionTarget

drowoseque:
## Description
`luigi.contrib.hive.HiveTableTarget` inherits `HivePartitionTarget`

## Motivation and Context
* `HiveTableTarget` is a particular case of `HivePartitionTarget` with no partition
* methods in `HiveTableTarget` are copy-pasted from `HivePartitionTarget`

Unit-tests are updated

### DIFF
--- a/luigi/contrib/hive.py
+++ b/luigi/contrib/hive.py
@@ -471,10 +471,20 @@ class HiveQueryRunner(luigi.contrib.hadoop.JobRunner):
 
 class HivePartitionTarget(luigi.Target):
     """
-    exists returns true if the table's partition exists.
+    Target representing Hive table or Hive partition
     """
 
     def __init__(self, table, partition, database='default', fail_missing_table=True, client=None):
+        """
+        @param table: Table name
+        @type table: str
+        @param partition: partition specificaton in form of
+        dict of {"partition_column_1": "partition_value_1", "partition_column_2": "partition_value_2", ... }
+        If `partition` is `None` or `{}` then target is Hive nonpartitioned table
+        @param database: Database name
+        @param fail_missing_table: flag to ignore errors raised due to table nonexistence
+        @param client: `HiveCommandClient` instance. Default if `client is None`
+        """
         self.database = database
         self.table = table
         self.partition = partition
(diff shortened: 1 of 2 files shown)


luigi/contrib/hive.py
```diff
@@ -469,34 +469,6 @@ def run_job(self, job, tracking_url_callback=None):
             return luigi.contrib.hadoop.run_and_track_hadoop_job(arglist, job.set_tracking_url)
 
 
-class HiveTableTarget(luigi.Target):
```
drowoseque: this class moved a little bit down to be visible for a child class

luigi/contrib/hive.py
```diff
@@ -543,7 +514,22 @@ def path(self):
         return location
 
     def open(self, mode):
```
drowoseque: actually I think we don't need to declare this method at all as it's not present in base class `luigi.Target` but it's present in 'FilySystemTarget' only
@Tarrasch @dlstadther @honnix 
WDYT?

luigi/contrib/hive.py
drowoseque: I moved this class declaration down to make new parent visible to it

drowoseque:
@honnix 
@dlstadther 
PTAL

luigi/contrib/hive.py
```diff
@@ -543,14 +514,28 @@ def path(self):
...
-        return NotImplementedError("open() is not supported for HivePartitionTarget")
+        return NotImplementedError("open() is not supported for {}".format(self.__class__.__name__))
+
+
+class HiveTableTarget(HivePartitionTarget):
+    """
+    exists returns true if the table exists.
```
dlstadther: could this docstring be improved?

drowoseque: @dlstadther done

honnix:
@drowoseque Sorry for being late. I saw this PR has been approved. Anything else you need from me apart from merging it?

drowoseque:
> @drowoseque Sorry for being late. I saw this PR has been approved. Anything else you need from me apart from merging it?

@honnix 
if one approve is enough then i need nothing from you except merging :)

honnix:
Yeah one is usually enough. I will take a look just to get informed and then merge it. Thanks for the PR.

Issue #2873, opened 2019-12-18
update luigi.contrib.lsf.LSFJobTask._dump for python 3 (similar to #1683)


DingXuefeng:
python version 3.6.8, luigi version 2.8.10.

Cannot pickle dump because the file type is not binary. can update to the version used by SGE. Creating a common abstract class would be even better.

```
Traceback (most recent call last):
  File "/storage/gpfs_data/Software/python3.6.8/lib64/python3.6/site-packages/luigi/worker.py", line 199, in run
    new_deps = self._run_get_new_deps()
  File "/storage/gpfs_data/Software/python3.6.8/lib64/python3.6/site-packages/luigi/worker.py", line 141, in _run_get_new_deps
    task_gen = self.task.run()
  File "/storage/gpfs_data/Software/python3.6.8/lib64/python3.6/site-packages/luigi/contrib/lsf.py", line 191, in run
    self._init_local()
  File "/storage/gpfs_data/Software/python3.6.8/lib64/python3.6/site-packages/luigi/contrib/lsf.py", line 162, in _init_local
    self._dump(self.tmp_dir)
  File "/storage/gpfs_data/Software/python3.6.8/lib64/python3.6/site-packages/luigi/contrib/lsf.py", line 217, in _dump
    pickle.dump(self, open(self.job_file, "w"))
TypeError: write() argument must be str, not bytes
```

similar to #1683

PR #2875, opened 2019-12-20
fix lsf module for python 3

DingXuefeng:
## Description
- use 'rb' mode for writing and reading pickle files. this is already implemented in sge module.
- add 'python' in the command sequence. otherwise receive "permission denied" error.

## Motivation and Context
- to update lsf module. see issue #2873 

- I ran my jobs with this code and it works for me.
- The job can be submitted on LSF queue and status are properly tracked.


honnix:
Could you please add test cases to cover this change?

PR #2877, opened 2019-12-27
tox cleanup

honnix:
## Description

* Removed `pypy-scheduler` that doesn't seem to do anything
* No conditional dep for `avro-python3`
* Force `flake8`, `isort` and `docs` to use `python3`


honnix:
cc @drowoseque

PR #2890, opened 2020-02-05
All configuration parameters in docs now use underscore in their names for consistency

svetlyak40wt:
Previously it was hard to search some parameters on the page because in some places a minus sign
was used and in others is underscore.

This commit fixes this issue.

Also, it fixes a few places where one backquote was used to denote a parameter name or value.
RestructuredText format uses two backquotes to denote an inline code block:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -42,7 +42,7 @@ Example cfg config:
 
     [hadoop]
     version=cdh4
-    streaming-jar=/usr/lib/hadoop-xyz/hadoop-streaming-xyz-123.jar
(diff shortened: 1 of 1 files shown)


svetlyak40wt:
Don't know, show to fix codecov/changes.

Are there any suggestions?

Tarrasch:
Thanks for the change @svetlyak40wt !!!! I'm sorry for the very slow merge, I try to check Luigi at least monthly, but I'm not maintaining it so actively now.

svetlyak40wt:
Isn't Spotify interested in Luigi anymore?

PR #2891, opened 2020-02-05
Make worker_timeout=0 override the default timeout

adaitche:
Sometimes it is useful to exclude certain tasks from the default worker
timeout (set in the Luigi config). This was possible before commit
af6cf07 by setting `worker_timeout = 0` for a task. Since this commit
there is no distinction between `0` and `None` anymore and this approach
doesn't work anymore. The present commit introduces this distinction
again.

I ran my jobs with this code and it works for me

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -131,7 +131,11 @@ class TaskProcess(multiprocessing.Process):
         self.worker_id = worker_id
         self.result_queue = result_queue
         self.status_reporter = status_reporter
-        self.worker_timeout = task.worker_timeout or worker_timeout
+        self.worker_timeout = (
+            # Don't override task.worker_timeout when it is 0, only when it is None. This allows to
+            # override the default timeout for individual tasks (by setting it to 0).
+            worker_timeout if task.worker_timeout is None else task.worker_timeout
+        )
         self.timeout_time = time.time() + self.worker_timeout if self.worker_timeout else None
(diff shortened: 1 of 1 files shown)


luigi/worker.py
```diff
@@ -131,7 +131,11 @@ def __init__(self, task, worker_id, result_queue, status_reporter,
         self.worker_id = worker_id
         self.result_queue = result_queue
         self.status_reporter = status_reporter
-        self.worker_timeout = task.worker_timeout or worker_timeout
+        self.worker_timeout = (
+            # Don't override task.worker_timeout when it is 0, only when it is None. This allows to
```
honnix: Thanks for the fix. I don't think we need this comment because the code itself is clear to read. It was a bug after all.

adaitche: If a comment like this one would have been there before, the bug probably wouldn't have been introduced in the first place. That was my motivation for adding it, to prevent the bug from reappearing in some future refactoring.

But I will remove the comment, if still think it is unnecessary.

honnix: I think a better way is to cover this with a test case than explaining how a potential bug could be made.

adaitche: Ok, done. Added a test and removed the comment.

adaitche:
@honnix Have implemented your request. Please take another look.

adaitche:
Opened a new PR #2968 as this one doesn't show the changes and remains closed.

PR #2893, opened 2020-02-07
[luigi.contrib.gcs] Add GCSFileWrapper

dasoran:
## Description
Added GCSFileWrapper and changed to use GCSFileWrapper instead of FileWrapper

## Motivation and Context
I cannot distinguish File-Instance of GCS and other by `type()` because of using FileWrapper.

I ran my job and I have confirmed that GCSFileWrapper and FileWrapper can be distinguished.


honnix:
I have a mixed feeling a this PR. It could potentially be a breaking change. What do you think, @Tarrasch @dlstadther ?

dasoran:
@honnix Thank you for your comments.
I changed commit-id because the ci-tests was failed by a timeout to github.

PR #2895, opened 2020-02-08
Import ABC from collections.abc instead of collections for Python 3.9 compatibility

tirkarthi:
## Description

Import ABC from collections.abc for Python 3.9 compatibility since Python 3.9 raises an ImportError.

## Motivation and Context

Improves Python 3.9 compatibility.


tirkarthi:
Thanks for the review and merge :)

PR #2896, opened 2020-02-12
Handle rare race condition in worker process terminate

brandonAIQ:
## Description
There's already a handler in the code for when the PID we're trying to kill no longer exists, but no handler for what happens one step further than that -- the worker process was already gone, but then another process on the same instance came in and took the same PID. This PR adds a catch for this situation so the whole worker doesn't die with `OSError 1: Operation not permitted`.

## Motivation and Context
We saw the above error in some of our production clusters, dug in a bit and found out that the `luigi-worker` process was attempting to kill PIDs it didn't even own.

Because of its random-chance nature, bug was hard to reproduce, but (a) we verified manually (in a Python shell) that this catches the right error in the right circumstances, and (b) we haven't seen it again since implementing this fix.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -30,6 +30,7 @@ ways between versions. The exception is the exception types and the
 
 import collections
 import datetime
+import errno
 import getpass
 import importlib
 import logging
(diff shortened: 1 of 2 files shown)


honnix:
Thanks for the PR. It's a nice fix. It's interesting on your instance you could spawn so many processes over the time and cause PID wrapping. :)

Could you please check why CI failed and add test cases to cover the change? Thanks again.

honnix:
Sorry the CI failure was not caused by this change. FYI #2897

brandonAIQ:
Hah, yeah we have a lot of other processes on that box so it makes sense. And sure, added tests!

honnix:
Thanks for adding the unit tests. Could you please double check whether the unit tests cover the changes? According to https://codecov.io/gh/spotify/luigi/compare/8b3205c0b500224bc557e69158e1b12e45cc06ab...b89ec9b14a43d9e2152e4a51b016fb27ac3a9f51/diff they don't seem to.

PR #2897, opened 2020-02-13
avro-python3 1.9.2 is broken

honnix:
## Description

https://jira.apache.org/jira/browse/AVRO-2737


honnix:
They are working on a fix, so maybe this would not be needed.

honnix:
fixed in 1.9.2.1

PR #2900, opened 2020-02-18
support _FrozenOrderedDict in flatten

nishiba:
## Description

- `luigi.task.flatten` supports instances of `_FrozenOrderedDict`.

## Motivation and Context

-  I would like to flatten  `DictParameter` values.

I have included unit tests.

### DIFF
--- a/test/task_test.py
+++ b/test/task_test.py
@@ -27,6 +27,8 @@ import luigi
 import luigi.task
 import luigi.util
 import collections
+
+from luigi.parameter import _FrozenOrderedDict
 from luigi.task_register import load_task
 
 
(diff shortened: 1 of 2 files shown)


test/task_test.py
```diff
@@ -112,6 +114,7 @@ def test_flatten(self):
         self.assertEqual(flatten('foo'), ['foo'])
         self.assertEqual(flatten(42), [42])
         self.assertEqual(flatten((len(i) for i in ["foo", "troll"])), [3, 5])
+        self.assertEqual(flatten(_FrozenOrderedDict({'a': 'foo', 'b': 'bar'})), ['foo', 'bar'])
```
Tarrasch, line 117: Can we avoid testing internal behavior? Also it's an import which is weird.

PR #2908, opened 2020-02-26
add 'encoding' argument

takecx:
Add 'encoding' argument to base_parser.py and cfg_parser.py

## Description

Add 'encoding' argument to specify config file encoding.

## Motivation and Context

Currently LuigiConfigParser.add_config_path() fails when config file(e.g. luigi.toml) contains japanese.
Since luigi config file may contains japanese, I think LuigiConfigParser.add_config_path() should accepts 'encoding' argument.


dlstadther:
Are there tests you can add?

Issue #2917, opened 2020-03-09
config parsing inconsistency between Python 2 and Python 3


thomasgilgenast:
I'm using luigi in a library that supports both Python 2 and Python 3 from the same code base, and I'm trying to ensure that the same luigi config file can be used with both Python 2 and Python 3. I'm encountering an inconsistency between how the Python 2 ConfigParser and Python 3 configparser handle escaping of "%" symbols in the "ini/cfg"-style config file. configparser requires that "%" symbols be escaped (by writing "%%"; the extra "%" gets removed when accessing values from the config file), while ConfigParser does not ("%" symbols in the config file are fine).

A minimal example to reproduce this is (run in Python 2 with configparser installed):

```
>>> import ConfigParser
>>> config2 = ConfigParser.ConfigParser()
>>> config2.add_section('section')
>>> config2.set('section', 'unescaped', '10%')
>>> config2.set('section', 'escaped', '10%%')
>>> config2.get('section', 'escaped')  # bad because extra % in parsed value
'10%%'
>>>
>>> import configparser
>>> config3 = configparser.ConfigParser()
>>> config3.add_section('section')
>>> config3.set('section', 'unescaped', '10%')  # bad because ValueError
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "...\backports\configparser\__init__.py", line 1305, in set
    super(ConfigParser, self).set(section, option, value)
  File "...\backports\configparser\__init__.py", line 978, in set
    value = self._interpolation.before_set(self, section, option, value)
  File "...\backports\configparser\__init__.py", line 439, in before_set
    "position %d" % (value, tmp_value.find('%'))
ValueError: invalid interpolation syntax in u'10%' at position 2
>>> config3.set('section', 'escaped', '10%%')
>>> config3.get('section', 'escaped')
u'10%'
```

Not escaping the % symbol is not an option, because it will lead to a ValueError in Python 3. Escaping the % symbol is also not an option, because it will lead to different results between Python 2 and Python 3 ("10%%" vs "10%").

A related error that I've seen when parsing "unescaped" config files in Python 3 is:

```
configparser.InterpolationSyntaxError: '%' must be followed by '%' or '('
```

As far as I can tell, this is not related to #2527 / #2354 since this difference is not "caused" by luigi - it is caused by differences in the underlying ConfigParser/configparser implementations.

[...]

PR #2918, opened 2020-03-09
changed order of ConfigParser import attempts to try configparser first

thomasgilgenast:
## Description

In all modules where ConfigParser/configparser is imported, reverse the order of the try-except so that importing configparser is tried first.

## Motivation and Context

This is intended to address #2917 by ensuring that luigi first tries to import configparser before attempting to import ConfigParser, allowing Python 2 clients that have the configparser backport installed to experience the same parsing behavior as Python 3 clients.

Smoke-tested in Python 2 (with configparser installed) and Python 3. I successfully ran luigi with a config file containing "%%" in both Python 2 and Python 3. I don't have a formal test for my specific use case yet (I think it reduces to proving that configparser is imported when available in Python 2), but I'm happy to add a test for this if needed.

### DIFF
--- a/luigi/configuration/cfg_parser.py
+++ b/luigi/configuration/cfg_parser.py
@@ -36,11 +36,11 @@ import warnings
 # In python3 ConfigParser was renamed
 # https://stackoverflow.com/a/41202010
 try:
-    from ConfigParser import ConfigParser, NoOptionError, NoSectionError, InterpolationError
-    Interpolation = object
-except ImportError:
     from configparser import ConfigParser, NoOptionError, NoSectionError, InterpolationError
     from configparser import Interpolation, BasicInterpolation
+except ImportError:
+    from ConfigParser import ConfigParser, NoOptionError, NoSectionError, InterpolationError
+    Interpolation = object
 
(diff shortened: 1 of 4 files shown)


dlstadther:
Shouldn't this be resolved by #2830 once

thomasgilgenast:
Yes, once the `py2-deprecation` branch addressing #2830 is merged, this PR and #2917 become irrelevant (e.g., the `py2-deprecation` branch's `luigi/configuration/cfg_parser.py` doesn't have this try-except - reversing the order of a try-except that doesn't exist anymore is pointless, and after `py2-deprecation` is merged it will be as if the commits in this PR never existed).

There are two commits on `master` since 2.8.12, so I guess I opened this PR on the off-chance that there might be plans to release a "2.8.13" version that would still support Python 2 and would include those newest commits on `master` (and potentially this PR if it was merged) right before merging `py2-deprecation` and releasing 3.0.0. However, if the plan is to merge the `py2-deprecation` branch before the next release, then it seems to me that it is "too late" to address this issue - in that case we can close this PR as well as #2917.

PR #2919, opened 2020-03-10
fix toml ListParameter loading

### LINKED ISSUE #2812: Luigi Passes List to json.loads When Using TOML Config and ListParameter
I'm using luigi 2.8.9 with toml 0.10.0 as extra. My luigi.toml contains a list. Accessing its value with a luigi.ListParameter causes the following error:

>   File "/Users/asd/Projects/smtc/pipeline/.venv/lib/python3.7/site-packages/luigi/parameter.py", line 1054, in parse
>     return list(json.loads(x, object_pairs_hook=_FrozenOrderedDict))
>   File "/Users/asd/.pyenv/versions/3.7.4/lib/python3.7/json/__init__.py", line 341, in loads
>     raise TypeError(f'the JSON object must be str, bytes or bytearray, '

The problem is that `x` already has type `<class 'list'>` while it should be a `str` instead.

H4dr1en:
Fixes #2812

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -1007,6 +1007,8 @@ class ListParameter(Parameter):
         :param str x: the value to parse.
         :return: the parsed value.
         """
+        if isinstance(x, list):
+            x = json.dumps(x)
         return list(json.loads(x, object_pairs_hook=FrozenOrderedDict))
 
     def serialize(self, x):


H4dr1en:
@Tarrasch @honnix @dlstadther Could you please review? CI fails for something not related I believe.

dlstadther:
restarted job which failed to start

luigi/parameter.py
```diff
@@ -1007,6 +1007,8 @@ def parse(self, x):
         :param str x: the value to parse.
         :return: the parsed value.
         """
+        if isinstance(x, list):
+            x = json.dumps(x)
```
Tarrasch, line 1011: It would be nice if we have some docs explaining this. Core luigi should avoid having hard to understand code.

PR #2921, opened 2020-03-18
Add ExternalProgramTask param debug_environment

LinqLover:
## Description
This PR adds a parameter named `debug_environment` to the class `ExternalProgramTask`. If it is set to `False`, environment variables won't be printed to stderr when an error occurs.

## Motivation and Context
There are several scenarios in which you would not like to see all your environment variables being printed onto the stderr stream. For example, it is possible that they include private access tokens. It should be possible to run your pipeline on Travis CI using Shippable or something similar without revealing this information to the public. 

I have included unit tests, but I still need to check whether Travis will accept them. :D

### DIFF
--- a/test/contrib/external_program_test.py
+++ b/test/contrib/external_program_test.py
@@ -104,6 +104,29 @@ class ExternalProgramTaskTest(unittest.TestCase):
         else:
             self.fail('Should have thrown ExternalProgramRunError')
 
+    @patch('luigi.contrib.external_program.tempfile.TemporaryFile')
+    @patch('luigi.contrib.external_program.subprocess.Popen')
+    @patch.dict('os.environ', {'SECRET_TOKEN': '1l0v3ca75'}, clear=True)
+    def test_print_environment_on_failure_if_debug_environment(self, proc, file):
+        proc.return_value.returncode = 1
+        with self.assertRaisesRegex(ExternalProgramRunError, 'ENVIRONMENT: SECRET_TOKEN=\'1l0veca75\''):
+            job = TestExternalProgramTask(debug_environment=True)
+            job.run()
+
+    @patch('luigi.contrib.external_program.tempfile.TemporaryFile')
+    @patch('luigi.contrib.external_program.subprocess.Popen')
+    @patch.dict('os.environ', {'SECRET_TOKEN': '1l0v3ca75'}, clear=True)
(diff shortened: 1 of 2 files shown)


LinqLover:
Wait, why doesn't Travis CI run the contrib tests?

twollnik:
These changes would also be really useful for me. I would love to see this merged soon, if possible. @honnix @Tarrasch @dlstadther

test/contrib/external_program_test.py
```diff
@@ -104,6 +104,29 @@ def test_handle_failed_job(self, proc, file, logger):
         else:
             self.fail('Should have thrown ExternalProgramRunError')
 
+    @patch('luigi.contrib.external_program.tempfile.TemporaryFile')
+    @patch('luigi.contrib.external_program.subprocess.Popen')
+    @patch.dict('os.environ', {'SECRET_TOKEN': '1l0v3ca75'}, clear=True)
+    def test_print_environment_on_failure_if_debug_environment(self, proc, file):
```
narape: Can you add one more test to verify that the default behavior is still print the environment variables

LinqLover: See a185763. But what is wrong with the codecov checks?

LinqLover:
I am a bit confused that none of the travis logs contains the string `debug_environment`. Are the added tests actually run anywhere?

LinqLover:
@Tarrasch Thanks for the review! `print_cmdline_on_failures` is not actually what this PR allows to change. It only allows to disable printing of environment variables. So what's about `print_environment_on_failures`? (But that's a quite long name ...)

LinqLover:
@Tarrasch Any updates on this? :-)

Tarrasch:
@LinqLover , sorry I only look through the PRs on a very infrequent basis these days. Sorry!

That said, I still don't get the name `debug_environment`. It doesn't sound like it is a `bool` to begin with. Further, the Parameter description says something about "debug variables", afaik "debug" isn't a verb that means "print to screen". I understand one uses the verb like that but it's not universal among programmers at all.

LinqLover:
@Tarrasch Okay, would you agree with the name `print_environment_on_failures`? :)

LinqLover:
@Tarrasch This PR has been ready for a long time and is only lacking properly naming. I would like to finally merge it, how would you like the parameter to be named?

PR #2923, opened 2020-03-30
Add ability to run Kubernetes CronJobs from Luigi UI

mohamagdy:
## Description
To extend the usability of Luigi UI, added the ability to read the CronJobs
from Kubernetes cluster in a specific namespace and the ability to run the
CronJobs manually from Luigi UI.

## Motivation and Context
The Luigi UI is limited and does not have the ability to list and run jobs as in other scheduling tools. As a starting point, added interaction with Kubernetes API that can list the CronJobs in a Kubernetes cluster (which most probably will be scheduled jobs that uses Luigi) and add the ability to run those jobs manually. This reduces the overhead of authorization of the Kubernetes cluster, hopping to the CronJobs page and triggering the job manually.

The changes are tested with:

1. Existing valid Kubernetes configurations
2. Invalid Kubernetes configrations
3. `kubeconfig` authorization method
4. `service-account` authorization` method

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -22,6 +22,8 @@ See :doc:`/central_scheduler` for more info.
 """
 
 import collections
+import datetime
+
 try:
     from collections.abc import MutableSet
 except ImportError:
@@ -29,6 +31,11 @@ except ImportError:
 import json
 
 from luigi.batch_notifier import BatchNotifier
+
+try:
+    from kubernetes import client, config
+except ImportError:
+    print("No Kubernetes environment")
 
 try:
     import cPickle as pickle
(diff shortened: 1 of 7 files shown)


luigi/scheduler.py
```diff
@@ -1487,6 +1494,55 @@ def _first_task_display_name(self, worker):
...
+    def trigger_job(self, job_name):
+        try:
+            if configuration.get_config().get("kubernetes", "auth_method") == "kubeconfig":
+                config.load_kube_config()
+            else:
+                config.load_incluster_config()
+
+            batch_api_v1 = client.BatchV1beta1Api()
+
+            namespace = configuration.get_config().get("kubernetes", "namespace", "default")
```
irvifa, line 1531: I think it's better to make the namespace option as more dynamic instead of only using `default` namespace. However we can still put the `default` as the default value.

mohamagdy: @irvifa the current version reads the `namespace` value from the Luigi configuration https://luigi.readthedocs.io/en/stable/configuration.html#kubernetes. If the `namespace` configration is not set, it uses the `default` namespace. Do you suggest that the namespaces could be displayed in the page and one can choose a namespace and list the jobs there?

irvifa: Sorry for not making myself clear enough, yes, that's what I mean..

luigi/scheduler.py
```diff
@@ -1487,6 +1494,55 @@ def _first_task_display_name(self, worker):
...
+    def jobs(self):
+        try:
+            if configuration.get_config().get("kubernetes", "auth_method") == "kubeconfig":
+                config.load_kube_config()
+            else:
+                config.load_incluster_config()
+
+            batch_api_v1 = client.BatchV1beta1Api()
+
+            namespace = configuration.get_config().get("kubernetes", "namespace", "default")
```
irvifa, line 1507: For this part is there any possibility of using non default namespace? I think create a dynamic config can be helpful as well

mohamagdy: @irvifa maybe the answer is here https://github.com/spotify/luigi/pull/2923#discussion_r413960107? One can set the `namespace` cluster value in Luigi configurations and then it won't be the `default` value. Does this answer your comment?

PR #2929, opened 2020-04-10
remove StreamingBodyAdaptor that didn't allow choosing the chunk size

vbarbaresi:
### Context

The StreamingBodyAdaptor caused performance issues:
Calling `__iter__()` on a StreamingBodyAdaptor object defaulted to chunks of 1 byte, which was really slowing down reads.
We weren't able to pass the `size` parameter to change the chunk size while iterating on the streaming body.

### Solution

botocore StreamingBody implementation already provides a complete implementation of streaming by chunks, that behaves like a file.
https://github.com/boto/botocore/blob/master/botocore/response.py#L90

It supersedes luigi's StreamingBodyAdaptor
It defaults to chunks of 1024 bytes, and can be adjusted. 
It also offers iter_lines() and iter_chunks() methods.

I guess that maybe it didn't exist at the time the `StreamingBodyAdaptor` was written.

### Testing

We are using this change in production and it works as expected

### DIFF
--- a/luigi/contrib/s3.py
+++ b/luigi/contrib/s3.py
@@ -76,21 +76,6 @@ class DeprecatedBotoClientException(Exception):
     pass
 
 
-class _StreamingBodyAdaptor(io.IOBase):
-    """
-    Adapter class wrapping botocore's StreamingBody to make a file like iterable
-    """
-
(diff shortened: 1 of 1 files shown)


Tarrasch:
Thanks!

Issue #2905, opened 2020-02-23
Byte writing with localTarget


ifiddes:
I raised this issue on stackoverflow recently:

https://stackoverflow.com/questions/59312056/using-luigi-localtarget-with-pdfpages-in-python3/60309832#60309832

Is it possible to allow for byte write mode for localTargets?

hirolau:
I am no expert in the working of LocalTarget, so i do not know if there is a reason for the removal of the b-flag, or if this is a bug.

A way to work around this is to wrap the code using the `temporary_path` function:

```Python
import luigi

class BinaryFileExample(luigi.Task):
    def output(self):
        return luigi.LocalTarget("simple_binary_file.extension")

    def run(self):
        with self.output().temporary_path() as my_binary_file_path:
            with open(my_binary_file_path, 'wb') as y:
                newFileBytes = [123, 3, 255, 0, 100]
                for byte in newFileBytes:
                    y.write(byte.to_bytes(1, byteorder='big'))
```

ihowell:
We're getting the band back together!

For those wondering, that don't want extra clicks, the problem is here:

https://github.com/spotify/luigi/blob/master/luigi/local_target.py#L159

```
def open(self, mode='r'):
    rwmode = mode.replace('b', '').replace('t', '')
    ...
```

Opening just zeroes out byte and text flags, which causes everything to be treated as text.

ifiddes:
Is there a reason for that line to be there? What would be the implications of just removing it?

hirolau:
I do not have the time to run the tests on a linux setup, I can create a pull request in a week or two, need to investigate that everything works as it should when reading/writing from binary files.

But the code could look something like:
```Python
VALID_MODES = ('r', 'rb', 'w', 'wb', 'rt', 'wt')

    def open(self, mode='r'):
        
        if mode not in VALID_MODES:
            raise Exception("mode must be in %, (got: %s)" % (VALID_MODES, mode))

        if mode[0] == 'w':
            self.makedirs()
            return self.format.pipe_writer(atomic_file(self.path))

        elif mode[0] == 'r':
            fileobj = FileWrapper(io.BufferedReader(io.FileIO(self.path, mode)))
            return self.format.pipe_reader(fileobj)
```

charlesoblack:
I just opened #2931 which adds stuff to the docs about this. I believe it's related to how atomic writes/reads work.

ThePletch:
[...]

PR #2931, opened 2020-04-14
Fix docs explaining write modes for Luigi Targets. Closes #2783

### LINKED ISSUE #2783: Feature Request: binary write mode for LocalTarget
I'd like to use `pickle.dump` to write to a LocalTarget after calling `open` in write mode, kind of like this: 

```
def run(self):
    some_object = object()
    local_target = self.output()  # An instance of LocalTarget
    with local_target.open(mode="w") as f:  # <-- binary write not supported
        pickle.dump(some_object, f)
```

Unfortunately this causes an error: `TypeError: write() argument must be str, not bytes` because the LocalTarget open method doesn't allow binary write mode. This appears to be a deliberate decision, as the [source code](https://luigi.readthedocs.io/en/stable/_modules/luigi/local_target.html#LocalTarget) for the `open` method just strips the `b` out of the mode argument. 

It's easy enough to work around this by using `self.output().path` as suggested by [this kind stranger](https://stackoverflow.com/a/44418914). Nonetheless I think it would be nice either to include this capability as part of the LocalTarget `open` method, or to throw a more informative error instead of just stripping out the `b`.

charlesoblack:
## Description
I added some documentation under `Task.run()` and under the API docs for `target.FileSystemTarget` explaining not to use the binary `b` flag, but instead use `format=Nop`. This closes #2783 , at least for now. In the future, we might want to instead throw a warning explaining not to use `b`.

## Motivation and Context
Closes #2783 .

Built + opened documentation locally and it seems to have worked.

### DIFF
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -127,6 +127,31 @@ An example:
                                 )
                             )
 
+It's useful to note that if you're writing to a binary file, Luigi automatically
+strips the ``'b'`` flag due to how atomic writes/reads work. In order to write a binary
+file, such as a pickle file, you should instead use ``format=Nop`` when calling
+LocalTarget. Following the above example:
+
+.. code:: python
+
+    class GenerateWords(luigi.Task):
+
+        def output(self):
+            return luigi.LocalTarget('words.pckl', format=Nop)
+
+        def run(self):
+            import pickle
+
+            # write a dummy list of words to output file
+            words = [
+                    'apple',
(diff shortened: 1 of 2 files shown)


mivade:
Late to the party, but I came here from #2905. What is `Nop` and where do I import it from?

charlesoblack:
```python
from luigi.format import Nop
```

It's just a no-op.

PR #2932, opened 2020-04-18
Show task progress in visualizer workers tab.

riga:
## Description

This PR adds the task progress percentage as a bootstrap progress bar in a new column to the table of running tasks in the "Workers" tab:

progress

The value is even updated when the message modal of the specific task is open (which already had the ability to update task messages and progress via polling).

I didn't include the new column in the actual "Task List" tab, because it is probably only important for currently running tasks.

## Motivation and Context

When dealing with a couple long-running tasks, it is a nice feature to see the progress per task already in the worker tab.

The behavior is rather static, and I couldn't find existing tests that check similar features. However, I verified that the automatic updating via the `refreshInterval` of the opened modal is working properly and handles also `null` values.

### DIFF
--- a/luigi/static/visualiser/index.html
+++ b/luigi/static/visualiser/index.html
@@ -214,6 +214,7 @@
                         <th>Name</th>
                         <th>Priority</th>
                         <th>Resources</th>
+                        <th>Progress</th>
                         <th>Time</th>
                         <th>Actions</th>
                       </thead>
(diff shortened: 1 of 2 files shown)


dlstadther:
I've never dealt with the UI portion of Luigi. Do you have a colleague you can tag for review? Or perhaps a previous Luigi UI contributor?

I like the idea here.

riga:
I actually contributed a few UI related PRs such as task messages, worker processes customization, scheduler → worker communication, etc, but I think avoiding self-review is the point here ;)

@yrath @bfis Could you maybe give it a try? I prepared some test tasks to run in this [gist](https://gist.github.com/riga/cfaec1ffa2cadf373925fa89be0f33c5).

yrath:
Sorry for the late reply. I've checked out the PR based on the gist, everything worked as expected and looks good to me.

riga:
@yrath Thanks for testing!

riga:
@dlstadther Is this ok for you or should I provide additional tests?

dlstadther:
Fellow review and testing works for me here. I'll merge.

PR #2934, opened 2020-04-22
Use context managers for Psycopg tasks

LinqLover:
## Description
This PR refactors the Postgres-related contribution tasks to use context managers and increases the number of `try`-`finally` blocks. By this, we can increase their robustness. Plus, using these constructs is simply a proven idiom.

Note that the overall failure potential is quite high in this place because first of all, `CopyToTable` is an abstract class that provides several hooks for subclasses. If any subclass raises an error in such a hook, connections must be rolled back and cursors be released. At the moment, you are likely to receive tons of "database is already in use" failures as soon as a single task fails.

Further information on robust connection programming can be found on the [Psycopg2 documentation site](https://www.psycopg.org/docs/usage.html#with-statement).

Not explicitly tested, but I did not add any new features.

Provided that no existing tests will be broken, is it necessary to test such a refactoring?

(If required, it might be an option to create subclasses of the tasks that provoke errors.)

---

_Draft state will be resolved when the checks have completed._

I am looking forward to your review!


LinqLover:
The tests are not green, but many of the failures don't appear to be related to this PR. Can you tell me which of the failures I should consider?

dlstadther:
@LinqLover , I don't have the time to review thoroughly right now, but I did skim through the Travis test failures. They all deal with Postgres tests (and master is green), so these are all impacted by your code changes.

dlstadther:
Also, please be aware of the relationship between rdbms, postgres, and redshift as you're making changes here.

Issue #2752, opened 2019-07-29
current working directory not in sys.path causing "module not found" errors


jocassid:
I just found out about luigi this afternoon and started teaching myself how to use it.  

I set up an environment in `/home/john/Practice/luigi/`.  I created a hello_luigi.py based on an example in the documentation in `/home/john/Practice/luigi` While in `/home/john/Practice/luigi` I ran the command `luigi --local-scheduler --module hello_luigi MyTask --x 1 --y 2` and got a "hello_luigi" module not found error.  I modified  `/home/john/Practice/luigi/bin/luigi` to print the contents of `sys.path` and found out that whereas the `bin` directory was on the path it's parent (and my current working directory) `/home/john/Practice/luigi` was not.  By adding a symbolic in bin using the command `ln -s ../hello_luigi.py hello_luigi.py` I could get it to run.  This experiment with the symbolic link led the way to my workaround which is to add the following lines to  `/home/john/Practice/luigi/bin/luigi` just under `if __name__ == '__main__`

```
    cwd = getcwd()
    if cwd not in sys.path:
        sys.path.insert(0, cwd)
```
I see a `bin/luigi` file in the source code, but it differs from what I'm seeing.  I'm guessing this file is altered by the setup process.  The need for this hack was unexpected.  Skimming through the documentation I didn't see anything as to how a luigi project directory should be set up (i.e. where do the modules with my tasks go?).  Is there some `luigi.cfg` option that fixes this problem?

dgronskij:
This has nothing to do with `luigi` itself: when it sees '--module hello_luigi' it (roughly saying) tries to import it by means of standard python import system. It's just as if you wrote `from hello_luigi import MyTask` in your arbitrary script , so the `hello_luigi` module must be visible to Python interpreter. The list of locations where Python looks for packages is governed by `sys.path` list which can be configured via `PYTHONPATH` environment variable. Thus, doing

`export PYTHONPATH="$PYTHONPATH:<directory where hello_luigi.py resides>"` before runng scripts must help.

jocassid:
I think we need to include instructions about modifying `PYTHONPATH` right after the `pip install` command.  

[Some musings on technological work-arounds]
[...]

PR #2943, opened 2020-05-07
Monkey patching the sys.path for cmdline usage

shouldsee:
#2752


shouldsee:
I am mostly interested in the `--local-scheduler` where the cli interface will be mostly dealing with dev-like environment. If monkey-patching is not an option, we might as well add a command line argument saying `--import-cwd`.

PR #2945, opened 2020-05-13
Remove Asana from Luigi users

jdavidheiser:
Asana is no longer using Luigi, so for the sake of accuracy, it was suggested that we should be removed from the list of users.  The blog post that describes our use of Luigi did not include any implementation details or tips that would be useful to the broader community of Luigi users.


honnix:
Thanks for the update. 👍

honnix:
There is some problem in CI and we will fix it.

PR #2948, opened 2020-05-19
Fix build

ckiosidis:
## Description

The build fails on the latest flake8 version.
Also the azure-storage library changed on the latest version, pinning it to a previous working version.

## Motivation and Context

CI doesn't work.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -40,7 +40,7 @@ deps =
     pymongo==3.4.0
     toml<2.0.0
     responses<1.0.0
-    azure-storage
+    azure-storage==0.36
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -40,7 +40,7 @@ deps =
     pymongo==3.4.0
     toml<2.0.0
     responses<1.0.0
-    azure-storage
+    azure-storage==0.36
```
honnix, line 43: We can make it an upper boound.

Tarrasch:
Thanks!!

PR #2951, opened 2020-05-31
Fix task level disable_window

Obus:
## Description
1. added `task.Task.disable_window` & deprecated `task.Task.disable_window_seconds`  (for the sake of consistency with the rest of the code)
2. removed `scheduler.Failure` & moved its functionality to `scheduler.Task` (to fix the ignored task-level `disable_window` issue)

## Motivation and Context
Firstly, there is slight confusion in `disable_persist` & `disable_window` naming: sometimes excess `_seconds` prefix is used. As the 2nd & final step to removing it (1st one in https://github.com/spotify/luigi/pull/2950) `task.Task.disable_window_seconds` is renamed to `disable_window` (`disable_window_seconds` is still there for backwards-compatibility, but deprecated).

Secondly, while testing `task.Task.disable_window` I've found that its value is actually ignored due to the following:
* `scheduler.Task.retry_policy` value is updated twice: first in the [`scheduler.Task.__init__`](https://github.com/spotify/luigi/blob/eb97c4093edc7f49331d3855897dc90df4069800/luigi/scheduler.py#L318) (with the scheduler's retry policy) and later in the [`scheduler.Scheduler.add_task`](https://github.com/spotify/luigi/blob/eb97c4093edc7f49331d3855897dc90df4069800/luigi/scheduler.py#L958) method (with either the scheduler or the task retry policy).
* `scheduler.Task.failures` who keeps the `disable_window` value & related logic is updated just once in the [`scheduler.Task.__init__`](https://github.com/spotify/luigi/blob/eb97c4093edc7f49331d3855897dc90df4069800/luigi/scheduler.py#L319) thus it only uses the scheduler disable_window.

Since this "duplication" of the `disable_window` in both `scheduler.Task.failures` & `scheduler.Task.retry_policy` is error prone and the `scheduler.Failures` is not used outside of `scheduler.Task`, I've removed `scheduler.Failures` and moved its functionality to the `scheduler.Task`.

Added a unit test: [PerTaskRetryPolicyBehaviorTest.test_per_task_disable_persist_with_single_worker](https://github.com/spotify/luigi/compare/master...Obus:fix-task-level-disable-window?expand=1#diff-f1cd3288d80ac94f428311dbefa70c7fR2159)


luigi/scheduler.py
```diff
@@ -587,7 +568,7 @@ def set_status(self, task, new_status, config=None):
                         'disabled for {persist} seconds'.format(
                             failures=task.retry_policy.retry_count,
                             task=task.id,
-                            window=config.disable_window,
+                            window=task.retry_policy.disable_window,
```
dlstadther, line 571: if i followed your pr summary, this is the change necessary to catch the actual value of `disable_window`? but it will still catch the value provided to the scheduler's config disable_window?

Obus: @dlstadther on this line we just logging the up-to-date `disalbe_window` value, while the necessary change is getting rid of `task.failures` [here](https://github.com/spotify/luigi/pull/2951/files#diff-ec5c8f6c192cef5a08a5e22aa6273604L319) and using `task.retry_policy.disable_window` (which holds the "correct" value of `disable_window`) instead of `task.failures.window` (which can become stale). 

On the 2nd question: :100: correct, `task.retry_policy.disable_window` equalis a per-task `disable_window` value (if specififed) of the one specified if the scheduler's config.

Obus:
@dlstadther, many thanks for the review!

@honnix , @Tarrasch , would you like to take a glance?

PR #2955, opened 2020-06-04
Kerberos support in HiveThriftContext

SergeySmith:
## Description

added Kerberos (SASL) support in `HiveThriftContext`

## Motivation and Context

Current versions of `HiveThriftContext` and `MetastoreClient` cannot be used on Hadoop clusters with Kerberos. Proposed changes enable such functionality.

The new version of `HiveThriftContext` was tested on actual Hadoop cluster with Kerberos and it worked.


dlstadther:
Changes make sense to me and are non-impactful for users already using this functionality.
Two things:
1) could you tag a luigi-hive contributor for review? or someone you work with?
2) I see that we have some hive unittests. Could you review and see if there are meaningful test(s) you can add for this functionality?

SergeySmith:
@dlstadther thanks for your suggestions. I was searching for ways to test Kerberos support but no luck so far. I will try to work it out.

PR #2957, opened 2020-06-08
Return bq job id from biquery.run_job()

tomasaschan:
This ensures that the `job_id` is accessible to client code after running a BQ job; this is useful for getting logs or other metadata from the job, and/or logging the job id itself to other places than just what's done by this class.

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -787,6 +787,7 @@ BigqueryRunQueryTask = BigQueryRunQueryTask
 BigqueryCreateViewTask = BigQueryCreateViewTask
 ExternalBigqueryTask = ExternalBigQueryTask
 
+
 class BigQueryExecutionError(Exception):
     def __init__(self, job_id, error_message) -> None:
         super(BigQueryExecutionError, self).__init__('BigQuery job {} failed: {}'.format(job_id, error_message))
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -349,7 +349,7 @@ class BigQueryClient:
             if status['status']['state'] == 'DONE':
                 if status['status'].get('errorResult'):
                     raise Exception('BigQuery job failed: {}'.format(status['status']['errorResult']))
-                return
+                return job_id
 
             logger.info('Waiting for job %s:%s to complete...', project_id, job_id)
             time.sleep(5)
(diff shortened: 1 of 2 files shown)


luigi/contrib/bigquery.py
```diff
@@ -786,3 +786,10 @@ def run(self):
 BigqueryRunQueryTask = BigQueryRunQueryTask
 BigqueryCreateViewTask = BigQueryCreateViewTask
 ExternalBigqueryTask = ExternalBigQueryTask
+
+
+class BigQueryExecutionError(Exception):
+    def __init__(self, job_id, error_message) -> None:
```
honnix, line 792: It would be great we have docstring added with param and type description.

tomasaschan:
Thanks for the review!

Not sure how to properly test this, though - there are no tests for the `run_job` method currently, and I haven't found any docs on how to set up my local environment to properly run the tests (even the core ones; this would need running tests against GCP), so adding one is difficult.

However, a search through the project for `run_job` shows that all usages of this method (naturally) ignore its output, so I find it hard to imagine there would be any breakage due to this.

tomasaschan:
...added a couple of tests anyway, but I haven't been able to run these, so I have no idea what CI will say about them.

honnix:
It turns out we don't run anything for `gcloud` at all, not even `contrib` in general. Travis CI and test annotation seem to be a bit messy. Don't want to block the PR, so I'm fine we move on.

@dlstadther @Tarrasch What do you think? It seem some `contrib` related tests are not annotated properly and they run as part of `core`; also `contrib` is not enabled in TravisCI.

dlstadther:
Some tests annotations were removed/excluded from Travis while i was trying to contain test runtime [here](https://github.com/spotify/luigi/pull/2541). I definitely think we should try to include as many of the annotations as we can, but be considerate of the Travis resources we take and the runtime duration to wait for PR test response.

honnix:
@dlstadther What you think of this PR? Should we merge as it is or we'd better remove unexecuted/non-verified test cases?

dlstadther:
I think this is safe to merge as is. Then a separate PR can update Travis config and test annotations.

honnix:
Cool. I will create an issue first for Travis CI. It doesn't feel like something we can quickly get into place.

honnix:
Created https://github.com/spotify/luigi/issues/2959

tomasaschan:
Is there a time frame when we could expect a release that includes this?

No stress - we can solve it with inheritance and overriding in the meantime - but it would be nice to not have to carry that code longer than necessary.

honnix:
@tomasaschan Sorry I missed this. We will cut a release next week, according to release cycle.

PR #2968, opened 2020-07-01
Make worker_timeout=0 override the default timeout

adaitche:
This a resubmission of #2891, which was closed by stale-bot. The outstanding change request was to turn a comment into a test. This is implemented now.

Sometimes it is useful to exclude certain tasks from the default worker
timeout (set in the Luigi config). This was possible before commit
af6cf07 by setting `worker_timeout = 0` for a task. Since this commit
there is no distinction between `0` and `None` anymore and this approach
doesn't work anymore. The present commit introduces this distinction
again.

I ran my jobs with this code and it works for me. Also added a test case.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -123,7 +123,7 @@ class TaskProcess(multiprocessing.Process):
         self.worker_id = worker_id
         self.result_queue = result_queue
         self.status_reporter = status_reporter
-        self.worker_timeout = task.worker_timeout or worker_timeout
+        self.worker_timeout = task.worker_timeout if task.worker_timeout is not None else worker_timeout
         self.timeout_time = time.time() + self.worker_timeout if self.worker_timeout else None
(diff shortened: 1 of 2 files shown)


luigi/worker.py
```diff
@@ -123,7 +123,7 @@ def __init__(self, task, worker_id, result_queue, status_reporter,
         self.worker_id = worker_id
         self.result_queue = result_queue
         self.status_reporter = status_reporter
-        self.worker_timeout = task.worker_timeout or worker_timeout
+        self.worker_timeout = worker_timeout if task.worker_timeout is None else task.worker_timeout
```
honnix, line 126: I find `task.worker_timeout if task.worker_timeout is not None else worker_timeout` might be easier to read, but no strong preference.

adaitche:
Done (flake and if-expression reformatting).

PR #2971, opened 2020-07-15
Add failed_execution_summary_log_level config setting

ThePletch:
## Description

Adds a new config setting to allow failed execution summaries to be logged at a different log level than successful execution summaries. The default value of the setting maintains Luigi's current behavior: it logs all execution summaries as INFO.

## Motivation and Context

The test suite output of our application that uses Luigi is extremely verbose, and a big contributor is the logs for tasks that were successfully executed, which we generally don't care about. This change allows us to configure our test suite (and, perhaps, our production logs) to only include execution summaries for executions that failed.

I have included unit tests.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -374,6 +374,12 @@ check_complete_on_run
   missing.
   Defaults to false.
 
+failed_execution_summary_log_level
+  Controls the log level at which the execution summaries for failed tasks are
+  reported. Useful for trimming down the size of test suite or system log output,
+  but still including failure information for debugging purposes.
+  Defaults to INFO.
+
 
 [elasticsearch]
 ---------------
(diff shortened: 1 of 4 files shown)


ThePletch:
Looks like this breaks a test. I'll investigate the failing test later this week.

ThePletch:
Would love guidance on these tests if possible - I'm getting a "logging already configured" `INFO`-level log message in my tests, but no additional messages. Am I doing something incorrect?

patrick7kelly:
@ThePletch Just a heads up, your tests are failing because the `Worker()` is instantiated in [setUp()](https://github.com/ThePletch/luigi/blob/474a98631847daadf14dfb3c398cf6e9b5d16970/test/interface_test.py#L38), so the `@with_config` decorator has no effect.

I realized the issue after checking how `@with_config` is used in [other tests](https://github.com/ThePletch/luigi/blob/474a98631847daadf14dfb3c398cf6e9b5d16970/test/worker_external_task_test.py#L121).

PR #2973, opened 2020-07-20
Garbage collect task result queue when worker context exits

ThePletch:
## Description

Adds a call to explicitly close() the task result queue when a worker exits, rather than waiting for the worker to be garbage collected.

## Motivation and Context

There are certain contexts (e.g. during test runs on certain architectures) where a worker may not be garbage collected until significantly after it exits. In test suites affected by this, running many Luigi tasks via the local scheduler can cause the test suite to break due to exceeding the system's open files threshold (as each task result queue uses a separate file-based semaphore). This fails all tests after the threshold is reached. Explicitly closing the task result queue when a worker exits ensures that the queue is not held open unnecessarily. Due to the behavior of close(), this should not close the queue for other producers/consumers, only the worker in question.

After making this change and re-running my test suite, the 'too many open files' issue is no longer present.

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -606,6 +606,7 @@ class Worker:
         for task in self._running_tasks.values():
             if task.is_alive():
                 task.terminate()
+        self._task_result_queue.close()
         return False  # Don't suppress exception
 
     def _generate_worker_info(self):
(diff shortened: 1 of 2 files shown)


ThePletch:
Looks like this breaks a test. I'll investigate the failing test later this week.

ThePletch:
Looks like tests are passing but my PR is breaking due to upstream changes that aren't covered. I don't think there's much I can do about those.

ThePletch:
@dlstadther Any word on this getting released? It'll make running our full test suite locally possible, so I'm anxious for it to get out.

dlstadther:
@ThePletch I can merge it and make it available from `master`. But it'll be up to @honnix and the folks at Spotify to push out the next release. I haven't been as involved and so leaving releases up to them.

Issue #2947, opened 2020-05-15
[Question] How to store history indefinitely?


tashrifbillah:
According to `Section 8.6.2 Enabling Task History` in the [doc](https://readthedocs.org/projects/luigi/downloads/pdf/latest/), task history is stored for last 24 hours only. Is there a way to override that to make it indefinite?

There is one parameter in Section `8.9.4 [cors]`:
`max_age Content of Access-Control-Max-Age. Defaults to 86400 (24 hours)`
that seems relevant to the duration. Would changing that help?

tashrifbillah:
Hi @Tarrasch , can you give me a direction?

Tarrasch:
Are you talking about the "Task DB" feature? I've never used that actually.

tashrifbillah:
I am talking about 

```cfg
[task_history]
db_connection = sqlite:////usr/local/var/luigi-task-hist.db
```

not sure if you call that "Task DB".

Can we ping someone who might know?

drowoseque:
@tashrifbillah 
Actually the history db stores data infinitely
24 hours relates to the history displaying.

tashrifbillah:
Hi @drowoseque , in that case, do you know how to display history beyond 24 hours?

drowoseque:
unfortunately there is no way to do it right now as limitation of 24 hours is hardcoded
![image](https://user-images.githubusercontent.com/3604364/85408349-12d17100-b56d-11ea-8c83-1b6ea7127ce9.png)

tashrifbillah:
So I should be able to link `days=1` to a new configuration parameter and submit a PR :)

drowoseque:
@tashrifbillah btw if you're interested in some particular task you can use /history/by_name/ endpoint which doesn't have any restrictions on dates

tashrifbillah:
Hi @drowoseque , can you help a bit about `/history/by_name/` ? [This](https://gist.github.com/tashrifbillah/2bca276c7a6e2d5c85e9206d6522b37e) is my simple job, upon execution of which--

`/history` shows fine:

![image](https://user-images.githubusercontent.com/35086881/85422248-63cc6f80-b543-11ea-825d-e21caaf7cb20.png)

But `/history/by_name/:TaskB` does not:

![image](https://user-images.githubusercontent.com/35086881/85422351-83639800-b543-11ea-87c2-edaa662f7ddf.png)


Here is my `luigi.cfg`:
```vim
[scheduler]
record_task_history = True
[...]

PR #2977, opened 2020-08-06
Fix history uri

tashrifbillah:
Fixes https://github.com/spotify/luigi/issues/2947

### DIFF
--- a/doc/central_scheduler.rst
+++ b/doc/central_scheduler.rst
@@ -78,20 +78,20 @@ The task history has the following pages:
 
     .. figure:: history.png
        :alt: Recent history screenshot
-* ``/history/by_id/id``
+* ``/history/by_id/{id}``
   detailed information about a run, including:
   parameter values, the host on which it ran, and timing information.
   Example screenshot:
 
     .. figure:: history_by_id.png
        :alt: By id screenshot
-* ``/history/by_name/name``
-  a listing of all runs of a task with the given task ``name``.
+* ``/history/by_name/{name}``
+  a listing of all runs of a task with the given task ``{name}``.
   Example screenshot:
 
     .. figure:: history_by_name.png
        :alt: By name screenshot
-* ``/history/by_params/name?data={"param": "value"}``
-  a listing of all runs of the task ``name`` restricted to runs with ``param=value`` matching the given history.
-  The ``data`` is a json blob describing the parameters,
-  e.g. ``{"foo": "bar"}`` looks for a task with ``foo=bar``.
+* ``/history/by_params/{name}?data=params``
+  a listing of all runs of the task ``{name}`` restricted to runs with ``params`` matching the given history.
+  The ``params`` is a json blob describing the parameters,
+  e.g. ``data={"foo": "bar"}`` looks for a task with ``foo=bar``.
--- a/doc/central_scheduler.rst
+++ b/doc/central_scheduler.rst
@@ -78,20 +78,20 @@ The task history has the following pages:
 
     .. figure:: history.png
        :alt: Recent history screenshot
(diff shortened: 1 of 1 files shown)


doc/central_scheduler.rst
```diff
@@ -78,20 +78,20 @@ The task history has the following pages:
 
     .. figure:: history.png
        :alt: Recent history screenshot
-* ``/history/by_id/:id``
+* ``/history/by_id/id``
```
ckiosidis, line 81: The colon refers to the `id` being a path parameter, we should revert it to having the `:`

doc/central_scheduler.rst
```diff
@@ -78,20 +78,20 @@ The task history has the following pages:
...
   detailed information about a run, including:
   parameter values, the host on which it ran, and timing information.
   Example screenshot:
 
     .. figure:: history_by_id.png
        :alt: By id screenshot
-* ``/history/by_name/:name``
```
ckiosidis, line 88: The colon refers to the `name` being a path parameter, we should revert it to having the `:`

doc/central_scheduler.rst
```diff
@@ -78,20 +78,20 @@ The task history has the following pages:
...
-  a listing of all runs of a task with the given task name.
+* ``/history/by_name/name``
+  a listing of all runs of a task with the given task ``name``.
   Example screenshot:
 
     .. figure:: history_by_name.png
        :alt: By name screenshot
-* ``/history/by_params/:name?data=params``
```
ckiosidis, line 94: The colon refers to the `name` being a path parameter, we should revert it to having the `:`

doc/central_scheduler.rst
tashrifbillah: Did I mention plugging in the URI with that colon resolves to error 404?

doc/central_scheduler.rst
tashrifbillah: > we should revert it to having the :

Not sure what do you mean. The `:` is already in the current doc. Even if you reject my PR, no reversion is necessary.

doc/central_scheduler.rst
honnix: @tashrifbillah `:id` is used as a convention in Luigi doc to indicate this is a variable to be replaced, the same purpose as `{id}` that is more commonly used today.

tashrifbillah: > :id is used as a convention in Luigi doc to indicate this is a variable to be replaced

Did you mention it somewhere in the doc? I guess the question is why would we use an unfamiliar convention that confuses new users?

honnix: Well. This is a question that needs some archaeology work. ;)

I think the colon convention was not invented by Luigi and is used by other frameworks as well, such as express.js.

What would you suggest instead of colon? Would `{id}` for example be better in this case?

tashrifbillah: > Would {id} for example be better in this case?

I think so.

honnix: Cool. Would you like to revise this PR a bit? @tashrifbillah

tashrifbillah: Sure, the places where I changed, are those the only places where I should effect the `:` vs `{ }` change?

honnix: That seems to be the case. I didn't find anything else after a quick search.

tashrifbillah: Hi @honnix , sorry about the delay. I revised the PR as you suggested. How does it look now?

tashrifbillah: Ping @honnix ...

honnix (approved):

Issue #2949, opened 2020-05-19
Support Tornado>=6


maresb:
Sorry to duplicate #2805, but I'm forced to uninstall Luigi from my JupyterLab environment since Tornado < 6 is breaking my extensions. Would you reconsider supporting Tornado 6? Thanks!

honnix:
According to Tornado release notes: `Python 2.7 and 3.4 are no longer supported; the minimum supported Python version is 3.5.2.`

However Luigi still claims to support Python 3.3 and 3.4. So there are two options: 1) Luigi drops 3.3. and 3.4; 2) have more branches in [setup.py](https://github.com/spotify/luigi/blob/master/setup.py#L51). Wdyt @Tarrasch @dlstadther ? Both 3.3. and 3.4 have reached end of life, so it seems fair to drop them (we are not testing either of those anyway in CI).

dlstadther:
I'm cool to drop mentioned support of 3.3 and 3.4

honnix:
Cool. I will work on some PRs to follow this up.

stonebig:
hum, still no hope for tornado>=6 support ? waiting for https://github.com/tornadoweb/tornado/pull/2815 ?

honnix:
@stonebig Sorry, blame me on this. I will make a few PRs to move this forward.

PR #2978, opened 2020-08-10
Drop Python 3.3 and 3.4 support

honnix:
## Description

Drop Python 3.3 and 3.4 support. Also allows more recent tornado version (>=6,<7).

## Motivation and Context

Python 3.3 and 3.4 have reached end of life quite a while ago. This PR also addresses #2949.

I installed Luigi under Python 2.7 and 3.6 and that worked as expected.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,12 +46,11 @@ if sys.platform == 'nt':
 else:
     install_requires.append('python-daemon')
 
-# Tornado >=5 requires updated ssl module so we only allow it for recent enough
-# versions of python (3.4+).
-if sys.version_info[:2] >= (3, 4):
-    install_requires.append('tornado>=4.0,<6')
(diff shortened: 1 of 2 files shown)


honnix:
FYI @stonebig @maresb

drowoseque:
@honnix 
I would suggest dropping support in `tox.ini` and in `luigi/setup_logging.py` as well
btw, isn't `ImportError` handling with `collections` and `collections.abc` related to the issue?

honnix:
@drowoseque Good catch. I thought we have no 33/34 in tox because readme says 36/36 tested. :D

This PR intended to just officially drop the support declaration, but not clean up code base (which we can do in a follow-up PR). But yes this is good point. Thanks.

honnix:
OK. I see we don't run those as part of Travis CI. I now remember the change. :)

Issue #2879, opened 2020-01-07
Unable to configure logging in 2.8.11 using TOML.


hirolau:
In 2.8.11 the TOML parser freezes the config dict, this makes it no longer possible to load logging using that config.

```
Traceback (most recent call last):
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/logging/config.py", line 538, in configure
    formatters[name])
TypeError: 'FrozenOrderedDict' object does not support item assignment

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/app-root/bin/luigid", line 11, in <module>
    load_entry_point('luigi==2.8.11', 'console_scripts', 'luigid')()
  File "/opt/app-root/lib/python3.6/site-packages/luigi/cmdline.py", line 31, in luigid
    DaemonLogging.setup(opts)
  File "/opt/app-root/lib/python3.6/site-packages/luigi/setup_logging.py", line 83, in setup
    configured = cls._section(opts)
  File "/opt/app-root/lib/python3.6/site-packages/luigi/setup_logging.py", line 48, in _section
    logging.config.dictConfig(logging_config)
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/logging/config.py", line 795, in dictConfig
    dictConfigClass(config).configure()
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/logging/config.py", line 541, in configure
    'formatter %r: %s' % (name, e))
ValueError: Unable to configure formatter 'simple': 'FrozenOrderedDict' object does not support item assignment
```

I have not had time to look into possible solutions for now.

hirolau:
One way to fix this would be to not freeze the logging section in the parser:

In  luigi/luigi/configuration/toml_parser.py instead of:

```Python
        # freeze dict params
        for section, content in self.data.items():
            for key, value in content.items():
                if isinstance(value, dict):
                    self.data[section][key] = recursively_freeze(value)
```
It would be possible to do:

```Python
        # freeze dict params
        for section, content in self.data.items():
            if key.lower() == 'logging':
                 continue
            for key, value in content.items():
                if isinstance(value, dict):
                    self.data[section][key] = recursively_freeze(value)
```

DVlahovic:
Also experiencing this issue when attempting to configure logging via toml

mzpqnxow:
[...]

PR #2984, opened 2020-08-14
Fixing TOML logging configuration

lphk92:
## Description
Using the `logging` section of a TOML configuration file is currently not working, because the `logging` module attempts to modify the given configuration dictionary in-place, and all parsed TOML configuration sections are converted to the immutable `FrozenOrderedDict` after being read.

This change "thaws" the `FrozenOrderedDict` before passing it into the `logging` module.

## Motivation and Context
I was trying to use a TOML configuration file for logging, as shown in https://github.com/spotify/luigi/blob/master/examples/config.toml
however, trying to use it resulted in the following error

```
  File "/Users/kushnerl/miniconda3/lib/python3.7/logging/config.py", line 543, in configure                                                                                                                                                                   
    formatters[name])                                                                                                                                                                                                                                         
TypeError: 'FrozenOrderedDict' object does not support item assignment                                                                                                                                                                                        
                                                                                                                                                                                                                                                              
The above exception was the direct cause of the following exception:                                                                                                                                                                                          
                                                                                                                                                                                                                                                              
 ...                                                                                                                                                                                                               File "/Users/kushnerl/miniconda3/lib/python3.7/site-packages/luigi/interface.py", line 237, in build                         
    luigi_run_result = _schedule_and_run(tasks, worker_scheduler_factory, override_defaults=env_params)                        
  File "/Users/kushnerl/miniconda3/lib/python3.7/site-packages/luigi/interface.py", line 145, in _schedule_and_run             
    InterfaceLogging.setup(env_params)                                                                                                                                                                                                                        
  File "/Users/kushnerl/miniconda3/lib/python3.7/site-packages/luigi/setup_logging.py", line 78, in setup                                                                                                                                                     
    configured = cls._section(opts)                                                                                                                                                                                                                           
  File "/Users/kushnerl/miniconda3/lib/python3.7/site-packages/luigi/setup_logging.py", line 43, in _section                                                                                                                                                  
    logging.config.dictConfig(logging_config)                                                                                                                                                                                                                 
  File "/Users/kushnerl/miniconda3/lib/python3.7/logging/config.py", line 800, in dictConfig                                                                                                                                                                  
    dictConfigClass(config).configure()                                                                                                                                                                                                                       
  File "/Users/kushnerl/miniconda3/lib/python3.7/logging/config.py", line 546, in configure                                    
    'formatter %r' % name) from e                                                                                              
ValueError: Unable to configure formatter 'default' 
```

### DIFF
--- a/luigi/freezing.py
+++ b/luigi/freezing.py
@@ -56,3 +56,14 @@ def recursively_freeze(value):
     elif isinstance(value, list) or isinstance(value, tuple):
         return tuple(recursively_freeze(v) for v in value)
     return value
+
+
+def recursively_thaw(value):
+    """
+    Recursively walks the given ``FrozenOrderedDict``, converting all ``FrozenOrderedDict`` objects into ``dict``.
+    """
+    if isinstance(value, FrozenOrderedDict):
+        value = dict(value)
+    if isinstance(value, dict):
+        return dict(((k, recursively_thaw(v)) for k, v in value.items()))
+    return value
--- a/luigi/setup_logging.py
+++ b/luigi/setup_logging.py
@@ -24,6 +24,7 @@ import logging.config
 import os.path
 
 from luigi.configuration import get_config, LuigiConfigParser
+from luigi.freezing import recursively_thaw
 
 from configparser import NoSectionError
 
@@ -40,6 +41,11 @@ class BaseLogging:
             logging_config = cls.config['logging']
         except (TypeError, KeyError, NoSectionError):
             return False
+
+        # The logging module modifies the logging_config dictionary
+        # in place, so we need to make sure it is not frozen before passing in
+        logging_config = recursively_thaw(logging_config)
+
         logging.config.dictConfig(logging_config)
         return True
 


hirolau:
For convenience here is a link to my initial bug-report of this issue (#2879). This solution is better then the simple one i suggested in the original issue.

I would very much want this merged. Thank you for contributing.

lphk92:
Great catch @hirolau! I just pushed up another commit that fixes the issue.

hirolau:
One more very very small thing. When I read the function the first time to added the line for tuples my mind read the conditions as *elif* instead of *if*. This was because I just assumed it would be a more pure recursive function, and my mind has a habit of assuming things. Thus my first attempt to change the function was a failure and I had to go back and debug. If I were to write this function I would do it like this with a single condition so the mind can be sure only one statement is run per recursion.

I am fine with either implementation, but if you like my take better feel free to change it.

```Python
def recursively_thaw(value):
    """
    Recursively walks the given ``FrozenOrderedDict``, converting all ``FrozenOrderedDict`` objects into ``dict``. 
    Tuples will be converted to lists, even if they were initially tuples when converting to ``FrozenOrderedDict``.
    """
    if isinstance(value, FrozenOrderedDict):
        return recursively_thaw(dict(value))
    
    elif isinstance(value, dict):
        return {key: recursively_thaw(val) for key, val in value.items()}
    
    elif isinstance(value, tuple):
        return [recursively_thaw(item) for item in value]
    
    else:
        return value
```

dlstadther (commented):
changes make sense to me. however, are there any tests which could be added to display the fix to the current issue at hand?

PR #2988, opened 2020-08-26
Add error handler of WrapperTask

hiro-o918:
## Description
- Add error handler for wrapper task named `on_upstream_failure`
- Make interface run `on_upstream_failure` when upstream tasks of a `WrapperTask` are failed

## Motivation and Context
Currently, `luigi.Task`  can be made atomic based on the `on_failure` function.
On the other hand, `luigi.WrapperTask` cannot, because it does not have its core `run` function and must not fail itself.
So luigi cannot run rolling back processes for a set of tasks.

This PR adds failure handler to `luigi.WrapperTask`, so that achieves the above.

I have included unit tests

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -432,13 +432,13 @@ traceback_max_length
   Maximum length for traceback included in error email. Default is 5000.
 
 
-[batch_notifier]
+[batch_email]
 ----------------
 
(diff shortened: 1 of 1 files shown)


dlstadther (approved):
These changes make sense to me

PR #2998, opened 2020-09-17
Use updated uri for gcs batch reqs

jackdingilian:
## Description
The default batch_uri used by BatchHttpRequest is deprecated: https://developers.googleblog.com/2018/03/discontinuing-support-for-json-rpc-and.html

This update uses the gcs specific batch uri in gcs_client and bumps google-api-python-client's lower bound to make sure we use a compatible version.  Currently batch removals using the client are calling the deprecated uri (https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/http.py#L1175).

I have run the gcs_client test which covers batch removal and I also tested end-to-end with an actual luigi job we use this in.

### DIFF
--- a/luigi/contrib/gcs.py
+++ b/luigi/contrib/gcs.py
@@ -60,6 +60,9 @@ EVENTUAL_CONSISTENCY_SLEEP_INTERVAL = 0.1
 # Maximum number of sleeps for eventual consistency.
 EVENTUAL_CONSISTENCY_MAX_SLEEPS = 300
 
+# Uri for batch requests
+GCS_BATCH_URI = 'https://storage.googleapis.com/batch/storage/v1'
+
 
 def _wait_for_consistency(checker):
     """Eventual consistency: wait until GCS reports something is true.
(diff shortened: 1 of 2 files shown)


luigi/contrib/gcs.py
```diff
@@ -231,7 +234,7 @@ def remove(self, path, recursive=True):
                 raise InvalidDeleteException(
                     'Path {} is a directory. Must use recursive delete'.format(path))
 
-            req = http.BatchHttpRequest()
+            req = http.BatchHttpRequest(batch_uri=GCS_BATCH_URI)
```
honnix, line 237: So google api client doesn't set this as default? It's a bit surprising.

jackdingilian: The default still goes to the global batch uri (https://www.googleapis.com/batch) - now there's a different batch endpoint for each api (but BatchHttpRequest is still generic to request type) so you need to set it to the uri for whatever api you are using

honnix (approved):

Issue #2999, opened 2020-09-18
ExternalProgramTask: stream_for_searching_tracking_url is not flushed, causing program output to be lost.


jarno-r:
When `stream_for_searching_tracking_url` is set, `_proc_with_tracking_url_context` is called to scan the output from the external program, but it doesn't properly flush the output file, which causes some or all output to be lost. *This is particularly a problem for tasks deriving from `SparkSubmitTask`, which enables the feature by default.*

As a workaround, setting `stream_for_searching_tracking_url = 'none'` in a task disables the feature.

PR #3000, opened 2020-09-18
Flush stream_for_searching_tracking_url

jarno-r:
A fix for issue [2999](https://github.com/spotify/luigi/issues/2999)

## Description
Flush the stream

## Motivation and Context
 [Issue 2999 - ExternalProgramTask: stream_for_searching_tracking_url is not flushed, causing program output to be lost.](https://github.com/spotify/luigi/issues/2999)

### DIFF
--- a/luigi/contrib/external_program.py
+++ b/luigi/contrib/external_program.py
@@ -192,6 +192,7 @@ class ExternalProgramTask(luigi.Task):
                             self.build_tracking_url(match.group(1))
                         )
                 else:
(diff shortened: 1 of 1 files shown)


dlstadther (approved):
makes sense to me

harpaj:
@dlstadther, do you think this could be merged? It's really a quite annoying regression for debugging Spark tasks.
I can also confirm that the change fixes the problem.

PR #3001, opened 2020-09-19
Add pickle protocol option support for Spark

gunnsoo:
## Description

Using spark module may become not working if python version is different between client and spark cluster. pickle module uses default protocol changed by python version in dump and dumps method. If Luigi task which inherits PySparkTask is picked in client and then it's unpicked in spark cluster and the pickle default protocol is different, pickle.load() causes AttributeError.
Default pickle protocol is defined here https://docs.python.org/3/library/pickle.html#pickle-protocols

This change enables users to specify pickle protocol version in luigi configuration file according to their spark cluster environment.

## Motivation and Context

I was trying to run Sample task inheriting PySparkTask from client(python3.8 is installed) to spark cluster(python3.6 is installed).
```
import luigi
from luigi.contrib.spark import PySparkTask


class Sample(PySparkTask):

    def main(self, sc, *args):
        print(luigi.__version__)


if __name__ == '__main__':
    luigi.run()
```
Then I got the following error.
```
  File "pyspark_runner.py", line 135, in <module>
    _get_runner_class()(*sys.argv[1:]).run()
  File "pyspark_runner.py", line 103, in __init__
    self.job = pickle.load(fd)
AttributeError: Can't get attribute 'Sample' on <module '__main__' from 'pyspark_runner.py'>
```
This can happen if pickle default protocol is different between client and cluster like client using python3 but cluster using python2 or like client 3.7 and cluster 3.8.
In my case I can't change cluster's python version, and think some people is in similar situation like me.


luigi/contrib/spark.py
```diff
@@ -292,6 +292,10 @@ def files(self):
         if self.deploy_mode == "cluster":
             return [self.run_pickle]
 
+    @property
+    def pickle_protocol(self):
+        return configuration.get_config().getint(self.spark_version, "pickle-protocol", pickle.DEFAULT_PROTOCOL)
```
dlstadther, line 297: why is this pulling from `self.spark_version` config section rather than the `spark` config section (`py-packages` appears to pull from a config section called `spark`)

gunnsoo: Thanks for your review. Updated it just like `py-packages`.

dlstadther (approved):
lgtm

dlstadther:

PR #3002, opened 2020-09-24
Relativize unnecessary absolute URL

maresb:
## Description

Remove the leading slash in the redirect to `/static/visualiser/index.html` defined in `luigi.server.RootPathHandler.get`.

## Motivation and Context

This makes Luigi more proxy-friendly, namely so that it can be proxied to have a nonempty base path. (In my case, I use this patch to successfully run Luigi behind [jupyter-server-proxy](https://github.com/jupyterhub/jupyter-server-proxy)).

Yes, I have been running a patched version of Luigi for a while now, and I've never had a problem.

### DIFF
--- a/luigi/server.py
+++ b/luigi/server.py
@@ -271,7 +271,7 @@ class ByParamsHandler(BaseTaskHistoryHandler):
 
 class RootPathHandler(BaseTaskHistoryHandler):
     def get(self):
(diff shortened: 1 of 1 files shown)


honnix:
This is addressed by #2785 that I should have merged a while ago. Thanks for the PR.

PR #3004, opened 2020-10-05
update hive.py

cn-aman:
The missing semicolon prevents to run hive command in local environment.

## Description
The missing semicolon was preventing the hive command task to run in the local environment while checking for partition.

## Motivation and Context
This change is required because due to this people making use of Luigi spends too much time debugging this error and I feel should be fixed in the Luigi repo. it self


honnix:
LGTM. Could you please add a test case to cover the change? Thanks for your contribution!

PR #3008, opened 2020-10-12
upgrade azure-storage and pass token_credential argument (#3005)

### LINKED ISSUE #3005: Pass token_credential to BlockBlobService
In `azureblob.py` the `token_credential` param (optional) is not passed to`BlockBlobService`. 

```
@property
    def connection(self):
        return blockblobservice.BlockBlobService(account_name=self.options.get("account_name"),
                                                 account_key=self.options.get("account_key"),
                                                 sas_token=self.options.get("sas_token"),
                                                 protocol=self.kwargs.get("protocol"),
                                                 connection_string=self.kwargs.get("connection_string"),
                                                 endpoint_suffix=self.kwargs.get("endpoint_suffix"),
                                                 custom_domain=self.kwargs.get("custom_domain"),
                                                 is_emulated=self.kwargs.get("is_emulated") or False)
```
According to the [docs](https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.blockblobservice.blockblobservice?view=azure-python-previous) it's a valid input param

vbarbaresi:
### Motivation and context

azure-storage==0.36 didn't support `token_credential` argument in `BlockBlobService` init.
Upgrade to the new package `azure-storage-blob` and use the last version that still supports `BlockBlobService` (it's deprecated, but that can be upgraded another time)

### Description

Fixes #3005 by passing `token_credential` argument.

I also had to explicitely default the protocol to "https", otherwise the initialization fails when we provide a token_credential argument.

I successfully ran azureblob unit tests

### DIFF
--- a/luigi/contrib/azureblob.py
+++ b/luigi/contrib/azureblob.py
@@ -63,6 +63,7 @@ class AzureBlobClient(FileSystem):
             * `token_credential` - A token credential used to authenticate HTTPS requests. The token value should be updated before its expiration.
         """
         self.options = {"account_name": account_name, "account_key": account_key, "sas_token": sas_token}
+        kwargs["protocol"] = kwargs.get("protocol") or "https"  # Default protocol to https if it's not set
         self.kwargs = kwargs
 
     @property
@@ -74,7 +75,8 @@ class AzureBlobClient(FileSystem):
                                                  connection_string=self.kwargs.get("connection_string"),
                                                  endpoint_suffix=self.kwargs.get("endpoint_suffix"),
                                                  custom_domain=self.kwargs.get("custom_domain"),
-                                                 is_emulated=self.kwargs.get("is_emulated") or False)
+                                                 is_emulated=self.kwargs.get("is_emulated") or False,
+                                                 token_credential=self.kwargs.get("token_credential"))
 
     def upload(self, tmp_path, container, blob, **kwargs):
         logging.debug("Uploading file '{tmp_path}' to container '{container}' and blob '{blob}'".format(
--- a/tox.ini
+++ b/tox.ini
@@ -38,7 +38,8 @@ deps =
     pymongo==3.4.0
     toml<2.0.0
     responses<1.0.0
-    azure-storage<=0.36
+    azure-storage-common==2.1.0
+    azure-storage-blob==2.1.0
     datadog==0.22.0
     prometheus-client==0.5.0
     dropbox: dropbox>=9.3.0<10.0
@@ -124,7 +125,8 @@ deps =
     Sphinx>=1.4.4,<1.5
     sphinx_rtd_theme
     enum34>1.1.0
-    azure-storage<=0.36
+    azure-storage-common==2.1.0
+    azure-storage-blob==2.1.0
     prometheus-client==0.5.0
 commands =
 # build API docs


luigi/contrib/azureblob.py
```diff
@@ -63,6 +63,7 @@ def __init__(self, account_name=None, account_key=None, sas_token=None, **kwargs
             * `token_credential` - A token credential used to authenticate HTTPS requests. The token value should be updated before its expiration.
         """
         self.options = {"account_name": account_name, "account_key": account_key, "sas_token": sas_token}
+        kwargs["protocol"] = kwargs.get("protocol") or "https"  # Default protocol to https if it's not set
```
dlstadther, line 66: if using dict.get, this could just be `kwargs.get('protocol', 'https')`

luigi/contrib/azureblob.py
```diff
@@ -74,7 +75,8 @@ def connection(self):
                                                  connection_string=self.kwargs.get("connection_string"),
                                                  endpoint_suffix=self.kwargs.get("endpoint_suffix"),
                                                  custom_domain=self.kwargs.get("custom_domain"),
-                                                 is_emulated=self.kwargs.get("is_emulated") or False)
+                                                 is_emulated=self.kwargs.get("is_emulated") or False,
```
dlstadther, line 78: same here

dlstadther (commented):
I would have asked for some test cases added, but iirc the Azure stuff was integration tested, not unit tested due to the need to connect to and interact with an actual azure account.

So for now, could you just address the two syntax items below?

luigi/contrib/azureblob.py
vbarbaresi: ✅

luigi/contrib/azureblob.py
vbarbaresi: ✅

dlstadther (approved):

vbarbaresi:
gentle bump, could it be merged?

luigi/contrib/azureblob.py
```diff
@@ -74,7 +75,8 @@ def connection(self):
                                                  connection_string=self.kwargs.get("connection_string"),
                                                  endpoint_suffix=self.kwargs.get("endpoint_suffix"),
                                                  custom_domain=self.kwargs.get("custom_domain"),
-                                                 is_emulated=self.kwargs.get("is_emulated") or False)
+                                                 is_emulated=self.kwargs.get("is_emulated", False),
+                                                 token_credential=self.kwargs.get("token_credential"))
```
dlstadther, line 79: will `azure-storage<=0.36` allow for the passing of `token_credential` without fail? (even if without impact)

vbarbaresi: no, it immediately fails
```
>>> blockblobservice.BlockBlobService(account_name="a", token_credential="a")
TypeError: __init__() got an unexpected keyword argument 'token_credential'
```
do you prefer to keep some compatibility in here in case users aren't able to upgrade their packages?

dlstadther: Oh wait, i see the hard update on az dependencies anyway.

So either we make this change, as is, and include it as an explicit breaking change. Or we explore options for enabling both.

I'm not immediately thinking of an easy way to keep compatibility. 

azure-storage==0.36 is also from 2017...

vbarbaresi: I can refactor these keyword arguments in a dict, and add `token_credential` argument only if we're using the new azure package:

```python
from pkg_resources import parse_version
new_azure_package = parse_version(azure.storage.blob.__version__) >= parse_version("2.1.0")
```

(Under a try/except AttributeError because this attribute didn't exist in the old package, it's `azure.storage.__version__` )

vbarbaresi: I pushed it in another branch https://github.com/spotify/luigi/commit/0566938d6b36283378cbfe1f9c0004699c79f570
I can cherry-pick this here, it's up to you

PR #3019, opened 2020-11-26
Adding configurable BigQuery ProjectID on contrib jobs

MaximilianoFelice:
## Description
Adding a way to configure BigQuery Project ID on both `BigQueryLoadTask` and `BigQueryRunQueryTask`.

They are both added as properties of the class, keeping the same default as in previous versions.

## Motivation and Context
BigQuery jobs can be run on different Project IDs than the ones configured by the output Dataset. This is especially useful for shared BigQuery Datasets, or when you usually need to defer the slot allocation between projects.

I ran a local test using this feature.

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -527,6 +527,12 @@ class BigQueryLoadTask(MixinBigQueryBulkComplete, luigi.Task):
         """	Indicates if BigQuery should allow quoted data sections that contain newline characters in a CSV file. The default value is false."""
         return False
 
+    @property
+    def project_id(self):
+        """Project ID on which to run the BigQuery Job
+        """
+        return self.output().table.project_id
+
     def run(self):
         output = self.output()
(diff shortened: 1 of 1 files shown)


luigi/contrib/bigquery.py
```diff
@@ -527,6 +527,12 @@ def allow_quoted_new_lines(self):
         """	Indicates if BigQuery should allow quoted data sections that contain newline characters in a CSV file. The default value is false."""
         return False
 
+    @property
+    def project_id(self):
+        """Project ID on which to run the BigQuery Job
```
honnix, line 532: Could you please add short description for the default value as well?

luigi/contrib/bigquery.py
```diff
@@ -610,6 +616,12 @@ def use_legacy_sql(self):
         """
         return True
 
+    @property
+    def project_id(self):
+        """Project ID on which to run the BigQuery Job
```
honnix, line 621: Same here.

honnix (commented):
Thanks for the PR. Could you please add test cases if possible?

PR #3022, opened 2020-12-05
Adding retry on _obj_exists

hirosassa:
## Description
Since GCS API sometimes returns 503 when calling `_obj_exists` and the task fails, I added retry at the fucntion.

## Motivation and Context
GCS API sometimes returns 503 error, even though there is nothing wrong, and the task fails.
In my use case, it frequently fails at `_obj_exists` call.
To prevent above problem, I added retry on the function.

P.S.
@lallea gave me a great suggestion https://github.com/spotify/luigi/pull/3022#issuecomment-744037904,
so I added tenacity library on every retries on gcs.py.

I ran test_gcs.py locally.


hirosassa:
feature work: apply _retry on every retry procedure on gcs.py

lallea:
There is a nice library that implements retries as function decorators: https://github.com/rholder/retrying

I have also seen frequent intermittent 5xx from cloud APIs, so it seems like retry functionality would be applicable in many places in Luigi. Perhaps it is worth taking a dependency on the retrying library? If we decide not to, I suggest taking inspiration and converting your code to a function decorator instead.

hirosassa:
@lallea  Thanks for your comment! I agree with your opinion. I'll try to implement using a decorator library, but your suggested one is not in active development. So I'll try https://github.com/jd/tenacity.

hirosassa:
@lallea I implemented decorator retries, and passed tests. Please check.
cc @honnix @Tarrasch @dlstadther

setup.py
```diff
@@ -37,7 +37,7 @@ def get_static_files(path):
 with open('README.rst') as fobj:
     long_description = "\n\n" + readme_note + "\n\n" + fobj.read()
 
-install_requires = ['python-dateutil>=2.7.5,<3']
+install_requires = ['python-dateutil>=2.7.5,<3', 'tenacity>=6.3.0']
```
lallea, line 40: I think tacking on `,<7` would be a wise precaution in case of future incompatible changes in tenacity.

luigi/contrib/gcs.py
```diff
@@ -400,29 +398,11 @@ def download(self, path, chunksize=None, chunk_callback=lambda _: False):
...
-                if error:
-                    attempts += 1
-                    if attempts >= NUM_RETRIES:
-                        raise error
-                else:
-                    attempts = 0
+            try:
+                _, done = downloader.next_chunk()
+            except errors.HttpError as err:
+                if err.resp.status < 500:
```
lallea, line 404: I don't think this check is in the right place. You will silently swallow 5xx errors. IIUC, you'll want to retry on those.

If you remove the `try .. except` wrapper here, and replace `retry_if_exception_type` with a custom function that checks the exception type and status code, I think that we should be fine.

setup.py
hirosassa: Thank you for your suggestion. I added `<7` on here and `tox.ini`.

luigi/contrib/gcs.py
hirosassa: I implemented `custom_retry` function to evaluate exception and status code. Please review.
Thank you for your comment, again!

luigi/contrib/gcs.py
```diff
@@ -64,6 +67,29 @@
 GCS_BATCH_URI = 'https://storage.googleapis.com/batch/storage/v1'
 
 
+# Retry configuration. For more details, see https://tenacity.readthedocs.io/en/latest/
+def custom_retry(retry_state):
+    """custom_retry returns true if the outcome produced the function is retryable,
+    false otherwise.
+    """
+    try:
+        retry_state.outcome.result()
```
lallea, line 76: I haven't used tenacity before, but I think this will return the actual result of the wrapped call, whereas you below return a boolean.

I think it can be expressed easier with retry_if_exception with a custom predicate, e.g.

```
def is_error_5xx(err: Exception):
    return isinstance(err, errors.HttpError) and e.resp >= 500

gcs_retry = retry(retry=retry_if_exception(is_error_5xx),
    ...

```

lallea:
Thanks for persisting with your work. BTW, do notice that I am not a maintainer, so someone else has to take a look as well. I am just helping out, since we are invested in Luigi, and want to use your contribution. :-)

luigi/contrib/gcs.py
hirosassa: I applied this suggestion 8d88eb0. Thanks!

dlstadther (approved):
lgtm

i only wonder if the `retry` should be configurable. That said, since this is new functionality, i'm ok if we say "these are reasonable defaults and can be easily updated to support configuration later".

hirosassa:
@dlstadther 
Thanks for your suggestion!

> these are reasonable defaults and can be easily updated to support configuration later

I completely agree with your opinion, but the implementation will affects a lot of parts in Luigi.
How about a following idea?; at first, merge this PR, and then I will try to implement it in next PR.

honnix:
Thanks for the PR, looks great. Just one thing, is the intention only to address 503 in gcs contrib or we want to and will use this lib in core? If yes, it would be great if we can create a few follow-up issues so anyone can pick up, otherwise adding this new lib breaks the contract between core and contrib.

hirosassa:
@honnix Thank you for your comment!

> we want to and will use this lib in core?

Yes. I think we can improve codes in core using this lib, for example [here](https://github.com/spotify/luigi/blob/cf2abbdb998f9e0d72801b8c8e9b2f8cfc16b7a1/luigi/rpc.py#L152-L177).

(I believe we can use this lib in other parts contrib, too)

So, I'll create follow-up issues.

hirosassa:
@dlstadther @honnix Thank you for your great comments and discussions! I think everything that I should do is done in this PR.
Cloud you please merge this PR?

tox.ini
```diff
@@ -126,6 +127,7 @@ deps =
     enum34>1.1.0
     azure-storage<=0.36
     prometheus-client==0.5.0
+    tenacity>=6.3.0<7
```
honnix, line 130:
```suggestion
    tenacity>=6.3.0,<7
```

tox.ini
```diff
@@ -42,6 +42,7 @@ deps =
     datadog==0.22.0
     prometheus-client==0.5.0
     dropbox: dropbox>=9.3.0<10.0
+    tenacity>=6.3.0<7
```
honnix, line 45:
```suggestion
    tenacity>=6.3.0,<7
```

honnix: Is this change needed? Shouldn't be this pulled in by `install_requires` in `setup.py`?

tox.ini
honnix: Is this change needed? Shouldn't be this pulled in by `install_requires` in `setup.py`?

tox.ini
hirosassa: This is unnecessary. Fixed!

tox.ini
hirosassa: This is unnecessary. Fixed!

hirosassa:
Hi, this PR branch is now using in our production and it works well.
Please let me know if you have any comments. Thanks!

jamescooke:
Thanks for this fix @hirosassa - we've been suffering with the 503 errors on GCS when checking for consistency. The number of errors seems to have recently increased so we're considering running this in production before the next official release.

hirosassa:
@jamescooke Thanks for your comment! I'm really happy to hear that!!!!

Issue #3025, opened 2020-12-21
Retry procedure is naively implemented


hirosassa:
Related to https://github.com/spotify/luigi/pull/3022, there are some naive retry implementations in Luigi.
Here's an example, https://github.com/spotify/luigi/blob/cf2abbdb998f9e0d72801b8c8e9b2f8cfc16b7a1/luigi/rpc.py#L156-L170

I think we can improve these codes using some libraries.
There are several good libraries implement retries as function decorators;
- https://github.com/rholder/retrying
- https://github.com/jd/tenacity

Both are very flexible to configure "retry-related" conditions. Retrying is a good one but it's not in active development currently. Tenacity, a fork of retrying, is in active. 
I think this dependency is worth adding on Luigi to simplify codes.

PR #3026, opened 2020-12-28
replace naive retry with tenacity

hirosassa:
## Description

Related to https://github.com/spotify/luigi/issues/3025

I implemented retry procedure using tenacity library.
https://github.com/jd/tenacity

## Motivation and Context

Related to https://github.com/spotify/luigi/pull/3022 , there are some naive implementations of retries in luigi.
To simplify such codes, I introduce tenacity library and implement retries using function decorator.

I fixed and ran tests locally

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -154,7 +154,7 @@ class RemoteScheduler:
         return Retrying(wait=wait_fixed(self._rpc_retry_wait),
                         stop=stop_after_attempt(self._rpc_retry_attempts),
                         reraise=True,
-                        before=retry_logging)
+                        after=retry_logging)
 
     def _fetch(self, url_suffix, body):
         full_url = _urljoin(self._url, url_suffix)
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -146,17 +146,14 @@ class RemoteScheduler:
 
     def _get_retry_decorator(self):
         def retry_logging(retry_state):
-            logger.warning("Failed connecting to remote scheduler %r", self._url, exc_info=True)
-            logger.info("Retrying attempt %r of %r (max)" % (retry_state.attempt_number, self._rpc_retry_attempts))
+            if self._rpc_log_retries:
+                logger.warning("Failed connecting to remote scheduler %r", self._url, exc_info=True)
+                logger.info("Retrying attempt %r of %r (max)" % (retry_state.attempt_number, self._rpc_retry_attempts))
 
-        if self._rpc_log_retries:
-            return retry(wait=wait_fixed(self._rpc_retry_wait),
-                         stop=stop_after_attempt(self._rpc_retry_attempts),
-                         reraise=True,
-                         after=retry_logging)
         return retry(wait=wait_fixed(self._rpc_retry_wait),
                      stop=stop_after_attempt(self._rpc_retry_attempts),
-                     reraise=True)
+                     reraise=True,
+                     after=retry_logging)
 
     def _fetch(self, url_suffix, body):
         full_url = _urljoin(self._url, url_suffix)
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -159,13 +159,8 @@ class RemoteScheduler:
         full_url = _urljoin(self._url, url_suffix)
         scheduler_retry = self._get_retry_decorator()
 
-        @scheduler_retry
-        def __fetch(full_url, body):
-            response = self._fetcher.fetch(full_url, body, self._connect_timeout)
-            return response
-
         try:
-            response = __fetch(full_url, body)
+            response = scheduler_retry.call(RemoteScheduler._call(self._fetcher, full_url, body, self._connect_timeout))
         except self._fetcher.raises as e:
             raise RPCError(
                 "Errors (%d attempts) when connecting to remote scheduler %r" %
@@ -173,6 +168,10 @@ class RemoteScheduler:
                 e
             )
         return response
+
+    @staticmethod
+    def _call(fetcher, full_url, body, timeout):
+        return fetcher.fetch(full_url, body, timeout)
 
     def _request(self, url, data, attempts=3, allow_null=True):
         body = {'data': json.dumps(data)}
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -23,13 +23,15 @@ import os
 import json
 import logging
 import socket
-import time
 import base64
 
 from urllib.parse import urljoin, urlencode, urlparse
 from urllib.request import urlopen, Request
 from urllib.error import URLError
 
+from tenacity import retry
+from tenacity import wait_fixed
+from tenacity import stop_after_attempt
 from luigi import configuration
 from luigi.scheduler import RPC_METHODS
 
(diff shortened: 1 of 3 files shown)


luigi/rpc.py
```diff
@@ -23,13 +23,15 @@
...
 import logging
 import socket
-import time
 import base64
 
 from urllib.parse import urljoin, urlencode, urlparse
 from urllib.request import urlopen, Request
 from urllib.error import URLError
 
+from tenacity import retry
```
honnix, line 32: We can import all of them in one line.

honnix:
@hirosassa I merged #3022 so I guess you would like to rebase. Is this PR ready for review?

luigi/rpc.py
```diff
@@ -144,35 +146,36 @@ def __init__(self, url='http://localhost:8082/', connect_timeout=None):
...
             self._fetcher = URLLibFetcher()
 
-    def _wait(self):
+    def _get_retry_decorator(self):
+        def retry_logging(retry_state):
+            logger.warning("Failed connecting to remote scheduler %r", self._url, exc_info=True)
+            logger.info("Retrying attempt %r of %r (max)" % (retry_state.attempt_number, self._rpc_retry_attempts))
+
         if self._rpc_log_retries:
-            logger.info("Wait for %d seconds" % self._rpc_retry_wait)
```
honnix, line 149: I think it is better to keep this log. Never know what users would do with it.

luigi/rpc.py
```diff
@@ -144,35 +146,36 @@ def __init__(self, url='http://localhost:8082/', connect_timeout=None):
         else:
             self._fetcher = URLLibFetcher()
 
-    def _wait(self):
+    def _get_retry_decorator(self):
```
honnix, line 149: Could this be simplified as:
```
@retry(wait=wait_fixed(self._rpc_retry_wait),
             stop=stop_after_attempt(self._rpc_retry_attempts),
             reraise=True,
             after=_retry_logging))
```
and we check the flag in `_retry_logging`?

luigi/rpc.py
```diff
@@ -144,35 +146,36 @@ def __init__(self, url='http://localhost:8082/', connect_timeout=None):
...
-                last_exception = e
-                if self._rpc_log_retries:
-                    logger.warning("Failed connecting to remote scheduler %r", self._url,
-                                   exc_info=True)
-                continue
-        else:
+        scheduler_retry = self._get_retry_decorator()
+
+        @scheduler_retry
+        def __fetch(full_url, body):
```
honnix, line 168: Can we avoid creating this inner function every time we can `_fetch`? Maybe just make it a normal static method. Also `__` as prefix doesn't seem to be needed: https://en.wikipedia.org/wiki/Name_mangling#Python

test/rpc_test.py
```diff
@@ -82,9 +74,9 @@ def test_log_rpc_retries_enabled(self, mock_logger):
         self.get_work(fetch_results)
         self.assertEqual([
             mock.call.warning('Failed connecting to remote scheduler %r', 'http://zorg.com', exc_info=True),
-            mock.call.info('Retrying attempt 2 of 3 (max)'),
+            mock.call.info('Retrying attempt 1 of 3 (max)'),
```
honnix, line 77: What caused this different behaviour?

test/worker_test.py
```diff
@@ -1167,8 +1161,8 @@ class A(DummyTask):
         with Worker(scheduler=sch) as worker:
             try:
                 worker.add(a)
-            except RPCError:
-                self.assertEqual(self.waits, 2)  # should attempt to add it 3 times
+            except RPCError as e:
+                self.assertTrue(e.args[0].find("Errors (3 attempts)") != -1)
```
honnix, line 1165: `str(e)` can give you the message.

hirosassa:
@honnix Thank you for your review. I'll rebase and fix it!

luigi/rpc.py
hirosassa: Thanks! fixed in bda80d5

luigi/rpc.py
hirosassa: I think this is good suggestion! fixed in a4d46a0

luigi/rpc.py
hirosassa: @honnix I tried to fix this in 5150504 
Which do you prefer original one or fixed?

test/rpc_test.py
hirosassa: This log shows the number of "retry" attempt, but in original one it shows the number of total try (including 1st attempt, not "retry" attempt). So I fixed.

test/worker_test.py
hirosassa: Thanks! fixed in 715ed32

luigi/rpc.py
hirosassa: added wait log and fixed test in 8f36797

honnix:
Thanks for making all the changes and improvement. Could you please rebase from main branch?

hirosassa:
@honnix Done!

honnix:
This has brought in all commits in main branch, could you please rebase instead?

hirosassa:
@honnix Sorry for my misunderstanding. Fixed

test/rpc_test.py
honnix: I see. With this change, `Retrying attempt n of 3 (max)` would be wrong because `n` has already been tried. If we still prefer log being issued after an attempt, the original counting looks more accurate.

hirosassa: @honnix Thanks for your comment! Make sense!
I fixed this. Please check.

luigi/rpc.py
```diff
@@ -144,35 +144,29 @@ def __init__(self, url='http://localhost:8082/', connect_timeout=None):
...
 
-    def _wait(self):
-        if self._rpc_log_retries:
-            logger.info("Wait for %d seconds" % self._rpc_retry_wait)
-        time.sleep(self._rpc_retry_wait)
+    def _get_retryer(self):
+        def retry_logging(retry_state):
+            if self._rpc_log_retries:
+                logger.warning("Failed connecting to remote scheduler %r", self._url, exc_info=True)
+                logger.info("Retrying attempt %r of %r (max)" % (retry_state.attempt_number+1, self._rpc_retry_attempts))
```
honnix, line 151:
```suggestion
                logger.info("Retrying attempt %r of %r (max)" % (retry_state.attempt_number + 1, self._rpc_retry_attempts))
```
nit

hirosassa: applied. Thanks!

honnix (approved):
Thanks for the PR. Looks great!

honnix:
@Tarrasch @dlstadther Would you like to take a look at this as well? Thanks.

luigi/rpc.py
```diff
@@ -144,35 +144,29 @@ def __init__(self, url='http://localhost:8082/', connect_timeout=None):
...
-                last_exception = e
-                if self._rpc_log_retries:
-                    logger.warning("Failed connecting to remote scheduler %r", self._url,
-                                   exc_info=True)
-                continue
-        else:
+        scheduler_retry = self._get_retryer()
+
+        try:
+            response = scheduler_retry(self._fetcher.fetch, full_url, body, self._connect_timeout)
```
dlstadther, line 164: for some reason i'm struggling to see where it's defined that `full_uri`, `body`, and `self._connect_timeout` are being passed to `self._fetcher.fetch()` here.

hirosassa: Thank you for your comment!
Here is a definition of `Retrying`, first argument of this class' `__call__` is function, and followings are its arguments.
https://github.com/jd/tenacity/blob/3e2244535ccfbb6a4b7fdd77bfc9aa61a1302302/tenacity/__init__.py#L422-L442

dlstadther: Ah, i see. A bit obfuscated, but it is being used as tenacity expects

dlstadther (approved):

hirosassa:
Thanks!

lallea:
Thanks @hirosassa for persisting with this tangent that I set you off on, and for being so responsive to all suggestions. :+1:

hirosassa:
Thank you for your great suggestion, too @lallea !

PR #3027, opened 2020-12-31
Fix string and byte error in contrib/hadoop.py

rudaoshi:
## Description
The  code  d = d.replace(b'(c__main__', "(c" + module_name)   is a mixed usage of str and byte and cannot run in python3. 

It is not related to the business logic, just a language issue. 

Yes,

### DIFF
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -971,7 +971,7 @@ class JobTask(BaseHadoopJobTask):
             if self.__module__ == '__main__':
                 d = pickle.dumps(self)
                 module_name = os.path.basename(sys.argv[0]).rsplit('.', 1)[0]
-                d = d.replace(b'(c__main__', "(c" + module_name)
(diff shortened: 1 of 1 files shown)


luigi/contrib/hadoop.py
```diff
@@ -971,7 +971,7 @@ def dump(self, directory=''):
             if self.__module__ == '__main__':
                 d = pickle.dumps(self)
                 module_name = os.path.basename(sys.argv[0]).rsplit('.', 1)[0]
-                d = d.replace(b'(c__main__', "(c" + module_name)
+                d = d.replace(b'c__main__', b'c' + module_name.encode('ascii'))
```
honnix, line 974: I think it's better we keep `(` and encode with default encoding (utf-8).

rudaoshi: I've made the revision according to your suggestion. New submission is committed. 

Thank you

PR #3032, opened 2021-01-16
Fix azureblob tests: use json instead of numpy

vbarbaresi:
Numpy is not installed in the current setup.
I replaced Numpy data by a simple JSON structure

It fixes these errors (https://travis-ci.org/github/spotify/luigi/jobs/754449469):

```
  File "/home/travis/build/spotify/luigi/test/contrib/azureblob_test.py", line 138, in run

    import numpy

ModuleNotFoundError: No module named 'numpy'
```

The test silently failed, asserts didn't run because the task was skipped
I added an assert checking the output of _FinalTask_ to make sure it runs

Tested using Azurite and using a real Azure storage account

### DIFF
--- a/test/contrib/azureblob_test.py
+++ b/test/contrib/azureblob_test.py
@@ -20,6 +20,7 @@ Integration tests for azureblob module.
 
 import os
 import unittest
+import json
 
 from nose.plugins.attrib import attr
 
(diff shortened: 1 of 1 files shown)


test/contrib/azureblob_test.py
```diff
@@ -169,4 +166,7 @@ def tearDown(self):
         pass
 
     def test_AzureBlobTarget(self):
-        luigi.build([FinalTask()], local_scheduler=True, log_level='NOTSET')
+        final_task = FinalTask()
+        luigi.build([final_task], local_scheduler=True, log_level='NOTSET')
+        output = final_task.output().open("r").read()
+        assert "Toto" in output
```
honnix, line 172: :+1:

honnix (approved):

PR #3033, opened 2021-01-17
Add a UTC date to the marker document and venv/ to the .gitignore

ptindall:
## Description
Added an additional field to the marker entry for the ElasticSearch target so we also have the UTC based date.

I also use PyCharm which sets up the local virtual as `venv/` so I wanted to be sure that was ignored as part of .gitignore

## Motivation and Context
We use UTC dates to control incremental processes and knowing when this marker was last updated based upon a UTC
date is needed for us.

I have run "tox -e py39-core test/contrib/esindex_test.py" locally on this change.

### DIFF
--- a/luigi/contrib/esindex.py
+++ b/luigi/contrib/esindex.py
@@ -169,7 +169,9 @@ class ElasticsearchTarget(luigi.Target):
 
         The document id would be sufficient but,
         for documentation,
-        we index the parameters `update_id`, `target_index`, `target_doc_type` and `date` as well.
+        we index the parameters `update_id`, `target_index`, `target_doc_type` and `date` as well.  `date_utc` added
+        so we can be sure to get the actual date and time based upon UTC and not the client date and time based on the
+        client machine.
         """
         self.create_marker_index()
         self.es.index(index=self.marker_index, doc_type=self.marker_doc_type,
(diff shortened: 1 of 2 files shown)


luigi/contrib/esindex.py
```diff
@@ -169,15 +169,18 @@ def touch(self):
...
         """
         self.create_marker_index()
         self.es.index(index=self.marker_index, doc_type=self.marker_doc_type,
                       id=self.marker_index_document_id(), body={
                           'update_id': self.update_id,
                           'target_index': self.index,
                           'target_doc_type': self.doc_type,
-                          'date': datetime.datetime.now()})
+                          'date': datetime.datetime.now(),
+                          'date_utc': datetime.datetime.utcnow()})
```
honnix, line 183: Not an expert of ES, so not sure of the impact here.

luigi/contrib/esindex.py
```diff
@@ -169,15 +169,18 @@ def touch(self):
 
         The document id would be sufficient but,
         for documentation,
-        we index the parameters `update_id`, `target_index`, `target_doc_type` and `date` as well.
+        we index the parameters `update_id`, `target_index`, `target_doc_type` and `date` as well.  `date_utc` added
```
honnix, line 172: I think the added comment fits better as a PR comment instead of docstring to track the reasoning. Maybe something as simple as: we index the parameters `update_id`, `target_index`, `target_doc_type` , `date` and `date_utc`.

Issue #2212, opened 2017-08-18
securing the scheduler


mattvonrocketstein:
I'm not sure, but it seems like workers communicate with schedulers via the same HTTP port as an admin would use to check out the task visualizer.  Is there any authentication involved in this process?  Can I front the scheduler with nginx to add SSL or basic auth, without disrupting communication with workers?

Tarrasch:
There's two issues here right? I think the HTTP and not a secure protocol issue is already fixed. You can use unix sockets somehow (I've not done it myself).

The second issue of having authentication, different users seeing different set of tasks, only admins being able to turn of the scheduler etc is not implemented tough. I don't find it realistic to be done anytime soon. :)

dckc:
My first thought was to add a big notice near this in the docs...

> By default, the server starts on AF_INET and AF_INET6 port 8082 (which can be changed with the --port flag) and listens on all IPs.

But then what would the notice say? I can't think of anything that doesn't come down to "The default configuration for luigid is insecure" which seems more like a bug to be fixed than a feature to be documented.

How about listening only on the localhost interface by default?

mattvonrocketstein:
> The second issue of having authentication, different users seeing different set of tasks, only admins being able to turn of the scheduler etc

I don't think anything like a full blown user model is necessary (or even useful), maybe just a simple shared secret between the workers/scheduler?

> "The default configuration for luigid is insecure" which seems more like a bug to be fixed than a feature to be documented.

IMHO, not necessarily a bug as long it's mentioned up front that the only recommended deployment strategy is into a VPC.

> How about listening only on the localhost interface by default?

Definitely more secure, but also less useful by default, so it doesn't seem very friendly.  Obviously this would prevent remote workers from talking to the scheduler unless we're going to get into SSH tunnels and such to allow this, which would also complicate deployment best practice and call for some extra documentation

dckc:
Ok, that's a useful notice for the docs:
> the only recommended deployment strategy is into a VPC.

PR #3054, opened 2021-03-23
Update central_scheduler.rst

justinsousa:
## Description
We operate in a datacenter that was complaining about luigi running without authentication (called out in https://github.com/spotify/luigi/issues/2212) and listening for traffic on external interfaces (IP address). To get around this I passed `--address 127.0.0.1` to the luigi call in the systemd service file to allow the webserver to listen only on localhost. I noticed this flag wasn't documented so I figured I'd add it to this file so that others see it mentioned alongside the port flag. Someone else at my company noticed this issue cited above and didn't realize that the `--address` flag existed. 

## Motivation and Context
Basically, this is just to bring into the docs what I found by just running `luigi --help`. 

docs only update

### DIFF
--- a/doc/central_scheduler.rst
+++ b/doc/central_scheduler.rst
@@ -31,8 +31,8 @@ To run the server as a daemon run:
 
 Note that this requires ``python-daemon``.
 By default, the server starts on AF_INET and AF_INET6 port ``8082``
-(which can be changed with the ``--port`` flag) and listens on all IPs.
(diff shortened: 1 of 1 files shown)


dlstadther (approved):
Thanks for contributing back and making Luigi's documentation more complete!

dlstadther:
Test failures are unrelated. Merging...

PR #3057, opened 2021-03-31
Keep orm session open during unit test

narape:
## Description
Create explicit orm session in unit test to control the scope/live cycle of the session in unit test for `DbTaskHistory`

## Motivation and Context
Unit test were failing because Task were not bound to a session as sessions were created inside of `DbTaskHistory`

* Changes were only to unit test, no production code was touched


honnix (approved):
LGTM if it works.

Issue #3016, opened 2020-11-06
Migrate CI Away from travis-ci.org


dlstadther:
# tl;dr;
The Luigi project needs to migrate off of Travis-CI.org before the end of the year. Where should we go? Travis-CI.com? Github Actions? Other suggestions?

---
# Background
Currently (as of 2020-11-06), Luigi uses travis-ci.org for builds. They have announced transition to travis-ci.com awhile back and recently set a read-only date for travis-ci.org for December 31 2020 [here](https://docs.travis-ci.com/user/migrate/open-source-repository-migration/#frequently-asked-questions).

Additionally, they have altered their billing model (of which i haven't read all the details yet) [here](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing).

The Luigi project will need to migrate their CI away from travis-ci.org before the end of the year. I'm happy to help with this process, but would like to get feedback (especially from @honnix and other Spotify folks) regarding _where_ to integrate.

My immediate thought is Github Actions given that is a thing now and has seemed to be reasonable popular.

(For those interested, I found out about this from [this Jeff Gerling blog post](https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works))

lallea:
Gitlab CI has a unique property that might be interesting for Luigi - it allows BYO local workers/runners to be coordinated by a hosted CI/CD pipeline. For most tests, we would want to use shared hosted runners, but supporting local runners would enable companies to contribute runners that run in-situ tests e.g. for validating cloud service integrations, and have the results integrated in the CI pipeline report.

Gitlab CI can be used with a Github repository. I have had a pleasant experience of Gitlab CI so far, but I haven't used it for OSS.

dlstadther:
Tagging some Spotify folks who have committed / released (sorry if i miss some people)
@honnix @ckiosidis @jesperps @NatashaL @narape @zatine

dlstadther:
Looks like Spotify's [newest open source project](https://github.com/spotify/klio) is using Github Actions. For that reason, i'll also tag @econchick to see if there is Spotify push for their open source projects to use Github Actions, or any option is fair game.

econchick:
hey @dlstadther - it's the maintainers' choice. If you'd like to move to GH actions, go for it. (I, for one, am enjoying GH actions FWIW). As spotify GH owner, feel free to ping me if you need help moving forward w/r/t permissions on GH or travis.

[...]

PR #3062, opened 2021-04-08
add postgres service and build matrix

ckiosidis:
## Description

In order to run the postgres tests, we need to create a postgres service and a database.

Added a postgres service that gets created on *every* build unfortunately.
Then on the specific postgres builds I added a step to create the database.

```


test/contrib/postgres_test.py::DailyCopyToTableTest::test_bulk_complete PASSED [ 16%]
test/contrib/postgres_test.py::PostgresQueryTest::test_bulk_complete PASSED [ 33%]
test/contrib/postgres_test.py::PostgresQueryTest::test_override_port PASSED [ 50%]
test/contrib/postgres_test.py::PostgresQueryTest::test_port_encoded_in_host PASSED [ 66%]
test/contrib/postgres_test.py::TestCopyToTableWithMetaColumns::test_copy_with_metadata_columns_disabled PASSED [ 83%]
test/contrib/postgres_test.py::TestCopyToTableWithMetaColumns::test_copy_with_metadata_columns_enabled PASSED **[100%]

```

## Motivation and Context

Moving away from Travis CI https://github.com/spotify/luigi/issues/3016

### DIFF
--- a/.github/workflows/pythonbuild.yml
+++ b/.github/workflows/pythonbuild.yml
@@ -9,6 +9,22 @@ on:
 jobs:
   build:
     runs-on: ubuntu-16.04
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: postgres
+        ports:
+        - 5432:5432
+        # Set health checks to wait until postgres has started
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
     strategy:
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


.github/workflows/pythonbuild.yml
```diff
@@ -65,6 +91,10 @@ jobs:
           mysql -e 'create database IF NOT EXISTS luigi_test;' -uroot -proot || true
           mysql -e 'create user 'travis'@'localhost';' -uroot -proot || true
           mysql -e 'grant all privileges ON *.* TO 'travis'@'localhost';' -uroot -proot || true
+      - name: Create PSQL database
```
narape, line 94: Can you change the name of the step in line 88? it says `Setup DB` but now we have another DB, maybe we can call it Setup MySQL DB

narape (approved):

.github/workflows/pythonbuild.yml
```diff
@@ -65,6 +91,10 @@ jobs:
...
+        if: ${{ contains(matrix.tox-env, 'postgres') }} 
+        run: |
+          PGPASSWORD=postgres psql -h localhost -p 5432 -c 'create database spotify;' -U postgres
```
ckiosidis, line 97: https://github.com/actions/example-services/blob/5ba52cd792abf85235f6f446e56ea428519c60ec/.github/workflows/postgres-service.yml#L15-L18

Since there is no container involved in the builds, postgres is available through localhost

PR #3069, opened 2021-04-09
Codecov ga

ckiosidis:
## Description

Moved codecov from tox to github action step.
There has been some flakiness and not sure if it's expected to work like that.
It seems that when the different environments register the coverage, it might prematurely throw an error on the step and we need to re-run it

Added the notify flag in order for codecov to wait for all the CI builds to finish before reporting the status.

```
  notify:
    wait_for_ci: yes
```


tnsetting (dismissed):
Let's merge it for now and figure out how to fix the coverage in a later pr

ckiosidis:
Merge this so we can make codecov required and we can fix the small discrepancy moving on

tnsetting (approved):

codecov.yml
```diff
@@ -1,3 +1,9 @@
+codecov:
+  require_ci_to_pass: yes
+  notify:
+    after_n_builds: 24
+    wait_for_ci: yes
```
ckiosidis, line 5: this didn't work as expected. But at least codecov will update the status multiple times as more coverages are being reported

PR #3072, opened 2021-04-15
postgres.CopyToTable: Robustize resources cleanup

LinqLover:
This PR aims to make connections using the `contrib.postgres` module more robust. If an error occurs during copying the rows, psycopg2 objects previously could be left opened without a proper clean-up, causing further connection attempts to the database to fail as well (`psycopg2.errors.ObjectInUse`).

By ensuring proper use of try...catch blocks and context managers, issues of this kind should be fixed now. Also, this PR includes regression tests.

For more information about releasing psycopg2 objects, visit: https://www.psycopg.org/docs/usage.html#with-statement

LBNL, a more precise exception class is used in the default `copy()` implementation of `CopyToTable` (f66534c6ae018d0cfddb7d09aa50cd1fcf6d36d1).

Kind of replaces #2934, which I unfortunately never had managed to complete.

Preferably please squash when merging.


test/contrib/postgres_test.py
```diff
@@ -82,6 +117,44 @@ def test_bulk_complete(self, mock_connect):
...
+        def get_mock_cursor():
+            original_cursor()
+            return mock_cursor
+
+        mock_connection.cursor = mock.MagicMock(side_effect=get_mock_cursor)
+
+        task = DummyPostgresImporter(date=datetime.datetime(2021, 4, 15))
+        task.columns = [(42,)]  # inject defect
+
+        #import pdb; pdb.set_trace()
```
LinqLover, line 146:
```suggestion
```

Sorry for that.

luigi/contrib/postgres.py
```diff
@@ -299,52 +299,54 @@ def run(self):
...
-
-        # mark as complete in same transaction
-        self.output().touch(connection)
-
-        # commit and clean up
-        connection.commit()
-        connection.close()
-        tmp_file.close()
+        try:
+            connection = self.output().connect()
```
lallea, line 303: This is quite far away from the finally statement. I suggest `with contextlib.closing(self.output().connect()) as connection:`

luigi/contrib/postgres.py
```diff
@@ -299,52 +299,54 @@ def run(self):
...
+            # transform all data generated by rows() using map_column and write data
+            # to a temporary file for import using postgres COPY
+            tmp_dir = luigi.configuration.get_config().get('postgres', 'local-tmp-dir', None)
+            tmp_file = tempfile.TemporaryFile(dir=tmp_dir)
```
lallea, line 307: Likewise here: `with tempfile.TemporaryFile(dir=tmp_dir) as tmp_file:`

Note: I am not a maintainer, just a community member with random opinions and happy to occasionally help with code reviews. :-)

luigi/contrib/postgres.py
LinqLover: Great tip! I wasn't aware of this utility. Done :)

luigi/contrib/postgres.py
LinqLover: Done 👍

PR #3073, opened 2021-04-15
Version 3.0.3

sonjaer:

### DIFF
--- a/luigi/__meta__.py
+++ b/luigi/__meta__.py
@@ -7,5 +7,5 @@ Workflow mgmgt + task scheduling + dependency resolution.
 __author__ = 'The Luigi Authors'
 __contact__ = 'https://github.com/spotify/luigi'
 __license__ = 'Apache License 2.0'
-__version__ = '3.0.2'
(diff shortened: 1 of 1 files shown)


ckiosidis (approved):

lallea:
IMHO, we should bump minor version number for all changes that are not pure bug fixes, since there is a risk of breaking things for users. I have experienced Luigi patch releases break our code multiple times, and people tend to get surprised thinking that a patch release should be low risk.

In this case, there is a new dependency added, which may very well break container builds or pip install scripts for downstream users.

sonjaer:
> IMHO, we should bump minor version number for all changes that are not pure bug fixes, since there is a risk of breaking things for users. I have experienced Luigi patch releases break our code multiple times, and people tend to get surprised thinking that a patch release should be low risk.
> 
> In this case, there is a new dependency added, which may very well break container builds or pip install scripts for downstream users.

Makes sense, will try and do this in the future. Added something to the release process about it https://github.com/spotify/luigi/pull/3074

PR #3077, opened 2021-04-22
Add support for naming tasks in @requires

IanCal:
## Description

This change lets a user name the tasks in the requires method like so:

    @requires(daily=DailySummaryTask, weekly=WeeklySummaryTask)
    class SummaryTask(Task):
        def run(self):
            with self.input()["daily"].open() as daily_results:
                ...

## Motivation and Context

When we use the requires decorator, we can save a lot of time and hassle passing parameters around. One issue we have hit is that the run method only sees a list of targets, with no context as to which earlier task generated the results.

Luigi already supports named inputs, by returning a dictionary from the requires function. This change just allows passing in kwargs to the decorator and creates a requires method using those. I added this to the inherits decorator as well, 

There wasn't a sensible option in my mind if you try and pass in both, so I made it throw an error in that case.

This doesn't include an update to the documentation but I can add some docs if the approach is OK.

This doesn't handle any fancier nested versions, just a simple dictionary.

I added some basic unit tests, but can expand these, and have put an example program below.

It's largely taken from demo code we're working on, and is I hope a relatively simple change.

Example:

    from luigi.util import requires
    from luigi import Task, Parameter, build
    from luigi.local_target import LocalTarget

    class DependencyA(Task):
        def run(self):
            with self.output().open("w") as f_out:
                f_out.write("DependencyA file output")

        def output(self):
            return LocalTarget(f"output/{self.get_task_family()}")

    class DependencyB(Task):
        def run(self):
            with self.output().open("w") as f_out:
                f_out.write("DependencyB file output")

        def output(self):
            return LocalTarget(f"output/{self.get_task_family()}")

    @requires(dep_a=DependencyA, dep_b=DependencyB)
    class ExampleTask(Task):
        def run(self):
            dep_a_data = self.input()["dep_a"].open().read()
            dep_b_data = self.input()["dep_b"].open().read()
            print("Data", dep_a_data, dep_b_data)

    build([
        ExampleTask()
    ], local_scheduler=True)

### DIFF
--- a/luigi/util.py
+++ b/luigi/util.py
@@ -279,16 +279,19 @@ class inherits:
                # ...
     """
 
-    def __init__(self, *tasks_to_inherit):
+    def __init__(self, *tasks_to_inherit, **kw_tasks_to_inherit):
         super(inherits, self).__init__()
-        if not tasks_to_inherit:
-            raise TypeError("tasks_to_inherit cannot be empty")
-
+        if not tasks_to_inherit and not kw_tasks_to_inherit:
+            raise TypeError("tasks_to_inherit or kw_tasks_to_inherit must contain at least one task")
+        if tasks_to_inherit and kw_tasks_to_inherit:
+            raise TypeError("Only one of tasks_to_inherit or kw_tasks_to_inherit may be present")
         self.tasks_to_inherit = tasks_to_inherit
+        self.kw_tasks_to_inherit = kw_tasks_to_inherit
 
     def __call__(self, task_that_inherits):
         # Get all parameter objects from each of the underlying tasks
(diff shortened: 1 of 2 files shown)


luigi/util.py
```diff
@@ -318,15 +332,14 @@ class requires:
 
     """
 
-    def __init__(self, *tasks_to_require):
+    def __init__(self, *tasks_to_require, **kw_tasks_to_require):
         super(requires, self).__init__()
-        if not tasks_to_require:
-            raise TypeError("tasks_to_require cannot be empty")
```
IanCal, line 324: This error isn't required as inherits will be called, and inherits throws the same errors.

dlstadther (approved):

PR #3079, opened 2021-05-05
Improve optional parameters

adrien-berchet:
## Description
Update OptionalParameter and add several other parameter classes for other types.

## Motivation and Context
Parameters raise warnings when their default value is None. As far as I know, it is only possible to use OptionalParameter for string type so I added optional classes for other types.

I have included unit tests.

### DIFF
--- a/luigi/__init__.py
+++ b/luigi/__init__.py
@@ -38,7 +38,10 @@ from luigi.parameter import (
     DateIntervalParameter, TimeDeltaParameter,
     IntParameter, FloatParameter, BoolParameter,
     TaskParameter, EnumParameter, DictParameter, ListParameter, TupleParameter,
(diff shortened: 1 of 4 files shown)


adrien-berchet:
Do you think that the fail in CI is due to this PR? I don't think so but I am not sure.

dlstadther:
> Do you think that the fail in CI is due to this PR? I don't think so but I am not sure.

The errors listed don't look related to these changes. @dataex do you have insights? This was the first time i've looked at PR-specific unittests for Luigi since the move to Github Actions. I assume `master` is passing?

adrien-berchet:
Hello
I come back to this since it seems the CI is working now. Can you try to trigger it please?

dlstadther (approved):

PR #3081, opened 2021-05-18
Make task.disable_window be default source of window int

### LINKED ISSUE #3029: Throwing warning for deprecated `disable_window_seconds`
Currently the `disable_window` function https://github.com/spotify/luigi/blob/92fe69e022989ebac6a5197fb8396042851e0497/luigi/task.py#L216 returns a reference to `disable_window_seconds` https://github.com/spotify/luigi/blob/92fe69e022989ebac6a5197fb8396042851e0497/luigi/task.py#L224

This function throws a warning which states that `disable_window_seconds` is deprecated stating that `disable_window` should be used.

This does not make much sense and causes errors when you treat warnings as errors :) 
Could we either remove the warning or actually implement the `disable_window` properly?

jamescooke:
## Description

When accessing `task.disable_window` property, use the property directly. Do not call the deprecated `task.disable_window_seconds`.

## Motivation and Context

Fixes #3029 

Current implementation gives a deprecation warning even when accessing the correct `task.disable_window` property, which is confusing. Fixing this means that the deprecation warning becomes more meaningful - it will only appear when `task.disable_window_seconds` is incorrectly accessed.

Have ran this change locally for me in my employers' Luigi test suite and removed 21 warnings when running a small test on a single task. However, that test suite does not use `disable_window`.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -218,12 +218,12 @@ class Task(metaclass=Register):
         Override this positive integer to have different ``disable_window`` at task level.
         Check :ref:`scheduler-config`
         """
-        return self.disable_window_seconds
+        return None
 
     @property
     def disable_window_seconds(self):
         warnings.warn("Use of `disable_window_seconds` has been deprecated, use `disable_window` instead", DeprecationWarning)
-        return None
+        return self.disable_window
 
     @property
     def owner_email(self):
--- a/test/task_test.py
+++ b/test/task_test.py
@@ -152,6 +152,15 @@ class TaskTest(unittest.TestCase):
         with self.assertWarnsRegex(UserWarning, 'Parameter "timedelta_param" with value ".*" is not of type timedelta.'):
             DummyTask(**params)
 
+    def test_disable_window_seconds(self):
+        """
+        Deprecated disable_window_seconds param uses disable_window value
+        """
+        class ATask(luigi.Task):
+            disable_window = 17
+        task = ATask()
+        self.assertEqual(task.disable_window_seconds, 17)
+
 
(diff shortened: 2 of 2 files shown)


jamescooke:
Please could one of @dlstadther , @Tarrasch or another maintainer approve the workflow so that tests run?

dlstadther (dismissed):
I agree this is what is should have been all along.

luigi/task.py
```diff
@@ -218,12 +218,12 @@ def disable_window(self):
...
         Check :ref:`scheduler-config`
         """
-        return self.disable_window_seconds
+        return None
 
     @property
     def disable_window_seconds(self):
         warnings.warn("Use of `disable_window_seconds` has been deprecated, use `disable_window` instead", DeprecationWarning)
-        return None
+        return self.disable_window
```
jamescooke, line 226: Does this line need a test added to get Codecov to go green? Happy to add it if required.

tnsetting: I think this is what `Codecov` is complains about. If you can fix it, it would awesome.

dlstadther: which skim looks like only the scheduler's `disable_window` param is unittested. I don't immediately see any of the Task's. Having test coverage would be great if there's valuable logic that be added to verify. I won't block approval and merge without it. Let me know what you want to do.

jamescooke: Have added a very basic test on `disable_window_seconds` which (hopefully) will give the required coverage 🤞🏻

dlstadther (approved):
lgtm

Issue #3020, opened 2020-11-27
AttributeError: 'LocalTarget' object has no attribute 'is_tmp'


joseph-luketelo:
I am using `LocalTarget` as my `FINALTARGET_BACKEND_CLASS` in .env file.
But when I run my model(s) I get the above-mentioned attribute error. 
When I use s3 as my `FINALTARGET_BACKEND_CLASS` all works fine.

```
File "/opt/conda/lib/python3.7/site-packages/luigi/local_target.py", line 189, in __del__
    if self.is_tmp and self.exists():

```
Above is the line of code that throws this error.

Regards

giusidp:
I have a similar error, but in my case it's:
AttributeError: 'LocalTarget' object has no attribute 'touch'

This is the rest of the error description:   
File "C:\Users\20225500\AppData\Local\Programs\Python\Python311\Lib\site-packages\luigi\worker.py", line 203, in run
    new_deps = self._run_get_new_deps()
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\myname\AppData\Local\Programs\Python\Python311\Lib\site-packages\luigi\worker.py", line 138, in _run_get_new_deps
    task_gen = self.task.run()
               ^^^^^^^^^^^^^^^
  File "C:\Users\myname\mypath\myscript.py", line 23, in run
    self.output().touch()

PR #3085, opened 2021-06-14
Destructor of LocalTarget sometimes fails with AttributeError: 'LocalTarget' object has no attribute 'is_tmp'

umartin:
## Description
This PR makes the destructor of LocalTarget more robust in case of partially initialised objects. 

## Motivation and Context
LocalTarget might not be fully initialised if an exception is thrown in the constructor of LocalTarget or a subclass.
Python will call the destructor anyway which will fail since is_tmp is not initialized.
Initialising is_tmp earlier in the constructor solves the problem in case of exceptions in LocalTarget but not for subclasses.
Possibly related to https://github.com/spotify/luigi/issues/3020

Unit test added.

### DIFF
--- a/luigi/local_target.py
+++ b/luigi/local_target.py
@@ -186,5 +186,5 @@ class LocalTarget(FileSystemTarget):
         return self.path
 
     def __del__(self):
-        if self.is_tmp and self.exists():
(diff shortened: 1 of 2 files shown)


dlstadther (approved):

dlstadther:

PR #3086, opened 2021-06-15
Truncate traceback for failure email

DoriRunyon:
## Description
Add traceback_max_length config option which truncates the traceback included in task failure email (default 5000). 

## Motivation and Context

Suggesting this change because of an error I encountered: "ClientError: An Error occurred (InvalidParameterValue) when calling the SendRawEmail operation: message length is more than 10485760 bytes long". The root cause was a very long sql query which led to a mysql error. The error traceback included the super long query and due to its length, the luigi "task failure" email failed causing the worker to error during "_handle_next_task". The issue was then disguised, the scheduler continued to show the task as "running". The email that failed was ~17m bytes long. This adds a traceback_max_length config option for the failure email which truncates the traceback included in task failure email (default 5000 which is roughly where Rollbar truncated it). 

I included a unit test

### DIFF
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -84,6 +84,9 @@ class email(luigi.Config):
         default='',
         config_path=dict(section='core', name='error-email'),
         description='Address to send error e-mails to')
+    traceback_max_length = luigi.parameter.IntParameter(
+        default=5000,
+        description='Max length for error traceback')
     sender = luigi.parameter.Parameter(
         default=DEFAULT_CLIENT_EMAIL,
         config_path=dict(section='core', name='email-sender'),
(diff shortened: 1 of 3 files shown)


dlstadther (commented):
can we move the `traceback_max_length` from a constant to a luigi config options? otherwise, i'm good here

DoriRunyon:
Updated @dlstadther-pl. Please let me know if this needs any other changes.

dlstadther (dismissed):
lgtm

idk why some of the Github Action tests fail

luigi/notifications.py
```diff
@@ -373,6 +376,11 @@ def format_task_error(headline, task, command, formatted_exception=None):
     :return: message body
     """
 
+    if formatted_exception:
+        if len(formatted_exception) > email().traceback_max_length:
+                truncated_exception = formatted_exception[:email().traceback_max_length]
```
tnsetting, line 381: Flake8 complains that it is `./luigi/notifications.py:381:17: E117 over-indented`

DoriRunyon: Ah, I will fix this

dlstadther (approved):

PR #3088, opened 2021-06-23
Patch BigQueryClient to retry on network error

jamesmcm:
## Description

This patch retries by re-creating the client object if it fails (up to a retry limit), so that the client will not be timed out when calling `complete()` on the BigQueryTarget after a long job.

## Motivation and Context

At the moment sometimes long jobs can cause the BigQueryClient to timeout since the connection is created at initialisation of the target.

Yes, passing this patched client worked for jobs that would have timeout issues previously.

### DIFF
--- a/luigi/contrib/bigquery.py
+++ b/luigi/contrib/bigquery.py
@@ -48,11 +48,11 @@ def is_error_5xx(err):
 
 
 bq_retry = retry(retry=(retry_if_exception(is_error_5xx) | retry_if_exception_type(RETRYABLE_ERRORS)),
-                  wait=wait_exponential(multiplier=1, min=1, max=10),
-                  stop=stop_after_attempt(3),
-                  reraise=True,
-                  after=lambda x: x.args[0].__initialise_client()
-                  )
+                 wait=wait_exponential(multiplier=1, min=1, max=10),
+                 stop=stop_after_attempt(3),
+                 reraise=True,
+                 after=lambda x: x.args[0].__initialise_client()
+                 )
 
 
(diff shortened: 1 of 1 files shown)


luigi/contrib/bigquery.py
```diff
@@ -147,6 +158,12 @@ def dataset_exists(self, dataset):
                         fetched_location if fetched_location is not None else 'unspecified',
                         dataset.location))
 
+        except (TimeoutError, BrokenPipeError, IOError) as bq_connection_error:
+            if self.retry_count > self.retry_limit:
+                raise Exception(f"Exceeded max retries for BigQueryClient connection error: {bq_connection_error}") from bq_connection_error
+            self.retry_count += 1
+            self.__initialise_client()
+            self.dataset_exists(dataset)
```
hirosassa, line 166: How about using retry decorator to simplify the codes?
The implementation below will be useful for you.
https://github.com/spotify/luigi/blob/master/luigi/contrib/gcs.py#L72-L81

jamesmcm: Thanks, I'll make that change.

hirosassa: @jamesmcm Thank you! Looks good to me, but flake8 notifies some formatting problems.

jamesmcm: Thanks, I fixed that too.

hirosassa:
This PR looks great! 
@dlstadther @Tarrasch @dataex plz take a look.

dlstadther (approved):
lgtm

PR #3089, opened 2021-06-28
Kubernetes contrib : waiting for cluster scale up

dav009:
## Description
Makes sure Luigi waits for kubernetes to scale up when waiting for a job run.

## Motivation and Context
Sometimes scheduling a new job results in the cluster needing to scale up. 
In these scenarios current luigi throws an exception which exits the pipeline. In the meantime the cluster runs the job after scaling up.
Changes introduced in this PR makes it so that Luigi waits for Scale up to take place when waiting for a job to run. 

## Have you tested this? If so, how?
- unit tests included in this PR
- scheduling a task for my pipepelines

PR #3093, opened 2021-08-03
Replace deprecated pykube(-ng) library with official Kubernetes python library

AndrewFarley:
## Description
The pykube and pykube-ng libraries that Luigi depends on are flawed and don't seem to support more recent versions of Kubernetes.  Both libraries appear to be abandonware, so we should pivot to the official python library for Kubernetes.  This pivot will simplify some of the handlings of this library, as certain configuration values are already supported and configurable via environment variables, as is the widely accepted standard for Kubectl and its supported tools and toolchain.  This also adds the following...

* Better/more debug logging about what its doing to/in Kubernetes
* Event logging before the pod starts up (so you can see what its doing, eg Pulling Image, etc)
* Cleaned up code a bit
* Able to detect cluster scaling up (contribution from @dav009)
* Able to detect cluster scaling up won't work and failing
* WIP Adding more robust tests to ensure we're handling all success and failure conditions properly

## Motivation and Context
Luigi wasn't working for me on a number of my clusters, and after further investigation this was largely due to the outdated and deprecated libraries in-use.  I've found numerous other projects which have pivoted over to the official library with a lot of success and used those largely as examples for how to do this.

I've added extra comments and documentation in all places that make sense for others to be able to more-easily understand how/what to do to use and adopt Kubernetes support for Luigi.

I've updated the unit tests and run a few test pipelines successfully with all features.

## Items to complete
- [x] Implement core feature-set and pivot to the new library
- [x] Add documentation
- [ ] **WIP** Add/update the test coverage
- [ ] **WIP** Add ability to watch logs during a run, not just after a run
- [ ] Merge and publish a new version for all to enjoy

### DIFF
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -35,6 +35,7 @@ Written and maintained by Marco Capuccini (@mcapuccini)
 Pivoted to official kubernetes-client python module by Farley (@AndrewFarley)
 """
 import logging
+import re
 import time
 import uuid
 from datetime import datetime
@@ -45,15 +46,9 @@ logger = logging.getLogger('luigi-interface')
 
 try:
     import kubernetes as kubernetes_api
-except ImportError as i:
+except ImportError:
     logger.warning("WARNING: kubernetes is not installed. KubernetesJobTask requires kubernetes")
     logger.warning("  Please run 'pip install kubernetes' and try again")
-
-
-class kubernetes(luigi.Config):
-    kubernetes_namespace = luigi.OptionalParameter(
-        default=None,
-        description="K8s namespace in which the job will run")
 
 class KubernetesJobTask(luigi.Task):
     __DEFAULT_POLL_INTERVAL = 5  # see __track_job
@@ -62,8 +57,9 @@ class KubernetesJobTask(luigi.Task):
     def _init_kubernetes(self):
         self.__logger = logger
         self.__kubernetes_core_api = kubernetes_api.client.CoreV1Api()
+
         # Configs can be set in Configuration class directly or using helper utility, by default lets try to load in-cluster config
-        # TODO: Make library support forcing one of these instead of automatic cascading logic...?
+        # and if that fails cascade into using an kube config
         try:
            kubernetes_api.config.load_incluster_config()
         except Exception as e:
@@ -71,16 +67,16 @@ class KubernetesJobTask(luigi.Task):
               kubernetes_api.config.load_kube_config()
            except Exception as ex:
               raise ex
-        
+
         # Create our API instances for Kubernetes
         self.__kubernetes_api_instance = kubernetes_api.client.CoreV1Api()
         self.__kubernetes_batch_instance = kubernetes_api.client.BatchV1Api()
-        
+
         self.job_uuid = str(uuid.uuid4().hex)
         now = datetime.utcnow()
-        
+
         # Set a namespace if not specified because we run jobs in a specific namespace, always
-        if self.kubernetes_namespace is None:
+        if not self.kubernetes_namespace:
             self.kubernetes_namespace = "default"
 
         self.uu_name = "%s-%s-%s" % (self.name, now.strftime('%Y%m%d%H%M%S'), self.job_uuid[:16])
@@ -89,7 +85,21 @@ class KubernetesJobTask(luigi.Task):
     def kubernetes_namespace(self):
         """
         Namespace in Kubernetes where the job will run.
-        It defaults to the default namespace in Kubernetes
+        It defaults to the default namespace in Kubernetes, you can override this
+        by setting this property in your spec.
+
+        .. code-block:: python
+
+            class PerlPi(KubernetesJobTask):
+                name = "test"
+                kubernetes_namespace = "inserthere"   # <-- here
+                spec_schema = {
+                    "containers": [{
+                        "name": "pi",
+                        "image": "perl",
+                        "command": ["perl",  "-Mbignum=bpi", "-wle", "print bpi(2000)"]
+                    }]
+                }
 
         For more details, please refer to:
         https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/
@@ -100,9 +110,9 @@ class KubernetesJobTask(luigi.Task):
     def name(self):
         """
         A name for this job. This task will automatically append a UUID to the
-        name before to submit to Kubernetes.
+        name before to submit to Kubernetes.  This is not optional.
         """
-        raise NotImplementedError("subclass must define name")
+        raise NotImplementedError("You must define the name of this job, please override the `name` property")
 
     @property
     def labels(self):
@@ -116,7 +126,8 @@ class KubernetesJobTask(luigi.Task):
     @property
     def spec_schema(self):
         """
-        Kubernetes Job spec schema in JSON format, an example follows.
+        Kubernetes Job spec schema in JSON format, an example follows. This
+        is not optional.
 
         .. code-block:: javascript
 
@@ -138,7 +149,7 @@ class KubernetesJobTask(luigi.Task):
         For more informations please refer to:
         http://kubernetes.io/docs/user-guide/pods/multi-container/#the-spec-schema
         """
-        raise NotImplementedError("subclass must define spec_schema")
+        raise NotImplementedError("You must define the `spec_schema` of an Kubernetes Job")
 
     @property
     def backoff_limit(self):
@@ -151,34 +162,81 @@ class KubernetesJobTask(luigi.Task):
     @property
     def delete_on_success(self):
         """
-        Delete the Kubernetes workload if the job has ended successfully.
+        Delete the Kubernetes workload if the job has ended successfully.  True by default
         """
         return True
 
     @property
     def print_pod_logs_on_exit(self):
         """
-        Fetch and print the pod logs once the job is completed.
+        Fetch and print the pod logs once the job is completed.  False by default
+        """
+        return False
+
+    @property
+    def print_pod_logs_during_run(self):
+        """
+        Fetch and print the pod logs during the the job.  False by default
+        TODO MUST IMPLEMENT THIS BEFORE MERGING...
         """
         return False
 
     @property
     def active_deadline_seconds(self):
         """
-        Time allowed to successfully schedule pods.
+        Time allowed to successfully schedule AND run the pods.
         See: https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#job-termination-and-cleanup
         """
         return None
 
     @property
     def poll_interval(self):
-        """How often to poll Kubernetes for job status, in seconds."""
+        """How often to poll Kubernetes for job status, in seconds.  Default of 5"""
         return self.__DEFAULT_POLL_INTERVAL
 
     @property
     def pod_creation_wait_interal(self):
-        """Delay for initial pod creation for just submitted job in seconds"""
+        """Delay for initial pod creation for just submitted job in seconds.  Default of 5"""
         return self.__DEFAULT_POD_CREATION_INTERVAL
+
+
+    def __is_scaling_in_progress(self, condition, messages):
+        """Parses condition and messages, returns true if cluster is currently scaling up"""
+
+        # If we're not unschedulable then stop processing...
+        if condition.reason != 'Unschedulable':
+            return False
+
+        # Lets check the messages
+        for message in messages:
+            # Check if our status message is about node availability
+            match = re.match(r'(\d)\/(\d) nodes are available.*', message)
+            if match:
+                current_nodes = int(match.group(1))
+                target_nodes = int(match.group(2))
+                # Only if there's non-matching nodes and it is still not being scheduled
+                if current_nodes <= target_nodes:
+                    return True
+
+
+    def __has_scaling_failed(self, condition, messages):
+        """Parses messages from kubectl events to see if scaling up failed"""
+        try:
+            # If we're not unschedulable then stop processing...
+            if condition.reason != 'Unschedulable':
+                return False
+
+            # Check our messages for the can't scale up message
+            for message in messages:
+                # Check if our status message is about that we can't scale up, if so exit immediately
+                if "pod didn't trigger scale-up (it wouldn" in message:
+                    # We return here immediately on purpose, instead of the delayed return below
+                    return True
+        except:
+            pass
+
+        return False
+
 
     def __track_job(self):
         """Poll job status while active"""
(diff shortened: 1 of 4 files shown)


luigi/contrib/kubernetes.py
```diff
@@ -42,74 +44,46 @@
...
 class kubernetes(luigi.Config):
-    auth_method = luigi.Parameter(
-        default="kubeconfig",
-        description="Authorization method to access the cluster")
-    kubeconfig_path = luigi.Parameter(
-        default="~/.kube/config",
-        description="Path to kubeconfig file for cluster authentication")
-    max_retrials = luigi.IntParameter(
-        default=0,
-        description="Max retrials in event of job failure")
```
AndrewFarley, line 61: A note for reviewers.  These things are safe to remove, and no longer need to live in code, as the upstream Kubernetes library supports handling them automatically via Environment Variables.  The max_retrials also (unless I misunderstand) is completely unnecessary as it duplicates functionality with `backoff_limit`.  Arguably, one config option here I could keep is auth_method so an end-user can "force" which auth method to use.  But, the Kubernetes library should be able to automatically detect and cascade into the "mode" it needs to function.

luigi/contrib/kubernetes.py
```diff
@@ -42,74 +44,46 @@
...
+        self.__kubernetes_core_api = kubernetes_api.client.CoreV1Api()
+        # Configs can be set in Configuration class directly or using helper utility, by default lets try to load in-cluster config
+        # TODO: Make library support forcing one of these instead of automatic cascading logic...?
+        try:
+           kubernetes_api.config.load_incluster_config()
+        except Exception as e:
+           try:
+              kubernetes_api.config.load_kube_config()
+           except Exception as ex:
+              raise ex
```
AndrewFarley, line 73: Note for reviewers: This is the "new" cascading automated configuration handling.  If we're in Kubernetes, it will automatically detect this and use in-cluster configuration and API endpoints, if that fails, it will attempt to use the kubeconfig file and env vars that help support it.

luigi/contrib/kubernetes.py
```diff
@@ -120,7 +94,7 @@ def kubernetes_namespace(self):
         For more details, please refer to:
         https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/
         """
-        return self.kubernetes_config.kubernetes_namespace
+        return "default"
```
tnsetting, line 97: How do we set the `kubernetes_namespace` now? In the current implementation it is set [here](https://github.com/spotify/luigi/pull/3093/files#diff-00edfcf9a2971c757897a6fdb1a39816d7b5ada36e37a2315ed50003abde3d7eL209)

luigi/contrib/kubernetes.py
```diff
@@ -42,74 +44,46 @@
 logger = logging.getLogger('luigi-interface')
 
 try:
-    from pykube.config import KubeConfig
-    from pykube.http import HTTPClient
-    from pykube.objects import Job, Pod
-except ImportError:
-    logger.warning('pykube is not installed. KubernetesJobTask requires pykube.')
+    import kubernetes as kubernetes_api
+except ImportError as i:
```
dlstadther, line 48: no need to alias error `as i` since not using var `i`

luigi/contrib/kubernetes.py
```diff
@@ -291,101 +252,109 @@ def __verify_job_has_started(self):
...
-            "kind": "DeleteOptions",
-            "apiVersion": "v1",
-            "propagationPolicy": "Background"
-        }
-        r = self.__kube_api.delete(json=delete_options_cascade, **job.api_kwargs())
-        if r.status_code != 200:
-            self.__kube_api.raise_for_status(r)
+        api_response = self.__kubernetes_batch_instance.delete_namespaced_job(job.metadata.name, self.kubernetes_namespace, body={"grace_period_seconds": 0, "propagation_policy": "Background"})
+        # TODO: Check status of this request...?
+        print(api_response.status)
```
dlstadther, line 310: can we replace this print statement with the intent of your todo comment?

setup.py
```diff
@@ -37,7 +37,7 @@ def get_static_files(path):
 with open('README.rst') as fobj:
     long_description = "\n\n" + readme_note + "\n\n" + fobj.read()
 
-install_requires = ['python-dateutil>=2.7.5,<3', 'tenacity>=6.3.0,<7']
+install_requires = ['python-dateutil>=2.7.5,<3', 'tenacity>=6.3.0,<7', 'kubernetes==17.17.0']
```
dlstadther, line 40: given that the Kubernetes job type is an optional portion of Luigi, can we not make `kubernetes` package install required when installing `luigi`?

test/contrib/kubernetes_test.py
```diff
@@ -86,15 +85,16 @@ def test_success_job(self):
...
-        # Check for retrials
-        kube_api = HTTPClient(KubeConfig.from_file("~/.kube/config"))  # assumes minikube
-        jobs = Job.objects(kube_api).filter(selector="luigi_task_id="
-                                                     + fail.job_uuid)
-        self.assertEqual(len(jobs.response["items"]), 1)
-        job = Job(kube_api, jobs.response["items"][0])
-        self.assertTrue("failed" in job.obj["status"])
-        self.assertTrue(job.obj["status"]["failed"] > fail.max_retrials)
-        self.assertTrue(job.obj['spec']['template']['metadata']['labels'] == fail.labels())
+        # TODO: Discuss/consider what to re-implement here, possibly only keep the fail.labels check, possibly ad backoff limit and name check
```
dlstadther, line 88: in terms of what to test here, i'd suggest ensuring that a failing job behaves as expected. If that includes attempting a retry, go for it, etc.

dlstadther (commented):
left a few comments; once they're addressed, i'll approve

luigi/contrib/kubernetes.py
AndrewFarley: @tnsetting This is a built-in feature in the foundational library we're using.  The config argument can be provided to specify the namespace.  See the `examples/kubernetes.py` for an example.

AndrewFarley: The reason for "default" here, is that means as per how all of the kube utils work, if you don't specify anything you're in the "default" namespace by default.  With python, you can "override" this property, which is exactly what is happening in the examples/kubernetes.py

AndrewFarley: So by this change, we're making it so you don't "have" to set a kubernetes_namespace, it'll run in default by default, unless you override this property.  Which is the same as you do before, basically.  This self.kubernetes_config concept in this retval doesn't make a ton of sense.  Kubernetes_config is/was replaced entirely, it was from the dated library.  This property is still overridable when you  use it.  Technically, your configs do not need to change at all it will function exactly as it did before.

AndrewFarley: And technically, I could make this property not a function and just a property variable/value that you override.  I am happy to make that change for simplicity sake if desired?

setup.py
AndrewFarley: Yep, I'll happily realign to this if it's optional.  Will do in some upcoming free time.  Thanks for the feedback.

luigi/contrib/kubernetes.py
AndrewFarley: TL;DR: Everything works the same as it did before, you can override this property currently.  But, I can make this a property (variable) instead of a property (function) if desired for simplification.  I'll try to do that I think.

test/contrib/kubernetes_test.py
AndrewFarley: **TL;DR**: Luigi isn't "managing Kubernetes" so the above tests don't make sense to test.  We're telling Kubernetes contractually what to do (eg: number of retries) and it tries to do that.  No purpose in testing what Kubernetes actually did, the output of what it actually do can be wildly different dependant on many variables and situational aspects.  Kubernetes manages the aspect of retrials, which I think is a large contrast to how Luigi works with many other providers manually re-submitting/executing/scheduling things.

Well... so @dlstadther what you're asking is for me to write tests to ensure Kubernetes is executing what you tell it to. 
 I'm trying to think of a good metaphor... but in a sense, it's like someone going to a master at their craft, lets say a baker in this example, and going into their kitchen and setting up some timers and some cameras to inspect the process at every step of the way to ensure the inputs end up with the same outputs.  I think the way to explain this as well, is Luigi isn't "managing Kubernetes" because there's no such thing really (well, sure, controllers do this, but we aren't a controller for Kubernetes).  I stripped this silly retry code that was in this contrib that attempted to fiddle with some retry logic because all of that is 100% built-into Kubernetes and makes no sense.  And thus, because we are not shepherding things in Kubernetes, we aren't "managing" kubernetes any more, we also shouldn't be testing things which are managed by Kubernetes.
 
 By defining the yaml object properly and by Kubernetes accepting that object, it has committed to _trying_ to execute that contract to the best of its ability, within the constraints you give it.  And contractually it tries to do that.  If you tell it to retry x times, it will do everything it can to ensure that happens.  However, there are various various reasons (eg: resource limits/constraints, node selectors, taints, etc) that it can't fulfill the request contractually.  Since we're defining the yaml to include a specific number of retries, kubernetes will _attempt_ to do that.  But checking if Kubernetes _did_ retry, doesn't make a lot of sense.  It either did or it didn't, that's not our responsibility to test against.  In theory, those above tests are actually unstable depending on the stability of the cluster, nodes in the cluster, resource constraints (or lack of), oomkiller happening, unable to scale up the cluster, etc.  A job could absolutely be failed, and not retry at all because it was never able to initialize because of a mismatch of nodelabels, or taints, etc.
 
 I think what might make sense here is to test that everything we tried to set, is actually set.  Eg: number of max retrials, image of the job, job definition details, etc.  That's the best thing I can think of, and I hope my explanation makes sense above.

AndrewFarley: But that's why I left the comment that said this... input needed here.   :)  Shall I just add some tests to check if what we set is actually set in the returned yaml object?  That seems... pedantic, but might be valuable at some point in the future if the API changes in Kubernetes that might make our expectations change (eg: renaming or moving where we set a variable, such as the retries)

dlstadther: @AndrewFarley , i appreciate your thoroughness of response. Yes, i'm familiar with Kubernetes (less so with how we're using Luigi here to interact; although looks moderately straight forwards as a k8s job that we then just poll for its status). So when i say we should "ensure that a failing job behaves as expected", I'm really referring to ensuring that Luigi acknowledges a failing job and responds to it appropriately. What I don't want to see is that a Luigi job thinks it was successful when its respective Kubernetes job actually exited in an erroneous state (in state phase error, or possibly even evicted by the control plane and thus an unknown status).

In the case that a job is not found, it looks like `__get_job` will raise an AssertionError; whereas if the job state is "Failed" (from `__get_job_status`), then we'll still raise an AssertionError within `_track_job`).

So the place I think we're missing test coverage for (that we should) is to ensure we get an AssertionError when its job fails. That isn't testing K8s functions correctly; that's ensuring that our Luigi code identifies and responds to the k8s api responses correctly and doesn't falsely think we were successful or respond to error in a way we don't expect. 

The extent to which we do this can be a little checking that:
```
        fail = FailJob()
        self.assertRaises(RuntimeError, fail.run)
```
and leaving it at that. 

So effectively removing the commented out section and calling it a day.

AndrewFarley:
Note: I am merging and fixing up the MR #3089 in this MR.  Picture of proof of work, added tons of debug logging as well.  MR will be updated shortly I believe addressing all issues...

Screen Shot 2021-08-30 at 3 07 28 AM

AndrewFarley:
@dav009 I merged(ish) your branch into mine logicall, but added code to detect if a scale up failed (aka, it won't fit even with  scale up) which the previous code would infinite loop because it wasn't looking into the Kubernetes event stream.  Pic of proof...

Screen Shot 2021-08-30 at 3 10 12 AM

luigi/contrib/kubernetes.py
AndrewFarley: Cleaned up code, simplified

setup.py
AndrewFarley: Resolved in latest push

test/contrib/kubernetes_test.py
AndrewFarley: Resolved that above issue, simple failjob

AndrewFarley:
@dlstadther Please have another look and review code, but please don't merge yet, I've decided I need to add some further tests because it's really abysmal to not really have any decent ones here to ensure things work as they should, there's so many different kinds of failure conditions which are trying to be detected here and I think we should test them.  I'm not fantastic with Pytest so it might be a bit slow, but I'm working on adding tests which I think are critical before merging this.  Only able to work on this in free time, so if anyone else can/wants to help me with tests, please do.  I've left some of my hacking at tests commented out.

AndrewFarley:
And @dav009 please review the code here and make sure I encapsulated what you were trying to do with your MR.  Overall, this MR will have a lot better QoL support for Kubernetes via Luigi.  There's still tons more things I think could/should be done in this library but mostly pedantic and could be in future merges instead.  I'm still working on redoing some of the tests and will sorta include your tests once I fix them.

dav009:
@AndrewFarley amazing job, thank you. I was also quite frustrated with how outdated this is. I will try to review this soonish. 

😅 I made an internal library for my organisation that addresses these (old pykube) and also take into account other issues (logging). I am testing it at the moment. We were planning on releasing it at some point.

dav009:
@AndrewFarley  In my team we ended up implementing this kubernetes contrib update as an independent package:
https://github.com/optimizely/kubeluigi

AndrewFarley:
Ping.  Folks, I'm not using Luigi, but this MR still really needs to happen.  If anyone has time to rebase and take over this PR, please let me know I'm happy to transfer my fork to you or add you permissions.  I know a few folks are using my fork already because it's necessary if you want Kubernetes support, so perhaps one of you can take the reins?

nlahmi-cymotive:
Hi, why was this not reviewed yet? I tested it locally and it seems to work great..

AndrewFarley:
Hah, I wrote this 2 years ago and still no one merged it.  Is this project dead?

AndrewFarley:
I just rebased and pushed my latest code.  Hopefully, in the hopes and dreams of fellow engineers, someone can see it in their heart to merge this.  Or, just don't use Kubernetes.  :P  Up to you

dlstadther:
Apologies for this not being reviewed.

As far "is this project dead?" - there is not consistent maintainer involvement (myself included). The project is still owned by Spotify, but not actively developed (to my knowledge, mostly due to their articles related to their [intent to replace Luigi with Flyte](https://engineering.atspotify.com/2022/03/why-we-switched-our-data-orchestration-service/) (note, this is just my personal assumption)). While I have permissions to merge PRs and push to PyPi, I have not used Luigi for development since changing employers a few years ago. I wish to see this project continue, but without an active community and additional maintainer participation, there's only so much I can do. I sympathize with your disappointment and I'm here to help as I reasonably can in my spare time.

dlstadther (commented):
In terms of review of this PR, I'm open to merging this code. However, 1 item prevents me and 1 item deters me:
1. [flake8 tests](https://github.com/spotify/luigi/actions/runs/4013965395/jobs/6933065857) are failing
2. a bunch of the unittests are commented out with TODO notes

I am willing to let point # 2 to slide so that `master` has a functional k8s module. But # 1 will need to be fixed before I can click the merge button.

AndrewFarley:
@dlstadther I've just fixed the code styling to satisfy that need.  I indeed also don't use Luigi, and I don't really have tons of time to write/update/etc the tests to validate that this works.  I can confirm it did work for me once upon a time, and others above have recently checked its (basic) functionality.  I can/should leave the TODOs in the test code commented out, in hopes that someone else can take up the reigns and add the tests.

At least the flake tests pass now - https://github.com/spotify/luigi/actions/runs/4034116089/jobs/6935056885

But it appears there are some other errors now?  I don't have much time this weekend, but I will try to review this further sometime soon.

dlstadther:
@AndrewFarley , thank you for the quick turnaround to address the failing flake8 tests. I'm good with getting this included once test results allow me to merge. I hope whomever will come use this updated k8s support within luigi will contribute back any improvements and test coverage items.

I just checked the current failures and they're all 1 failure:
* the test class tries to load a kubeconfig and cannot find one - `kubernetes_api.config.load_kube_config()`

So looks to me like once a kube config is mocked (or however it is expected) for tests, that we'd be good to go for now.

AndrewFarley:
Well, in order for the Kubernetes stuff to work, it needs to initialize the library and config somehow.  I will need to have a think and a google around of some creative way to workaround this for testing only which won't have K8s support.

PR #3101, opened 2021-08-30
Fix the kubernetes test

abjorkl5:
I was trying to run the kubernetes test, but it didn't pass.

With these changes, it was able to pass using minikube.

----

- When a job fails, it asserts rather than exits

- There are other labels besides the custom ones

- Need to mock out the verify_job_has_started()

- Showing the pods requires access to the cluster

### DIFF
--- a/luigi/contrib/kubernetes.py
+++ b/luigi/contrib/kubernetes.py
@@ -33,6 +33,7 @@ Requires:
 Written and maintained by Marco Capuccini (@mcapuccini).
 """
 import logging
+import os
 import time
 import uuid
 from datetime import datetime
@@ -54,7 +55,7 @@ class kubernetes(luigi.Config):
         default="kubeconfig",
         description="Authorization method to access the cluster")
     kubeconfig_path = luigi.Parameter(
-        default="~/.kube/config",
+        default=os.getenv("KUBECONFIG", "~/.kube/config"),
         description="Path to kubeconfig file for cluster authentication")
     max_retrials = luigi.IntParameter(
         default=0,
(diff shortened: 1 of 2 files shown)


abjorkl5:
I noticed there were some much bigger changes done in #3093

These were more for testing the old code, rather than change it

AndrewFarley:
@abjorkl5 Can I ask you what command(s) are you using to run the tests?  Even with the previous generation of tests, I was never able to run the tests successfully for a variety of reasons, and that is delaying the tests on my big MR upgrading the k8s library of Luigi.  And/or, can you maybe jump into my MR and help me fix up all the tests and make some new ones?  I want to test every failure condition ideally, and I'm not good at pytest, having never really used it before.

abjorkl5:
I was using this command:

```
PYTHONPATH=. luigi --module examples.kubernetes PerlPi --local-scheduler
```

But I didn't want to change the example instructions in _this_ PR (for the test)
I think it said "kubernetes_job" by mistake ? Basically, I was doing Nomad...

And for the unit test, I used:

```
pytest-3 test/contrib/kubernetes_test.py
```

I had to comment out some unsupported `--cov` option, though. (in tox.ini)

```
ERROR: usage: pytest-3 [options] [file_or_dir] [file_or_dir] [...]
pytest-3: error: unrecognized arguments: --cov=luigi
```

This was on Ubuntu 20.04

AndrewFarley:
Thanks for that @abjorkl5, when I have some time on the weekend I'll hopefully take a look at this MR and merge it into my #3093 refactor and improve the tests there.  This critical feature really needs some good tests.  Cheers!

PR #3117, opened 2021-11-03
Add rerun command snippet in "show error" modal

leifdenby:
## Description
When running very complex tasks with many nested dependencies I find myself wanting to re-run just a few tasks that failed instead of the parent task (so as to avoid the time-consuming checks for completed/passed/etc required tasks). 
Because of how the task information is rendered in the visualiser there wasn't an easy way to copy-paste the information there. I've added the rendering of a snippet in the "show error" modal which contains the exact command to copy-paste into a terminal to re-run a specific task.

## Motivation and Context
To save time when debugging failing tasks

Yes, I tested with the example task class below

```python
import luigi                                    
                                                
class FailTask(luigi.Task):                     
    arg_1 = luigi.Parameter()                   
    arg_2 = luigi.FloatParameter()              
    arg_3 = luigi.OptionalParameter(default=0.2)
                                                
    def run(self):                              
        raise Exception(42)                     
```

Screenshot of the new element:
Screenshot 2021-11-03 at 09 07 51


dlstadther (approved):
lgtm

PR #3122, opened 2021-11-17
group tasks by task class in SVG visualizer

joooeey:
## Description

I adjusted the SVG task visualizer so that it groups tasks by task class. I.e. it doesn't but tasks of different classes at the same depth level anymore.

There is one exception: if at least one task requires a task of the same task class, it falls back to the old way of visualizing the tree.

In addition, to avoid overlapping edges, a different (random) horizontal offset is added to each depth level.

## Motivation and Context

The main goal is to make the task graph quicker to interpret by grouping tasks of the same task class.

I have included two new unit tests but haven't run them yet. Counting on Travis here.

I also ran my jobs with this code and got the following image:

original:

task_visualizer_old

with my changes applied:

task_visualizer_new

## Request

Feel free to test the new visualizer with your workflows and let me know how you like it.

All you'd have to do is replace `/site-packages/luigi/static/visualiser/js/graph.js`, wait a few seconds and reload `localhost:8082`.

### DIFF
--- a/luigi/static/visualiser/js/graph.js
+++ b/luigi/static/visualiser/js/graph.js
@@ -65,21 +65,72 @@ Graph = (function() {
         });
         return edges;
     }
-
-    /* Compute the maximum depth of each node for layout purposes, returns the number
-       of nodes at each depth level (for layout purposes) */
+    /* Compute the depth of each node for layout purposes */
     function computeDepth(nodes, nodeIndex) {
+        var selfDependencies = false
         function descend(n, depth) {
             if (n.depth === undefined || depth > n.depth) {
                 n.depth = depth;
                 $.each(n.deps, function(i, dep) {
                     if (nodeIndex[dep]) {
-                        descend(nodes[nodeIndex[dep]], depth + 1);
+                        var child_node = nodes[nodeIndex[dep]]
+                        descend(child_node, depth + 1);
+                        if (!selfDependencies && n.name == child_node.name) {
+                            selfDependencies = true;
+                        }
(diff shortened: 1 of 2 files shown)


lallea:
This looks much better, thanks for the contribution. I rarely use Luigi's UI in production, but I hold data engineering courses, and the visualiser is useful to explain what is going on. You just made it much clearer. Thanks!

luigi/static/visualiser/js/graph.js
```diff
@@ -91,35 +145,46 @@ Graph = (function() {
...
+                        var depth = (selfDependencies ? depth + 1 : classDepths[next_node.name])
+                        placeNodes(next_node, depth);
                     }
                 });
             }
         }
         placeNodes(nodes[0], 0);
+        console.log("nodes with rows")
+        console.log(nodes)
+        console.log(rowSizes)
```
dlstadther, line 158: Do we want to leave these console log statement in? I'm not aware if this is desired to do in js code

dlstadther (commented):
I'm not a js dev, but i think this lgtm. I left one question, but otherwise, i'm good here. I'll approve once my question is answered.

luigi/static/visualiser/js/graph.js
joooeey: I must have overlooked those. I had way more log statements in and deleted most of them. They are deleted now. I'm not really a JS dev either so can't tell you if we *should* have console.log statements.

dlstadther (approved):

PR #3125, opened 2021-11-23
add possiblity of input in seconds to TimeDeltaParameter

### LINKED ISSUE #3123: format for config times is inconsistent
My `luigi.cfg` looked like this:
```
[core]
# Hide DeprecationWarning:
autoload_range: false

[scheduler]
# Time before a finished task graph is deleted (in seconds):
remove_delay: 10

[worker]
# Keep workers alive as long as they have pending jobs
keep_alive: true
max_keep_alive_idle_duration: 3600
```
Now luigi can parse the 10 seconds but it can't pass the 3600 seconds. That's surprising.

Turns out I had to specify `max_keep_alive_idle_duration: 1 hour` but that can't be found in the documentation without inspecting the source code. Yes, it's in the docs for the TimedeltaParameter but nowhere does it say that this option is interpreted as a TimedeltaParameter.

However, it would be better if all the config values specifying a time can be given in the same format. I'll see if I can get in a PR to solve this. The idea would be for all times to accept all formats accepted by TimeDeltaParameter plus to interpret an integer as seconds.

joooeey:
## Description

Allow input as number in seconds to `TimeDeltaParameter` to make `max_keep_alive_idle_duation` except the same format as most other config options (i.e. seconds). This means that any `TimeDeltaParameter` used in a `luigi.Task` can also be given as a bare number and interpreted as seconds.

Plus two small changes:
* The word "seconds" is added to all the descriptions where it was missing.
* `batch_notifier.email_interval` gets a new alias `batch_notifier.email_interval_minutes` to make it obvious that a different unit is used there.

## Motivation and Context

Closes #3123. Making duration config parameters as consistent as possible. (`batch_notifier.email_interval` is still given in minutes but renamed to `batch_notifier.email_interval_minutes`).

I have included two new unit tests.


doc/configuration.rst
```diff
@@ -447,7 +447,7 @@ traceback_max_length
 Parameters controlling the contents of batch notifications sent from the
 scheduler
 
-email_interval
+email_interval_minutes
```
dlstadther, line 450: this config variable should remain `email_interval`, that's the name which is searched for by the code https://github.com/spotify/luigi/blob/857deff3c7ade4bfd57f9acfb0e4816814cce8c5/luigi/batch_notifier.py#L17

dlstadther (commented):
other than the one comment about incorrectly named variable, i'm good here

doc/configuration.rst
joooeey: I made a change in that file. But looks like `tox` through it out. I didn't know I have to commit before running `tox`. Please check the commit I'll make in a few minutes.

joooeey: Committed. I'm not sure if it solves the issue though. And I don't know how I'd go about writing a test for it. Do you have any tips?
https://github.com/spotify/luigi/pull/3125/files#diff-b5f8cc1a0854bdb1861ce3c50dea58a7b7759c4a36ca6a65a1516fa47d39e7ffL18-R20

dlstadther: Cool; yeah, this is what i was expecting.

As for the link/code preview, i think it has to be a code link from `master`. I didn't do anything special other than past the link.

dlstadther (approved):

dlstadther:

PR #3126, opened 2021-11-25
Add PathParameter and OptionalPathParameter

adrien-berchet:
## Description
Add two classes: PathParameter and OptionalPathParameter.

## Motivation and Context
When dealing with file or folder paths given as parameters to a Task, we always end up to cast it to a more convenient object than a string. These new classes automatically convert the given string to a `pathlib.Path` which allows to use it directly as a path object inside the Task.

I have included unit tests.

## Side effect
The OptionalPathParameter has to deal with both str and pathlib.path objects as inputs, so I had to update the class OptionalParameter so that it is possible to define a list of types in the `expected_type` attribute.


krabo0om:
Thanks for the PR. But what about the parse method? Does it not have to be implemented? I am currently using the following: 

```python
class PathParameter(luigi.Parameter):
    def parse(self, x):
        return pathlib.Path(x)

class MyTask(luigi.Task):
    my_path: pathlib.Path = PathParameter()
```

adrien-berchet:
Hi @krabo0om 
I use `normalize()` because, as far as I know, it is always called, while `parse()` is only called to process inputs from the CLI and CFG files.
For example, consider the following example:

```python
import pathlib

import luigi


class PathParameter(luigi.Parameter):
    def parse(self, x):
        return pathlib.Path(x)


class MyTask(luigi.Task):
    my_path: pathlib.Path = PathParameter(default="the default path")

    def run(self):
        print(type(self.my_path), self.my_path)

    def output(self):
        return luigi.LocalTarget("NOT_A_FILE")


def main():
    task = MyTask("a new path")
    luigi.build([task], local_scheduler=True)


if __name__ == "__main__":
    main()
```

Here are the different possible cases:
1. **default value**: if you run it with the following command without any entry for `MyTask` in `luigi.cfg`:
    ```bash
    python -m luigi --module test_path_param MyTask --local-scheduler
    ```
    The default value will be used and you will see this: `<class 'str'> the default path`
    So the default value is not processed.
2. **constructor value**: if you run it with the following command:
    ```bash
    python test_path_param.py
    ```
    The value given to the constructor is used and you will see this: `<class 'str'> a new path`
    So the constructor value is also not processed.
3. **CLI value**: if you run it with the following command:
    ```bash
    python -m luigi --module test_path_param MyTask --local-scheduler --MyTask-my-path a_cli_path
    ```
    The CLI value will be used and you will see this: `<class 'pathlib.PosixPath'> cli_path`
    So the constructor value is processed.
4. **value from CFG file**: if you run it with the following command with an entry for `MyTask` in `luigi.cfg`:
    ```bash
    python -m luigi --module test_path_param MyTask --local-scheduler
    ```
    The value from the CFG file is used and you will see this: `<class 'pathlib.PosixPath'> path_from_cfg`
    So the CFG value is processed.

While using the `normalize()` method the parameter is processed in all these cases.

krabo0om:
Thank you @adrien-berchet for this detailed example. I was never quite sure about those methods and now I see that I only used it with a value from CFG file (case 4).

adrien-berchet:
Yeah, CFG files and CLI are clearly the most common use cases.

dlstadther (approved):
lgtm. Thanks!

adrien-berchet:
Thanks :-)

PR #3130, opened 2021-12-09
OptionalParameter in ExternalPythonProgramTask

oinegue:
As of now, `ExternalPythonProgramTask` parameters are simple `Parameter`.
But the default for both of them is `None` and they are used only if they are not None.
If the value for any of them is not provided, a warning is raised, for example:
`UserWarning: Parameter "extra_pythonpath" with value "None" is not of type string.`

Therefore I think that the appropriate type is OptionalParameter instead.

PS: First pull request for me, I hope I'm doing it right. Otherwise please help me understand what's wrong. Thanks


dlstadther (approved):
lgtm

PR #3135, opened 2021-12-30
Enable updating the default config parser even after loading luigi

tarohi24:
## Description

This PR enables updating the default config parser by updating `LUIGI_CONFIG_PARSER`.

## Motivation and Context

You can change the default config parser by updating the environment variable `LUIGI_CONFIG_PARSER`.
However, you cannot change it after importing `luigi.configuration` because the environment variable is evaluated when loading the module.
`get_config` is used in many modules, many of which don't allow passing the `parser` argument to it.
So I changed `luigi.configuration` to allow users to change the default parser even after the module is loaded.

I added a test : https://github.com/spotify/luigi/compare/master...tarohi24:enable_changing_default_parser?expand=1#diff-057c25ffc6418870df3cf372c8de2cf04dd57256dad785ca96ff3eb3046e3ee3R101


dlstadther (approved):
lgtm

PR #3137, opened 2022-01-14
Testing Editing of Github Action Config Checks from PR

dlstadther:
## Description

Testing how Github Actions behaves when PR checks are changed

## Motivation and Context

Encountering issues with other PRs which remove or rename part of the Tox env matrix which cause some of the "checks" are never returned and blocks the PR from merge.

(this pr)

### DIFF
--- a/.github/workflows/pythonbuild.yml
+++ b/.github/workflows/pythonbuild.yml
@@ -76,12 +76,12 @@ jobs:
     strategy:
       matrix:
         include:
-          - python-version: 3.6
-            tox-env: py36-postgres
-          - python-version: 3.7
-            tox-env: py37-postgres
-          - python-version: 3.8
-            tox-env: py38-postgres
+          #- python-version: 3.6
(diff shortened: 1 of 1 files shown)


dlstadther:
When a Github Action "check" is removed from workflow config (or renamed), PR checks continue to wait to their status to be reported. Will wait indefinitely. See screenshot for current state of PR:

luigi_gh_actions_waiting_on_removed_master_checks

---

How do we configure PRs so that their checks are based on the workflow file from their branch only?

dlstadther:
Based on google searching, it sounds like the `15 expected checks` were manually configured by a repo admin within the Branch Protection Rules section of the repo settings. In order to remove (including renaming) required checks, an admin will have to update those settings. [ref](https://github.community/t/expected-waiting-for-status-to-be-reported/18001/2)

---

And while i have permissions to merge PRs, I do not have repo settings access. There is no more i can do in this area.

PR #3140, opened 2022-01-21
support Python 3.10

hirosassa:
## Description

This PR adds Python 3.10 CI workflow to support Python 3.10.

## Motivation and Context

To follow latest Python version.

Ran tox on local Mac and Docker (ubuntu 20.04)

### DIFF
--- a/test/lock_test.py
+++ b/test/lock_test.py
@@ -18,7 +18,6 @@
 import os
 import subprocess
 import tempfile
-import time
 import mock
 from helpers import unittest
 
@@ -32,13 +31,12 @@ luigi.notifications.DEBUG = True
 pcmd_retry = retry(retry)
 
 
-def _is_empyty(s):
-    return s == ""
-
-
 class TestCmd(unittest.TestCase):
 
     def test_getpcmd(self):
+        def _is_empyty(cmd):
+            return cmd == ""
+
         # for CI stability, add retring
         @retry(retry=retry_if_result(_is_empyty), wait=wait_exponential(multiplier=0.2, min=0.1, max=3), stop=stop_after_attempt(3))
         def _getpcmd(pid):
@@ -70,6 +68,9 @@ class LockTest(unittest.TestCase):
         os.rmdir(self.pid_dir)
 
     def test_get_info(self):
+        def _is_empyty(pid, cmd, pid_file):
+            return cmd == ""
+
         # for CI stability, add retring
         @retry(retry=retry_if_result(_is_empyty), wait=wait_exponential(multiplier=0.2, min=0.1, max=3), stop=stop_after_attempt(3))
         def _get_info(pid_dir, pid):
--- a/test/lock_test.py
+++ b/test/lock_test.py
@@ -22,24 +22,35 @@ import time
 import mock
 from helpers import unittest
 
+from tenacity import retry, retry_if_result, stop_after_attempt, wait_exponential
 import luigi
 import luigi.lock
 import luigi.notifications
 
 luigi.notifications.DEBUG = True
 
+pcmd_retry = retry(retry)
+
+
+def _is_empyty(s):
+    return s == ""
+
 
 class TestCmd(unittest.TestCase):
 
     def test_getpcmd(self):
+        # for CI stability, add retring
+        @retry(retry=retry_if_result(_is_empyty), wait=wait_exponential(multiplier=0.2, min=0.1, max=3), stop=stop_after_attempt(3))
+        def _getpcmd(pid):
+            return luigi.lock.getpcmd(pid)
+
         if os.name == 'nt':
             command = ["ping", "1.1.1.1", "-w", "1000"]
         else:
             command = ["sleep", "1"]
 
         external_process = subprocess.Popen(command)
-        time.sleep(0.42)  # need this sleep for CI running on GitHub from Python 3.10
-        result = luigi.lock.getpcmd(external_process.pid)
+        result = _getpcmd(external_process.pid)
 
         self.assertTrue(
             result.strip() in ["sleep 1", '[sleep]', 'ping 1.1.1.1 -w 1000']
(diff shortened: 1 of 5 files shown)


dlstadther:
@hirosassa what do we need to solve these 3.10 errors? There have been a couple 3.10 related PRs that i've merged in the past few days (or so) - a deprecation warning and the tenacity version upgrade. Are there additional changes that we need to make before updating this branch and retesting?

So sorry for my infrequent review and feedback here

hirosassa:
@dlstadther Thanks for your comment! I would like to discuss about the CI failure.
The CI error on Python 3.10 occurred only on GitHub Actions (not reproduced on local Docker and Mac).

I found the problem comes from the code below:
https://github.com/spotify/luigi/blob/master/luigi/lock.py#L65-L66
actual `fh` contains nothing, it is expected that it contains current executed command like `echo hello`.

I tried adding retry using `tenacity` like here, and CI sometimes pass but it also fails "sometime".
https://github.com/hirosassa/luigi/pull/1/files#diff-e126af1be65ab218abe136e1248d7ec7fca5e00f0c3322d3c1eef50ff4389b20R82-R85

Do you have any ideas?

rschmidtner:
Are there any plans to make luigi Python 3.10 compatible in the future?

ravwojdyla:
@hirosassa thanks for starting this PR! Regarding the cmdline issue from https://github.com/spotify/luigi/pull/3140#issuecomment-1065871000, have you tried to add a short sleep between these two lines (e.g. `time.sleep(0.42)`):

https://github.com/spotify/luigi/blob/38b0c2b8d400f6fc57ebbdc4064aadd4f3d7b612/test/lock_test.py#L39-L40

hirosassa:
@ravwojdyla Thanks for your comment! I added sleep at [c5ff943](https://github.com/spotify/luigi/pull/3140/commits/c5ff9439a19e6e71d8a5e64d326b50509971bee9) but it failed.

ravwojdyla:
> I added sleep at [c5ff943](https://github.com/spotify/luigi/pull/3140/commits/c5ff9439a19e6e71d8a5e64d326b50509971bee9) but it failed.

`test_getpcmd` now did NOT fail, `test_get_info` failed, which is the other test that fail in the past, probably the same reason. Can you add the sleep in the other test as well, here:

https://github.com/spotify/luigi/blob/38b0c2b8d400f6fc57ebbdc4064aadd4f3d7b612/test/lock_test.py#L61-L62

hirosassa:
@ravwojdyla Wow! Thank you. Finally all the tests are succeed.

ravwojdyla:
> Finally all the tests are succeed.

@hirosassa nice, good job! `time.sleep(0.42)` was somewhat arbitrary and very likely an overkill, some milliseconds would probably be sufficient. If you want to, in the comment you can mention that this sleep is "necessary" to make sure the test can read populated `/proc/*/cmdline` on linux.

lallea:
The sleep is fragile. On a bad day, it won't be enough and a longer sleep slows down tests. I suggest moving the `getpcmd` call to a separate function and wrapping it with `tenacity.retry` with a timeout of tens of seconds.

hirosassa:
@lallea @ravwojdyla Thanks for your comment. I agree with your opinion. I'll try it.

ravwojdyla:
> I suggest moving the getpcmd call to a separate function and wrapping it with tenacity.retry with a timeout of tens of seconds.

@lallea that sounds good. afaiu you are suggesting to retry until you can get the cmd line from `/proc/*/cmdline`, just want to point out that there can be multiple reasons why `cmdline` file exists BUT it's empty. At least two reasons come to my mind: it can be because the cmdline has not be written yet (might be buffered, in process etc), or the process is a zombie. If you add `retry` with "tens of seconds", in the case of the zombie process (which I do not know how often that will be the case), you might actually wait for those "tens of seconds", and that would be in the "prod code", not test. So there's a potential downside to that solution.

lallea:
> > I suggest moving the getpcmd call to a separate function and wrapping it with tenacity.retry with a timeout of tens of seconds.
> 
> @lallea that sounds good. afaiu you are suggesting to retry until you can get the cmd line from `/proc/*/cmdline`, just want to point out that there can be multiple reasons why `cmdline` file exists BUT it's empty. At least two reasons come to my mind: it can be because the cmdline has not be written yet (might be buffered, in process etc), or the process is a zombie. If you add `retry` with "tens of seconds", in the case of the zombie process (which I do not know how often that will be the case), you might actually wait for those "tens of seconds", and that would be in the "prod code", not test. So there's a potential downside to that solution.

I'm suggesting breaking out the call to a new function in `lock_test.py` and put the retry there, not in production code. I agree that it would be appropriate in prod code.

hirosassa:
@lallea @ravwojdyla I added retry on `getpcmd` and `get_info` in test code. Please check.

test/lock_test.py
```diff
@@ -21,23 +21,34 @@
...
 from helpers import unittest
 
+from tenacity import retry, retry_if_result, stop_after_attempt, wait_exponential
 import luigi
 import luigi.lock
 import luigi.notifications
 
 luigi.notifications.DEBUG = True
 
+pcmd_retry = retry(retry)
```
ravwojdyla, line 31: Is this used for anything?

ravwojdyla (dismissed):

test/lock_test.py
hirosassa: Removed in [e8a957c](https://github.com/spotify/luigi/pull/3140/commits/e8a957cfa01f4f62f4df42359420e9f4815f437f)

ravwojdyla (approved):

hirosassa:
@dlstadther Hi Dillon! Could you review this PR?

dlstadther (approved):

hirosassa:
@dlstadther Could you release newer version on pypi? (there's many new feature released)

dlstadther:
Apologies @hirosassa ; I have been a bit consumed as of late. @honnix , could Spotify devs prep and release a new Luigi version?

honnix:
@dlstadther We will take care of it. Looking at https://github.com/spotify/luigi/pull/3220, it seems we should get that in before dropping a new release.

dlstadther:
> @dlstadther We will take care of it. Looking at #3220, it seems we should get that in before dropping a new release.

Thanks @honnix !

PR #3142, opened 2022-01-29
Added support for pg8000 as the PostgreSQL driver

rscarvalho:
## Description

Added support for the `pg8000` library as the PostgreSQL driver.

## Motivation and Context

I'm trying to ingest data into a Gooel Cloud SQL instance without having to use `cloud_sql_proxy` and found this library: https://github.com/GoogleCloudPlatform/cloud-sql-python-connector#cloud-sql-connector-for-python-drivers, however that only has support for `pg8000` to connect to PostgreSQL.

Both `pg8000` an `psycopg2` implement [DB API 2.0 (a.k.a. PEP-249)](https://www.python.org/dev/peps/pep-0249/) so most of the operation should be supported; I've just made a few changes to make sure we handle exceptions in a cross-library way.

I ran my jobs with these changes and they worked fine.


narape (approved):

narape (dismissed):

dlstadther (commented):
lgtm; i'm going to update with master and let tests run. if all pass, i'll merge

dlstadther (approved):
only check failure was codecov on patch additions. Not worried about that. Merging!

PR #3143, opened 2022-02-02
Fix Python 3 incompatible code in LSF contrib module

chrschorn:
## Description
The LSF contrib module does not seem to work on Python 3 (tested on 3.8) due to what looks like a Python 2.7-focussed implementation.

The current implementation treats outputs of `subprocess.Popen` and outputs of `pickle.dumps` as strings, which throws errors when doing string modifications later in the same file. Both of these output in bytes by default though, since at least Python 3.5 (`pickle.dumps` returned a string in Python 2.x). Adding `universal_newlines=True` makes Popen output strings again.

I also removed code that seemed to do some pickle-string (again, Python 2.7 behavior) manipulation for which I couldn't find documentation and which does not work in 3.x

This should also fix this (stale, closed) issue: https://github.com/spotify/luigi/issues/2873

I ran a few LSF test jobs with this change on Python 3.8 and it works.

### DIFF
--- a/luigi/contrib/lsf.py
+++ b/luigi/contrib/lsf.py
@@ -83,7 +83,7 @@ def track_job(job_id):
     """
     cmd = "bjobs -noheader -o stat {}".format(job_id)
     track_job_proc = subprocess.Popen(
-        cmd, stdout=subprocess.PIPE, shell=True, text=True)
+        cmd, stdout=subprocess.PIPE, shell=True, universal_newlines=True)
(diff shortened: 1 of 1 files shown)


chrschorn:
The LSF module has additional issues that would need additional work (e.g. luigi import fails after loading by pickle, running with `luigi.build` in `__main__` is broken in Python 3.x). In addition, tests for interaction with the LSF system and different versions would be needed to truly make this maintainable.

PR #3146, opened 2022-02-18
Allow for a custom MetricsCollector class to be loaded at runtime

nipper:
## Description
This adds in the ability to load a custom subclass of `MetricsCollector` at startup. This is so that if you want to have a private implementation of MetricsCollector you can.

It does this by adding the value of `-1` to MetricsCollectors which represents `Custom` and an optional `metrics_custom_import` that is an optional string in the form of `module.sub_module.ClassName`. When both are provided it imports the class and returns an instance of it. 

## Motivation and Context
I need to have a private custom metrics collector that I'd like to use and this provides a backwards compatible way to do it.

I've included a new unit test: `test_custom_metrics_collector` in `test/scheduler_test.py`

Since checks are blocked w/o maintainer approval I ran them here: [https://github.com/nipper/luigi/pull/1/checks](https://github.com/nipper/luigi/pull/1/checks)


dlstadther (approved):
Nice and clever! Sorry for such delay in review and merge

nipper:
Thanks! No worries!

PR #3152, opened 2022-03-23
oracle support

ceyhunkerti:
This adds support for oracle database

## Description
- Added `luigi/contrib/oracle.py` and `test/contrib/oracle_test.py`

## Motivation and Context
This adds yet another database support to `luigi` and extends its functionality

## Test
```sh
tox -e py39 test/oracle_test.py
```

### DIFF
--- a/luigi/contrib/oracle.py
+++ b/luigi/contrib/oracle.py
@@ -1,0 +1,299 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2012-2015 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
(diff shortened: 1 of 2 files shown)


luigi/contrib/oracle.py
```diff
@@ -0,0 +1,299 @@
...
+
+        connection = self.output().connect()
+
+        # attempt to copy the data into oracle
+        # if it fails because the target table doesn't exist
+        # try to create it by running self.create_table
+        for attempt in range(2):
+            try:
+                cursor = connection.cursor()
+                print("caling init copy...")
```
dlstadther, line 266: can we remove this print statement? or move to a logging statement?

luigi/contrib/oracle.py
```diff
@@ -0,0 +1,299 @@
...
+                break
+
+        # mark as complete in same transaction
+        self.output().touch(connection)
+        connection.commit()
+        connection.close()
+
+    @property
+    def bulk_size(self):
+        return 10000
```
dlstadther, line 287: Should this default be configurable from the `[oracle]` cfg section? I'm fine if you think it's best to leave here and override by inheriting this class.

luigi/contrib/oracle.py
ceyhunkerti: ✅  removed

luigi/contrib/oracle.py
ceyhunkerti: actually took it from `mysql` connector its like this in that connector. changed to 
```py
@property
    def bulk_size(self):
        return int(luigi.configuration.get_config().get("oracle", "bulk-size", 10000))
```

PR #3154, opened 2022-04-12
Add support for __init_subclass in Task metaclass.

umartin:
## Description
Since python 3.6 and PEP 487 metaclasses must pass kwargs to super. Otherwise the __init_subclass__ hook fails.

## Motivation and Context

For a detailed discussion regarding the same issue in abc see:
https://bugs.python.org/issue29581
And the patch for the issue above:
https://github.com/python/cpython/commit/6fb12b5c43945f61f3da82e33eafb4ddae2296ee

Unit test added

### DIFF
--- a/luigi/task_register.py
+++ b/luigi/task_register.py
@@ -54,7 +54,7 @@ class Register(abc.ABCMeta):
     ambiguous task name (two :py:class:`Task` have the same name). This denotes
     an error."""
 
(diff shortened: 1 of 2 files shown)


umartin:
All checks now pass after master was fixed

dlstadther (approved):
LGTM. Thanks!

Issue #3167, opened 2022-05-27
Misnamed parameter "include_significant" on method Task.get_param_names 


MartinHeimsoth-V:
The classmethod `Task.get_param_names` has either a really confusingly misnamed parameter `include_significant`, or the implementation has a bug.
Example:
```python
>>> class T(Task):
...     significant = Parameter(significant=True)
...     insignificant = Parameter(significant=False)
... 
>>> T.get_param_names()
['significant']
>>> T.get_param_names(include_significant=True)
['significant', 'insignificant']
```

So, I'd suggest to rename the parameter to `include_insignificant` (*in*significant), or to change the implementation that on default it only returns the insignificant parameters, which would be weird though.

teranao-ai:
@MartinHeimsoth-V 
 I was wondering the same thing.

Regarding your suggestions,
> So, I'd suggest to rename the parameter to include_insignificant (insignificant), or to change the implementation that on default it only returns the insignificant parameters, which would be weird though.

The former suggestion sounds better. I see the `include_significant` is just a typo and expected behavior is to return only significant parameters on default.
The later suggestion of changing default behavior is a major change that can cause some problems in existing codes.

@daveFNbuck 
I saw you were involved in the latest revision of this method. Do you have any idea on the expected behavior?

I made a related pull-request (https://github.com/spotify/luigi/pull/3168) before seeing this issue.
This request is accidently equivalent to the former suggestion of renaming `include_significant`.

PR #3168, opened 2022-05-29
Fix typos in Task.get_param_names: include_significant -> include_insignificant

nayopu:
Fix typo `include_significant`  -> `include_insignificant` in `Task.get_param_names`

## Description

Fix typos  `include_significant`  -> `include_insignificant` in `Task.get_param_names` so the names match the actual behavior.

## Motivation and Context

No context. 
The typos are inconsistent with the actual behavior.

Not tested.
But there will be no side effects as just fixing typos and `include_significant` only appears in the fixed method.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -373,8 +373,8 @@ class Task(metaclass=Register):
         return [name for name, p in cls.get_params() if p._is_batchable()]
 
     @classmethod
-    def get_param_names(cls, include_significant=False):
-        return [name for name, p in cls.get_params() if include_significant or p.significant]
+    def get_param_names(cls, include_insignificant=False):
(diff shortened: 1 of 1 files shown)


teranao-ai:
This problem is also mentioned in #3167 suggesting:
> So, I'd suggest to rename the parameter to include_insignificant (insignificant), or to change the implementation that on default it only returns the insignificant parameters, which would be weird though.

This request is equivalent to the former suggestion of *renaming* in #3167. 
I see the `include_significant` is just a typo and expected behavior is to return only significant parameters on default, as the my code in this request does.
I also add the later suggestion of *changing default behavior* in #3167 is a major change that can cause some problems in existing codes.

PR #3170, opened 2022-06-03
Issue a warning when a parameter is not consumed by a task

adrien-berchet:
## Description
After consuming the parameters I add a check to ensure that the task consumed all given parameters. If it is not the case a warning is raised.

## Motivation and Context

In one of our workflows, several tasks have similar parameters but not all. This confused some users that tried to use such parameter with irrelevant tasks and the users were confused. Adding warnings could help users to understand what's happening. It can also help finding typos in parameter names (if they have a default value `luigi` does not complain and just use the default value).

I added a unit test.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -95,6 +95,10 @@ class OptionalParameterTypeWarning(UserWarning):
     pass
 
 
+class UnconsumedParameterWarning(UserWarning):
+    """Warning class for parameters that are not consumed by the task."""
+
+
 class Parameter:
     """
     Parameter whose value is a ``str``, and a base class for other parameter types.
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -32,9 +32,11 @@ import functools
 
 import luigi
 
+from luigi import configuration
 from luigi import parameter
 from luigi.task_register import Register
 from luigi.parameter import ParameterVisibility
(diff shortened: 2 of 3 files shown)


dlstadther (approved):
Handy! Thanks for the contribution!

adrien-berchet:
With pleasure :-)

Do you know when the next release is planned? The last one is pretty old and there are several things in the master branch we are interested in.

dlstadther:
@adrien-berchet , the current release is quite old. I haven't done a Luigi release myself in years, but I can try to get one out. If so, it'll likely be 3.1.0.

Out of curiosity, are y'all currently using `3.0.3`, `master`, or a fork?

adrien-berchet:
Ah cool, that would be great, thanks!

Currently we use the 3.0.3. But we have a repository (https://github.com/BlueBrain/luigi-tools) in which we add everything we need and then we submit PRs here for the generic features that might be interesting for others.

dlstadther:
I've gotten the release started at https://github.com/spotify/luigi/pull/3171

I'll have to get a Spotify employee to approve the version bump before i can merge. At that point, i'll proceed with updating PyPi (hopefully i still have access to do so).

adrien-berchet:
Ok, thank you very much!

PR #3172, opened 2022-06-08
Use PEP593 to support typing of parameters

adriangb:
This uses PEP593 Annotations to support typing of parameters in a way that will make mypy & co treat the instance variables correctly.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -29,6 +29,7 @@ import hashlib
 import re
 import copy
 import functools
+import sys
 
 import luigi
 
@@ -47,6 +48,30 @@ TASK_ID_TRUNCATE_PARAMS = 16
 TASK_ID_TRUNCATE_HASH = 10
 TASK_ID_INVALID_CHAR_REGEX = re.compile(r'[^A-Za-z0-9_]')
 _SAME_AS_PYTHON_MODULE = '_same_as_python_module'
+
+
+if sys.version_info < (3, 9):
+    try:
+        from typing_extensions import Annotated, get_args, get_origin, get_type_hints
+    except ImportError:
+        Annotated = get_args = get_origin = get_type_hints = None
+else:
+    from typing import Annotated, get_args, get_origin, get_type_hints
+
+
+def _get_pep539_annotations(tsk_cls):
+    if Annotated is None:
+        return []
+    params = []
+    type_hints = get_type_hints(tsk_cls, include_extras=True)
+    for param_name, annotation in type_hints.items():
+        if get_origin(annotation) is Annotated:
+            args_iter = iter(get_args(annotation))
+            next(args_iter)
+            for metadata in args_iter:
+                if isinstance(metadata, Parameter):
+                    params.append((param_name, metadata))
+    return params
 
 
 def namespace(namespace=None, scope=''):
@@ -363,7 +388,15 @@ class Task(metaclass=Register):
             param_obj = getattr(cls, param_name)
             if not isinstance(param_obj, Parameter):
                 continue
+            params.append((param_name, param_obj))
 
+        pep593_params = _get_pep539_annotations(cls)
+        for param_name, param_obj in pep593_params:
+            try:
+                default_val = getattr(cls, param_name)
+                param_obj._default = default_val
+            except AttributeError:  # no default value
+                pass
             params.append((param_name, param_obj))
 
         # The order the parameters are created matters. See Parameter class
--- a/test/parameter_test.py
+++ b/test/parameter_test.py
@@ -16,6 +16,7 @@
 #
 
 import datetime
+from typing import Annotated
 
 from helpers import with_config, LuigiTestCase, parsing, in_parse, RunOnceTask
 from datetime import timedelta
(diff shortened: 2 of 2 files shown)


test/parameter_test.py
```diff
@@ -147,6 +153,15 @@ class NoopTask(luigi.Task):
     pass
 
 
+class PEP593Task(luigi.Task):
+    param: Annotated[int, luigi.IntParameter()]
```
honnix, line 157: What happens if a user specifies `Annotated[bool, luigi.IntParameter()]`? It seems we will just ignore it. Is that right?

honnix: I meant accidentally using `bool`  with `luigi.IntParameter`.

adriangb: Yeah sorry mistyped above. Type checkers will only care about the type annotation but Luigi/runtime will only care about the Luigi parameter. So in your example that would be an integer at runtime, meaning the run time behavior would be exactly the same as it is now. And because bool is a subtype of int there isn't any expression the type checker would flag. Now if the user did `Annotated[str, IntegerParameter]` the value would once again be an integer at runtime but their type checker would be very upset when they try to do `self.parameter / 1`.

adriangb:
Unrelated to this PR. You may want to change your workflow approval settings in Settings -> General -> Actions to this so that users like me that have never contributed to this repo but are not new to GitHub itself don't require approval.

Screen Shot 2022-06-08 at 8 58 53 AM

luigi/task.py
```diff
@@ -363,7 +388,15 @@ def get_params(cls):
             param_obj = getattr(cls, param_name)
             if not isinstance(param_obj, Parameter):
                 continue
+            params.append((param_name, param_obj))
 
+        pep593_params = _get_pep539_annotations(cls)
+        for param_name, param_obj in pep593_params:
+            try:
+                default_val = getattr(cls, param_name)
+                param_obj._default = default_val
```
adriangb, line 397: I do want to verify that this is safe to do. Do I need to do `param_obj = copy.copy(param_obj)` first?

The reason for doing this is that the `luigi.Parameter(default=...)` option only exists because the default is taken up by the parameter itself. But with this new syntax you can set a default value directly.

adriangb:
@honnix I merged in master, tests are passing now. Could you take another look at this?

luigi/task.py
```diff
@@ -49,6 +50,30 @@
...
+if sys.version_info < (3, 9):
+    try:
+        from typing_extensions import Annotated, get_args, get_origin, get_type_hints
+    except ImportError:
+        Annotated = get_args = get_origin = get_type_hints = None
+else:
+    from typing import Annotated, get_args, get_origin, get_type_hints
+
+
+def _get_pep539_annotations(tsk_cls):
```
rafaelraposospot, line 62: I'm not a fan of having the pep593 in the method name and also in the case there's a typo:

```suggestion
def _get_pep593_annotations(tsk_cls):
```

luigi/task.py
```diff
@@ -49,6 +50,30 @@
...
+    if Annotated is None:
+        return []
+    params = []
+    type_hints = get_type_hints(tsk_cls, include_extras=True)
+    for param_name, annotation in type_hints.items():
+        if get_origin(annotation) is Annotated:
+            args_iter = iter(get_args(annotation))
+            next(args_iter)
+            for metadata in args_iter:
+                if isinstance(metadata, Parameter):
```
lupasarin, line 72: In case it is not a Parameter, we just ignore the type hint. Shouldn't we instead give feedback to the user at least as a warning telling them that this hint won't be used because it is not valid?

luigi/task.py
```diff
@@ -363,7 +388,15 @@ def get_params(cls):
...
                 continue
+            params.append((param_name, param_obj))
 
+        pep593_params = _get_pep539_annotations(cls)
+        for param_name, param_obj in pep593_params:
+            try:
+                default_val = getattr(cls, param_name)
+                param_obj._default = default_val
+            except AttributeError:  # no default value
+                pass
```
lupasarin, line 399: Should we notify the user, e.g. as a warning, that something went wring and the parameter won't be used?

test/parameter_test.py
rafaelraposospot: Thank you for the explanation @adriangb. I'm concerned about not checking the type earlier and only having errors when trying to use the parameter.

Should we check earlier to enforce types if they're being annotated, concerned that if we check later when things are actually used it will difficult the debug process for the developer?

test/parameter_test.py
```diff
@@ -1329,3 +1335,29 @@ def test_exists(self, path_parameter):
                 luigi.build([path_parameter["cls"]()], local_scheduler=True)
         else:
             assert luigi.build([path_parameter["cls"]()], local_scheduler=True)
+
+
+if Annotated is None:
+    PEP593Task = luigi.Task
+else:
+    class PEP593Task(luigi.Task):
+        param: Annotated[int, luigi.IntParameter()]
```
lupasarin, line 1344: I think it would be nice to check cases like `param: Annotated[int, luigi.BoolParameter()]` like suggested in another comment

tgy:
@adriangb why was this closed?

adriangb:
Because I no longer work at the company that used Luigi and don't have time to push this forward given the pace it was going at. Feel free to take it over.

tgy:
@adriangb ok thanks!

PR #3175, opened 2022-06-10
Close requests.Socket in RemoteScheduler before exiting (#3173)

### LINKED ISSUE #3173: Unclosed socket from luigi.rpc module
**Description**
Luigi does not close all connections to remote scheduler:
```
ResourceWarning: unclosed <socket.socket fd=10, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=(...), raddr=(...)>
```
This is probably connected to not closed `requests.Session` objects in `RequestsFetcher`.
https://github.com/spotify/luigi/blob/78145d669cd0658a52e1158a3fd1ca10e7e3b028/luigi/rpc.py#L100-L118

**Packages**
```
urllib3==1.26.9
requests==2.27.1
luigi==3.0.3 (same code on master)
```

starhel:
## Description
If using `luigi.build` and `sys.exit`, `ResourceWarning` on unclosed socket is raised. As it can be easily fixed, I've added `requests.Session.close` call before exiting. 

## Motivation and Context
Fixes #3173

I've tested it using code below under py38. Adding solid unit/integration test requires running remote scheduler and I will need guidance if one is needed. 
```
import sys
from time import sleep

import luigi


class Task(luigi.Task):
    p = luigi.IntParameter()

    def run(self):
            sleep(1)
            f.write(str(self.p))

    def output(self):
        return luigi.LocalTarget(f'file_{self.p}.txt')


if __name__ == '__main__':
    sys.exit(0 if luigi.build([Task(p=2)], local_scheduler=False) else 1)
```

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -173,6 +173,8 @@ def _schedule_and_run(tasks, worker_scheduler_factory=None, override_defaults=No
         success &= worker.run()
     luigi_run_result = LuigiRunResult(worker, success)
     logger.info(luigi_run_result.summary_text)
+    if hasattr(sch, 'close'):
+        sch.close()
     return luigi_run_result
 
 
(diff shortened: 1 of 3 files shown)


dlstadther (approved):
lgtm! Thanks!

PR #3176, opened 2022-06-10
Silent UnconsumedParameterWarning for core.autoload_range

starhel:
## Description
`autoload_range` set in `luigi.cfg` emits `UnconsumedParameterWarning` as value is used only in `luigi.__init__`. I've added internal field `_ignore_unconsumed` to ignore this parameter in `core` class. 

## Motivation and Context
My motivation in to silent all warnings emitted by luigi in my developed code as it's add some noise in tests. 

I've run simple graph with this change to ensure that's warning is no longer emitted.

### DIFF
--- a/luigi/interface.py
+++ b/luigi/interface.py
@@ -49,6 +49,7 @@ class core(task.Config):
     This is arguably a bit of a hack.
     '''
     use_cmdline_section = False
+    _ignore_unconsumed = {'autoload_range'}
 
     local_scheduler = parameter.BoolParameter(
         default=False,
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -439,6 +439,8 @@ class Task(metaclass=Register):
         if task_family in conf.sections():
             for key, value in conf[task_family].items():
                 composite_key = f"{task_family}_{key}"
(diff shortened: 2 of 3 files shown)


poofeg:
Potentially the same is required for the no_configure_logging option.

dlstadther:
Where do you feel would be a good place to document this functionality? Are there other core-related parameters that should also have warnings suppressed?

starhel:
There are probably more of them: 
* `core.hdfs_tmp_dir` (deprecated)
* `core.history_filename` (probably shouldn't be in core)
* `core.max_shown_tasks` (it's mentioned in docs in core, but it's in scheduler config on master)
* `core.max_graph_nodes` (same)
* `core.rpc_connect_timeout` (used in RemoteScheduler class, should be rewritten to use core class)
* `core.rpc_retry_attempts` (same)
* `core.rpc_retry_wait` (same)
* `core.no_configure_logging` (should be rewritten to use core class)

This is only `core` analysis, so there are probably some other config classes with options defined only in the config file. Unfortunately, for now, I have no time to prepare a proper refactor, so I can only add more options to the ignored list.

PR #3178, opened 2022-06-21
Add worker config to cache task completion results.

riga:
## Description

This PR adds a new worker config `cache_task_completion` which, when enabled (default is `False` of course), leads to tasks' successful `complete()` results being cached by the worker. In short, this can save resources in case tasks have a large number of dynamic dependencies that are yielded in several stages (see below), and the tasks' `complete()` calls are somewhat expensive (e.g. when file targets are remote).

## Motivation and Context

We sometimes have k's of tasks being yielded through dynamic dependencies, and we only use files on remote locations. Therefore we want to limit the number of API calls and would really benefit from this option. Also, in our case we are sure that once a task is complete, it never toggles to being incomplete again, so that this kind of caching is 100% safe (and I'm sure that the opposite is rather rare). Example:

```python
class MyTask(luigi.Task):
    ...
    def run(self):
        # first round
        deps_1 = [DepA1(), DepB1(), ..., DepZ1()]
        outputs_1 = yield deps_1

        # second round
        # (with some artificial relation to outputs_1 to justify that two yields are required ...)
        deps_1 = [dep(param=x) for dep, x in zip([DepA2, DepB2, ..., DepZ2], outputs_1)]
        yield deps_2
```

When running, `deps_1` are yielded twice, and the completion checks are triggered multiple times:

1. when `yield`ed the first time and reaching [this point](https://github.com/spotify/luigi/blob/a295041d9c344135ef7f8d41e26f6fdbe6e55c58/luigi/worker.py#L149),
2. [before being added to the tree](https://github.com/spotify/luigi/blob/a295041d9c344135ef7f8d41e26f6fdbe6e55c58/luigi/worker.py#L748)
3. after each dep ran, in the [post-run check](https://github.com/spotify/luigi/blob/a295041d9c344135ef7f8d41e26f6fdbe6e55c58/luigi/worker.py#L195),
4. when `yield`ed the second time at the same position as 1.5. 

And for every additional round of yielding, there will be another check (same as 4.). With the new config enabled, the checks would only run for 1. and 2..

I added a test to `worker_test.py` that checks the number of `complete()` calls of tasks being yielded as dynamic dependencies, with and without the new config. The documentation is updated as well.

### DIFF
--- a/test/worker_test.py
+++ b/test/worker_test.py
@@ -460,8 +460,7 @@ class WorkerTest(LuigiTestCase):
                 self.has_run = True
 
         # test the enabled feature
-        with Worker(scheduler=self.sch, worker_id='2') as w:
-            w._config.cache_task_completion = True
+        with Worker(scheduler=self.sch, worker_id='2', cache_task_completion=True) as w:
             b0 = B(i=0)
             a0 = A(i=0)
             a1 = A(i=1)
@@ -480,8 +479,7 @@ class WorkerTest(LuigiTestCase):
             self.assertEqual(a2.complete_count, 3)
 
         # test the disabled feature
-        with Worker(scheduler=self.sch, worker_id='2') as w:
-            w._config.cache_task_completion = False
+        with Worker(scheduler=self.sch, worker_id='2', cache_task_completion=False) as w:
             b10 = B(i=10)
             a10 = A(i=10)
             a11 = A(i=11)
--- a/test/worker_test.py
+++ b/test/worker_test.py
@@ -459,7 +459,7 @@ class WorkerTest(LuigiTestCase):
                 yield A(i=self.i + 2)
                 self.has_run = True
 
-        # test the enabled feature
+        # test with enabled cache_task_completion
         with Worker(scheduler=self.sch, worker_id='2', cache_task_completion=True) as w:
             b0 = B(i=0)
             a0 = A(i=0)
@@ -478,7 +478,7 @@ class WorkerTest(LuigiTestCase):
             self.assertEqual(a1.complete_count, 2)
             self.assertEqual(a2.complete_count, 2)
 
-        # test the disabled feature
+        # test with disabled cache_task_completion
         with Worker(scheduler=self.sch, worker_id='2', cache_task_completion=False) as w:
             b10 = B(i=10)
             a10 = A(i=10)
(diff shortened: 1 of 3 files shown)


test/worker_test.py
```diff
@@ -434,6 +434,88 @@ def requires(self):
...
+            self.assertEqual(b10.complete_count, 1)
+            self.assertEqual(a10.complete_count, 0)
+            self.assertEqual(a11.complete_count, 0)
+            self.assertEqual(a12.complete_count, 0)
+            w.run()
+            # the complete methods of a's yielded first in b's run method were called more often
+            self.assertEqual(b10.complete_count, 1)
+            self.assertEqual(a10.complete_count, 5)
+            self.assertEqual(a11.complete_count, 4)
+            self.assertEqual(a12.complete_count, 3)
```
dlstadther, line 498: i'm a little unclear why these `complete_count` values for each of the tasks differ in quantity. Could you clarify that for me?

dlstadther: Similarly, i'm confused as to why the assertions below (with `check_complete_on_run=True`) resulted in larger `complete_count` quantities.

dlstadther (commented):
I understand the purpose of this change and I think it's great! However, I'm a bit confused by some of the unittest assertions. Hoping to get a little clarity in my understanding of those values.

test/worker_test.py
riga: > i'm a little unclear why these complete_count values for each of the tasks differ in quantity. Could you clarify that for me?

Sure!

The complete count with `cache_task_completion=True` of (5, 4, 3) for (`a10`, `a11`, `a12`) is due to luigi's assumption of idempotence of `run()` methods that yield dynamic dependencies in that the worker invokes `run()` and in case it's a generator, it get's the next result and

- if it's a bunch of already complete tasks (some of the completeness checks is happening here), it gets the next generator result, or
- if it's a bunch of tasks of which at least one is not complete yet, it adds all of them to the tree and forgets about the state of the generator.

([code here](https://github.com/spotify/luigi/blob/a295041d9c344135ef7f8d41e26f6fdbe6e55c58/luigi/worker.py#L139-L154)) The yielding task is placed back to the tree in PENDING state, too. And when it's started again later on, the entire procedure is triggered again, leading to a new generator, but now with the previously incomplete bunch being complete. Therefore, completion checks of tasks of a certain bunch are always performed at least once more than those in the next bunch.

> Similarly, i'm confused as to why the assertions below (with check_complete_on_run=True) resulted in larger complete_count quantities.

With `check_complete_on_run=True` there is a single, additional call happening [here](https://github.com/spotify/luigi/blob/a295041d9c344135ef7f8d41e26f6fdbe6e55c58/luigi/worker.py#L195).

dlstadther: Oh, i see now. This makes much more sense now! Thank you for the thorough response!

dlstadther (approved):
Code changes, documentation additions, and unittest coverage LGTM!

Thanks for your contribution!

PR #3179, opened 2022-06-26
Improve control over dynamic requirements

riga:
This PR is meant as a continuation of #3178 and further improves the control over dynamic task requirements handled by the worker.

## Description

I added a shallow class `DynamicRequirements` which is intended to wrap a batch of tasks being yielded as dyn. reqs. and optionally to define a custom, probably optimized completeness check for that batch. The new class is understood by the `TaskProcess` as a valid yield value of run() methods, which required changes to only a few lines.

## Motivation and Context

As already mentioned in #3178, we are sometimes dealing with k's of tasks yielded as dynamic requirements in some wrapper task. These tasks store their outputs in a remote location and we can make the assumption that they are located in the same directory (presumably not unusual). For this matter we don't want to "stat" all output files separately, but just do a "listdir" on the common base directory, followed by a local comparison of basenames, saving us k's of interactions with remote resources.  Although we make use of the caching implemented in the referenced PR, we would like to further reduce remote API calls but the workers (or rather TaskProcess) internally flatten all requirements and perform separate completeness checks:

https://github.com/spotify/luigi/blob/master/luigi/worker.py#L154

*Which* requirements exist and which don't is not of interest at this point, so *batched* completeness checks as the one suggested above would be fully compatible with the current logic :)

Yep, I added a new test case, update the docs, and amended the `dynamic_requirements.py` example.

### DIFF
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -209,7 +209,7 @@ the Task.run_ method will resume from scratch each time a new task is yielded.
 In other words, you should make sure your Task.run_ method is idempotent.
 (This is good practice for all Tasks in Luigi, but especially so for tasks with dynamic dependencies).
 As this might entail redundant calls to tasks' :func:`~luigi.task.Task.complete` methods,
-you should consider setting the "cache_task_completion" option int the :ref:`worker-config`.
+you should consider setting the "cache_task_completion" option in the :ref:`worker-config`.
 To further control how dynamic task requirements are handled internally by worker nodes,
 there is also the option to wrap dependent tasks by :class:`~luigi.task.DynamicRequirements`.
 
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -208,9 +208,9 @@ It does come with some constraints:
 the Task.run_ method will resume from scratch each time a new task is yielded.
 In other words, you should make sure your Task.run_ method is idempotent.
 (This is good practice for all Tasks in Luigi, but especially so for tasks with dynamic dependencies).
-As this might entail redundant calls to tasks' Task.complete_ method, you should consider
-using the "cache_task_completion" :ref:`worker-config`.
-To further control how dynamic task requirements handled internally by worker nodes,
+As this might entail redundant calls to tasks' Task.complete_ methods, you should consider
+setting the "cache_task_completion" option int the :ref:`worker-config`.
+To further control how dynamic task requirements are handled internally by worker nodes,
 there is also the option to wrap dependent tasks by :class:`~luigi.task.DynamicRequirements`.
 
 For an example of a workflow using dynamic dependencies, see
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -198,8 +198,8 @@ You can also yield a list of tasks.
         def run(self):
             other_target = yield OtherTask()
 
-        # dynamic dependencies resolve into targets
-        f = other_target.open('r')
+            # dynamic dependencies resolve into targets
+            f = other_target.open('r')
 
 
 This mechanism is an alternative to Task.requires_ in case
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -208,8 +208,8 @@ It does come with some constraints:
 the Task.run_ method will resume from scratch each time a new task is yielded.
 In other words, you should make sure your Task.run_ method is idempotent.
 (This is good practice for all Tasks in Luigi, but especially so for tasks with dynamic dependencies).
-As this might entail redundant calls to tasks' Task.complete_ methods, you should consider
-setting the "cache_task_completion" option int the :ref:`worker-config`.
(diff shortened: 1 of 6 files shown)


riga:
The failing test reports

```shell
ModuleNotFoundError: No module named 'openapi_spec_validator'

# and connected to that

_pytest.nodes.Collector.CollectError: ImportError while importing test module '/home/runner/work/luigi/luigi/test/contrib/ecs_test.py'.
```

Is it possible that there's something wrong with `contrib/ecs_test.py`? I don't think the changes in this PR affect that one.

riga:
Seemed like a temporary glitch, all tests pass now 🎉

lallea:
This could likely improve many of our jobs. Thanks!

I have considered implementing something similar, but in the storage interface layer, e.g. a GCSClient that batches lookups, potentially backed by a cache shared by workers. We already have a GCSClient with an in-memory cache, which improves things, but doesn't go all the way. Our cache implementation is very simple and makes some assumptions, so we haven't shared it.

What would the pros and cons be of DynamicRequirements vs similar functionality in the storage layer? DynamicRequirements can be used for multiple types of storage. A solution in the storage layer would be opaque to the worker core, and might arguably better separate concerns by keeping storage interface complexity within the storage layer. I am not sure how valuable a shared or persistent cache would be.

I do not doubt that this is a good solution. Just taking the opportunity to reason. Open source architecture is often accidental. :-)

riga:
Thanks for the feedback @lallea !

> What would the pros and cons be of DynamicRequirements vs similar functionality in the storage layer?

Interesting, we also work with in-memory / local caching on SSDs plus batched lookups. However, I think having these DynamicRequirements on-top can be beneficial 

a) to reduce the load even further, and
b) to make the first step of savings accessible for projects that can't or perhaps haven't yet invested time into custom caching solutions.

Some more background on our case (Physics research):

We sometimes have dynamic tasks with O(k) yielded tasks, and on some of our file systems, saving O(k) cache lookups can speed things up a lot. Depending on where we process things (we can't always choose that), we cannot control the worker infrastructure so we often end up on machines that only have shared, slow NFS's where caching via local disks isn't an option. Low-hanging fruits such as defining DynamicRequirements could be really helpful there.

lallea:
Yes, I agree it looks useful even in the presence of other caching.

Could it make the bulk complete code in the Range classes `missing_datetimes` obsolete? I think that code has never worked properly. I have looked at it, but failed to understand how it could work. :-)

riga:
@dlstadther Kind ping :)

doc/tasks.rst
```diff
@@ -198,8 +198,8 @@ You can also yield a list of tasks.
         def run(self):
             other_target = yield OtherTask()
 
-	    # dynamic dependencies resolve into targets
-	    f = other_target.open('r')
+        # dynamic dependencies resolve into targets
+        f = other_target.open('r')
```
dlstadther, line 202: for the documentation code example here, should these two lines be indented _inside_ of the `run()`? Looks like you just bumped them outside.

examples/dynamic_requirements.py
```diff
@@ -91,6 +92,20 @@ def run(self):
         with self.output().open('w') as f:
             f.write('Tada!')
 
+        # and in case data is rather long, consider wrapping the requirements
+        # in DynamicRequirements and optionally define a custom complete method
+        def custom_complete(complete_fn):
+            # example: Data() stores all outputs in the same directory, so avoid doing len(data) fs
+            # calls but rather check the only the first, and compare basenames for the rest
```
dlstadther, line 99: perhaps typo? should this read `...but rather check only the first,...`? (too many `the`s)

luigi/task.py
```diff
@@ -765,6 +765,87 @@ def bulk_complete(cls, parameter_tuples):
...
+
+        Outputs of the requirements in the identical structure (via :py:func:`getpaths`). Read only.
+
+    .. py:attribute:: custom_complete
+
+       The optional, custom function performing the completeness check of the wrapped requirements.
+    """
+
+    def __init__(self, requirements, custom_complete=None):
+        super(DynamicRequirements, self).__init__()
```
dlstadther, line 813: since Luigi only officially supports Python 3 now, this could be written `super().__init__()`. No need to change. Just pointing out.

dlstadther (commented):
Thanks for the ping @riga !

Changes LGTM; only left 2 small documentation questions and 1 optional code comment.

doc/tasks.rst
riga: True! Corrected in 8c3e3a54.

examples/dynamic_requirements.py
riga: Fixed in 8c3e3a54.

luigi/task.py
riga: 👍 Added in 8c3e3a54.

dlstadther (approved):
LGTM! Thank you for your hard work and contribution!

PR #3183, opened 2022-07-11
Add sections() method to LuigiTomlParser class (fix #3182)

mzpqnxow:
This addresses the issue described in #3182 

EDIT: Referenced the wrong issue number, d'oh!

Issue #3182, opened 2022-07-11
LuigiTomlParser object has no attribute 'sections'


mzpqnxow:
## Missing/incorrect method when loading TOML configuration data

When loading a TOML configuration file on Luigi==3.1.0, I get the following backtrace:

```
Traceback (most recent call last):
  File "/home/user/luigi-test-proj/venv/bin/luigi", line 8, in <module>
    sys.exit(luigi_run())
  File "/home/user/luigi-test-proj/venv/lib/python3.9/site-packages/luigi/cmdline.py", line 9, in luigi_run
    run_with_retcodes(argv)
  File "/home/user/luigi-test-proj/venv/lib/python3.9/site-packages/luigi/retcodes.py", line 71, in run_with_retcodes
    retcodes = retcode()
  File "/home/user/luigi-test-proj/venv/lib/python3.9/site-packages/luigi/task_register.py", line 87, in __call__
    param_values = cls.get_param_values(params, args, kwargs)
  File "/home/user/luigi-test-proj/venv/lib/python3.9/site-packages/luigi/task.py", line 439, in get_param_values
    if task_family in conf.sections():
AttributeError: 'LuigiTomlParser' object has no attribute 'sections'
```

When I take a look at the `LuigiTomlParser`, it's pretty easy to see a way to fix this with minimal effort, by adding a `sections()` method that just returns a simple list of the sections- 

```diff
+    def sections(self):
+        """Return a list of sections as list of strings"""
+        return [section for section in self.data.keys() if section.lower() != 'logging']
+
```

I will attach my configuration file, though it's worth mentioning the example toml from this repo also causes the issue

PR #3185, opened 2022-07-11
Alternate (probably more correct) fix for #3182 (supercedes #3183)

mzpqnxow:
This addresses issue #3182 (and supersedes the previous PR #3183)

It fixes the issue in a (probably) more correct way, by adding `ConfigParser` to the list of parent classes for `LuigiTomlParser`

Because `ConfigParser` implements `sections()`, the issue goes away (and things seem to work properly)

I wonder if leaving this out was just an oversight, as I only noticed this as a clean solution when I reviewed how the `LuigiConfigParser` worked- which is by inheriting from `BaseParser` and `ConfigParser` as well

### DIFF
--- a/luigi/configuration/toml_parser.py
+++ b/luigi/configuration/toml_parser.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 #
 import os.path
+from configparser import ConfigParser
 
 try:
     import toml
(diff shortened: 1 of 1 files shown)


dlstadther (approved):

mzpqnxow:
Thanks!

PR #3196, opened 2022-09-01
Suggested fixes by iCR, OpenRefactory, Inc.

openrefactory:
This issue was detected in branch `master` of `luigi` project on the version with commit hash `3611b0`. This is an instance of a weak cryptography issues.

**Fixes for weak cryptography issue:**
In file: `ftp.py`, method:` _ftp_connect`, a clear-text protocol such as `FTP, Telnet or SMTP` is used. These protocols [transfer data without any encryption, ](https://cwe.mitre.org/data/definitions/319.html)which expose applications to a large range of risks. **iCR** suggested that data should be transferred over only secure transport channels.

This issue was detected by **OpenRefactory's Intelligent Code Repair (iCR)**. We are running **iCR** on libraries in the `PyPI` repository to identify issues and fix them. More info at: [pypi.openrefactory.com](https://pypi.openrefactory.com/)

### DIFF
--- a/luigi/contrib/ftp.py
+++ b/luigi/contrib/ftp.py
@@ -85,7 +85,9 @@ class RemoteFileSystem(luigi.target.FileSystem):
         if self.tls:
             self.conn = ftplib.FTP_TLS()
         else:
-            self.conn = ftplib.FTP()
+            # OpenRefactory Warning: The 'ftplib.FTP' method uses 'FTP' protocol which transfers data without any encryption.
+            # Using SFTP/SCP/FTPS over FTP is recommended. Use 'ftplib.FTP_TLS' method instead.
+            self.conn = ftplib.FTP_TLS()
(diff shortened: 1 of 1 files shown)


openrefactory:
@dlstadther

dlstadther (changes requested):
This conditional is initially _not_ using TLS. If you want to use tls, then the `self.tls` instance attribute would be true.

If we accept this patch, the `self.tls` option becomes meaningless.

What goal are you trying to achieve here then? If this was attempted using an automated tool, it made this suggested change in error.

lallea:
https://twitter.com/lalleal/status/1565718271088803842

I received no response. I found no information about how to opt-out on their web. There seems to be a way to block users from creating PRs: https://docs.github.com/en/communities/maintaining-your-safety-on-github/blocking-a-user-from-your-organization

Issue #3202, opened 2022-10-27
Worker fails to run with Python 3.11 


saschamarkus-uberresearch:
When migrationg a batch pipeline with multiple workers to Python 3.11 the workers fails to run.

Stacktrace:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/pack/app/venv/lib/python3.11/site-packages/luigi/worker.py", line 297, in run
    super(ContextManagedTaskProcess, self).run()
  File "/pack/app/venv/lib/python3.11/site-packages/luigi/worker.py", line 168, in run
    random.seed((processID, currentTime))
  File "/usr/local/lib/python3.11/random.py", line 160, in seed
    raise TypeError('The only supported seed types are: None,\n'
TypeError: The only supported seed types are:

```
The error is caused by a change in [random.seed()](https://docs.python.org/3/library/random.html#random.seed) for Python 3.11

> Changed in version 3.11: The seed must be one of the following types: NoneType, [int](https://docs.python.org/3/library/functions.html#int), [float](https://docs.python.org/3/library/functions.html#float), [str](https://docs.python.org/3/library/stdtypes.html#str), [bytes](https://docs.python.org/3/library/stdtypes.html#bytes), or [bytearray](https://docs.python.org/3/library/stdtypes.html#bytearray).

The issue can be reproduced by running 
`pytest -c /dev/null test/worker_task_process_test.py`
which fails on `test_context_manager_with_multiprocessing`

One possible fix would be to replace 
`   random.seed((processID, currentTime))`
with something like
`   random.seed(processID * currentTime)`
or
`   random.seed(str((processID, currentTime)))`

jmrnilsson:
Hmm, I hit this too. Let's check for new version..

jmrnilsson:
No reverting back to Python 3.9 works though. So it's a work-around..

dlstadther:
@jmrnilsson , the patch that @davidesba submitted in currently in `master` if you want to try it. It'll be included in a future (date tbd) versioned release of the project.

saschamarkus-uberresearch:
@jmrnilsson 
If you use the pyproject.toml/setup.cfg way to set up your project you can use the master branch from grid instead of the release from pypy with this in your setup.cfg. There ist probably a similar way for the requirements.txt 

```
[options]
packages = find:
install_requires =
  luigi @ git+https://github.com/spotify/luigi.git 

```

PR #3204, opened 2022-11-03
Update random seed for python 3.11 compatibility

davidesba:
Fix compatibility error with python 3.11: https://github.com/spotify/luigi/issues/3202

## Description

Update worker process random seed to be compatible with python3.11
## Motivation and Context

Fixes issue: https://github.com/spotify/luigi/issues/3202
 

Run tests and tested on local project

### DIFF
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -170,7 +170,7 @@ class TaskProcess(multiprocessing.Process):
             # Need to have different random seeds if running in separate processes
             processID = os.getpid()
             currentTime = time.time()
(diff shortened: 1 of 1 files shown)


dlstadther (approved):
A benign change and tests still look good; Thanks for the contribution!

(Please note that Luigi's test suite only covers up to 3.9; you may find other attributes within 3.10 or 3.11 which might cause problems. There was an attempt a few months ago to update for 3.10 support and testing, but I am unsure of the current state of that work.)

PR #3207, opened 2022-11-10
s3: Fix docstring of listdir() and add to list()

progval:
## Description

1. listdir yields absolute paths, not relative
2. empty line was missing before first `:param`

This is what I get when querying it with an ARN.


dlstadther (approved):
Thanks for the documentation updates!

PR #3209, opened 2022-11-24
Fix md5 hash error for FIPS in python 3.9

davidesba:
## Description

Add  usedforsecurity=False to all md5 hashes to ensure FIPS compatibility in python >= 3.9
## Motivation and Context

As python 3.9 documentation for hashlib states:
>All hashlib constructors take a keyword-only argument usedforsecurity with default value True. A false value allows the use of insecure and blocked hashing algorithms in restricted environments. False indicates that the hashing algorithm is not used in a security context, e.g. as a non-cryptographic one-way compression function.
 

Tested in local project with python 3.11

### DIFF
--- a/luigi/contrib/hadoop.py
+++ b/luigi/contrib/hadoop.py
@@ -25,6 +25,7 @@ an example of how to run a Hadoop job.
 import abc
 import datetime
 import glob
+import hashlib
 import logging
 import os
 import pickle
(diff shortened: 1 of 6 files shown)


dlstadther (approved):
Thanks!

mjomora:
When will we have this tagged?

PR #3214, opened 2022-12-13
fix: using default_scheduler_url as a mounting point for not root path

### LINKED ISSUE #3213: using default_scheduler_url as a mounting point (not root `http://address/mount`)  behind proxy not working
I need to mount `luigi` behind a nginx, for example luigi at `http://address/mount`.
For that I configure:
```ini
[core]
default_scheduler_url=http://address/mount
....
```

GUI is ok and works but, CLI not due to url resolution. it happens there https://github.com/spotify/luigi/blob/c13566418c92de3e4d8d33ead4e7c936511afae1/luigi/rpc.py#L54

To understand what happened:
```python
parsed=urlparse('http://address/mount')
url='/api/add_task'
urljoin(parsed.geturl(), url)
# ==> give 'http://address/api/add_task'
# expected http://address/mount/api/add_task
```
What I must do for working - slash at the end of mount point, no slash for url -:
```python
parsed=urlparse('http://address/mount/')
url='api/add_task'
urljoin(parsed.geturl(), url)
# ==> http://address/mount/api/add_task
```

franckOL:
(http://address/mount) behind proxy works. fix #3213

Signed-off-by: ROUDET Franck INNOV/IT-S <franck.roudet@orange.com>

## Description

2 files modified to change the build of URL.
`api` path not at root now, it is in fact the same strategy that use the GUI.

## Motivation and Context

Fixes:
 * https://github.com/spotify/luigi/issues/3213
 * and https://groups.google.com/g/luigi-user/c/kJ7aI2GNfcE
 
I'm able to test:
* With or Without scheduler-url set
* With or Without scheduler-url not at root

### DIFF
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -145,7 +145,7 @@ class RemoteScheduler:
             'You need to install requests-unixsocket for Unix socket support.'
         )
 
-        self._url = url.rstrip('/')
+        self._url = url
         config = configuration.get_config()
 
         if connect_timeout is None:
--- a/luigi/rpc.py
+++ b/luigi/rpc.py
@@ -145,7 +145,7 @@ class RemoteScheduler:
             'You need to install requests-unixsocket for Unix socket support.'
         )
 
-        self._url = url
+        self._url =  url if url[-1] == '/' else url + '/'
         config = configuration.get_config()
 
         if connect_timeout is None:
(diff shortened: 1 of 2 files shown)


franckOL:
I don't know how to test with tox.
I test locally GUI and API both with mounting point and at root.

franckOL:
@dlstadther What do you think about that PR ?

dlstadther:
@franckOL , you'll need to look into and fix any failing tests as a result of this change.

franckOL:
Not seeing how to fix codeconv. I think my changes are not visible in codecov.
and for failed @github-actions Build / core (3.9, py39-core) (pull_request), Not understanding what happens ! network issue :-)

PR #3217, opened 2023-01-12
Can pass a JSON schema to DictParameter and ListParameter

adrien-berchet:
## Description

Add an optional parameter to `DictParameter` and `ListParameter` so the loaded value can be validated against a JSON schema.

## Motivation and Context

Adding a simple validation step reduces the amount of code in the `run` method of the tasks. Also, the arguments are checked at the beginning of the workflow so it fails faster, which is always better than failing during the workflow.

I added simple tests.


adrien-berchet:
Note that this PR adds the `jsonschema` dependency. If you prefer I can make it optional.

dlstadther (commented):
This is a great addition!

However, I'm leaning towards suggesting `jsonschema` be an optional installation since this is an optional feature usage.

dlstadther (approved):
Thanks!

PR #3220, opened 2023-01-17
Fix schema validation and add custom validators

adrien-berchet:
## Description

Fixes #3217 that breaks when `jsonschema` is not installed (sorry about that :facepalm: ).
Also add the possibility to provide a custom validator as schema instead of just relying on `jsonschema.validate`.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -1137,15 +1137,14 @@ class DictParameter(Parameter):
         """
         Ensure that dictionary parameter is converted to a FrozenOrderedDict so it can be hashed.
         """
-        frozen_value = recursively_freeze(value)
         if self.schema is not None:
-            unfrozen_value = recursively_unfreeze(frozen_value)
+            unfrozen_value = recursively_unfreeze(value)
             try:
-                self.schema.validate(unfrozen_value)  # Validators may update the instance inplace
-                frozen_value = super().normalize(unfrozen_value)
+                self.schema.validate(unfrozen_value)
+                value = unfrozen_value  # Validators may update the instance inplace
             except AttributeError:
                 jsonschema.validate(instance=unfrozen_value, schema=self.schema)
-        return frozen_value
+        return recursively_freeze(value)
 
     def parse(self, source):
         """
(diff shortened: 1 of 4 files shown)


adrien-berchet:
Sorry @dlstadther , I went too fast in #3217 . This time it should be fine.

luigi/parameter.py
```diff
@@ -1117,7 +1139,12 @@ def normalize(self, value):
         """
         frozen_value = recursively_freeze(value)
         if self.schema is not None:
-            jsonschema.validate(instance=recursively_unfreeze(frozen_value), schema=self.schema)
+            unfrozen_value = recursively_unfreeze(frozen_value)
+            try:
+                self.schema.validate(unfrozen_value)  # Validators may update the instance inplace
+                frozen_value = super().normalize(unfrozen_value)
```
dlstadther, line 1145: I understand the try/except here to handle for when `self.schema` is a custom validator, but could you explain why you re-freeze the unfrozen value again? Why/how would the validator modify the value of the instance such that we cannot return the `frozen_value` from the beginning of this method?

dlstadther (commented):
Thank you for catching and fixing your error.

I do have a clarifying question before approving

luigi/parameter.py
adrien-berchet: It is possible to define custom validators to handle specific schemas or do extra stuffs: https://python-jsonschema.readthedocs.io/en/stable/creating/
For example, it is possible to fill missing values with default values: https://python-jsonschema.readthedocs.io/en/stable/faq/#why-doesn-t-my-schema-s-default-property-set-the-default-on-my-instance
In this example the given instance is updated with default values during the validation, so the new instance must be frozen again and returned.
Nevertheless, we can indeed simplify this a bit: unfreeze the given value when a schema is provided and just freeze the result, instead of freezing, unfreezing and freezing again. I pushed a new commit for this.

luigi/parameter.py
```diff
@@ -1115,10 +1137,14 @@ def normalize(self, value):
...
         Ensure that dictionary parameter is converted to a FrozenOrderedDict so it can be hashed.
         """
-        frozen_value = recursively_freeze(value)
         if self.schema is not None:
-            jsonschema.validate(instance=recursively_unfreeze(frozen_value), schema=self.schema)
-        return frozen_value
+            unfrozen_value = recursively_unfreeze(value)
+            try:
+                self.schema.validate(unfrozen_value)
+                value = unfrozen_value  # Validators may update the instance inplace
```
adrien-berchet, line 1144: I defined the `unfrozen_value` variable for clarity but we could just overwrite the `value` variable at line `1141`. What do you prefer?

honnix:
Let me try to understand the situation a bit better. Does it mean the main branch of Luigi is broken? We plan to drop a new release and if that is the case, we would suggest reverting the breaking PR so we have a healthy main branch, and you will also be able to work on this without a rush @adrien-berchet .

adrien-berchet:
Hi @honnix 
Well, the main branch is indeed broken but it will not break any existing code, it will only break if someone tries to use the new feature without `jsonschema` installed.
With this new PR everything looks good in my tests.

adrien-berchet:
Note that the CI passed, only the coverage reported a failure but I am pretty sure it is not due to this PR (2 lines in `luigi.contrib.hive` were not covered).

dlstadther (approved):
I understand now! Thank you!

adrien-berchet:
You're welcome :blush:

PR #3225, opened 2023-02-09
allow setting a "custom disconnect delay" of a worker, by clicking on…

koo5:
… a button in worker list. The custom delay is taken into account when pruning workers.

## Description
a new dropdown in the workers page, and a new variable custom_disconnect_delay in scheduler.py/Worker

## Motivation and Context
My usecase are long running tasks with possibly intermittent connectivity. Therefore i configure luigi something like this:

```
[scheduler]
remove_delay = 88888888
worker_disconnect_delay = 88888888

[core]
rpc-retry-attempts = 88888888
```

But when i am certain that a worker is not coming back, i need to inform the scheduler about this. Rather than adding a special "remove" button, this PR lets you override the disconnect delay value, and automatic pruning does the rest.

This PR is somewhat preliminary:
* the gui needs some tweaks, specifically, the null value is displayed as an empty string upon page load. 
* `on('click', '#btn-set-worker-custom-disconnect-delay-****` could probably be abstracted
* maybe Worker.custom_disconnect_delay would be better stored in Worker.info?

I played with this code and it seems to do the job.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -353,9 +353,13 @@ class Worker:
         self.info = {}
         self.disabled = False
         self.rpc_messages = []
+        self.custom_disconnect_delay = None
 
     def add_info(self, info):
         self.info.update(info)
+
+    def set_custom_disconnect_delay(self, delay):
+        self.custom_disconnect_delay = delay
 
     def update(self, worker_reference, get_work=False):
         if worker_reference:
@@ -366,7 +370,12 @@ class Worker:
 
     def prune(self, config):
         # Delete workers that haven't said anything for a while (probably killed)
-        if self.last_active + config.worker_disconnect_delay < time.time():
+        if self.custom_disconnect_delay:
+            disconnect_delay = self.custom_disconnect_delay
+        else:
+            disconnect_delay = config.worker_disconnect_delay
+        if self.last_active + disconnect_delay < time.time():
+            logger.debug("Worker %s timed out (no contact for >=%ss)", self, disconnect_delay)
             return True
 
     def get_tasks(self, state, *statuses):
@@ -718,7 +727,6 @@ class Scheduler:
         remove_workers = []
         for worker in self._state.get_active_workers():
             if worker.prune(self._config):
-                logger.debug("Worker %s timed out (no contact for >=%ss)", worker, self._config.worker_disconnect_delay)
                 remove_workers.append(worker.id)
 
         self._state.inactivate_workers(remove_workers)
@@ -962,6 +970,10 @@ class Scheduler:
     @rpc_method()
     def set_worker_processes(self, worker, n):
         self._state.get_worker(worker).add_rpc_message('set_worker_processes', n=n)
+
+    @rpc_method()
+    def set_worker_custom_disconnect_delay(self, worker, n):
+        self._state.get_worker(worker).set_custom_disconnect_delay(n)
 
     @rpc_method()
     def send_scheduler_message(self, worker, task, content):
@@ -1471,6 +1483,7 @@ class Scheduler:
                 state=worker.state,
                 first_task_display_name=self._first_task_display_name(worker),
                 num_unread_rpc_messages=len(worker.rpc_messages),
+                custom_disconnect_delay=worker.custom_disconnect_delay,
                 **worker.info
             ) for worker in self._state.get_active_workers()]
         workers.sort(key=lambda worker: worker['started'], reverse=True)
--- a/luigi/static/visualiser/index.html
+++ b/luigi/static/visualiser/index.html
@@ -167,35 +167,53 @@
             {{/is_disabled}}
                 <div class="box-header with-border">
                     <h3 class="box-title">{{name}}</h3>
-                    <div class="box-tools pull-right">
-                      {{#num_unread_rpc_messages}}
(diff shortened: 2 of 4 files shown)


dlstadther:
I like the idea here and I'm happy to review the python. However, I'm not much use when it comes to the concerns you mention with the UI

> * the gui needs some tweaks, specifically, the null value is displayed as an empty string upon page load.
> * on('click', '#btn-set-worker-custom-disconnect-delay-**** could probably be abstracted
> * maybe Worker.custom_disconnect_delay would be better stored in Worker.info?

luigi/scheduler.py
```diff
@@ -366,7 +370,12 @@ def update(self, worker_reference, get_work=False):
 
     def prune(self, config):
         # Delete workers that haven't said anything for a while (probably killed)
-        if self.last_active + config.worker_disconnect_delay < time.time():
+        if self.custom_disconnect_delay:
+            disconnect_delay = self.custom_disconnect_delay
+        else:
+            disconnect_delay = config.worker_disconnect_delay
+        if self.last_active + disconnect_delay < time.time():
+            logger.debug("Worker %s timed out (no contact for >=%ss)", self, disconnect_delay)
```
dlstadther, line 378: Now that Luigi is Python 3.6+, consider using `.format()` strings or `f-strings`, instead of the legacy `%-strings`.

luigi/scheduler.py
```diff
@@ -353,10 +353,14 @@ def __init__(self, worker_id, last_active=None):
         self.info = {}
         self.disabled = False
         self.rpc_messages = []
+        self.custom_disconnect_delay = None
```
dlstadther, line 356: Related to you question of if `custom_disconnect_delay` should live as an attribute of Worker... It does seem a little odd here since all the other attributes of this class are related to tracking work.

Given that the value this would be overriding, `config.worker_disconnect_delay`, is a scheduler config value, that perhaps this configuration should be a scheduler rpc update. Thoughts? (it is already a bit confusing the overlap between scheduler and worker config).

luigi/static/visualiser/index.html
```diff
@@ -167,35 +167,53 @@ <h4 class="modal-title" id="disiableWorkerLabel">Disable worker?</h4>
...
-                      {{/num_unread_rpc_messages}}
-                      {{^is_disabled}}
-                      {{#workers}}
-                      <div class="btn-group">
-                        <button type="button" class="btn btn-sm btn-default dropdown-toggle btn-set-workers" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-                          Workers: <span id="label-n-workers" data-worker="{{name}}">{{workers}}</span> <span class="caret"></span>
+                    <div class="pull-right">
+                      <span class="btn-group">
+                        <button type="button" class="btn btn-sm btn-default dropdown-toggle btn-set-worker-custom-disconnect-delay" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                          Custom disconnect delay: <span id="label-custom-disconnect-delay" data-worker="{{name}}">{{custom_disconnect_delay}}</span> <span class="caret"></span>
```
dlstadther, line 173: thoughts on `custom disconnect delay` defaulting to the scheduler config, such that this value isn't ever null? (assuming this is the null value you mention was displaying as an empty string)

luigi/scheduler.py
koo5: You lost me at 
> this configuration should be a scheduler rpc update.

However, i found a reasonably elegant way to save the frontend from dealing with the possible Null value. `worker_list` can obtain the current value. 

I don't see a better way right now. I'm not sure if there's a good way to make Scheduler construct Workers with a configuration value. Also, it seems that it would be complicated to make Worker weakref its Scheduler (to obtain scheduler config in, say, a property getter). 

https://github.com/koo5/luigi/commit/4c220d67728bd9ed91917d08c327dec37cc0584e

Also, nevermind about using `Worker.info`, as that seems to be for passing information from workers to scheduler.

luigi/static/visualiser/index.html
koo5: You lost me at 
> this configuration should be a scheduler rpc update.

However, i found a reasonably elegant way to save the frontend from dealing with the possible Null value. `worker_list` can obtain the current value. 

I don't see a better way right now. I'm not sure if there's a good way to make Scheduler construct Workers with a configuration value. Also, it seems that it would be complicated to make Worker weakref its Scheduler (to obtain scheduler config in, say, a property getter). 

https://github.com/koo5/luigi/commit/4c220d67728bd9ed91917d08c327dec37cc0584e

Also, nevermind about using `Worker.info`, as that seems to be for passing information from workers to scheduler.

koo5:
You lost me at 
> this configuration should be a scheduler rpc update.

However, i found a reasonably elegant way to save the frontend from dealing with the possible Null value. `worker_list` can obtain the current value. 

I don't see a better way right now. I'm not sure if there's a good way to make Scheduler construct Workers with a configuration value. Also, it seems that it would be complicated to make Worker weakref its Scheduler (to obtain scheduler config in, say, a property getter). 

https://github.com/koo5/luigi/commit/4c220d67728bd9ed91917d08c327dec37cc0584e

Also, nevermind about using `Worker.info`, as that seems to be for passing information from workers to scheduler.

PR #3230, opened 2023-03-13
Fix XSS vulnerability

honnix:
## Description

Fix XSS vulnerability

## Motivation and Context

This PR escapes HTML content to address XSS vulnerability. A previous attempt did not fully resolve it: https://github.com/spotify/luigi/pull/2391

I ran a task and verified that HTML content is properly escape.

### DIFF
--- a/luigi/static/visualiser/js/graph.js
+++ b/luigi/static/visualiser/js/graph.js
@@ -242,6 +242,15 @@ Graph = (function() {
         return element;
     }
 
+    function escapeHtml(unsafe) {
+        return unsafe
+          .replace(/&/g, "&amp;")
+          .replace(/</g, "&lt;")
+          .replace(/>/g, "&gt;")
+          .replace(/"/g, "&quot;")
+          .replace(/'/g, "&#039;");
+    }
+
     DependencyGraph.prototype.renderGraph = function() {
         var self = this;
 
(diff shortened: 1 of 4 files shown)


luigi/static/visualiser/js/graph.js
```diff
@@ -242,6 +242,15 @@ Graph = (function() {
         return element;
     }
 
+    function escapeHtml(unsafe) {
+        return unsafe
```
narape, line 246: can `unsafe` be null? 
```suggestion
        if (unsafe == null) {
          return unsafe;
        }
        return unsafe
```

luigi/static/visualiser/js/visualiserApp.js
```diff
@@ -20,6 +20,15 @@ function visualiserApp(luigi) {
     };
     var VISTYPE_DEFAULT = 'svg';
 
+    function escapeHtml(unsafe) {
```
narape, line 23: why not have this function in a separate file and imported from both places?

luigi/static/visualiser/js/graph.js
honnix: Not really, it should better be a bug on the caller side, instead of hiding it I believe.

luigi/static/visualiser/js/visualiserApp.js
honnix: The introduced complexity is not worth to me. I don't really know what JS standard the code base uses.

narape (dismissed):

luigi/static/visualiser/js/visualiserApp.js
honnix: OK I used the old ugly way to address this.

narape (approved):

PR #3237, opened 2023-04-25
Fix for applying logging configurations using the a TOML configuration file

JoseHJBlanco:
## Description
Configuring logging using the "logging" session of the TOML configuration file throws an exception because the `logging` library tries to write to a frozen dictionary. This PR solves the issue by unfreezing the dictionary before passing it to `logging.dictConfig`.

## Motivation and Context

Starting from version `2.8.11`, the logging section of the configuration file is loaded into frozen a dictionary which is then passed to the `logging.dictConfig` when using the TOML parser. However, `logging.dictConfig` expects a mutable dictionary and tries to overwrites some of its keys, resulting in an exception. In practice that means we cannot use the TOML configuration file to configure logging.

This problem has been mentioned in the issue https://github.com/spotify/luigi/issues/2879

Manually tested the changes and added a new test case [here](https://github.com/JoseHJBlanco/luigi/blob/93d51636e8e6e1cd0df7d68c3b8d3368ad079f1a/test/setup_logging_test.py#L66) to replicate the scenario in which the `master` branch would throw an error.

### DIFF
--- a/luigi/setup_logging.py
+++ b/luigi/setup_logging.py
@@ -24,6 +24,7 @@ import logging.config
 import os.path
 
 from luigi.configuration import get_config, LuigiConfigParser
+from luigi.freezing import recursively_unfreeze
 
 from configparser import NoSectionError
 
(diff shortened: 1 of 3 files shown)


dlstadther (approved):
lgtm. Thanks for the contribution and unittest update!

PR #3258, opened 2023-10-17
worker: Log which outputs are missing when task is unexpectedly incomplete

progval:
## Motivation and Context

When a task ran but did not write all its outputs, Luigi errors with:

> RuntimeError: Unfulfilled dependency at run time: NodeProperties_graph__tmp_pytest_of_d_cnt_dir_rev_rel__02a93cea21

which can be hard to debug for tasks with many outputs. After this PR, it lists which outputs are missing. For example:

> RuntimeError: Unfulfilled dependency at run time: NodeProperties_graph__tmp_pytest_of_d_cnt_dir_rev_rel__02a93cea21 (/tmp/pytest-of-dev/pytest-512/test_compressgraph_None_0/compressed_graph/graph.property.committer_timestamp.bin, /tmp/pytest-of-dev/pytest-512/test_compressgraph_None_0/compressed_graph/graph.property.message.bin, /tmp/pytest-of-dev/pytest-512/test_compressgraph_None_0/compressed_graph/graph.property.tag_name.bin, /tmp/pytest-of-dev/pytest-512/test_compressgraph_None_0/compressed_graph/graph.property.content.is_skipped.bin, /tmp/pytest-of-dev/pytest-512/test_compressgraph_None_0/compressed_graph/graph.property.tag_name.offset.bin, /tmp/pytest-of-dev/pytest-512/test_compressgraph_None_0/compressed_graph/graph.property.committer_id.bin, /tmp/pytest-of-dev/pytest-512/test_compressgraph_None_0/compressed_graph/graph.property.content.length.bin, /tmp/pytest-of-dev/pytest-512/test_compressgraph_None_0/compressed_graph/graph.property.author_id.bin, /tmp/pytest-of-dev/pytest-512/test_compressgraph_None_0/compressed_graph/graph.property.message.offset.bin, /tmp/pytest-of-dev/pytest-512/test_compressgraph_None_0/compressed_graph/graph.property.committer_timestamp_offset.bin)

## Description

I implemented a basic `__str__` for all target types, using either the path if it's filesystem-like, or a table name for database-like targets (to be consistent with `BigQueryTarget`).

Unit tests are not broken, and this works for me (only tested with local filesystem)

### DIFF
--- a/luigi/contrib/mongodb.py
+++ b/luigi/contrib/mongodb.py
@@ -35,6 +35,9 @@ class MongoTarget(Target):
         self._index = index
         self._collection = collection
 
+    def __str__(self):
+        return '%s/%s' % (self._index, self._collection)
+
     def get_collection(self):
         """
         Return targeted mongo collection to query on
--- a/luigi/contrib/mongodb.py
+++ b/luigi/contrib/mongodb.py
@@ -36,7 +36,7 @@ class MongoTarget(Target):
         self._collection = collection
 
     def __str__(self):
-        return '%s/%s' % (self._index, self._collection)
+        return f'{self._index}/{self._collection}'
 
     def get_collection(self):
         """
--- a/luigi/contrib/sqla.py
+++ b/luigi/contrib/sqla.py
@@ -189,6 +189,9 @@ class SQLAlchemyTarget(luigi.Target):
         self.connect_args = connect_args
         self.marker_table_bound = None
 
+    def __str__(self):
+        return self.marker_table_bound or self.connection_string
+
     @property
     def engine(self):
         """
--- a/luigi/contrib/sqla.py
+++ b/luigi/contrib/sqla.py
@@ -190,7 +190,7 @@ class SQLAlchemyTarget(luigi.Target):
         self.marker_table_bound = None
 
     def __str__(self):
-        return self.marker_table_bound or self.connection_string
+        return self.target_table
 
     @property
     def engine(self):
--- a/luigi/worker.py
+++ b/luigi/worker.py
@@ -187,7 +187,7 @@ class TaskProcess(multiprocessing.Process):
                     if not self.check_complete(dep):
                         nonexistent_outputs = [output for output in dep.output() if not output.exists()]
                         if nonexistent_outputs:
-                            missing.append('%s (%s)' % (dep.task_id, ', '.join(map(str, nonexistent_outputs))))
+                            missing.append(f'{dep.task_id} ({", ".join(map(str, nonexistent_outputs))})')
                         else:
                             missing.append(dep.task_id)
                 if missing:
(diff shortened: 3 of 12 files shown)


luigi/contrib/sqla.py
```diff
@@ -189,6 +189,9 @@ def __init__(self, connection_string, target_table, update_id, echo=False, conne
         self.connect_args = connect_args
         self.marker_table_bound = None
 
+    def __str__(self):
+        return self.marker_table_bound or self.connection_string
```
dlstadther, line 193: i don't follow why you prioritize the use of `self.marker_table_bound` here. if it's non-null, it's a sqlalchemy table object referencing the marker table, not the `target_table`

actually, why wouldn't this `__str__` representation not just be `self.target_table`?

luigi/contrib/mongodb.py
```diff
@@ -35,6 +35,9 @@ def __init__(self, mongo_client, index, collection):
         self._index = index
         self._collection = collection
 
+    def __str__(self):
+        return '%s/%s' % (self._index, self._collection)
```
dlstadther, line 39: personally, i'd prefer the use of `str.format()` or f-strings

luigi/worker.py
```diff
@@ -182,7 +182,14 @@ def run(self):
             # checking completeness of self.task so outputs of dependencies are
             # irrelevant.
             if self.check_unfulfilled_deps and not _is_external(self.task):
-                missing = [dep.task_id for dep in self.task.deps() if not self.check_complete(dep)]
+                missing = []
+                for dep in self.task.deps():
+                    if not self.check_complete(dep):
+                        nonexistent_outputs = [output for output in dep.output() if not output.exists()]
+                        if nonexistent_outputs:
+                            missing.append('%s (%s)' % (dep.task_id, ', '.join(map(str, nonexistent_outputs))))
```
dlstadther, line 190: would really prefer `str.format()` or f-strings for readability

dlstadther (commented):
Thanks for submitting this quality of life improvement.
I do have a couple comments/questions.

luigi/contrib/mongodb.py
progval:
```suggestion
        return f'{self._index}/{self._collection}'
```

luigi/contrib/sqla.py
progval:
```suggestion
        return self.target_table
```

You're right, I missed it

luigi/worker.py
progval:
```suggestion
                            missing.append(f'{dep.task_id} ({", ".join(map(str, nonexistent_outputs))})')
```

Ok, I was only following the existing style

dlstadther (approved):
LGTM. Thanks!

PR #3260, opened 2023-10-28
Make Dropbox listdir optionally recursive

smrohrer:
## Description
Make the Dropbox `listdir` function accept an optional keyword argument `recursive`, True by default to preserve the current behavior.

## Motivation and Context
Currently, `listdir` calls the Dropbox SDK `files_list_folder` function with `recursive=True` hard-coded. `recursive` should be optional, so we can replicate the functionality of `LocalFileSystem.listdir` which is not recursive.

- Added a unit test for non-recursive behavior
- Modified the other `listdir` tests to verify that by default, the list includes the nested file `DROPBOX_TEST_FILE_IN_DIR`
- Ran `tox -e py37-dropbox` locally

### DIFF
--- a/test/contrib/dropbox_test.py
+++ b/test/contrib/dropbox_test.py
@@ -87,19 +87,33 @@ class TestClientDropbox(unittest.TestCase):
         list_of_dirs = self.luigiconn.listdir(DROPBOX_TEST_PATH)
         self.assertTrue('/' not in list_of_dirs)
         self.assertTrue(DROPBOX_TEST_PATH in list_of_dirs)
-        self.assertTrue(DROPBOX_TEST_SIMPLE_FILE in list_of_dirs)  # we verify recursivity
+        self.assertTrue(DROPBOX_TEST_SIMPLE_DIR in list_of_dirs)
+        self.assertTrue(DROPBOX_TEST_SIMPLE_FILE in list_of_dirs)
+        self.assertTrue(DROPBOX_TEST_FILE_IN_DIR in list_of_dirs)  # we verify recursivity
 
     def test_listdir_simple_with_one_slash(self):
         list_of_dirs = self.luigiconn.listdir(DROPBOX_TEST_PATH + '/')
         self.assertTrue('/' not in list_of_dirs)
         self.assertTrue(DROPBOX_TEST_PATH in list_of_dirs)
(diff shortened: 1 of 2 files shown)


test/contrib/dropbox_test.py
```diff
@@ -87,19 +87,33 @@ def test_listdir_simple(self):
...
         self.assertTrue(DROPBOX_TEST_PATH in list_of_dirs)
-        self.assertTrue(DROPBOX_TEST_SIMPLE_FILE in list_of_dirs)  # we verify recursivity
+        self.assertTrue(DROPBOX_TEST_SIMPLE_DIR in list_of_dirs)
+        self.assertTrue(DROPBOX_TEST_SIMPLE_FILE in list_of_dirs)
+        self.assertTrue(DROPBOX_TEST_FILE_IN_DIR in list_of_dirs)  # we verify recursivity
+
+    def test_listdir_non_recursive(self):
+        list_of_dirs = self.luigiconn.listdir(DROPBOX_TEST_PATH, recursive=False)
+        self.assertTrue('/' not in list_of_dirs)
+        self.assertTrue(DROPBOX_TEST_PATH not in list_of_dirs)
```
dlstadther, line 113: I'm a little confused why the input path appears in the results during `recursive=True`, but not when `recursive=False`. Is that expected behavior for the underlying api call?

dlstadther (commented):
Code change and tests lgtm, with 1 confusing test assertion. I assume it's an underlying sdk implementation, but i would like to doublecheck.

smrohrer:
Thank you for the quick and diligent review.

I am closing this pull request because of a misunderstanding on my part: I confused the intended behavior of `luigi.target.FileSystem.listdir` with `os.listdir` - the former _is_ meant to be recursive, and the latter is not.

However, I think the inclusion of the input path in the Dropbox `listdir` result _does_ contradict the expected behavior of `luigi.target.FileSystem.listdir`. This is a behavior of the Dropbox `files_list_folder` function which is not described in the [documentation](https://dropbox-sdk-python.readthedocs.io/en/latest/api/dropbox.html#dropbox.dropbox_client.Dropbox.files_list_folder).

If you think `luigi.contrib.dropbox.DropboxClient.listdir` should filter out the input path, I'd be happy to open another pull request.

PR #3265, opened 2023-12-14
Add python 3.11 support

ukky17:
## Description

* Add support and CI workflow for Python 3.11.
* Replace `getargspec`, which has been deprecated since Python 3.0 and was removed in Python 3.11. https://docs.python.org/3/whatsnew/3.11.html#removed


dlstadther (approved):
The only code change (`inspect.getfullargspec()` is documented to be supported in our lowest (3.6) python version as well.

All CI tests are passing.

Thanks!

PR #3266, opened 2023-12-20
Support Python3.12

ukky17:
## Description
* Add support and CI workflow for Python 3.12.
* Skip a unit test that uses `boto`, because `boto` is not compatible with Python 3.12.
  * https://github.com/boto/boto/issues/3951
  * cf. `boto` is an old package of `boto3` and is no longer maintained.
* `assertEquals`, `assertNotEquals`, and `assertAlmostEquals` are replaced, which have been deprecated since Python 3.1 and are removed in Python 3.12.
  * https://docs.python.org/3/whatsnew/3.12.html#id3

## Motivation and Context

Support Python 3.12

All CI tests have passed.

### DIFF
--- a/test/contrib/s3_test.py
+++ b/test/contrib/s3_test.py
@@ -343,7 +343,7 @@ class TestS3Client(unittest.TestCase):
         s3_client.get('s3://mybucket/putMe', tmp_file_path)
         with open(tmp_file_path, 'r') as f:
             content = f.read()
-        self.assertEquals(content, self.tempFileContents.decode("utf-8"))
+        self.assertEqual(content, self.tempFileContents.decode("utf-8"))
         tmp_file.close()
 
     def test_get_as_bytes(self):
@@ -353,7 +353,7 @@ class TestS3Client(unittest.TestCase):
 
         contents = s3_client.get_as_bytes('s3://mybucket/putMe')
 
-        self.assertEquals(contents, self.tempFileContents)
+        self.assertEqual(contents, self.tempFileContents)
 
     def test_get_as_string(self):
         create_bucket()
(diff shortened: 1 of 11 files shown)


test/contrib/s3_test.py
```diff
@@ -107,6 +108,7 @@ def test_read_no_file_sse(self):
         t = self.create_target(encrypt_key=True)
         self.assertRaises(FileNotFoundException, t.open)
 
+    @unittest.skipIf(tuple(sys.version_info) >= (3, 12), "boto is not supported on Python 3.12+")
```
dlstadther, line 111: I think the better thing to do here would be to make the s3 module compatible with boto3

dlstadther (commented):
I'm not a fan of skipping the 1 boto unit test for the 3.12 version. The correct thing would be to ensure that the s3 module is 3.12 ready by replacing boto usage with boto3 equivalent functionality.

test/contrib/s3_test.py
starhel: Can the s3 module be improved in another PR? Ubuntu 24 came out with Python 3.12 as default, and the lack of support in Luigi will block updates for many users (although fortunately, the changes only affect tests, so it's not critical).

dlstadther: Given the need for 3.12 Luigi support, i'm ok getting this PR out the door.

@ukky17 , if you can resolve the merge conflict, i can merge if tests still pass.

ukky17: @dlstadther I have resolved the conflict. Please review again and merge if it looks good to you.

dlstadther (approved):

PR #3269, opened 2024-01-05
Migrate readthedocs to version 2

pablocasares:
## Description

Migrate readthedocs to version 2.

## Motivation and Context

Readthedocs docs generation stopped working because old setup was deprecated.


honnix:
Is this verified for PR build? Is it possible to show that in the PR checks?

.readthedocs.yaml
```diff
@@ -0,0 +1,18 @@
...
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.9"
+
+sphinx:
+  configuration: doc/conf.py
+
+formats:
+  - pdf
```
honnix, line 12: Do we need these format?

pablocasares: they were enabled in the old setup. We can disable them if not needed.

honnix: I see. Let's keep them then.

honnix (dismissed):

pablocasares:
> Is this verified for PR build? Is it possible to show that in the PR checks?

I updated the setup and it should now show the step in PR builds.

honnix (approved):

PR #3275, opened 2024-01-22
don't convert str to tuple in TupleParameter

kitagry:
## Description

TupleParameter may convert str to tuple, and I fixed it.

## Motivation and Context

Currently, TupleParameter returns unexpected value when to parse json string list.

```
>>> x = '["hoge", "fuga"]'
>>> TupleParameter().parse(x)
(('h', 'o', 'g', 'e'), ('f', 'u', 'g', 'a'))
```

I added unit test.

### DIFF
--- a/test/list_parameter_test.py
+++ b/test/list_parameter_test.py
@@ -81,7 +81,7 @@ class ListParameterTest(unittest.TestCase):
             a.normalize(["INVALID_ATTRIBUTE"])
 
         # Check that empty list is not valid
-        with pytest.raises(ValidationError, match=r"\[\] is too short"):
+        with pytest.raises(ValidationError):
             a.normalize([])
 
         # Check that valid lists work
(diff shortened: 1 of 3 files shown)


test/list_parameter_test.py
```diff
@@ -81,7 +81,7 @@ def test_schema(self):
             a.normalize(["INVALID_ATTRIBUTE"])
 
         # Check that empty list is not valid
-        with pytest.raises(ValidationError, match=r"\[\] is too short"):
+        with pytest.raises(ValidationError):
```
kitagry, line 84: CI failed because jsonschema validation error is changed.

I think the message should not be compared because it will be very flaky test. Do you think about this?

```
# before
with pytest.raises(ValidationError, match=r"\[\] is too short"):
    a.normalize([])

# after
with pytest.raises(ValidationError, '[] should be non-empty'):
    a.normalize([])
```

andresgomezfrr (approved):
LGTM

PR #3277, opened 2024-02-15
Fix outdated link to buffer article in README.rst

radeklat:
## Description

Fix outdated link to buffer article in the README.rst file.

## Motivation and Context

The current link is broken and no longer works.

The slug part of the URL is the same and the article mentions Luigi. So I assume it is the same.


dlstadther (approved):
Thanks for finding and fixing this!

dlstadther:
Not sure why some of the ci tests failed. I've triggered them to rerun to see if the issues were transient.

PR #3279, opened 2024-02-28
Update server_test.py

eltociear:
## Description

beging -> beginning


dlstadther (changes requested):
Pretty sure the comment should be `begin`, not `beginning`.

Also, appears that something CI dependencies is causing tests to fail (saw this with another PR a few days ago too). Unsure when I'll have the time to investigate further.

pablocasares (approved):

dlstadther (approved):

dlstadther:
Rerunning failed tests to see if transient

Issue #3280, opened 2024-03-05
TypeError: <Target> object is not iterable when check_complete(dep) is False


GianlucaFicarelli:
When a Task is run, and one of the dependencies is not fulfilled, and this dependency has a single Target as output (instead of a list of targets, or a dict of targets), then Luigi fails with:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/luigi/worker.py", line 188, in run
    nonexistent_outputs = [output for output in dep.output() if not output.exists()]
TypeError: 'RemoteTarget' object is not iterable
```
The failing code is at [this line](https://github.com/spotify/luigi/blob/25d179b70cbfaf8d78f6cee6c153fcf098babeec/luigi/worker.py#L188) and it seems related to the changes in https://github.com/spotify/luigi/pull/3258.

I think that the code should be fixed to handle the case when `dep.output()` isn't a list of targets.

winniehell:
this was was fixed in https://github.com/spotify/luigi/pull/3281 but not yet released.

@andresgomezfrr do you have a rough estimate on when the next version will be released?

winniehell:
@andresgomezfrr thank you! :heart:

dmyersturnbull:
This is fixed, right? Should this be closed?

PR #3281, opened 2024-03-12
Fix error message in case of unfulfilled dependencies with single output

GianlucaFicarelli:
## Description
Use `flatten()` to handle the outputs of any unfulfilled dependencies when a TaskProcess is run.

## Motivation and Context
When a Task is run, and one of the dependencies is not fulfilled, and this dependency has a single Target as output (instead of a list of targets, or a dict of targets), then Luigi fails and prints the wrong error `TypeError: 'RemoteTarget' object is not iterable`.

See https://github.com/spotify/luigi/issues/3280

Included unit test for `TaskProcess.run()`


winniehell:
@andresgomezfrr can you please take a look at this?

andresgomezfrr (approved):

Issue #2965, opened 2020-06-24
luigi.contrib.azureblob needs updating


marhoy:
The `luigi.contrib.azureblob` depends on the old version of the `azure.storage.blob` package.
It needs to be rewritten/updated to support the 12.x.y-series of `azure.storage.blob`.

@manish, @ckiosidis : You are listed as contributers. Any plans for updating? If you need to continue to support the old interface, then maybe make a new module that supports the new interface?

Old (aka Legacy):
https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob?view=azure-python-previous

New:
https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob?view=azure-python

icfly2:
Is the bot marking as won't fix accurate?

The old azure setup is pretty outdated now and a change / new contrib module shouldn't be too hard. What is the best way forward? Update the contrib.azure, make contrib.azure12 or switch to airflow?

kaydee-edb:
Any update on this ? Are there any plans to update the blob storage API's per new SDK ?

lallea:
This is a community driven open source project. Hence, there are no plans and no roadmap. If you care about updated azure integration, you can drive the issue forward by submitting a patch or valuable information, or you can offer to pay someone to do the necessary work.

kaydee-edb:
Thank you. I can pick it up. Will try to create a new contrib module for azure blob v12.

bcamel:
I am going to bump this up because I intented to use a luigi task to upload some objects into Azure Blob, but noticed that the dependency required for the regarding contrib module are outdated as mentioned in this issue.

For now, I have only used the changes of the linked PR against a local instance of Azurite. Maybe unit tests need extension to cover various parametrisations for the `AzureBlobClient`?

PR #3290, opened 2024-05-19
Updated luigi.contrib.azureblob to 12.x.y series of azure.storage.blob

bcamel:
## Description
Updated `luigi.contrib.azureblob` to adhere to API changes in the Azure Storage Blobs client library for Python.

Related changes:
- Updated the CI script to start an Azurite service via Docker. Uses the official image `mcr.microsoft.com/azure-storage/azurite`
- Updated the related environment variables in the unit test runner to be able to connect to the local Azurite instance.

## Motivation and Context
This PR addresses issue #2965.

- Ran the GitHub Actions including the matrix items `pyXYZ-azureblob`.
- Extended test case to cover the `mkdir` function of `AzureBlobClient`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -48,7 +48,7 @@ deps =
     google-compute-engine
     coverage>=5.0,<6
     codecov>=1.4.0
-    requests>=2.20.0,<3.0
+    requests>=2.20.0,<=2.31.0
     unixsocket: requests-unixsocket<1.0
     pygments
     hypothesis>=6.7.0,<7.0.0
@@ -58,7 +58,7 @@ deps =
     responses<1.0.0
     azure-storage-blob<=12.20.0
     datadog==0.22.0
-    prometheus-client>=0.5.0<0.15
+    prometheus-client>=0.5.0,<0.15
     dropbox: dropbox>=11.0.0
     jsonschema
 passenv =
--- a/tox.ini
+++ b/tox.ini
@@ -76,7 +76,7 @@ setenv =
     AWS_ACCESS_KEY_ID=accesskey
     AWS_SECRET_ACCESS_KEY=secretkey
     AZURITE_ACCOUNT_NAME=devstoreaccount1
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -75,6 +75,9 @@ setenv =
     AWS_DEFAULT_REGION=us-east-1
     AWS_ACCESS_KEY_ID=accesskey
     AWS_SECRET_ACCESS_KEY=secretkey
+    AZURITE_ACCOUNT_NAME=devstoreaccount1
+    AZURITE_ACCOUNT_KEY=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==
```
dlstadther, line 79: To confirm, this account key value is a fake value, right? If so, would Azurite allow being supplied a shorter string that can be more clearly a dumb value, such as `azuritesecretkey`?

dlstadther (commented):
1 question, otherwise, once CI tests pass, i'm good here.

tox.ini
bcamel: Bit puzzled about the unixsockets test suite failing, because seemingly unrelated to this PR, but let me see.

About your comment: Apparently, key has to be base64 encoded. In the added commit, I used a short string that decodes to "azurite".

bcamel: CI failure: Might be the `requests` version.

`master` build on CI used requests-2.31.0, CI build triggered with this PR used requests-2.32.3. 
A (local) run of `tox -e py311-unixsocket` on `master` fails, I forced it to use 2.31.0 version of `requests` and it passed. Can dig deeper, if I have the time. Because `master` fails already, this might not be induced by transitive dependencies of the azure-storage lib introduced here but motivates a separate PR.

dlstadther (approved):

Issue #3294, opened 2024-07-11
In the configuration docs, batch_notifier should be batch_email


ccmaymay:
In the [Configuration documentation](https://luigi.readthedocs.io/en/stable/configuration.html), the batch email section is listed as `batch_notifier`.  However, in [the code](https://github.com/spotify/luigi/blob/3cafd155bbc35918cfce442f9cef445eaa97f13c/luigi/batch_notifier.py#L16), the respective config class is `batch_email`.

ccmaymay:
Fix was merged in 618d777f, released in 3.5.2.  Thank you!

PR #3296, opened 2024-07-11
Fix batch email configuration docs

### LINKED ISSUE #3293: In configuration docs, email_interval_minutes should be email_interval
In the [Configuration documentation](https://luigi.readthedocs.io/en/stable/configuration.html), the email notification interval parameter is listed as `email_interval_minutes`.  However, in [the code](https://github.com/spotify/luigi/blob/3cafd155bbc35918cfce442f9cef445eaa97f13c/luigi/batch_notifier.py#L17), the parameter is `email_interval`.

ccmaymay:
Fix batch_email section name, email_interval parameter name in configuration docs.

I don't think there's much to test; I have been trying to get email notifications working in an application, and they only started working as expected once I discovered these documentation errors and corrected the section/parameter names in my code.

Closes #3293, #3294.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -438,7 +438,7 @@ traceback_max_length
 Parameters controlling the contents of batch notifications sent from the
 scheduler
 
-email_interval_minutes
+email_interval
   Number of minutes between e-mail sends. Making this larger results in
   fewer, bigger e-mails.
   Defaults to 60.
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -432,7 +432,7 @@ traceback_max_length
   Maximum length for traceback included in error email. Default is 5000.
 
 
-[batch_notifier]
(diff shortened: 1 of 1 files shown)


dlstadther (approved):
The docs referenced the config section and name that would have been used by the original configuration system (which was deprecated but not fully cleaned up; [I had actually cleaned up this deprecation in the scheduler](https://github.com/spotify/luigi/pull/2491)). You can see [here](https://github.com/spotify/luigi/blob/master/luigi/batch_notifier.py#L17-L19) that the `email_interval` had a `config_path` which referenced the section and name from the docs. However, in [2015 the original Luigi author updated](https://github.com/spotify/luigi/pull/726) the config section parsing to default to the config class name and its property names.

Once the CI completes (i updated this branch with master to resolve an issue with CI), I'll merge.

PR #3297, opened 2024-07-21
chore: add py.typed for type checking

hiro-o918:
## Description

Add py.typed marker
https://peps.python.org/pep-0561/#packaging-type-information

## Motivation and Context

I want to run type checking on my project with lugi.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ def get_static_files(path):
                         "*.eot", "*.svg", "*.ttf", "*.woff", "*.woff2"]]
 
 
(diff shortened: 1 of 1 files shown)


dlstadther (approved):
I'm not sure if you'll run into issue while type checking. But I'll approve to include the ability.

hiro-o918:
Thank you for your merging.
In detail, if there is no `py.typed`, all the classes are handled as `Any` and no type checking works.
This can makes us write invalid typing as type annotations, because it is not checked by type checkers like mypy

PR #3298, opened 2024-07-21
Replace pipes.quote with shlex.quote

shadchin:
## Description

The `shlex.quote()` API is available from Python 3.3 on; `pipes.quote()` was never documented, and is removed in Python 3.13.

## Motivation and Context
Support Python 3.13

### DIFF
--- a/luigi/contrib/hadoop_jar.py
+++ b/luigi/contrib/hadoop_jar.py
@@ -20,8 +20,8 @@ Provides functionality to run a Hadoop job using a Jar
 
 import logging
 import os
-import pipes
 import random
+import shlex
 import warnings
 
 import luigi.contrib.hadoop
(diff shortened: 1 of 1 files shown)


dlstadther (approved):
`pipes.quote()` was [documented in the Python 2.7 docs](https://docs.python.org/2.7/library/pipes.html#pipes.quote) with the note
> Deprecated since version 2.7: Prior to Python 2.7, this function was not publicly documented. It is finally exposed publicly in Python 3.3 as the quote function in the [shlex](https://docs.python.org/2.7/library/shlex.html#module-shlex) module.

So even though `shlex.quote` comes with the warning `Warning The shlex module is only designed for Unix shells.`, given that it is equivalent to `pipes.quote`, this should have no change of behavior. (also `shlex.quote` was added in Python 3.3 and Luigi only "officially" supports 3.6+, we're still good here).

Thanks for this contribution!

PR #3305, opened 2024-09-02
add ChoiceListParameter

kitagry:
## Description

A parameter type complementary to ChoiceParameter that allows for an arbitrarily sized list (0 to n).

## Motivation and Context

Convenient addition to the Luigi parameter type inventory.

I have included unit tests.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -1540,6 +1540,52 @@ class ChoiceParameter(Parameter):
                 var=var, choices=self._choices))
 
 
+class ChoiceListParameter(ChoiceParameter):
+    """
+    A parameter which takes two values:
+        1. an instance of :class:`~collections.Iterable` and
+        2. the class of the variables to convert to.
+
+    Values are taken to be a list, i.e. order is preserved, duplicates may occur, and empty list is possible.
+
+    In the task definition, use
+
+    .. code-block:: python
+
+        class MyTask(luigi.Task):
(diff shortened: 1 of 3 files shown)


test/choice_list_parameter_test.py
```diff
@@ -0,0 +1,53 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from helpers import unittest
+
+import luigi
+
+
+class ChoiceListParameterTest(unittest.TestCase):
```
dlstadther, line 23: These tests should be put into `test/parameter_test.py`

luigi/parameter.py
```diff
@@ -1540,6 +1540,66 @@ def normalize(self, var):
                 var=var, choices=self._choices))
 
 
+class ChoiceListParameter(Parameter):
```
dlstadther, line 1543: This is fine with me with one exception. If this class is intended to be a multi-select of `ChoiceParameter`, it's counter to object-oriented for it to not inherit from `ChoiceParameter` (or a common Choice-based parent) such for them to maintain alignment of their choice parsing and validation.

What are your thoughts?

dlstadther (changes requested):

kitagry:
@dlstadther Thank you for the review! and I'm sorry for the late response.

I fixed test and ListChoiceParameter inheritance.

dlstadther (approved):

kitagry:
Thank you!

Issue #3301, opened 2024-08-24
Arbitrary file write during tarfile extraction in `luigi/contrib/sge_runner.py`


Ali-Razmjoo:
Hi,

I am reporting a potential security issue with arbitrary file write during tarfile extraction in 

https://github.com/spotify/luigi/blob/master/luigi/contrib/sge_runner.py#L67-L70

Extracting files from a malicious tar archive without validating that the destination file path is within the destination directory can cause files outside the destination directory to be overwritten, due to the possible presence of directory traversal elements (..) in archive paths.

- `..\malicious_file`
- `c:\malicious_file`
- `/etc/passwd`

## Recommendation 
```python
with tarfile.open(sys.argv[1]) as tar:
    for entry in tar:
        #GOOD: Check that entry is safe
        if os.path.isabs(entry.name) or ".." in entry.name:
            raise ValueError("Illegal tar archive entry")
        tar.extract(entry, "/tmp/unpack/")
```

## References
- Snyk: [Zip Slip Vulnerability](https://snyk.io/research/zip-slip-vulnerability)
- OWASP: [Path Traversal](https://owasp.org/www-community/attacks/Path_Traversal)
- Python Library Reference: [TarFile.extract](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extract)
- Python Library Reference: [TarFile.extractall](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall)
- Common Weakness Enumeration: [CWE-22](https://cwe.mitre.org/data/definitions/22.html)

Ali-Razmjoo:
fixed in #3309

PR #3309, opened 2024-09-04
Fix arbitrary file write during tarfile extraction in luigi/contrib/lsf_runner.py and luigi/contrib/sge_runner.py

### LINKED ISSUE #3302: Arbitrary file write during tarfile extraction in `luigi/contrib/lsf_runner.py`
Hi,

I am reporting a potential security issue with arbitrary file write during tarfile extraction in 
https://github.com/spotify/luigi/blob/master/luigi/contrib/lsf_runner.py#L55-L58

Extracting files from a malicious tar archive without validating that the destination file path is within the destination directory can cause files outside the destination directory to be overwritten, due to the possible presence of directory traversal elements (..) in archive paths.

- `..\malicious_file`
- `c:\malicious_file`
- `/etc/passwd`

## Recommendation 
```python
with tarfile.open(sys.argv[1]) as tar:
    for entry in tar:
        #GOOD: Check that entry is safe
        if os.path.isabs(entry.name) or ".." in entry.name:
            raise ValueError("Illegal tar archive entry")
        tar.extract(entry, "/tmp/unpack/")
```

## References
- Snyk: [Zip Slip Vulnerability](https://snyk.io/research/zip-slip-vulnerability)
- OWASP: [Path Traversal](https://owasp.org/www-community/attacks/Path_Traversal)
- Python Library Reference: [TarFile.extract](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extract)
- Python Library Reference: [TarFile.extractall](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall)
- Common Weakness Enumeration: [CWE-22](https://cwe.mitre.org/data/definitions/22.html)

Ali-Razmjoo:
Fixes #3302 and #3301

This pull request introduces a new `SafeExtractor` class to securely handle tar file extraction and replaces the existing tar extraction logic in `lsf_runner.py` and `sge_runner.py` with this new class. Additionally, it includes tests to validate the functionality of `SafeExtractor`.

### Introduction of `SafeExtractor` class:

* [`luigi/safe_extractor.py`](diffhunk://#diff-3514a056843c45119d752414924be04eb1d65d0a6ecf92305a644d19b892d909R1-R96): Added a new `SafeExtractor` class to securely extract tar files and prevent path traversal vulnerabilities. This class ensures that extracted file paths are validated before extraction.

### Refactoring to use `SafeExtractor`:

* [`luigi/contrib/lsf_runner.py`](diffhunk://#diff-e5ef6c9d8f4fb5e1803f70eb9d6cb3e3730c685855f912db8af676822c4dc2a1L31-R31): Replaced the existing tar extraction logic with the `SafeExtractor` class in the `do_work_on_compute_node` and `extract_packages_archive` functions. [[1]](diffhunk://#diff-e5ef6c9d8f4fb5e1803f70eb9d6cb3e3730c685855f912db8af676822c4dc2a1L31-R31) [[2]](diffhunk://#diff-e5ef6c9d8f4fb5e1803f70eb9d6cb3e3730c685855f912db8af676822c4dc2a1L55-R56)
* [`luigi/contrib/sge_runner.py`](diffhunk://#diff-28a14a43f58740fcb079da75257a7211791212a373bbffc51ff8a2701d5a8bf7L39-R39): Updated the `_do_work_on_compute_node` and `_extract_packages_archive` functions to use the `SafeExtractor` class for tar extraction. [[1]](diffhunk://#diff-28a14a43f58740fcb079da75257a7211791212a373bbffc51ff8a2701d5a8bf7L39-R39) [[2]](diffhunk://#diff-28a14a43f58740fcb079da75257a7211791212a373bbffc51ff8a2701d5a8bf7L67-R68)

### Addition of tests for `SafeExtractor`:

* [`test/contrib/lsf_test.py`](diffhunk://#diff-8d3703cd288185dfc522bfba016899d5cab6b9dfb8783f9cae0769865e7e9859R36-R41): Added unit tests for the `SafeExtractor` class to ensure it correctly extracts files and handles path traversal attempts. [[1]](diffhunk://#diff-8d3703cd288185dfc522bfba016899d5cab6b9dfb8783f9cae0769865e7e9859R36-R41) [[2]](diffhunk://#diff-8d3703cd288185dfc522bfba016899d5cab6b9dfb8783f9cae0769865e7e9859R110-R149)

### DIFF
--- a/luigi/safe_extractor.py
+++ b/luigi/safe_extractor.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2012-2015 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""
+This module provides a class `SafeExtractor` that offers a secure way to extract tar files while
+mitigating path traversal vulnerabilities, which can occur when files inside the archive are
+crafted to escape the intended extraction directory.
+
+The `SafeExtractor` ensures that the extracted file paths are validated before extraction to
+prevent malicious archives from extracting files outside the intended directory.
+
+Classes:
+    SafeExtractor: A class to securely extract tar files with protection against path traversal attacks.
+
+Usage Example:
+    extractor = SafeExtractor("/desired/directory")
+    extractor.safe_extract("archive.tar")
+"""
+
+import os
+import tarfile
+
+
+class SafeExtractor:
+    """
+    A class to safely extract tar files, ensuring that no path traversal
+    vulnerabilities are exploited.
+
+    Attributes:
+        path (str): The directory to extract files into.
+
+    Methods:
+        _is_within_directory(directory, target):
+            Checks if a target path is within a given directory.
+
+        safe_extract(tar_path, members=None, \\*, numeric_owner=False):
+            Safely extracts the contents of a tar file to the specified directory.
+    """
+
+    def __init__(self, path="."):
+        """
+        Initializes the SafeExtractor with the specified directory path.
+
+        Args:
+            path (str): The directory to extract files into. Defaults to the current directory.
+        """
+        self.path = path
+
+    @staticmethod
+    def _is_within_directory(directory, target):
+        """
+        Checks if a target path is within a given directory.
+
+        Args:
+            directory (str): The directory to check against.
+            target (str): The target path to check.
+
+        Returns:
+            bool: True if the target path is within the directory, False otherwise.
+        """
+        abs_directory = os.path.abspath(directory)
+        abs_target = os.path.abspath(target)
+        prefix = os.path.commonprefix([abs_directory, abs_target])
+        return prefix == abs_directory
+
+    def safe_extract(self, tar_path, members=None, *, numeric_owner=False):
+        """
+        Safely extracts the contents of a tar file to the specified directory.
+
+        Args:
+            tar_path (str): The path to the tar file to extract.
+            members (list, optional): A list of members to extract. Defaults to None.
+            numeric_owner (bool, optional): If True, only the numeric owner will be used. Defaults to False.
+
+        Raises:
+            RuntimeError: If a path traversal attempt is detected.
+        """
+        with tarfile.open(tar_path, 'r') as tar:
+            for member in tar.getmembers():
+                member_path = os.path.join(self.path, member.name)
+                if not self._is_within_directory(self.path, member_path):
+                    raise RuntimeError("Attempted Path Traversal in Tar File")
+            tar.extractall(self.path, members, numeric_owner=numeric_owner)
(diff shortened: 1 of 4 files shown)


test/contrib/lsf_test.py
```diff
@@ -103,5 +107,45 @@ def tearDown(self):
                 pass
 
 
+class TestSafeExtract(unittest.TestCase):
```
dlstadther, line 110: Any test that is specific to only the SafeExtract should be contained within a test class (and file) specific to its source. In other words, these unit tests for SafeExtract belong in a `test/safe_extractor_test.py` file, not in a contrib location for a module which happens to use the SafeExtract functionality (unless you were wanting to test specific interactions between LSF and SafeExtract.

test/contrib/lsf_test.py
```diff
@@ -103,5 +107,45 @@ def tearDown(self):
...
+        shutil.rmtree(self.temp_dir)
+
+    def test_safe_extract(self):
+        tar_path = os.path.join(self.temp_dir, 'test.tar')
+        with tarfile.open(tar_path, 'w') as tar:
+            for i in range(3):
+                file_path = os.path.join(self.temp_dir, f'test_file_{i}.txt')
+                with open(file_path, 'w') as f:
+                    f.write(f'This is test file {i}')
+                tar.add(file_path, arcname=f'test_file_{i}.txt')
```
dlstadther, line 125: In many places, you define the f-string `f'test_file_{i}.txt'` multiple times rather than defining as a variable and referencing it. This is minor, but would be cleaner and more resilient to content changes.

luigi/safe_extractor.py
```diff
@@ -0,0 +1,97 @@
...
+            numeric_owner (bool, optional): If True, only the numeric owner will be used. Defaults to False.
+
+        Raises:
+            ValueError: If a path traversal attempt is detected.
+        """
+        with tarfile.open(tar_path, 'r') as tar:
+            for member in tar.getmembers():
+                member_path = os.path.join(self.path, member.name)
+                if not self._is_within_directory(self.path, member_path):
+                    raise ValueError("Attempted Path Traversal in Tar File")
```
dlstadther, line 96: Is `ValueError` the appropriate error type here? Is it an issue due to user input?

dlstadther (changes requested):

test/contrib/lsf_test.py
Ali-Razmjoo: correct, will fix it now

test/contrib/lsf_test.py
Ali-Razmjoo: I will send a fix 👍

luigi/safe_extractor.py
Ali-Razmjoo: I am unsure if this happens in scenario of the file being tampered with or replaced on purpose. So I CARPE DIEM(ed) and went with `ValueError`. 

We can do `SecurityError` or `OSError`

```python
class SecurityError(Exception):
    pass

# something like this?
raise SecurityError("Attempted Path Traversal in Tar File")

# or something like this?
raise OSError("Attempted Path Traversal in Tar File")
```

Ali-Razmjoo: Do you have any suggestion?

dlstadther: I'm not sure we need to define a custom exception type for this.

Based on the [docs](https://docs.python.org/3/library/exceptions.html), ValueError nor OSError's description fit the issue here. Therefore, if sticking with a built-in exception time, i'd suggest the fallback option of [`RuntimeError`](https://docs.python.org/3/library/exceptions.html#RuntimeError).

Ali-Razmjoo: done

dlstadther (approved):
Thank you!

PR #3333, opened 2025-01-13
Allow more control over Prometheus metrics collection

mateka:
## Description
I have added possibility to define prometheus metrics labels based on task parameters.

## Motivation and Context
I wanted to have more grouping options for prometheus monitoring. We use one scheduler for many graphs which share tasks and we wanted to have possibility to discern different graphs.

I ran example jobs with this code and it works for me.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -637,6 +637,18 @@ marker_table
   created if it doesn't already exist. Defaults to "table_updates".
 
 
+[prometheus]
+------------
+
+use_task_family_in_labels
+  Should task family be used as a prometheus bucket label.
+  Default value is true.
+
+task_parameters_to_use_in_labels
+  List of task arguments' names used as additional prometheus bucket labels.
+  Passed in a form of a json list.
+
+
 [redshift]
 ----------
 
@@ -1044,6 +1056,7 @@ statsd_port
 metric_namespace
   Optional prefix to add to the beginning of every metric sent to Datadog.
   Default value is "luigi".
+
 
 Per Task Retry-Policy
 ---------------------
--- a/luigi/contrib/prometheus_metric.py
+++ b/luigi/contrib/prometheus_metric.py
@@ -1,40 +1,61 @@
 from prometheus_client import CollectorRegistry, Counter, Gauge, generate_latest, CONTENT_TYPE_LATEST
+from luigi import parameter
 from luigi.metrics import MetricsCollector
+from luigi.task import Config
+
+
+class prometheus(Config):
+    use_task_family_in_labels = parameter.BoolParameter(
+        default=True, parsing=parameter.BoolParameter.EXPLICIT_PARSING
+    )
+    task_parameters_to_use_in_labels = parameter.ListParameter(default=[])
 
 
 class PrometheusMetricsCollector(MetricsCollector):
 
-    def __init__(self):
+    def _generate_task_labels(self, task):
+        return {
+            label: task.family if label == "family" else task.params.get(label)
+            for label in self.labels
+        }
+
+    def __init__(self, *args, **kwargs):
         super(PrometheusMetricsCollector, self).__init__()
         self.registry = CollectorRegistry()
+        config = prometheus(**kwargs)
+        self.labels = list(config.task_parameters_to_use_in_labels)
+        if config.use_task_family_in_labels:
+            self.labels += ["family"]
+        if not self.labels:
+            raise ValueError("Prometheus labels cannot be empty (see prometheus configuration)")
         self.task_started_counter = Counter(
             'luigi_task_started_total',
             'number of started luigi tasks',
-            ['family'],
+            self.labels,
             registry=self.registry
         )
         self.task_failed_counter = Counter(
             'luigi_task_failed_total',
             'number of failed luigi tasks',
-            ['family'],
+            self.labels,
             registry=self.registry
         )
         self.task_disabled_counter = Counter(
             'luigi_task_disabled_total',
             'number of disabled luigi tasks',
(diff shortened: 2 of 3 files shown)


luigi/contrib/prometheus_metric.py
```diff
@@ -1,63 +1,82 @@
...
+    def _generate_task_labels(self, task):
+        return {
+            label: task.family if label == "family" else task.params.get(label)
+            for label in self.labels
+        }
+
     def __init__(self):
         super(PrometheusMetricsCollector, self).__init__()
         self.registry = CollectorRegistry()
+        config = prometheus()
```
dlstadther, line 23: I'm not sure why this is the case (probably a miss during PR review), but the initial implementation of this `PrometheusMetricsCollector` forgot to allow for `*args, **kwargs` to its `__init__()`.

In most other classes in Luigi, we allow for args/kwargs and pass `**kwargs` to the config class initializations to facilitate config overrides. e.g. https://github.com/spotify/luigi/blob/master/luigi/contrib/datadog_metric.py#L27

doc/configuration.rst
```diff
@@ -1045,6 +1045,18 @@ metric_namespace
   Optional prefix to add to the beginning of every metric sent to Datadog.
   Default value is "luigi".
 
+[prometheus]
+---------
+
+use_task_family
```
dlstadther, line 1051: The naming of this config value isn't very self-documenting. Is there an adjustment you can make to the name of this option such to be clearer to the user without requiring consulting documentation for its intent?

dlstadther: same goes for `task_parameters`.

Both of these options refer to bucket labels, but neither name give this context.

A quick glance at the prometheus client for `Counter`, looks like the labels that are generated are passed as `labelnames` https://github.com/prometheus/client_python/blob/5926a7c0ddb8755a5e504e6f077e081f5f4c7674/prometheus_client/metrics.py#L105

dlstadther (changes requested):
Thanks for your contribution. I've left some comments mostly about naming and organization. Please take a look and address them or provide response to the feedback.

Also, please see if there are valuable tests you can add to [the existing prometheus test file](https://github.com/spotify/luigi/blob/master/test/contrib/prometheus_metric_test.py) such that any future contributions avoid causing undesirable changes of behavior.

doc/configuration.rst
```diff
@@ -1045,6 +1045,18 @@ metric_namespace
   Optional prefix to add to the beginning of every metric sent to Datadog.
   Default value is "luigi".
 
+[prometheus]
```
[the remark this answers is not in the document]
mateka: I moved this section in other place.

doc/configuration.rst
```diff
@@ -1045,6 +1045,18 @@ metric_namespace
...
+---------
```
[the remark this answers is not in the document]
mateka: Should be fixed now

luigi/contrib/prometheus_metric.py
mateka: I added init params and it was super useful in unit tests!

doc/configuration.rst
mateka: II changed names to: use_task_family_in_labels and task_parameters_to_use_in_labels. Hope that will make fields meaning more clear.

mateka:
> Thanks for your contribution. I've left some comments mostly about naming and organization. Please take a look and address them or provide response to the feedback.
> 

Wow. Thank you for quick feedback! I addressed your comments.

> Also, please see if there are valuable tests you can add to [the existing prometheus test file](https://github.com/spotify/luigi/blob/master/test/contrib/prometheus_metric_test.py) such that any future contributions avoid causing undesirable changes of behavior.

Yes, after pushing my changes I realized that I had not added unit tests. It took a while, because I was not so sure how to make unit tests code nice and without repetitions. Now all cases should be covered.

I am also thinking about extending scheduler `Task` class by adding metadata dictionary, which could be somehow updated from user provided context in `run` method of `ContextManagedTaskProcess` class. This metadata could be used for even more metrics in custom `MetricsCollector` classes (for instance to measure used RAM or reporting statistics from cloud or map-reduce frameworks). Do you think this is a good direction? Probably it would be even possible to define additional Prometheus metrics based on above mentioned metadata in configuration without custom `Prometheus Metrics Collector` provided by users!

dlstadther (approved):
Thanks for addressing my feedback as well as adding tests!

dlstadther:
> I am also thinking about extending scheduler Task class by adding metadata dictionary, which could be somehow updated from user provided context in run method of ContextManagedTaskProcess class. This metadata could be used for even more metrics in custom MetricsCollector classes (for instance to measure used RAM or reporting statistics from cloud or map-reduce frameworks). Do you think this is a good direction? Probably it would be even possible to define additional Prometheus metrics based on above mentioned metadata in configuration without custom Prometheus Metrics Collector provided by users!

I don't fully picture this extension of `Task`. However, my mind hasn't been deep into the development of Luigi internals for a few years. I'm mostly here to provide PR reviews and try to maintain code quality and test coverage. If you feel you would like some feedback on your idea, feel free to post an `Issue` with your idea and try to elicit feedback from various folks who you have seen are active in this project or who have contributed in the past to related features.

Given that the Luigi project doesn't see the same usage and interaction as it did many years ago (in favor or other tools such as Airflow, Prefect, Dagster, and others), I do have some caution about adding too much complexity to Luigi. I want this project to continue to fulfill the needs of its community, but I also don't want to see it become overly complex to maintain (given the few folks still managing the project).

All that said, I welcome any contribution that fills a need and which will become generally applicable to the community.

PR #3335, opened 2025-01-17
Fixes #3334: Add support for custom parameter classes in mypy plugin

### LINKED ISSUE #3334: Mypy errors with user-defined parameter class
**Description**
The mypy plugin does not work with parameter classes that are user-defined rather than built into luigi. For example:
```python
import luigi
from uuid import UUID


class UUIDParameter(luigi.Parameter):
    def parse(self, s):
        return UUID(s)


class DummyTask(luigi.Task):
    x: int = luigi.IntParameter()
    y: UUID = UUIDParameter()


dummy_task = DummyTask(x=10, y="9b0591d7-a167-4978-bc6d-41f7d84a288c")
reveal_type(dummy_task.x)
reveal_type(dummy_task.y)
```

```
main.py:12: error: Incompatible types in assignment (expression has type "UUIDParameter", variable has type "UUID")  [assignment]
main.py:15: error: Unexpected keyword argument "y" for "DummyTask"  [call-arg]
main.py:16: note: Revealed type is "builtins.int"
main.py:17: note: Revealed type is "uuid.UUID"
```

Revealed types are correct, but mypy gives 2 errors connected with custom parameter class.

**Packages**
```
lockfile==0.12.2
luigi==3.6.0
mypy==1.14.1
mypy-extensions==1.0.0
python-daemon==3.1.2
python-dateutil==2.9.0.post0
six==1.17.0
tenacity==8.5.0
tomli==2.2.1
tornado==6.4.2
typing_extensions==4.12.2
```

starhel:
## Description
Fix for custom parameter classes in mypy plugin.

## Motivation and Context
I'm using custom parameter classes in my code, and I'd like to benefit from the new mypy plugin. The change is not the prettiest one but the smallest I've found.

I've updated unittests.

### DIFF
--- a/luigi/mypy.py
+++ b/luigi/mypy.py
@@ -6,7 +6,6 @@ https://github.com/python/mypy/blob/0753e2a82dad35034e000609b6e8daa37238bfaa/myp
 
 from __future__ import annotations
 
-import re
 import sys
 from typing import Callable, Dict, Final, Iterator, List, Literal, Optional
 
@@ -61,10 +60,6 @@ from mypy.typevars import fill_typevars
 
 METADATA_TAG: Final[str] = "task"
 
-PARAMETER_FULLNAME_MATCHER: Final[re.Pattern] = re.compile(
-    r"^luigi(\.parameter)?\.\w*Parameter$"
-)
-
 if sys.version_info[:2] < (3, 8):
     # This plugin uses the walrus operator, which is only available in Python 3.8+
     raise RuntimeError("This plugin requires Python 3.8+")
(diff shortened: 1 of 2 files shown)


luigi/mypy.py
```diff
@@ -109,6 +109,7 @@ def _task_parameter_field_callback(self, ctx: FunctionContext) -> Type:
         # if no `default` argument is found, return AnyType with unannotated type.
         except ValueError:
             return AnyType(TypeOfAny.unannotated)
+        print(ctx.context)
```
dlstadther, line 112: Let's remove `print()` statements before merging.

dlstadther (commented):
I don't have issue with this change. Tests continue to pass.

Left 1 request.

starhel:
I fixed an accidental print in the code and rebased onto the current master.

dlstadther (approved):

Issue #3342, opened 2025-02-14
Possibility to report more metrics in MetricsCollector


mateka:
I would like to monitor more metrics, than collected by standard `PrometheusMetricsCollector`. In order to do that, I would prepare custom `MetricsCollector` and `TaskContext`. The latter would gather metrics for each task and report it to the scheduler, which would pass them to some subclass of `MetricsCollector` class.

For example, I would like to monitor local tasks running in subprocesses. To do that, I had created classes:
```
class TaskContext:
    """Calculates time and memory for task's process"""
    def __init__(self, task_process):
        self._task_process = task_process
        self._start = None

    def __enter__(self):
        self._start = time.perf_counter()
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        assert self._start is not None
        elapsed = time.perf_counter() - self._start
        used_memory = max(
            resource.getrusage(resource.RUSAGE_SELF).ru_maxrss, resource.getrusage(resource.RUSAGE_CHILDREN).ru_maxrss
        )
        self._task_process.status_reporter.report_task_statistics({"memory": used_memory / 1024, "elapsed": elapsed})
```
and 
```
class MetricCollector(PrometheusMetricsCollector):
   """PrometheusMetricsCollector extended to collect task's elapsed time and used RAM"""
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.task_run_execution_time = Gauge(
            'luigi_task_run_execution_time_seconds',
            'luigi task run method execution time in seconds',
            self.labels,
            registry=self.registry
        )
        self.task_execution_memory = Gauge(
            'luigi_task_max_memory_megabytes',
            'luigi task run method max memory usage in megabytes',
            self.labels,
            registry=self.registry
        )
    
    def handle_task_statistics(self, task, statistics):
        if "elapsed" in statistics:
            self.task_run_execution_time.labels(**self._generate_task_labels(task)).set(statistics["elapsed"])
        if "memory" in statistics:
            self.task_execution_memory.labels(**self._generate_task_labels(task)).set(statistics["memory"])
```

Proper luigi.cfg:
```
[scheduler]
metrics_collector = custom
metrics_custom_import = metrics_collector.MetricCollector

[worker]
task_process_context = task_context.TaskContext

[prometheus]
use_task_family_in_labels = true
task_parameters_to_use_in_labels = ["x","y"]
```

[...]

PR #3343, opened 2025-02-14
Make reporting more metrics for tasks possible

mateka:
I would like to collect and monitor more metrics from luigi graphs runs.

## Description
I added method for passing tasks' run statistics from TaskContext into scheduler and metrics collector.

## Motivation and Context
Example and motivation is in Issue https://github.com/spotify/luigi/issues/3342. 

I tested in on example graph and it worked.

### DIFF
--- a/luigi/metrics.py
+++ b/luigi/metrics.py
@@ -66,6 +66,9 @@ class MetricsCollector(metaclass=abc.ABCMeta):
     def handle_task_done(self, task):
         pass
 
+    def handle_task_statistics(self, task, statistics):
+        pass
+
     def generate_latest(self):
         return
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -271,6 +271,10 @@ task_limit
   thus far. Prevents incidents due to spamming of the scheduler, usually
   accidental. Default: no limit.
 
+task_process_context
+  An optional setting allowing Luigi to import a custom context manager
+  used to wrap the execution of tasks' run methods. Default: no context manager.
+
 timeout
   .. versionadded:: 1.0.20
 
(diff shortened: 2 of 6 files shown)


luigi/metrics.py
```diff
@@ -66,6 +66,9 @@ def handle_task_disabled(self, task, config):
     def handle_task_done(self, task):
         pass
 
+    def handle_task_statistics(self, task, statistics):
+        pass
```
dlstadther, line 70: Is there a common implementation that can be introduced here? Or within the Datadog and/or Prometheus implementations of `MetricsCollector`?

If not, and this method needs to remain as an optionally-implemented method, i would ask at a minimum for there to be appropriately informative and discoverable documentation written surrounding its intended use. (I don't believe there are currently any `docs/` content around metrics collectors).

mateka: This method is entirely optional. I had not provided any documentation and tests yet, because I was hoping to gather some feedback in https://github.com/spotify/luigi/issues/3342 as you suggested in my earlier pull request (https://github.com/spotify/luigi/pull/3333#issuecomment-2591417339).

I will gladly add tests and documentation if you are convinced this change is small enough.

dlstadther:
@mateka , a few PRs were merged today which alter the build and testing process for Luigi, We tried to update this branch with those modification. However, the CI trigger didn't appear to actually kickoff the Github Action workflow.

Would you be able to make an adjustment to this branch and push a change so that we can see the Github Action behavior?

mateka:
> @mateka , a few PRs were merged today which alter the build and testing process for Luigi, We tried to update this branch with those modification. However, the CI trigger didn't appear to actually kickoff the Github Action workflow.
> 
> Would you be able to make an adjustment to this branch and push a change so that we can see the Github Action behavior?

I pushed new version of this branch: rebased on current master with cherry-picked commit and it works now. I also added some documentation and test for new feature.

dlstadther (approved):
Thanks for the documentation and rebase!

Issue #3050, opened 2021-03-09
Lock module: getpcmd on Windows is brittle to changes in wmic


tundish:
I'm running Luigi 3.0.0 on Windows 10 in a MINGW64 terminal (installed via Git Bash 2.9.2.windows.1).

A Windows update at the weekend seems to have changed the behaviour of `wmic` and this breaks code in _lock.py_ (`getpcmd`).

I have a quick fix for it and when it's tidy I'll post a patch in this ticket.

tundish:
Here is a patch for the fix:
```
diff --git a/luigi/lock.py b/luigi/lock.py
index ae5fc67a..7dc5532b 100644
--- a/luigi/lock.py
+++ b/luigi/lock.py
@@ -37,10 +37,12 @@ def getpcmd(pid):
         # Use wmic command instead of ps on Windows.
         cmd = 'wmic path win32_process where ProcessID=%s get Commandline 2> nul' % (pid, )
         with os.popen(cmd, 'r') as p:
-            lines = [line for line in p.readlines() if line.strip("\r\n ") != ""]
-            if lines:
-                _, val = lines
-                return val
+            lines = list(filter(None, (i.strip() for i in p.readlines())))
+            try:
+                return lines[-1]
+            except IndexError:
+                # Fall back to default
+                pass
     elif sys.platform == "darwin":
         # Use pgrep instead of /proc on macOS.
         pidfile = ".%d.pid" % (pid, )
```

I have no internet access from this system and limited opportunity for testing.
Would someone check this for regressions please?

PR #3359, opened 2025-08-26
fixed windows wmic result parsing

bainewedlock:
## Description

In lock.py: changed 
```
                _, val = lines
```
to
```
                val = lines[-1]
```
because it works with the current WMIC version and is generally less brittle.

## Motivation and Context
https://github.com/spotify/luigi/issues/3050

I got the following error everytime I tried to run a luigi task:
```
ERROR: Uncaught exception in luigi
  (...)
  File "(...)\.venv\Lib\site-packages\luigi\lock.py", line 42, in getpcmd
    _, val = lines
    ^^^^^^
ValueError: too many values to unpack (expected 2)
```

Tested on Windows 11.

### DIFF
--- a/luigi/lock.py
+++ b/luigi/lock.py
@@ -39,7 +39,7 @@ def getpcmd(pid):
         with os.popen(cmd, 'r') as p:
             lines = [line for line in p.readlines() if line.strip("\r\n ") != ""]
             if lines:
-                _, val = lines
+                val = lines[-1]
                 return val
     elif sys.platform == "darwin":
         # Use pgrep instead of /proc on macOS.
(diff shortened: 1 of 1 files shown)


dlstadther:
Quick googling uncovers that `wmic` has been deprecated since `Windows 10, version 21H1` and replaced by a powershell command. When asking an AI assistant for the powershell equivalent of this command, it gave:

```powershell
Get-CimInstance Win32_Process -Filter "ProcessId={pid}" | Select-Object CommandLine
```

Perhaps see if the following can be a replacement command:
```python
cmd = f"powershell.exe -Command Get-CimInstance Win32_Process -Filter "ProcessId={pid}" | Select-Object CommandLine"
```

**Note**: this is untested.

dlstadther (commented):
Sorry for my absence here. This lgtm, but does need to be adjusted to pass the flake8 linting test.

PR #3381, opened 2026-01-27
fix: correct sys platform markers for windows; update lockfile

### LINKED ISSUE #3378: 3.7.1 - sys_platform "nt" marker
```
  "python-daemon<2.2.0; sys_platform == 'nt'",
  "python-daemon; sys_platform != 'nt'",
```

i believe the value of this is "win32" on windows.

I think this should be `os_name == "nt"` or `sys_platform == "win32"` or `platform_system == "Windows"`

dlstadther:
## Description

The `sys_platform` markers in the pyproject.toml file were incorrect. The `nt` value isn't a valid value for `sys_platform`; allowed values can be found in [sys.platform](https://docs.python.org/3/library/sys.html#sys.platform) docs. ("nt" was a valid selection from `os.name`).

## Motivation and Context

Resolves https://github.com/spotify/luigi/issues/3378

`uv lock --upgrade` finds no issues with the configuration (though it didn't with the invalid value either)

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -6,10 +6,14 @@ resolution-markers = [
     "python_full_version == '3.9.*' and sys_platform == 'nt'",
     "python_full_version >= '3.8.1' and python_full_version < '3.9' and sys_platform == 'nt'",
     "python_full_version < '3.8.1' and sys_platform == 'nt'",
-    "python_full_version >= '3.10' and sys_platform != 'nt'",
-    "python_full_version == '3.9.*' and sys_platform != 'nt'",
-    "python_full_version >= '3.8.1' and python_full_version < '3.9' and sys_platform != 'nt'",
-    "python_full_version < '3.8.1' and sys_platform != 'nt'",
+    "python_full_version >= '3.10' and sys_platform == 'win32'",
+    "python_full_version >= '3.10' and sys_platform != 'nt' and sys_platform != 'win32'",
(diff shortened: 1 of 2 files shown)


uv.lock
```diff
@@ -6,10 +6,14 @@ resolution-markers = [
     "python_full_version == '3.9.*' and sys_platform == 'nt'",
     "python_full_version >= '3.8.1' and python_full_version < '3.9' and sys_platform == 'nt'",
     "python_full_version < '3.8.1' and sys_platform == 'nt'",
-    "python_full_version >= '3.10' and sys_platform != 'nt'",
-    "python_full_version == '3.9.*' and sys_platform != 'nt'",
-    "python_full_version >= '3.8.1' and python_full_version < '3.9' and sys_platform != 'nt'",
-    "python_full_version < '3.8.1' and sys_platform != 'nt'",
+    "python_full_version >= '3.10' and sys_platform == 'win32'",
+    "python_full_version >= '3.10' and sys_platform != 'nt' and sys_platform != 'win32'",
```
dlstadther: It's weird that `uv lock` kept the `sys_platform != 'nt'` references. But doing `uv lock --upgrade` breaks CI because unbounded dependencies get included which break CI for certain python versions.

I don't have the time to debug further. This lock file will be fine for these conditions where `sys_platform != 'nt' and sys_platform != 'win32'`. Looks ugly, but functionally ok.

RRap0so (approved):

Issue #3382, opened 2026-01-29
mypy plugin incorrectly computes return type for some parameters


anlambert:
After upgrading to `luigi 3.7.2`, `mypy` reports a lot of errors related to luigi task parameters in some of our projects, see [swh-export.log](https://jenkins.softwareheritage.org/job/DENV/job/tests/2820/artifact/swh-export.log), [swh-datasets.log](https://jenkins.softwareheritage.org/job/DENV/job/tests/2820/artifact/swh-datasets.log) or [swh-graph.log](https://jenkins.softwareheritage.org/job/DENV/job/tests/2820/artifact/swh-graph.log).

While some of these errors must be fixed from our side, I noticed the luigi mypy plugin does not correctly compute type for task parameters that use a `TypeVar` in their definition. The concerned parameters types are: `ChoiceParameter`, `ChoiceListParameter`, `EnumParameter` and `EnumListParameter`.

You can reproduce the issue using that `test_mypy_luigi_parameters.py` file.
```python
from enum import Enum
from typing import Tuple

import luigi


class MyEnum(Enum):
    A = 1
    B = 2
    C = 3

class MyTask(luigi.Task):
    choice_int_p: int = luigi.parameter.ChoiceParameter(choices=[1, 2, 3])
    choice_list_int_p: Tuple[int, ...] = luigi.parameter.ChoiceListParameter(choices=[1, 2, 3])
    choice_list_str_p: Tuple[str, ...] = luigi.parameter.ChoiceListParameter(choices=["foo", "bar", "baz"])
    choice_str_p: str = luigi.parameter.ChoiceParameter(choices=["foo", "bar", "baz"])
    enum_p: MyEnum = luigi.parameter.EnumParameter(enum=MyEnum)
    enums_p: Tuple[MyEnum, ...] = luigi.parameter.EnumListParameter(enum=MyEnum)

MyTask(
    choice_int_p=1,
    choice_list_int_p=(1, 3),
    choice_list_str_p=("foo", "bar"),
    choice_str_p="toto",
    enum_p=MyEnum.A,
    enums_p=(MyEnum.B, MyEnum.C)
)
```

```shell
$ mypy --version
mypy 1.19.1 (compiled: yes)

$ mypy --config-file pyproject.toml test_mypy_luigi_parameters 
luigi_mypy_choiceparameter_test.py:13: error: Incompatible types in assignment (expression has type "ChoiceType", variable has type "int")  [assignment]
[...]

PR #3383, opened 2026-01-29
fix(mypy): Ensure all task parameters get correctly typed

anlambert:
This pull request is a fix for https://github.com/spotify/luigi/issues/3382.

Some special parameters using a `TypeVar` failed to be correctly typed so handle these edge cases.

Update test to check proper typing for all parameter types.


dlstadther (approved):
Looks reasonable to me. Thanks for submitting the solution to your problems!

PR #3387, opened 2026-02-04
Use `__set__` / `__get__` descriptors for type checking (replaces `__new__` approach)

### LINKED ISSUE #3386: Proposal: Use `Parameter[T]` type annotation instead of `T` for better type checker compatibility
## Summary

Following the recent type improvements in #3369 and #3370, I'd like to propose updating the recommended type annotation style for Luigi parameters.

## Background

In #3370, I added `@dataclass_transform` support and a `.pyi` stub file. At that time, I was influenced by the existing mypy plugin (#3315) which uses the `s: str = Parameter()` style, so I followed the same pattern in the documentation examples.

However, after further investigation, I found that the `s: Parameter[str] = Parameter()` style works better across all major type checkers.

However, after further investigation, I found that the `s: Parameter[str] = Parameter()` style works  better and **doesn't require any of these workarounds** - the `__get__` and `__set__` descriptor protocol handles  everything.                                                                                           

## Current Style

```python
class MyTask(luigi.Task):
    s: str = luigi.Parameter()
    n: int = luigi.IntParameter()
```

## Proposed Style

```python
class MyTask(luigi.Task):
    s: Parameter[str] = luigi.Parameter()
    n: IntParameter = luigi.IntParameter()
```

## Type Checker Comparison

Assignment Error at Class Definition

| Type Checker | Style 1 (`s: str`) | Style 2 (`s: Parameter[str]`) |                                 
| ------------ | ------------------ | ----------------------------- |                                 
| pyright      | ❌ Error           | ✅ No error                   |                                 
| mypy         | ❌ Error           | ✅ No error                   |                                 
| pyrefly      | ❌ Error           | ✅ No error                   |                                 
| ty           | ❌ Error           | ✅ No error                   |                                 

### Type Inference (Both styles work correctly)

| Access           | Style 1          | Style 2                       |                               
| ---------------- | ---------------- | ----------------------------- |                               
| `task.s`         | `str` ✅         | `str` ✅                      |                               
| `task.n`         | `int` ✅         | `int` ✅                      |                               
| `Task.__init__`  | `(s: str, n: int)` ✅ | `(s: str, n: int)` ✅    |                               
| `Task.s`         | `str`            | `Parameter[str]` (more accurate) |                            

Wrong Type Detection in __init__ (Both styles work correctly)

# Both styles correctly detect this error:
task = MyTask(s="hello", n="42")  # Error: n should be int!

| Type Checker | Style 1          | Style 2          |                                                
| ------------ | ---------------- | ---------------- |                                                
| pyright      | ✅ Detects error | ✅ Detects error |                                                
| mypy         | ✅ Detects error | ✅ Detects error |                                                
| pyrefly      | ✅ Detects error | ✅ Detects error |                                                
| ty           | ✅ Detects error | ✅ Detects error |                                                

## Why Style 2 is Better                                                                              
                                                                                                      
1. **No type errors** - Works out of the box with all major type checkers                             
2. **No `__new__` override needed** - Style 1 requires `__new__` to return `T` to trick type checkers (#3369)                                                                                               
3. **No `.pyi` stub file needed** - Style 1 requires `parameter.pyi` for proper type inference (#3370)
4. **No mypy plugin needed** - Style 1 relies on custom mypy plugin (#3315)                           
5. **More accurate class access** - `Task.s` correctly shows `Parameter[str]`                         
6. **Same type safety** - `__init__` type checking works identically                                  
                                                                                                      
## Simplification Opportunity                                                                         
                                                                                                      
If we adopt Style 2 as the recommended approach, we could potentially:                                
                                                                                                      
- Remove the `__new__` overrides added in #3369                                                       
- Simplify or remove `parameter.pyi` added in #3370                                                   
- Deprecate the mypy plugin (#3315) in favor of native `@dataclass_transform` support                 
                                                                                                      
## Suggested Actions                                                                                  
                                                                                                      
1. Update documentation examples to use `Parameter[T]` style                                          
2. Consider simplifying the type infrastructure (remove `__new__` hacks, etc.)                        
3. Mention both styles in docs (with Style 2 as recommended)                                          
                                                                                                      
## Related                                                                                            
                                                                                                      
- #3369 - Add generic type support to Parameter classes (adds `__new__` overrides)                    
- #3370 - Add parameter.pyi for type hinting parameters                                               
- #3315 - mypy plugin (uses Style 1)

kitagry:
fix #3386 

## Description

This PR addresses #3386 by implementing the descriptor protocol ( `__set__` and `__get__` ) in `luigi.Parameter`. This replaces the previous experimental approach that relied on modifying `__new__` , offering a more standard and robust solution for static type checking.

## Motivation and Context

As discussed in #3386, we want static analysis tools (like Pyright) to correctly validate types passed to the Task constructor.
Previously, we explored using `__new__` to manipulate the class instantiation, but that approach was invasive and prone to side effects.

This PR adopts a cleaner strategy: making Parameter a data descriptor.

### Implementation Details

1. Added `__set__` (The Type Hint Hook):

By defining `__set__,` we satisfy PEP 681 compliant type checkers. They now recognize that MyTask(param=value) expects a specific type matching the descriptor's setter.

Note: Luigi's `Task.__init__ `still writes directly to `self.__dict__`, bypassing this method at runtime during initialization. This ensures no performance penalty.

2. Added `__get__` (Preserving Behavior):

Adding `__set__` promotes Parameter to a data descriptor. This changes Python's attribute lookup priority (descriptors are now checked before `instance.__dict__`).

To maintain Luigi's existing behavior (where values are stored in `instance.__dict__`), I implemented `__get__` to explicitly retrieve the value from the instance's dictionary.

This fixes regressions seen in tests (e.g., TaskAsParameterName1335Test) where the Parameter instance itself was being returned instead of its value when shadowing occurred.

### Benefits

* Cleaner Code: Removes the need for complex `__new__` hacks.
* Type Safety: Enables strict type checking for Task parameters in IDEs.
* Backward Compatibility: Runtime behavior remains unchanged as `__get__` proxies to `__dict__`.


luigi/mypy.py
```diff
@@ -107,29 +108,55 @@ def _task_parameter_field_callback(self, ctx: FunctionContext) -> Type:
...
+                                element_type = choices_type.args[0]
+
+                is_list = default_type.type.fullname in (
+                    "luigi.parameter.ChoiceListParameter",
+                    "luigi.parameter.EnumListParameter",
+                )
+                if is_list:
+                    return ctx.api.named_generic_type("builtins.tuple", [element_type])
+                else:
+                    return element_type
```
dlstadther: Thoughts on keeping separate conditional branches for `ChoiceList / EnumList` and `Choice / Enum` and creating a dedication function that encapsulates the shared logic that each can call? With them both in the same condition (with nested loop and conditional), it gets hard to see what's really happening.
Open to other refactors.

dlstadther (commented):
2 comments:
* complex conditional logic that would be good to refine
* concern for `x: Parameter[str]` vs `y: IntParameter` being confusing due to misalignment

luigi/mypy.py
kitagry: fixed in [a20acc1](https://github.com/spotify/luigi/pull/3387/commits/a20acc143a9d920c2635706c44acdbad38621c60)

dlstadther (approved):
Thanks!

PR #3396, opened 2026-02-22
update tenacity

### LINKED ISSUE #3345: Tenacity version out of date
Hi - the current tenacity version is out of date - it's currently pinned to the 8.* major version and 9.0.0 is out - this is causing some compatibility issues with other libraries.

If it would be possible to update this, it would be really helpful! Thanks!

hirosassa:
## Description
SSIA 
fix #3345 

## Motivation and Context

follow latest major version of tenacity (v9)

FYI:
The latest tenacity dropped python 3.8 and 3.9 support.
ref: https://pypi.org/project/tenacity/9.1.4/

CI ran successfully

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ license = {file = "LICENSE"}
 requires-python = ">=3.8.0, <3.13"
 dependencies = [
   "python-dateutil>=2.7.5,<3",
-  "tenacity>=8,<9",
+  "tenacity>=9",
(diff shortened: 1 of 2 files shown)


dlstadther (approved):
Thanks; Regarding python versions, we can probably drop support for older python versions (as a dedicated PR). Looking at https://pypistats.org/packages/luigi , we can probably do 3.10+

PR #3398, opened 2026-02-22
add pytest-xdist for faster unit test

hirosassa:
## Description

In this PR, I added pytest-xdist for making CI faster.
I only added xdist for `core` test because it takes about 5 minutes and others take about 1 minute or less.

As a result, `pyXXX-core` workflows run in 2-4 minutes (reduces 1-2 minutes).

## Motivation and Context

For faster CI workflow and local test.

CI ran successfully

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -2,6 +2,25 @@ from typing import List
 
 import pytest
 
+import luigi.task_register
+
+
+@pytest.fixture(autouse=True)
+def reset_luigi_registry():
+    """Reset the Luigi task registry before and after each test.
+
+    Prevents registry pollution between tests when running with pytest-xdist,
+    where multiple tests execute sequentially within the same worker process.
+    This mirrors the behaviour of LuigiTestCase.setUp/tearDown and applies it
+    to all tests automatically, including those that inherit unittest.TestCase
+    directly without going through LuigiTestCase.
+    """
+    original = luigi.task_register.Register._get_reg()
+    luigi.task_register.Register.clear_instance_cache()
+    yield
+    luigi.task_register.Register._set_reg(original)
+    luigi.task_register.Register.clear_instance_cache()
+
 
 def pytest_collection_modifyitems(items: List[pytest.Item]) -> None:
     """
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -73,7 +73,7 @@ commands =
     unixsocket: {envpython} test/runtests.py -m unixsocket {posargs:}
     dropbox: {envpython} test/runtests.py -m dropbox {posargs:}
     azureblob: {envpython} test/runtests.py -m azureblob {posargs:}
-    core: {envpython} test/runtests.py --doctest-modules -m "not minicluster and not gcloud and not mysql and not postgres and not unixsocket and not contrib and not apache and not aws and not azureblob and not dropbox" {posargs:}
+    core: {envpython} test/runtests.py --doctest-modules -m "not minicluster and not gcloud and not mysql and not postgres and not unixsocket and not contrib and not apache and not aws and not azureblob and not dropbox" -n auto --dist=loadfile {posargs:}
```
hirosassa: `--dist loadfile` ensures all tests from the same file run on same worker, preventing registry pollution caused by module-level `luigi.Task` subclasses conflicting with identically named classes defined inside test methods of other files.

test/conftest.py
```diff
@@ -2,6 +2,25 @@
...
+    where multiple tests execute sequentially within the same worker process.
+    This mirrors the behaviour of LuigiTestCase.setUp/tearDown and applies it
+    to all tests automatically, including those that inherit unittest.TestCase
+    directly without going through LuigiTestCase.
+    """
+    original = luigi.task_register.Register._get_reg()
+    luigi.task_register.Register.clear_instance_cache()
+    yield
+    luigi.task_register.Register._set_reg(original)
+    luigi.task_register.Register.clear_instance_cache()
```
hirosassa: Each worker runs multiple tests sequentially in the same process, so task classes registered in on test can leak into the next. This fixture snapshots and restores `Register._reg` around every test, covering tests that do not inherit `LuigiTestCase`

dlstadther (approved):
The speed improvement is minimal likely because the slower tests are often in the same file. I was messing with tests for one of the core modules (maybe server py?) and recall the whole set of tests being very slow.

PR #3403, opened 2026-03-01
feat: introduce mypy type checking with gradual adoption

kitagry:
## Description
Add mypy static type checking to the CI pipeline using a gradual adoption strategy. Currently failing modules are added to a per-module `ignore_errors` list in `pyproject.toml`, allowing new code to be checked immediately while existing errors are fixed incrementally.
                                                                                          
This PR also fixes type errors in 13 modules that were easy to resolve, and corrects a long-standing type annotation bug in `parameter.py` where `config_path` was declared as `Optional[str]` but the actual API has always required `{"section": ..., "name": ...}`. A new `ConfigPath` TypedDict now makes this contract explicit.
                                                                                          
**Changes:**

- Add `[tool.mypy]` config to `pyproject.toml` with per-module `ignore_errors` overrides for the 12 remaining modules with errors
- Add `[testenv:mypy]` to `tox.ini` and `mypy` step to GitHub Actions CI
- Fix type errors in 13 modules (`batch_notifier`, `beam_dataflow`, `bigquery`, `external_daily_snapshot`, `hdfs/webhdfs_client`, `hive`, `opener`, `prometheus_metric`, `rdbms`, `sge`, `mock`, `server`, `tools/deps`)
- Add `ConfigPath` TypedDict and fix `config_path` annotation in `parameter.py`
                                                                                          
## Motivation and Context
mypy is already listed as a dev dependency but was never enforced. This change makes type checking a mandatory CI gate and establishes a path to gradually improve type coverage across the codebase.
                                                                                          
`tox run -e mypy` passes with no errors (`Success: no issues found in 102 source files`).


tox.ini
```diff
@@ -95,6 +95,11 @@ commands =
     ruff check .
     ruff format --check .
 
+[testenv:mypy]
```
dlstadther: Could we name this section something more generic? Like "typecheck".
That way we can swap other tools in/out without needing to change the name of the test or repo configuration on which test names are required vs optional.

dlstadther (commented):
minor comments

dlstadther (approved):

PR #3406, opened 2026-03-08
feat: type check parameter classes

kitagry:
## Description

Enable type checking for luigi/parameter.py.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -142,6 +142,7 @@ class UnconsumedParameterWarning(UserWarning):
 
 
 T = TypeVar("T", default=str)
+_OptT = TypeVar("_OptT")
 
 
 class ConfigPath(TypedDict):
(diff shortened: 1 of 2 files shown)


dlstadther (approved):
There are quite a number of type-ignores on here. Are they included because we you gave up trying to solve them? Or because the design of the parameter system (and typing) doesn't have a solution?
There still more value having a subset of typechecks than no type checks.

kitagry:
These `type: ignore` comments are due to a fundamental limitation of mypy's support for Generic mixins with cooperative multiple inheritance.

OptionalParameterMixin uses super() calls whose MRO is only resolved at runtime. Mypy cannot statically determine the type of super() in this context, which causes misc/call-arg/arg-type errors. The same applies to the subclass definitions (OptionalIntParameter etc.) where mypy detects conflicting `__get__` overloads from different base classes.

These are not "gave up" situations — they reflect the design of the existing mixin pattern, which  Python's type system cannot fully express today. Resolving this properly would require reworking  OptionalParameterMixin itself (e.g., replacing the mixin pattern with composition or a factory  approach), which is out of scope for this PR. The suppressed errors are intentionally scoped to this  mixin boundary.

PR #3411, opened 2026-03-25
Proposal: set default value of OptionalParameter to None

kitagry:
## Description

Currently, when using `OptionalParameter` (and its variants like OptionalStrParameter), users must explicitly set `default=None`. If omitted, Luigi throws a `MissingParameterException`.

Since the prefix `"Optional"` strongly implies that the parameter can be omitted and will default to a null state, the current behavior is slightly counter-intuitive. This PR proposes changing the default value of optional parameters from `_no_value` to `None`.

```python
class MyTask(luigi.Task):
    param1: luigi.Parameter[str | None] = luigi.OptionalStrParameter()

    def output(self):
        return luigi.LocalTarget(f"output_{self.param1}.txt")

    def run(self):
        with self.output().open("w") as f:
            f.write(f"Parameter 1: {self.param1}\n")


luigi.build([MyTask()], local_scheduler=True)
```

### expected behavior

```

Scheduled 1 tasks of which:
* 1 ran successfully:
    - 1 MyTask(param1=)
```

### actual behavior

```
luigi.parameter.MissingParameterException: MyTask[args=(), kwargs={}]: requires the 'param1' parameter to be set
```

## Motivation and Context

To improve the developer experience by aligning the behavior of OptionalParameter with standard Python conventions and user expectations. This change reduces boilerplate code (removing the need to write default=None everywhere) and prevents unexpected errors when an optional parameter is simply omitted.

Verified that the provided reproduction code now runs successfully and outputs as expected.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -436,7 +436,7 @@ class OptionalParameterMixin(Generic[_OptT]):
 
     def __init__(
         self,
-        default: Union[_OptT, None, _NoValueType] = _no_value,
+        default: Union[_OptT, None, _NoValueType] = None,
         **kwargs: Unpack[_ParameterKwargs],
     ):
         super().__init__(default=default, **kwargs)  # type: ignore[arg-type, call-arg, misc]
--- a/test/parameter_test.py
+++ b/test/parameter_test.py
@@ -415,6 +415,14 @@ class ParameterTest(LuigiTestCase):
             'OptionalParameter "param" with value "1" is not of type "str" or None.', luigi.parameter.OptionalParameterTypeWarning
         )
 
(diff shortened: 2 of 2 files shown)


kitagry:
@dlstadther Sorry. Could you review this?

dlstadther (dismissed):
I'm good here. Are there any test cases that need to be added/updated?

kitagry:
@dlstadther Thank you for the review! I added a test to verify that a Task with OptionalParameter() (no explicit default) can be instantiated without a value and defaults to None.

dlstadther (approved):
LGTM. Thanks for adding a test case!

PR #3412, opened 2026-04-05
feat: add type annotations to target and contrib modules

kitagry:
## Description
Add mypy type annotations to `luigi/target.py`, `luigi/format.py`, and several contrib modules (`s3`, `webhdfs`, `gcs`, `hadoop`). This is part of the gradual type-checking adoption introduced in #3403.

## Motivation and Context

Following the mypy infrastructure introduced in #3403 and the parameter class annotations in #3406, this PR extends type coverage to the Target layer and key contrib modules — enabling static analysis to catch type errors in the core I/O abstractions.

Changes per commit:
- `luigi/target.py`, `luigi/contrib/s3.py`, `luigi/contrib/webhdfs.py`, `luigi/format.py`: add Target type annotations
- `luigi/contrib/gcs.py`: enable GCS target type (unblock from ignore list)
- `luigi/contrib/hadoop.py`: add type annotations, add `types-requests` dep

I ran mypy on the modified modules and confirmed no new type errors are introduced.

### DIFF
--- a/luigi/target.py
+++ b/luigi/target.py
@@ -27,6 +27,8 @@ import random
 import tempfile
 import warnings
 from contextlib import contextmanager
+from types import TracebackType
+from typing import IO, Any, Generator, Iterator, Optional, Union
 
 logger = logging.getLogger("luigi-interface")
 
@@ -44,7 +46,7 @@ class Target(metaclass=abc.ABCMeta):
     """
 
     @abc.abstractmethod
-    def exists(self):
+    def exists(self) -> bool:
         """
         Returns ``True`` if the :py:class:`Target` exists and ``False`` otherwise.
         """
@@ -99,7 +101,7 @@ class FileSystem(metaclass=abc.ABCMeta):
     """
 
     @abc.abstractmethod
-    def exists(self, path):
+    def exists(self, path: str) -> bool:
         """
         Return ``True`` if file or directory at ``path`` exist, ``False`` otherwise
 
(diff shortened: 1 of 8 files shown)


dlstadther:
@kitagry , i'm sorry i lost track of this PR. Looks like there's a merge conflict. I'd be happy to go through the review of this. Could you resolve merge conflict and update?

kitagry:
@dlstadther Sorry for the late reply. I fixed conficts!

luigi/target.py
```diff
@@ -321,31 +322,36 @@ class AtomicLocalFile(io.BufferedWriter):
...
+        self,
+        exc_type: Optional[type[BaseException]],
+        exc: Optional[BaseException],
+        traceback: Optional[TracebackType],
+    ) -> None:
         "Close/commit the file if there are no exception"
         if exc_type:
             return
-        return super(AtomicLocalFile, self).__exit__(exc_type, exc, traceback)
+        super(AtomicLocalFile, self).__exit__(exc_type, exc, traceback)
```
dlstadther, line 357: Technically, this changes behavior. But moves it to the correct behavior.

pyproject.toml
```diff
@@ -89,6 +89,7 @@ common = [
   "types-python-dateutil",
   "types-requests",
   "types-toml",
+  "types-ujson>=5.10.0.20250822",
```
dlstadther, line 92: Help me understand why this additional dependency is needed. I didn't see any explicit import of it in this PR.

dlstadther (commented):
1 comment. 1 question

pyproject.toml
kitagry: Sorry for the late reply. `ujson` is used in `luigi/contrib/hadoop.py`, and mypy occurs error like the following if the library is not installed.

```
error: Library stubs not installed for "ujson" [import-untyped]
note: Hint: "python3 -m pip install types-ujson"
note: (or run "mypy --install-types" to install all missing stub packages)
```

dlstadther (approved):
Apologies for the significant delay!

PR #3414, opened 2026-04-08
Fix typerror in `_update_priority` in scheduler.py

### LINKED ISSUE #3353: TypeError in _update_priority in scheduler.py
### Description

Calling `Scheduler.add_task()` with mocked task dependencies causes a `TypeError` due to an implicit comparison between a `MagicMock` and `int`. This happens when `task.priority` is a `MagicMock`, and `_update_priority()` tries to call `max(prio, task.priority)`, which results in `TypeError: '>' not supported between instances of 'MagicMock' and 'int'`.

A potential fix is to add a check or default initialization for `task.priority` before using it in a numerical comparison.

### Steps to Reproduce
Minimal reproducible test:

```python
from helpers import unittest
import mock
from luigi.scheduler import Scheduler, PENDING

WORKER = 'myworker'

class AddTaskTest(unittest.TestCase):

    def setUp(self):
        self.s = Scheduler()
        self.s._state = mock.MagicMock()
        self.s._config = mock.MagicMock()
        self.s._config.stable_done_cooldown_secs = 300
        self.s._update_worker = mock.MagicMock()
        self.s._update_task_history = mock.MagicMock()
        self.s._generate_retry_policy = mock.MagicMock()
        self.worker = mock.MagicMock()
        self.worker.enabled = True
        self.s._update_worker.return_value = self.worker

    def test_task_dependencies(self):
        task = mock.MagicMock()
        task.status = PENDING
        task.deps = set()
        self.s._state.get_task.return_value = task
        self.s.add_task(worker=WORKER, task_id='test_task', deps=['dep1', 'dep2'], new_deps=['dep3'])

if __name__ == '__main__':
    unittest.main()
```

**Expected behavior:**
The task should perform type checking or raise a clear error if `task.priority` is not valid.

**Actual behavior:**
the stack trace is:
```
luigi/scheduler.py:927: in add_task 
  self._update_priority(task, priority, worker_id)

self = <luigi.scheduler.Scheduler object at 0x7f76f6666770> task = , prio = 0 worker = 'myworker'

def _update_priority(self, task, prio, worker):
    """
    Update priority of the given task.

    Priority can only be increased.
    If the task doesn't exist, a placeholder task is created to preserve priority when the task is later scheduled.
    """
  task.priority = prio = max(prio, task.priority)
E TypeError: '>' not supported between instances of 'MagicMock' and 'int'```

```

subhramit:
## Description
Add a check for `task.priority` as described in https://github.com/spotify/luigi/issues/3353 and default to 0.

## Motivation and Context
Closes https://github.com/spotify/luigi/issues/3353

Quoting from the issue:
>Calling `Scheduler.add_task()` with mocked task dependencies causes a `TypeError` due to an implicit comparison between a `MagicMock` and `int`. This happens when `task.priority` is a `MagicMock`, and `_update_priority()` tries to call `max(prio, task.priority)`, which results in `TypeError: '>' not supported between instances of 'MagicMock' and 'int'`.

>A potential fix is to add a check or default initialization for task.priority before using it in a numerical comparison.

This PR implements the fix.
 

I added unit tests replicating the typeerror scenarios.

### DIFF
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -761,6 +761,10 @@ class Scheduler:
         Priority can only be increased.
         If the task doesn't exist, a placeholder task is created to preserve priority when the task is later scheduled.
         """
+        existing_priority = (
+            task.priority if isinstance(task.priority, (int, float)) else 0
+        )
+        task.priority = prio = max(prio, existing_priority)
         task.priority = prio = max(prio, task.priority)
         for dep in task.deps or []:
             t = self._state.get_task(dep)
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -761,9 +761,7 @@ class Scheduler:
         Priority can only be increased.
         If the task doesn't exist, a placeholder task is created to preserve priority when the task is later scheduled.
         """
-        existing_priority = (
-            task.priority if isinstance(task.priority, (int, float)) else 0
-        )
+        existing_priority = task.priority if isinstance(task.priority, (int, float)) else 0
         task.priority = prio = max(prio, existing_priority)
         task.priority = prio = max(prio, task.priority)
         for dep in task.deps or []:
--- a/luigi/scheduler.py
+++ b/luigi/scheduler.py
@@ -225,13 +225,15 @@ class OrderedSet(MutableSet):
 
 
 class Task:
+    DEFAULT_PRIORITY = 0
+
     def __init__(
         self,
         task_id,
         status,
         deps,
         resources=None,
-        priority=0,
+        priority=DEFAULT_PRIORITY,
         family="",
         module=None,
         params=None,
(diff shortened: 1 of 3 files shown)


dlstadther (commented):
2 things:

1. DRY the default priority of zero. Currently, if we were to change the default value when initializing Task, there would be a disconnect between `Task.__init__()` and `Scheduler._update_priority()`
2. avoid mutating `task.priority` and `prio` back to back; it's necessary to set them as the max value, but do so once not twice

subhramit:
> 2 things:
> 
> 1. DRY the default priority of zero. Currently, if we were to change the default value when initializing Task, there would be a disconnect between `Task.__init__()` and `Scheduler._update_priority()`
> 2. avoid mutating `task.priority` and `prio` back to back; it's necessary to set them as the max value, but do so once not twice

Done. Thanks!

dlstadther (approved):

PR #3416, opened 2026-04-09
Fix BoolParameter to consider global parsing default.

riga:
## Description

As part of the latest changes for the 3.8.0 release, some type hints were added that break changing the parsing option of BoolParameters. This PR contains a small 2-line fix to consider the default as before.

## Motivation and Context

The docstring of BoolParameter states that the `parsing` default value can be set

> globally by

    .. code-block:: python

        luigi.BoolParameter.parsing = luigi.BoolParameter.EXPLICIT_PARSING

However, this is no longer possible and breaks code on our side since the value is never used. As before (I created the PR for that 8 years ago), it *should* be evaluated at init time.

## Have you tested this?

The two line fix should be self-explanatory, but we tested the fix and it works as expected.

---

Due to things breaking for everyone making use of the global setting, 3.8.0 should probably be removed from pypi and re-released.

### DIFF
--- a/luigi/parameter.py
+++ b/luigi/parameter.py
@@ -870,10 +870,10 @@ class BoolParameter(Parameter[bool]):
     def __init__(
         self,
         default: Union[bool, _NoValueType] = _no_value,
-        parsing: str = IMPLICIT_PARSING,
+        parsing: str | None = None,
         **kwargs: Unpack[_ParameterKwargs],
     ):
-        self.parsing = parsing
+        self.parsing = self.__class__.parsing if parsing is None else parsing
         super().__init__(default=default, **kwargs)
         if self._default == _no_value:
             self._default = False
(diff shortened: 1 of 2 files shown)


dlstadther (commented):
Hi @riga , Thanks for bringing this up and contributing an adjustment. I can get this change merged and a 3.8.1 patch release prepped soon.

Prior to merging, could you add some tests to ensure this doesn't get broken again? Luigi contains many features and lots of them don't include any test coverage. This makes it way too easy to break existing functionality that isn't known by the contributor.

riga:
Thanks @dlstadther 👍 

I extended the test cases and made sure they break *without* the fix in this PR.

dlstadther (approved):

dlstadther:
@riga , sorry for the delays; i have finally released 3.8.1
* https://github.com/spotify/luigi/releases/tag/v3.8.1
* https://pypi.org/project/luigi/3.8.1/

PR #3418, opened 2026-04-11
fix(server): resolve /history 500 error by correcting importlib.resources usage

### LINKED ISSUE #3415: Scheduler /history endpoint throws 500 Server Error
### Problem

The /history endpoint in the scheduler web UI fails with a 500 error.

### Steps to reproduce

Using Python 3.13:

```
> pip install 'luigi==3.8.0' 'sqlalchemy<1.4'
...

> cat << EOF > luigi.cfg
[scheduler]
record_task_history = True

[task_history]
db_connection = sqlite:///history.sqlite
EOF

> luigid
...
```

Then visit http://localhost:8082/history

### Expected behaviour

Task history UI is shown.

### Actual behaviour

"500: Internal Server Error" is shown.

Logs from luigid:
```
2026-04-08 15:22:19,391 tornado.application[87851] ERROR: Uncaught exception GET /history (::1)
HTTPServerRequest(protocol='http', host='localhost:8082', method='GET', uri='/history', version='HTTP/1.1', remote_ip='::1')
  File "/Users/arthurb/.virtualenvs/temp/lib/python3.13/site-packages/tornado/web.py", line 1878, in _execute
  File "/Users/arthurb/.virtualenvs/temp/lib/python3.13/site-packages/luigi/server.py", line 236, in get
    self.render("recent.html", tasks=tasks)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/arthurb/.virtualenvs/temp/lib/python3.13/site-packages/tornado/web.py", line 1025, in render
    html = self.render_string(template_name, **kwargs)
  File "/Users/arthurb/.virtualenvs/temp/lib/python3.13/site-packages/tornado/web.py", line 1157, in render_string
    template_path = self.get_template_path()
  File "/Users/arthurb/.virtualenvs/temp/lib/python3.13/site-packages/luigi/server.py", line 160, in get_template_path
    return importlib.resources.files("templates").name
           ^^^^^^^^^^^^^^^^^^^
AttributeError: module 'importlib' has no attribute 'resources'
2026-04-08 15:22:19,397 tornado.access[87851] ERROR: 500 GET /history (::1) 13.44ms
```

### Notes

It looks like the relevant change was fc62c364c76ed2ba3fc08e263712b6d9489dbde7, so this affects 3.7.3+. Should be an easy fix.

dlstadther:
## Description

Replaces the broken `importlib.resources` usage in `BaseTaskHistoryHandler.get_template_path()` with the simpler `os.path` approach already used for `static_path` in the same file.

The previous code had three problems introduced in fc62c364 when `pkg_resources` was removed:
1. `import importlib` does not make `importlib.resources` available — it must be explicitly imported as a submodule
2. `files("templates")` references a non-existent top-level package named `"templates"`; the correct anchor is `"luigi"`
3. `.name` returns only the basename ("templates"), not the full filesystem path that Tornado requires

The fix uses `os.path.join(os.path.dirname(__file__), "templates")`, which is already the established pattern for `static_path` on the line immediately below and requires no new imports.

## Motivation and Context

Fixes #3415 . Visiting `/history` on any luigi scheduler with `record_task_history = True` results in a 500 Internal Server Error on luigi 3.7.3+ due to the AttributeError described above. This affects all supported Python versions (3.10–3.13).

New test included.

I also manually reproduced the 500 error against a fresh Python 3.13 environment with luigi==3.8.0 and confirmed the fix resolves it (using the "Steps to reproduce" included in the linked Issue).

### DIFF
--- a/luigi/server.py
+++ b/luigi/server.py
@@ -37,7 +37,6 @@ See :doc:`/central_scheduler` for more info.
 
 import atexit
 import datetime
-import importlib
 import json
 import logging
 import os
@@ -157,7 +156,7 @@ class BaseTaskHistoryHandler(tornado.web.RequestHandler):
         self._scheduler = scheduler
 
     def get_template_path(self):
-        return importlib.resources.files("templates").name
+        return os.path.join(os.path.dirname(__file__), "templates")
 
 
 class AllRunHandler(BaseTaskHistoryHandler):
--- a/test/server_test.py
+++ b/test/server_test.py
@@ -483,6 +483,19 @@ class MetricsHandlerTest(unittest.TestCase):
             patched_write.assert_not_called()
 
 
+def test_get_template_path_returns_existing_directory():
+    """Regression test for https://github.com/spotify/luigi/issues/3415.
+
+    The returned path must point to the luigi/templates directory that
+    tornado uses to render task-history pages.
+    """
+    handler = luigi.server.BaseTaskHistoryHandler(tornado.web.Application(), mock.MagicMock(), scheduler=mock.MagicMock())
+    path = handler.get_template_path()
+    assert os.path.isdir(path), f"get_template_path() returned non-existent directory: {path!r}"
+    assert os.path.basename(path) == "templates"
+    assert os.path.isfile(os.path.join(path, "recent.html"))
+
+
 class FromUtcTest(unittest.TestCase):
     def test_with_microseconds(self):
         """Test parsing UTC time string with microseconds"""


dlstadther:
@RRap0so , could you review? Feel free to tag some other folk who i should be raises requests to.

dlstadther:
@cjordn , i've seen you help with updating some CI settings recently. Would you be comfortable helping with a couple reviews? This one and the linked release.

If not, would you be able to direct me towards some other folks who could help out? (While this project doesn't get the attention it once did, we have a maintainer bottleneck)

RRap0so (approved):

PR #3420, opened 2026-04-16
feat: relax outdated dependency constraints for gcloud and mock

hiro-o918:
## Description

Relax outdated version constraints in `pyproject.toml`.

| Package | Before | After |
|---|---|---|
| `google-api-python-client` | `>=1.6.6,<2.0` | `>=2.0` |
| `google-auth` | `==1.4.1` | `>=2.0` |
| `google-auth-httplib2` | `==0.0.3` | `>=0.2` |
| `httplib2` | (transitive) | `>=0.22` (added; used directly in `bigquery.py` / `gcs.py`) |
| `mock` | `<2.0` | unconstrained |

Pins like `google-auth==1.4.1` (2018) and `mock<2.0` blocked security updates and constrained downstream resolution. Upper bounds dropped — luigi is a library and CI will catch regressions.

Also fixes a latent test bug in `test/contrib/s3_test.py` that was masked by `mock<2.0`: `called_with` (without `assert_` prefix) is a no-op attribute access, and `sts_mock.client.assume_role` referenced the wrong child mock. With `mock>=2.0` this is detected and raises `AttributeError`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ common = [
   "HTTPretty==0.8.10",
   "hypothesis>=6.7.0,<7.0.0",
   "jsonschema",
-  "mock<2.0",
+  "mock",
   "moto>=1.3.10,<5.0",
   "mypy",
   "mysql-connector-python",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -106,9 +106,10 @@ dropbox = [
 ]
 
 gcloud = [
-  "google-api-python-client>=1.6.6,<2.0",
-  "google-auth==1.4.1",
-  "google-auth-httplib2==0.0.3",
+  "google-api-python-client>=2.0",
+  "google-auth>=2.0",
+  "google-auth-httplib2>=0.2",
+  "httplib2>=0.22",
```
dlstadther, line 112: I do worry that by these not defining upper bounds that major releases will be allowed through and break things without warning. However, it would be limited only to users installing `luigi[gcloud]`.

Would you be opposed to setting upper bounds on major version here?

PR #3421, opened 2026-05-04
Improve error for contrib when the dependency packages are not installed

hirosassa:
## Description

This PR improves error behavior when using `contrib` modules without their required dependencies installed.

## Motivation and Context
In the current implementation, when a dependency is missing, only a warning is logged at import time. This means users may not notice the issue until the code is actually executed. Furthermore, since the ImportError is silently swallowed in the except block, runtime errors manifest as NameError or AttributeError instead, making root cause investigation difficult.

In this change, I introduce an availability flag in each `contrib` module's try/except ImportError block and checks it in the __init__ of the relevant classes. When a required dependency is not installed, a clear ImportError is raised at instantiation time with a message indicating which package to install (e.g., pip install boto3). This allows users to identify the problem earlier and with a more actionable error message.

Add tests and passed CI successfully.


luigi/contrib/bigquery_avro.py
```diff
@@ -30,6 +32,11 @@ class BigQueryLoadAvro(BigQueryLoadTask):
 
     source_format = SourceFormat.AVRO
 
+    def __init__(self, *args, **kwargs):
+        if not _avro_enabled:
+            raise ImportError("avro is required for BigQueryLoadAvro. Install it with: pip install avro-python3")
```
dlstadther, line 37: Indirectly related to this PR `avro-python3` is deprecated in favor of `avro`. Something worth addressing separately.

luigi/contrib/dataproc.py
```diff
@@ -42,6 +40,11 @@ class _DataprocBaseTask(luigi.Task):
 
     dataproc_client = get_dataproc_client()
 
+    def __init__(self, *args, **kwargs):
+        if not _dataproc_enabled:
+            raise ImportError("google-api-python-client is required for Dataproc functionality. Install it with: pip install google-api-python-client")
```
dlstadther: This module's import also imports the separate `google-auth` library. The previous warning message referenced it, but the new one does not.

luigi/contrib/ftp.py
```diff
@@ -73,7 +73,7 @@ def _sftp_connect(self):
         try:
             import pysftp
         except ImportError:
-            logger.warning("Please install pysftp to use SFTP.")
+            raise ImportError("pysftp is required for SFTP functionality. Install it with: pip install pysftp")
```
dlstadther, line 76: Outside of scope - pysftp hasn't been updated in a very long time. Should probably be replaced with paramiko.

luigi/contrib/gcp.py
```diff
@@ -25,6 +24,8 @@ def get_authenticate_kwargs(oauth_credentials=None, http_=None):
 
     Used by `gcs.GCSClient` and `bigquery.BigQueryClient` to initiate the API Client
     """
+    if not _gcp_enabled:
+        raise ImportError("google-auth and httplib2 are required for GCP functionality. Install them with: pip install google-auth httplib2")
```
dlstadther: I think the warning meant `google-auth-httplib2` instead of `httplib2`.

luigi/contrib/redshift.py
```diff
@@ -32,7 +32,7 @@
     import psycopg2
     import psycopg2.errorcodes
 except ImportError:
-    logger.warning("Loading postgres module without psycopg2 installed. Will crash at runtime if postgres functionality is used.")
+    pass
```
dlstadther, line 36: Could we make an in-line code comment that this will fallback to using contrib/postgres for the check for dependency availability?

dlstadther (commented):
I worry a little about the non-dryness between defining the name of the dependent python package(s) as a string in a exception message, when there are a subset of dependency groups defined in the pyproject.toml . A more ideal pattern (IMO) would be to define 1 dependency group per contrib module (where needed) which defines the range of dependent packages and versions it requires. Then the contrib exception can mention something like "missing package dependencies, install with luigi[s3]".

luigi/contrib/redshift.py
hirosassa: fixed in [4aa8763](https://github.com/spotify/luigi/pull/3421/commits/4aa8763b079dc4311632393083b18efd67325ce9)

luigi/contrib/gcp.py
hirosassa: fixed in 1501bb6

luigi/contrib/dataproc.py
hirosassa: fixed in 1501bb6

hirosassa:
> A more ideal pattern (IMO) would be to define 1 dependency group per contrib module 

I agree with this idea. 
I think it would be best to first define contrib-specific extras (e.g., s3, gcloud, docker) in [project.optional-dependencies] in a separate PR, and then update the error messages to something like pip install luigi[gcloud].

hirosassa:
CI for Azure blob is flaky recently

dlstadther (approved):
I'm not sure why the codecov failure is here.

dlstadther:
@hirosassa , i'm not in a position to override the failing codecov check. It appears the codecov complaint is about the qty of new lines that aren't executed by any test.

hirosassa:
@dlstadther Thanks for flagging this. The new lines that codecov is complaining about are the if not _xxx_enabled: raise ImportError(...) checks in each contrib module's __init__. These paths only execute when the required dependency packages are not installed, but in CI, those dependencies are installed, so these lines are inherently uncoverable under normal test runs.

To cover them, we'd need to mock sys.modules or reload modules with dependencies removed, which adds test complexity without much real value — it would just verify that if not flag: raise works.

Would it be possible to override the codecov check and merge as-is, or is there a preferred way to handle this?

dlstadther:
> Would it be possible to override the codecov check and merge as-is, or is there a preferred way to handle this?

Technically, yes. But i don't have the repo permissions to be able to do this. I've historically needed someone like @RRap0so to help in these scenarios.

PR #3424, opened 2026-05-10
Add type annotations to Task.requires(), Task.output(), and Task.input()

hirosassa:
## Description

- Add return type annotations to Task.requires(), Task.output(), and Task.input() in luigi/task.py
- Add # type: ignore[assignment] to LSFJobTask.output parameter in luigi/contrib/lsf.py which shadows the Task.output method with a Parameter attribute

## Motivation and Context

Task.requires(), output(), and input() previously had no type annotations, causing type checkers (e.g. Pyright) to infer return types as list[Unknown], which led to false reportIncompatibleMethodOverride errors in downstream libraries that override these methods

Type check succeeded.

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -30,7 +30,10 @@ import traceback
 import warnings
 from collections import OrderedDict, deque
 from contextlib import contextmanager
-from typing import Any, Dict, Optional
+from typing import TYPE_CHECKING, Any, Dict, Optional, Union
+
+if TYPE_CHECKING:
+    from luigi.target import Target
 
 from typing_extensions import dataclass_transform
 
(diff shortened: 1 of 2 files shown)


hirosassa:
The CI failure doesn't seem to be related to my changes.

luigi/task.py
```diff
@@ -613,7 +616,7 @@ def bulk_complete(cls, parameter_tuples):
         """
         raise BulkCompleteNotImplementedError()
 
-    def output(self):
+    def output(self) -> Union["Target", list["Target"], dict[str, "Target"]]:
```
dlstadther: I believe 3.10 supports the `|` syntax as a replacement for `typing.Union`. They accomplish the same thing but without the import. `"Target" | list["Target"] | dict[str, "Target"]`.

Do you have a strong preference?

dlstadther (commented):
1 minor comment/question

luigi/task.py
hirosassa: Thank you for your comment! fixed in f8af1fd

dlstadther (approved):

PR #3426, opened 2026-05-10
add task_register to __all__ and type annotations to deps() / flatten()

hirosassa:
## Description

In this PR, I added task_register to __all__ and type annotations to deps() / flatten()

## Motivation and Context

- Add task_register to `luigi/__init__.py`'s `__all__` so that type checkers recognize luigi.task_register as a valid attribute
- Add return type annotation -> list["Task"] to Task.deps()
- Add type annotations (struct: Any) -> list[Any] to flatten()

Type check succeeded

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -676,7 +676,7 @@ class Task(metaclass=Register):
         """
         return getpaths(self.requires())
 
-    def deps(self):
+    def deps(self) -> list["Task"]:
         """
         Internal method used by the scheduler.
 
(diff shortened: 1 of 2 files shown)


luigi/task.py
```diff
@@ -676,7 +676,7 @@ def input(self):
         """
         return getpaths(self.requires())
 
-    def deps(self):
+    def deps(self) -> list["Task"]:
```
dlstadther, line 679: Perhaps doesn't really help right now (since we still officially support 3.10), but if we were 3.11+, we could change this to `list[typing.Self]`

dlstadther (approved):

luigi/task.py
hirosassa: I'll change this on this November 👍 (3.10 EoL)

PR #3427, opened 2026-05-12
Update notifications.py to send multiple images

RaquelDiazG:
## Description
Add support to send multiple images in notifications
Add try-catch in ses
Add tests

## Motivation and Context
I used this module to send notifications with multiple images

I ran this code in my production environment

### DIFF
--- a/luigi/notifications.py
+++ b/luigi/notifications.py
@@ -119,7 +119,7 @@ def generate_email(sender, subject, message, recipients, images_png):
             with open(image_png, "rb") as fp:
                 msg_image = MIMEImage(fp.read(), "png")
                 filename = image_png.split("/")[-1]
-                msg_image.add_header('Content-Disposition', "attachment; filename={}".format(filename))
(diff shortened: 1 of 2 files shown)


dlstadther (changes requested):
I'm good with the concept of the change here.

I have 2 concerns (1 new, 1 existing):
1. new concern - anyone calling `send_mail()` with `image_png` kwarg will begin seeing failures after this change - consider how to make your change backwards compatible
2. existing concern - there's a fair bit of repetition across these various "send mail" methods; i know there's some nuance in implementation for SNS vs sendgrid vs ... , but it does smell like a missing abstraction

Please address concern # 1.

RaquelDiazG:
I fixed your 2 concerns

dlstadther (approved):
Thanks for addressing my concerns.

The only concern that continues to exist is the single scenario where the end user is calling `send_email(..., image_png=...)`. With this change, they'll have a breaking change. But if they're providing as a positional arg, you've got them covered. I think we'll keep this like you have it and require this to be a minor bump.

PR #3428, opened 2026-05-16
Widen return types of Task.input/output/requires to include tuples and nested structures

hirosassa:
## Description
- Introduce a recursive `FlattenableItems` type alias and use it as the return type of `Task.input()`, `Task.output()`, and `Task.requires()`
- The previous type `T | list[T] | dict[str, T]` did not include `tuple` or nested structures (e.g. `list[list[T]]`), even though the runtime implementation (`flatten`, `getpaths`) already handles them via `isinstance(struct, (list, tuple))` and recursive calls

## Motivation and Context
Luigi's own `flatten()` and `getpaths()` already handle tuples and recursively nested structures at runtime. However, the current type annotations on `Task.input()`, `Task.output()`, and `Task.requires()` only declare `T | list[T] | dict[str, T]`, which is narrower than what the implementation actually supports. This gap causes false positive mypy `[override]` errors in subclasses that return tuples or nested collections — patterns that are perfectly valid at runtime.

Aligning the type annotations with the actual runtime behavior makes the type system more accurate and removes unnecessary friction for downstream subclasses.

CI succeeded

### DIFF
--- a/luigi/task.py
+++ b/luigi/task.py
@@ -32,7 +32,7 @@ import traceback
 import warnings
 from collections import OrderedDict, deque
 from contextlib import contextmanager
-from typing import TYPE_CHECKING, Any, Dict, Optional
+from typing import TYPE_CHECKING, Any, Dict, Optional, TypeAlias, TypeVar
 
 if TYPE_CHECKING:
     from luigi.target import Target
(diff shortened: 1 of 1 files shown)


dlstadther (approved):
Necessary due to complex return types.

Sorry for missing this notification a few days ago. I've restarted the CI; assuming it completes successfully, it'll merge.

PR #3267, opened 2023-12-20
Fix for sqlalchemy 2 and text upgrades

pjaol:
## Description
Quick fix for task history logging and sqlalchemy 2.x

## Motivation and Context
Using sqlalchemy 2.x breaks luigid if task history is enabled
e.g.
```toml
[task_history]
db_connection = sqlite:///./log/luigi-task-hist.sqlite
```
Results in luigid exiting 
```python
File ".../env/lib/python3.11/site-packages/luigi/db_task_history.py", line 271, in _upgrade_schema
    for i in conn.execute('PRAGMA table_info(task_parameters);').fetchall():
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../env/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1414, in execute
    raise exc.ObjectNotExecutableError(statement) from err
sqlalchemy.exc.ObjectNotExecutableError: Not an executable object: 'PRAGMA table_info(task_parameters);'

```

This is because sqlalchemy requires a sqlalchemy.text object vs string in conn.execute

## Have you tested this? If so, how?
Tested locally with sqlite3 and sqlalchemy 2.0.23 and luigi 3.4.0

PR #3429, opened 2026-05-29
Sqlalchemy 2 fix

pmgast:
## Description
As mentioned in the referenced PR luigi with enabled task_history is not compatible with sqlalchemy version >2 as raw sql is now supposed to be supplied with the `text()` construct (See https://docs.sqlalchemy.org/en/14/changelog/migration_20.html)
I have merged spotify/luigi master into the sqlalchemy_2_fix branch from @pjaol and updated the syntax changes.

## Motivation and Context
https://github.com/spotify/luigi/pull/3267

I tested with `tox run -e py310-core` locally.


pmgast:
I get strange result during testing with the CI. It reports test/contrib/docker_runner_test.py::TestDockerTask::test_fail_job_container as failed but on my local machine the test passes, so it is hard for me to figure out what is wrong...

dlstadther:
The docker contrib failure was transient. After rerunning CI, I now see a ruff format error. Please check and address.

pmgast:
From my point of view this PR is ready to be merged. If more work is required one the coverage I would need some directions...

dlstadther:
The failing check is CodeCov's patch coverage gate - it requires 50% of the new/changed lines in your PR to be covered by the test suite. This isn't a reflection on the quality of the fix itself.

To get this merged, there are two paths:
1. Expand the tests in `test/db_task_history_test.py` to cover more of the changed lines in `db_task_history.py` - even basic mocking of the DB session can be enough to clear the 50% threshold.
2. Update `.codecov.yml` to add the affected files to the ignore list (alongside the other hard-to-test database-backed modules already listed there). That would look something like:
```
ignore:
    - "luigi/db_task_history.py"
```

Either approach works. I've had mixed feelings about the codecov patch check and have almost removed it before; however, it does uncover areas of weaker test coverage in the project.

dlstadther (approved):
Thanks for your persistence here! And sorry for my delays!

PR #3433, opened 2026-06-18
Support python 3.14

liamlundy:
## Description
This PR adds test support for python 3.14 and updates the project requirements to reflect 3.14 adoption

## Motivation and Context
Currently you cannot install luigi on python 3.14 due to the version restrictions

I'm running this through the full CI test bed


dlstadther:
@liamlundy Take a look at the CI failures. On first glance, looks like incompatibility with some current dependency versions.

liamlundy:
> @liamlundy Take a look at the CI failures. On first glance, looks like incompatibility with some current dependency versions.

@dlstadther I was able to fix the dependency issues but also ran into a multiprocessing issue that stems from the [default start method changing in 3.14](https://docs.python.org/3.14/whatsnew/3.14.html). I set the method manually in the conftest.py so that the tests pass but I wasn't sure what the desired fix was here. Ideally, the code would be fixed to [work with forkserver](https://docs.python.org/3.14/library/multiprocessing.html#multiprocessing-programming-forkserver) but I didn't want to bloat this PR.

liamlundy:
@dlstadther is there anything else i should change before mergin this PR?

test/conftest.py
```diff
@@ -1,9 +1,13 @@
+import multiprocessing
 from typing import List
 
 import pytest
 
 import luigi.task_register
 
+# Set the multiprocessing start method to "fork" for tests to avoid pickling issues
+multiprocessing.set_start_method("fork", force=True)
```
dlstadther: @liamlundy , I haven't read the full implications of the links you shared regarding 3.14's default start method change. However, by changing the multiprocess start method here, you're changing it for all tested python versions - not only 3.14. Which wouldn't be the behavior we see for <3.14.

At a minimum, i think this conftest change should only take effect if the version of python being run is >=3.14.

---

Separately, what are you experiencing when running your luigi workflows on 3.14 (including the changes on this branch)? Are there other changes you require for this multiprocess start behavior to function correctly?

liamlundy: @dlstadther makes sense to me. I'll make that change apply only to the 3.14 tests.

A quick summary of the changes in python 3.14 and the impact on the luigi is as follows:
Python 3.14 changes the default multiprocessing start method on linux to be forkserver instead of fork. The implication of this that matters for luigi is that all objects must be pickable. There are at least a few exceptions to this that would need to be refactored for forkserver to work. The cases I looked at were mostly anonymous functions that couldn't be pickled.

I can create a separate ticket that enumerates the impacted classes that would need to change in order to be compatible with forkserver.

Another option would be to change the default pickling library to be something that can handle a wider range of objects.

liamlundy: This should be the only code change need to support python 3.14. The other changes were just version bumps for some dependencies in pyproject.toml.

dlstadther: Thanks for the adjustment and elaboration.

I'm good with the current state here to approve, though it sounds like it's possible there may still be areas where the new default multiprocessing start method may yet cause issues with areas of luigi's code.

dlstadther (approved):

PR #3434, opened 2026-06-29
feat: add OptionalDateParameter

inakam:
## Description

Add `OptionalDateParameter`, an optional variant of `DateParameter`, following the established pattern used by existing optional parameters (e.g. `OptionalIntParameter`, `OptionalFloatParameter`):

```python
class OptionalDateParameter(OptionalParameterMixin[datetime.date], DateParameter):
    """Class to parse optional date parameters."""
    expected_type = datetime.date
```

## Motivation and Context

Concrete use case: filtering rows from a database table from a given date onward, while applying no date filter when the parameter is `None`. Currently this requires defining a custom subclass via `OptionalParameterMixin`, despite the mixin being designed for exactly this purpose. See existing optional parameter additions in #3406.

Per review feedback, this PR is limited to the single type with an immediate use case (the other date/datetime/interval variants were dropped).

## Have you tested this?

I have included a unit test using the same `actual_test` helper as the other optional parameter tests. One difference: the bad-data warning check is skipped, because `DateParameter.normalize()` raises on non-date values before the type warning is emitted.


dlstadther:
@inakam , does this introduce parameter types you require? Or is it just to contribute towards "completeness"? I'm not inclined to add parameter types if there aren't explicit requested uses of them.

inakam:
@dlstadther 
Sorry for the late reply.
I do have a concrete use case for `OptionalDateParameter`: filtering rows from a database table from a given date onward, while applying no date filter when the parameter is `None`.

I currently define it using `OptionalParameterMixin`. I included the other date/time variants because, if we add one, it seemed reasonable to support the whole family. But I’m happy to narrow the PR to `OptionalDateParameter` if you’d prefer to add only types with an immediate use case.

dlstadther:
Thanks for the reply. For the time being, let's limit to only types with immediate use cases. Thanks!

dlstadther (approved):

PR #3449, opened 2026-08-23
Fix Windows compatibility: upgrade psutil constraint and default TextWrapper to LF newlines

VimalN2005:
## Description
This PR makes two key changes to fix development environment setup and core test failures on Windows machines:
1. **Upgrades the `psutil` constraint** in `pyproject.toml` from `<4.0` to `>=5.0.0`.
2. **Sets the default newline parameter in `TextWrapper`** in `luigi/format.py` to `'\n'` (LF).

## Motivation and Context
1. **`psutil` installation failure:** When setting up the development environment on Windows (specifically with Python 3.12+), installing `psutil<4.0` forces `uv` or `pip` to attempt to build `psutil 3.4.2` from source, which fails to compile on modern Python versions without MSVC Build Tools. Upgrading to `>=5.0.0` allows the installer to download a pre-compiled wheel.
2. **Test failures due to Carriage Returns (`\r\n`):** On Windows, `io.TextIOWrapper` (underneath `TextWrapper`) translates line endings to `\r\n` (CRLF) by default. This causes mock target test assertions that expect `\n` (LF) to fail (e.g., `AssertionError: b'55\r\n' != b'55\n'`). Standardizing `TextWrapper`'s default newline to `\n` ensures cross-platform consistency of output line endings and allows all tests to pass successfully on Windows.

I have run the unit tests. I verified that both `test/fib_test.py` and `test/decorator_test.py` run and pass successfully on a Windows machine after applying these changes.


VimalN2005:
All checks have passed successfully. Please let me know if any further changes or reviews are needed. Thanks!

